PR #12, opened 2017-05-20
node docker update

Goodsmileduck:


Dockerfile-mysterium-node
```diff
@@ -1,15 +1,21 @@
...
-ARG PACKAGE_FILE
-COPY $PACKAGE_FILE /tmp/mysterium-node.deb
+ENV VERSION 0.0.6
+ENV NODE 321
 
 # Install packages
 RUN apt-get update \
-    && dpkg --install /tmp/mysterium-node.deb
-    && apt-get install --fix-broken \
+    && apt install -y wget openvpn iptables\
```
Waldz: I would remove: "openvpn iptables", they are required in .deb package with more exact version

Dockerfile-mysterium-node
```diff
@@ -1,15 +1,21 @@
 FROM ubuntu:latest
 MAINTAINER Valdas Petrulis <petrulis.valdas@gmail.com>
 
-# e.g. PACKAGE_FILE="build/package/mysterium-node_0.0.1_linux_amd64.deb"
-ARG PACKAGE_FILE
-COPY $PACKAGE_FILE /tmp/mysterium-node.deb
+ENV VERSION 0.0.6
+ENV NODE 321
```
Waldz: Not good hardcoding NODE key, it should be injected with container environment variables

entrypoint-node.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+set -e
+
+mkdir -p /dev/net
```
Waldz: Can this be done In Dockerfile, not during startup?

Dockerfile-mysterium-node
```diff
@@ -1,15 +1,21 @@
...
+    && rm mysterium-node_linux_amd64.deb
 
-ENTRYPOINT ["/usr/bin/mysterium_server"]
\ No newline at end of file
+COPY entrypoint-node.sh /entrypoint-node.sh
+RUN chmod +x /entrypoint-node.sh
+
+ENTRYPOINT ["/entrypoint-node.sh"]
+
+CMD mysterium_server --config-dir=/etc/mysterium-node --node=${NODE} --runtime-dir=/tmp
```
Waldz: Exact starting command should be like this:
```
sudo /usr/bin/mysterium_server --node=NODE_KEY --config-dir=/etc/mysterium-node --runtime-dir=/var/run/mysterium-node
```

Dockerfile-mysterium-node
```diff
@@ -1,15 +1,21 @@
 FROM ubuntu:latest
```
ewrogers: `FROM debian:jessie-slim` yields a smaller container size.

entrypoint-node.sh
ewrogers: Yes, seems you can add
```
# Create OpenVPN tunnel device
RUN set -e \
    && mkdir -p /dev/net \
    && mknod /dev/net/tun c 10 200 \
    && chmod 600 /dev/net/tun
```

docker-compose.yml
```diff
@@ -6,12 +6,16 @@ services:
       context: .
       dockerfile: Dockerfile-mysterium-node
     container_name: mysterium-node
+    cap_add:
+      - NET_ADMIN
```
ewrogers: I haven't had luck with `--cap-add=NET_ADMIN`. I needed to use `--privileged` for IP forwarding to work.

Dockerfile-mysterium-node
Goodsmileduck: build not wotking without it, don't know why. but that probably about ubuntu version

Dockerfile-mysterium-node
Goodsmileduck: yes, you should, that just a example. there env in docker-compose.yml

docker-compose.yml
Goodsmileduck: privileged is not good approach, would recommend it for debug. it also didn't work for me if I create tun in RUN step on build, with entrypoint it works.

Dockerfile-mysterium-node
Goodsmileduck: so you need add RUN step to create /var/run/mysterium-node and chown

Dockerfile-mysterium-node
Goodsmileduck: I have checked it again, dpkg didn't install dependecies and give a error, so build will be failed.
Other way to use apt-get directly
`RUN apt-get update \
    && apt install -y wget\
    && wget https://github.com/MysteriumNetwork/node/releases/download/${VERSION}/mysterium-node_linux_amd64.deb \
    && apt-get install -y ./mysterium-node_linux_amd64.deb \
    && rm -rf /var/cache/apk/* \
    && rm mysterium-node_linux_amd64.deb`

entrypoint-node.sh
Goodsmileduck: I have tested it, it didn't works. Sorry, I know less about mknod. entrypoint way is not a hack, I would recommend it. Later you can preheat some other stuff in this script.

PR #30, opened 2017-09-05
Custom Communication Channel, Glide integration

Waldz:


cmd/mysterium_server/command_run/command.go
```diff
@@ -72,6 +76,15 @@ func (cmd *commandRun) Run(options CommandOptions) error {
 		}
 	}()
 
+	cmd.communication, err = cmd.communicationOptions.Connect()
+	if err != nil {
+		return err
+	}
+	err = cmd.communication.Publish("node-register", []byte(options.NodeKey))
```
ignasbernotas: What about using constants for event names?

Waldz: Implemented constants

PR #40, opened 2017-10-23
MYST-20 Communication dialog request/Response DTO models

Waldz:


Waldz:
Improvements done, review appreciated

tadovas:
Is it possible to extract hardcoded nats server address/port to options somehow? Or at least to declare inner constant in nats package. Now "127.0.0.1" is lying around everywhere.

communication/codec_bytes_test.go
```diff
@@ -0,0 +1,77 @@
+package communication
+
+import (
+	"github.com/pkg/errors"
+	"github.com/stretchr/testify/assert"
+	"testing"
+)
+
+func TestCodecBytesInterface(t *testing.T) {
+	var _ Codec = NewCodecBytes()
```
ignasbernotas: This test does nothing?

communication/codec_json_test.go
```diff
@@ -0,0 +1,106 @@
...
+	"github.com/stretchr/testify/assert"
+	"testing"
+)
+
+type customPayload struct {
+	Field int
+}
+
+func TestCodecJSONInterface(t *testing.T) {
+	var _ Codec = NewCodecJSON()
```
ignasbernotas: Same as above, does it do anything?

communication/nats_discovery/address.go
```diff
@@ -0,0 +1,77 @@
...
+	return &NatsAddress{
+		servers: []string{
+			fmt.Sprintf("nats://%s:%d", server, port),
+		},
+		topic: topic,
+	}
+}
+
+func NewAddressForIdentity(identity dto_discovery.Identity) *NatsAddress {
+	return NewAddress("127.0.0.1", 4222, string(identity))
```
ignasbernotas: Is it ok to hardcode "127.0.0.1" + port here?

ignasbernotas: Maybe worth changing into a variable we could modify during compilation?

communication/codec_bytes_test.go
Waldz: This test check if struct implements wanted interface

communication/codec_bytes.go
```diff
@@ -0,0 +1,43 @@
...
+		return []byte(payload), nil
+	}
+
+	return []byte{}, fmt.Errorf("Cant pack payload: %#v", payloadPtr)
+}
+
+func (codec *codecBytes) Unpack(data []byte, payloadPtr interface{}) error {
+	switch payload := payloadPtr.(type) {
+	case *[]byte:
+		*payload = data
```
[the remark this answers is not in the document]
Waldz: It's casting.
But we dont have byte streams yet, so I dont want to polish them without real case.

communication/nats_discovery/address.go
Waldz: It's the same how it was before, we expect queue server to run next to node&client.
We have a ticket to make it injectable

communication/codec_bytes.go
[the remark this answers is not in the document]
Waldz: What was the idea of story:
- you can write custom providers/consumers
- providers/consumers will provide DTOs of requests/responses
- DTOs can be any structs, what are able to serialise to JSON

PR #43, opened 2017-10-25
Added config variable replacement from .env file during builds

ignasbernotas:


.gitignore
```diff
@@ -1,3 +1,4 @@
 .idea
 /build
-/vendor
\ No newline at end of file
+/vendor
+.env
```
Waldz: `.env` -> `/.env` (in root directory only)

Waldz:
Could You create improvement ticket for Seelog level injecting, by environment

PR #50, opened 2017-11-22
Feature/myst 80 node generate identity

interro:


identity/identity.go
```diff
@@ -0,0 +1,77 @@
...
+func (idm *IdentityManager) GetIdentities() []string {
+	var ids []string
+	for _, account := range idm.keystoreManager.Accounts() {
+		ids = append(ids, account.Address.Hex())
+	}
+	return ids
+}
+
+func (idm *IdentityManager) IdentityExists() (bool) {
+	return len(idm.keystoreManager.Accounts()) > 0
```
shroomist: `IdentityExists` method sounds like it should lookup for an Account in keystore by string.
it now returns true if at least one id was registered.

PR #51, opened 2017-11-30
MYST-122 added identity generation and reuse by providing the ID. N…

ignasbernotas:
Added identity generation and reuse by providing the ID. Node uses it on boot if specified.

https://mysteriumnetwork.atlassian.net/browse/MYST-122

### DIFF
--- a/cmd/directory.go
+++ b/cmd/directory.go
@@ -1,0 +1,14 @@
+package cmd
+
+import (
+	"github.com/mitchellh/go-homedir"
+	"path/filepath"
+)
+
+func GetDirectory(paths ...string) string {
+	dir, _ := homedir.Dir()
+
+	dir = filepath.Join(dir, ".mysterium", filepath.Join(paths...))
+
+	return dir
+}
--- a/cmd/mysterium_server/command_run/command.go
+++ b/cmd/mysterium_server/command_run/command.go
@@ -14,6 +14,8 @@ import (
 	"github.com/mysterium/node/session"
 	"io"
 	"time"
+	"github.com/mysterium/node/identity"
+	"fmt"
 )
 
 type CommandRun struct {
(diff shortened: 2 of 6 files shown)


cmd/directory.go
```diff
@@ -0,0 +1,14 @@
+package cmd
+
+import (
+	"github.com/mitchellh/go-homedir"
+	"path/filepath"
+)
+
+func GetDirectory(paths ...string) string {
```
Waldz: This function does not need to be public

cmd/mysterium_server/command_run/options.go
```diff
@@ -1,14 +1,15 @@
...
-	"errors"
 	"flag"
+	"github.com/mysterium/node/cmd"
 )
 
 type CommandOptions struct {
-	NodeKey          string
-	DirectoryConfig  string
-	DirectoryRuntime string
+	NodeKey           string
```
Waldz: Should we call it `NodeKey`, maybe refactor to something like `Identity`

cmd/directory.go
```diff
@@ -0,0 +1,14 @@
...
+	dir, _ := homedir.Dir()
```
Waldz: Seems strange that error are intentionally unhandled

cmd/directory.go
```diff
@@ -0,0 +1,14 @@
...
+
+	dir = filepath.Join(dir, ".mysterium", filepath.Join(paths...))
```
Waldz: Does not seem that You need kwargs-like multiple variable

cmd/mysterium_server/command_run/command.go
```diff
@@ -32,7 +34,10 @@ type CommandRun struct {
 }
 
 func (cmd *CommandRun) Run(options CommandOptions) (err error) {
-	providerId := dto_discovery.Identity(options.NodeKey)
+	providerId, err := identity.SelectIdentity(options.DirectoryKeystore, options.NodeKey)
```
Waldz: What about more DI approach `cmd.identityManager.SelectIdentity(options.NodeKey)`

ignasbernotas:
Fixed in PR #53

PR #53, opened 2017-12-05
Feature/myst 81 reuse node identity

ignasbernotas:
Cache last stored identity

https://mysteriumnetwork.atlassian.net/browse/MYST-122
https://mysteriumnetwork.atlassian.net/browse/MYST-81


identity/cache_test.go
```diff
@@ -0,0 +1,48 @@
...
+func Test_IdentityCacheGetIdentity(t *testing.T) {
+	identity := dto.Identity("0x000000000000000000000000000000000000000A")
+	cache := identityCache{
+		File: file,
+	}
+
+	err := cache.StoreIdentity(&identity)
+	assert.Nil(t, err)
+
+	_, err = os.Stat(file)
```
tadovas: Purpose of this line? err is ignored anyway.

identity/controller.go
```diff
@@ -0,0 +1,41 @@
+package identity
+
+import "github.com/mysterium/node/service_discovery/dto"
+
+type identityController struct {
```
tadovas: Naming. Usually name Controller stands for C in MVC paradigm and it is something facing user, handling its actions and rendering views. Maybe indentityController can be renamed to something like identityManager to avoid confusion with widely adopted terms.

ignasbernotas: how about identityHandler?

identity/cache_test.go
ignasbernotas: It's for a test, i don't need the file info, I just want to check if it exists

identity/controller.go
ignasbernotas: renamed to identityHandler

ignasbernotas:
Can I merge? @tadovas @Waldz

PR #54, opened 2017-12-07
MYST-159 Node. Call API method on creating new identity

ignasbernotas:
Node registers identity via indentity API


server/client_rest.go
```diff
@@ -31,6 +31,19 @@ type clientRest struct {
...
 }
 
+func (client *clientRest) RegisterIdentity(identity *dto_discovery.Identity) (err error) {
+	response, err := client.doRequest("POST", "identities", dto.CreateIdentityRequest{
+		Identity: *identity,
+	})
+
+	if err == nil {
+		defer response.Body.Close()
+		log.Info(MYSTERIUM_API_LOG_PREFIX, "Identity created: ", *identity)
```
Waldz: Identity registered..

server/client_rest.go
```diff
@@ -31,6 +31,19 @@ type clientRest struct {
 	httpClient http.Client
 }
 
+func (client *clientRest) RegisterIdentity(identity *dto_discovery.Identity) (err error) {
```
Waldz: We should reconsider our pointer convention.
Because now You can easily pass `nil` which unwanted situation

server/client_rest.go
```diff
@@ -31,6 +31,19 @@ type clientRest struct {
...
+	response, err := client.doRequest("POST", "identities", dto.CreateIdentityRequest{
```
Waldz: I miss test, which freezes request JSON structure. I know we did not have tests for this, but we should start having

server/client_fake.go
```diff
@@ -25,6 +25,12 @@ func (client *clientFake) NodeRegister(proposal dto_discovery.ServiceProposal) (
 	return nil
 }
 
+func (client *clientFake) RegisterIdentity(identity *dto_discovery.Identity) (err error) {
+	log.Info(MYSTERIUM_API_LOG_PREFIX, "Identity created: ", *identity)
```
Waldz: Fake identity registered

glide.yaml
```diff
@@ -6,13 +6,13 @@ import:
...
 - package: github.com/satori/go.uuid
   version: ^1.1.0
-- package: github.com/mitchellh/go-homedir
 - package: github.com/ethereum/go-ethereum
   version: ^1.7.2
   subpackages:
   - accounts
 - package: github.com/julienschmidt/httprouter
   version: ^1.1.0
+- package: github.com/mitchellh/go-homedir
```
Waldz: Should not version be freezed?

identity/handler.go
```diff
@@ -30,6 +30,12 @@ func SelectIdentity(dir string, nodeKey string) (id *dto.Identity, err error) {
 		return
 	}
 
+	return
+}
+
+func CreateIdentity(dir string) (id *dto.Identity, err error) {
+	handler := NewIdentityHandler(dir)
```
Waldz: Use DI pattern. All dependencies should be injected for top layer, avoid constructing them in the middle of functions

Waldz: Same in `SelectIdentity()`, and we have situation then handler is constructed 2 times

cmd/mysterium_server/command_run/command.go
```diff
@@ -38,6 +38,18 @@ func (cmd *CommandRun) Run(options CommandOptions) (err error) {
 		return err
 	}
 
+	// if no identity can be selected, lets create a new one
+	if providerId == nil {
+		providerId, err = identity.CreateIdentity(options.DirectoryKeystore)
```
Waldz: What we decided on json metadata file about all local identities?
Could be like: 
```
[
    {
        "identity": "asdad12425",
        "default": false,
        "registered": true, 
    },
    {
        "identity": "asdad12425",
        "default": true,
        "registered": false, 
    }
]
```

cmd/mysterium_server/command_run/command.go
```diff
@@ -38,6 +38,18 @@ func (cmd *CommandRun) Run(options CommandOptions) (err error) {
 		return err
 	}
 
+	// if no identity can be selected, lets create a new one
+	if providerId == nil {
```
Waldz: This logic is very tight with selection process, and could easily fit in `identityHandler` aswel

Waldz:
Reformated PR from:
`MYST-159-node-api-identity`
To
`MYST-159 Node. Call API method on creating new identity`

cmd/mysterium_server/command_run/command.go
ignasbernotas: I thought that was just an idea. Is it confirmed now?

server/client_rest.go
Waldz: 1. Lets change signature of `IdentityManager.GetIdentity() *Identity` -> `IdentityManager.GetIdentity() Identity, error`
2. Same with all other `IdentityManager` methods, which returns pointer *Identity
3. Lets stop passing *Identity in all project which can easily become `nil`

cmd/mysterium_server/command_run/command.go
Waldz: 1. Most important is to start having internal structure:
```
type Identity struct {
    Name string
    IsDefault boolean
    IsRegistered boolean
}
```
2. Make `dto_discovery.Identity` private OR it's not needed anymore 
3. (Optional) Metadata storage it's nice to have (can be JSON, SQL lite, etc.) we can create anotjer ticket for that

server/client_rest.go
shroomist: Identities should be immutable, indestructible and persisted on myst network.

glide.yaml
ignasbernotas: Fixed.

identity/cache.go
```diff
@@ -22,22 +22,22 @@ func NewIdentityCache(dir string, jsonFile string) *identityCache {
 	}
 }
 
-func (ic *identityCache) GetIdentity() *dto.Identity {
```
Waldz: `GetIdentity() (dto.Identity, error)` to recognise error about that result is empty

identity/handler.go
```diff
@@ -0,0 +1,77 @@
...
+
+	return
+}
+
+func (ih *identityHandler) getIdentityByValue(id string) (dto.Identity, error) {
+	if ih.manager.HasIdentity(id) {
+		return ih.manager.GetIdentity(id), nil
+	}
+
+	return dto.Identity(""), errors.New("identity not found")
```
shroomist: empty string identity is still a valid identity, shouldn't this be
`return nil, error()`

identity/manager.go
```diff
@@ -31,10 +31,10 @@ func identityToAccount(identityStr string) accounts.Account {
 	}
 }
 
-func (idm *identityManager) CreateNewIdentity(passphrase string) (*dto.Identity, error) {
+func (idm *identityManager) CreateNewIdentity(passphrase string) (dto.Identity, error) {
 	account, err := idm.keystoreManager.NewAccount(passphrase)
 	if err != nil {
-		return nil, err
+		return "", err
```
Waldz: Ideally should be `"" -> dto.Identity("")`, now your string is casted

identity/cache.go
```diff
@@ -22,22 +22,22 @@ func NewIdentityCache(dir string, jsonFile string) *identityCache {
...
+func (ic *identityCache) GetIdentity() (identity dto.Identity) {
```
Waldz: `GetIdentity() (dto.Identity, error)` to recognise error about that result is empty

identity/manager.go
```diff
@@ -45,23 +45,23 @@ func (idm *identityManager) GetIdentities() []dto.Identity {
...
 	for i, account := range accountList {
-		ids[i] = *accountToIdentity(account)
+		ids[i] = accountToIdentity(account)
 	}
 
 	return ids
 }
 
-func (idm *identityManager) GetIdentity(identityString string) *dto.Identity {
+func (idm *identityManager) GetIdentity(identityString string) dto.Identity {
```
Waldz: `GetIdentity() (dto.Identity, error)` to recognise error about that result is empty

identity/manager.go
```diff
@@ -45,23 +45,23 @@ func (idm *identityManager) GetIdentities() []dto.Identity {
...
 		}
 	}
 
-	return nil
+	return dto.Identity("")
 }
 
 func (idm *identityManager) HasIdentity(identityString string) bool {
-	return idm.GetIdentity(identityString) != nil
+	return len(idm.GetIdentity(identityString)) != 0
```
Waldz: Here You could be using error from `GetIdentity`

server/client_fake.go
```diff
@@ -25,6 +25,12 @@ func (client *clientFake) NodeRegister(proposal dto_discovery.ServiceProposal) (
 	return nil
 }
 
+func (client *clientFake) RegisterIdentity(identity dto_discovery.Identity) (err error) {
+	log.Info(MYSTERIUM_API_LOG_PREFIX, "Identity registered: ", identity)
```
Waldz: Mention something in logs, that fake registration was used e.g. `Fake identity registered`

server/client_rest.go
```diff
@@ -31,6 +31,19 @@ type clientRest struct {
...
 }
 
+func (client *clientRest) RegisterIdentity(identity dto_discovery.Identity) (err error) {
+	response, err := client.doRequest("POST", "identities", dto.CreateIdentityRequest{
+		Identity: identity,
+	})
+
+	if err == nil {
+		defer response.Body.Close()
+		log.Info(MYSTERIUM_API_LOG_PREFIX, "Identity created: ", identity)
```
Waldz: Registered, not created :)

identity/handler.go
```diff
@@ -0,0 +1,77 @@
...
+}
+
+func (ih *identityHandler) getIdentityFromCache() (identity dto.Identity, err error) {
+	id := ih.cache.GetIdentity()
+
+	if len(id) > 0 && ih.manager.HasIdentity(string(id)) {
+		return id, nil
+	}
+
+	return identity, errors.New("identity not found in cache")
```
shroomist: is this a common practice to return `nil` like this?

identity/manager.go
shroomist: should return `dto.Identity, err` not `string`

identity/manager.go
```diff
@@ -45,23 +45,23 @@ func (idm *identityManager) GetIdentities() []dto.Identity {
...
 	identityString = strings.ToLower(identityString)
 	for _, id := range idm.GetIdentities() {
 		if strings.ToLower(string(id)) == identityString {
-			return &id
+			return id
 		}
 	}
 
-	return nil
+	return dto.Identity("")
```
shroomist: `dto.Identity("")` is a valid identity, should probably return `nil, error`

identity/handler.go
```diff
@@ -0,0 +1,77 @@
...
+
+type identityHandler struct {
+	manager IdentityManagerInterface
+	cache   *identityCache
+}
+
+func NewNodeIdentityHandler(keystore keystoreInterface, cacheDir string) *identityHandler {
+	return &identityHandler{
+		manager: NewIdentityManager(keystore),
+		cache:   NewIdentityCache(cacheDir, "remember.json"),
```
shroomist: no error handling on file read/creation?

server/client_rest.go
ignasbernotas: Weird, I thought I fixed it somewhere:)

identity/manager.go
Waldz: Yeah that's normal to return empty struct on error, if you return is by-value, not by-reference

PR #60, opened 2017-12-18
MYST-153 / Refactored dto.Identity string to identity.Identity struct

ignasbernotas:
Have fun 😂

### DIFF
--- a/client_connection/manager.go
+++ b/client_connection/manager.go
@@ -2,7 +2,7 @@ package client_connection
 
 import (
 	"errors"
-	id "github.com/mysterium/node/identity"
+    "github.com/mysterium/node/identity"
 )
 
 type State string
@@ -23,7 +23,7 @@ type ConnectionStatus struct {
 }
 
 type Manager interface {
-	Connect(identity id.Identity, NodeKey string) error
+	Connect(identity identity.Identity, NodeKey string) error
 	Status() ConnectionStatus
 	Disconnect() error
 	Wait() error
@@ -37,7 +37,7 @@ func NewManager() *fakeManager {
 	return &fakeManager{make(chan error)}
 }
 
-func (nm *fakeManager) Connect(identity id.Identity, NodeKey string) error {
+func (nm *fakeManager) Connect(identity identity.Identity, NodeKey string) error {
 	return nil
 }
 
--- a/client_connection/manager.go
+++ b/client_connection/manager.go
@@ -2,7 +2,7 @@ package client_connection
 
 import (
 	"errors"
-	"github.com/mysterium/node/service_discovery/dto"
+	id "github.com/mysterium/node/identity"
 )
 
 type State string
@@ -23,7 +23,7 @@ type ConnectionStatus struct {
 }
 
 type Manager interface {
-	Connect(identity dto.Identity, NodeKey string) error
+	Connect(identity id.Identity, NodeKey string) error
 	Status() ConnectionStatus
 	Disconnect() error
 	Wait() error
@@ -37,7 +37,7 @@ func NewManager() *fakeManager {
 	return &fakeManager{make(chan error)}
 }
 
-func (nm *fakeManager) Connect(identity dto.Identity, NodeKey string) error {
+func (nm *fakeManager) Connect(identity id.Identity, NodeKey string) error {
 	return nil
 }
 
--- a/communication/nats_dialog/dialog_establisher.go
+++ b/communication/nats_dialog/dialog_establisher.go
@@ -42,8 +42,7 @@ func (establisher *dialogEstablisher) CreateDialog(contact dto_discovery.Contact
 	if err != nil || response.(*dialogCreateResponse).Reason != 200 {
 		return nil, fmt.Errorf("Dialog creation rejected: %s", err)
 	}
-
-	dialogAddress := nats_discovery.NewAddressNested(contactAddress, string(establisher.myIdentity.Address))
+	dialogAddress := nats_discovery.NewAddressNested(contactAddress, establisher.myIdentity.Address)
 	dialog := &dialog{nats.NewSender(dialogAddress), nats.NewReceiver(dialogAddress)}
 
 	log.Info(establisherLogPrefix, fmt.Sprintf("Dialog established with: %#v", contact))
--- a/communication/nats_dialog/dialog_establisher.go
+++ b/communication/nats_dialog/dialog_establisher.go
@@ -7,10 +7,10 @@ import (
 	"github.com/mysterium/node/communication/nats"
 	"github.com/mysterium/node/communication/nats_discovery"
 	dto_discovery "github.com/mysterium/node/service_discovery/dto"
-	id "github.com/mysterium/node/identity"
+    "github.com/mysterium/node/identity"
 )
 
-func NewDialogEstablisher(identity id.Identity) *dialogEstablisher {
+func NewDialogEstablisher(identity identity.Identity) *dialogEstablisher {
 	return &dialogEstablisher{
 		myIdentity: identity,
 	}
@@ -19,7 +19,7 @@ func NewDialogEstablisher(identity id.Identity) *dialogEstablisher {
 const establisherLogPrefix = "[NATS.DialogEstablisher] "
 
 type dialogEstablisher struct {
-	myIdentity id.Identity
+	myIdentity identity.Identity
 }
 
 func (establisher *dialogEstablisher) CreateDialog(contact dto_discovery.Contact) (communication.Dialog, error) {
--- a/communication/nats_dialog/dialog_establisher.go
+++ b/communication/nats_dialog/dialog_establisher.go
@@ -36,14 +36,14 @@ func (establisher *dialogEstablisher) CreateDialog(contact dto_discovery.Contact
 
 	response, err := nats.NewSender(contactAddress).Request(&dialogCreateProducer{
 		&dialogCreateRequest{
-			IdentityId: establisher.myIdentity.Id,
+			IdentityId: establisher.myIdentity.Address,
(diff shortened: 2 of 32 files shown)


client_connection/manager.go
```diff
@@ -2,7 +2,7 @@ package client_connection
 
 import (
 	"errors"
-	"github.com/mysterium/node/service_discovery/dto"
+	id "github.com/mysterium/node/identity"
```
tadovas: confusing naming. Identity from "id" package :/ why we need alias here? (and in other imports)

identity/identity.go
```diff
@@ -0,0 +1,9 @@
+package identity
+
+type Identity struct {
+	Id string
+}
+
+func NewIdentity(id string) Identity {
```
tadovas: Are we really making "new" identity here or just creating some kind of Identity from string? What about identity.FromString(id string)?

ignasbernotas: Hmm, not sure. The "New" convention is what we usually go for, so maybe worth leaving this as is? Though I do get what you're saying. @tadovas @Waldz

client_connection/manager.go
ignasbernotas: I just wanted to shorten it, plus there are quite a few variables called "identity" so it kind of collides with the package name. Didn't want to refactor too much.

shroomist: I did actually use `id` for variable names to avoid this clash in my code xD

ignasbernotas: :D so should I change it? Already got a commit for it.

identity/identity.go
shroomist: new identity created here in manager, and being put in keystore: https://github.com/MysteriumNetwork/node/blob/39cc02af6ff5f79c9803e311cfdce3af4f0f927f/identity/manager.go#L35
I agree with @tadovas

shroomist: to be more specific, I think it goes like `identity.FromEthAddress(address)`

ignasbernotas: Maybe use FromAddress(), with a change of Identity struct Id property to Address?

client_connection/manager.go
Waldz: Yeah, looks strange for me also

ignasbernotas: Cool, this will be pushed in a bit.

identity/identity.go
ignasbernotas: Just need advice on this last bit. Problem with Id -> Address rename is that a lot of outside structures (DTOs) use IdentityId naming.

tadovas: It would make sense if it signature would be FromAddress(ethAddress Address) but now we are "creating" "identity" from string, looks like a simple converter to me, and not to identity but to some kind of Identifier which can be used with identityManager to load real identity

communication/nats_dialog/dto_request_test.go
```diff
@@ -2,20 +2,20 @@ package nats_dialog
...
 	"encoding/json"
-	dto_discovery "github.com/mysterium/node/service_discovery/dto"
 	"github.com/stretchr/testify/assert"
 	"testing"
+	"github.com/mysterium/node/identity"
 )
 
 func TestRequestSerialize(t *testing.T) {
-	var identity = dto_discovery.Identity("123")
+	var identity = identity.NewIdentity("123")
```
Waldz: Dont need identity struct, could be just string

communication/nats_dialog/dto_request_test.go
```diff
@@ -48,14 +48,14 @@ func TestRequestUnserialize(t *testing.T) {
 				"identity_id": "123"
 			}`,
 			dialogCreateRequest{
-				IdentityId: dto_discovery.Identity("123"),
+				IdentityId: identity.NewIdentity("123").Id,
```
Waldz: Dont need identity struct, could be just string

identity/cache.go
```diff
@@ -1,15 +1,14 @@
...
-	"github.com/mysterium/node/service_discovery/dto"
 	"io/ioutil"
 	"encoding/json"
 	"os"
 	"path/filepath"
 )
 
 type cacheData struct {
-	Identity dto.Identity `json:"identity"`
+	Identity Identity `json:"identity"`
```
Waldz: should not be string?

identity/identity.go
Waldz: Looks like good idea `identity.address`

identity/cache.go
ignasbernotas: Doesn't matter. We can store the full structure in the cache file.

openvpn/service_discovery/factory.go
```diff
@@ -39,7 +40,7 @@ func NewServiceProposalWithLocation(
 			Price:    money.NewMoney(0.125, money.CURRENCY_MYST),
 			Duration: 1 * time.Hour,
 		},
-		ProviderId:       dto_discovery.Identity(providerId),
+		ProviderId:       id.Identity(providerId).Id,
```
Waldz: Those casting looks strange. Should not it be just `providerId.Id`?

openvpn/service_discovery/service_poposal_test.go
```diff
@@ -38,7 +39,7 @@ func TestServiceProposalUnserialize(t *testing.T) {
 		ServiceDefinition: dto_openvpn.ServiceDefinition{},
 		PaymentMethodType: "PER_TIME",
 		PaymentMethod:     dto_openvpn.PaymentMethodPerTime{},
-		ProviderId:        dto_discovery.Identity("node"),
+		ProviderId:        id.NewIdentity("node").Id,
```
Waldz: just string `"node"`

openvpn/service_discovery/service_poposal_test.go
```diff
@@ -119,7 +120,7 @@ func TestServiceProposalSerialize(t *testing.T) {
...
```
Waldz: Just string `"node"`

service_discovery/dto/service_proposal_test.go
```diff
@@ -27,7 +28,7 @@ func TestServiceProposalSerialize(t *testing.T) {
 		ServiceDefinition: TestServiceDefinition{},
 		PaymentMethodType: "PER_TIME",
 		PaymentMethod:     TestPaymentMethod{},
-		ProviderId:        Identity("node"),
+		ProviderId:        identity.NewIdentity("node").Id,
```
Waldz: Just string `"node"`

tequilapi/endpoints/identities.go
```diff
@@ -26,7 +26,7 @@ func (endpoint *identitiesApi) List(writer http.ResponseWriter, request *http.Re
 	idsSerializable := make([]identityDto, len(idArry))
 	for i, id := range idArry {
 		idsSerializable[i] = identityDto{
-			Id: string(id),
+			Id: string(id.Id),
```
Waldz: Dont need to cast `string` -> `string`

communication/nats_dialog/dialog_establisher.go
```diff
@@ -35,14 +36,14 @@ func (establisher *dialogEstablisher) CreateDialog(contact dto_discovery.Contact
...
-			IdentityId: establisher.myIdentity,
+			IdentityId: establisher.myIdentity.Id,
 		},
 	})
 	if err != nil || response.(*dialogCreateResponse).Reason != 200 {
 		return nil, fmt.Errorf("Dialog creation rejected: %s", err)
 	}
 
-	dialogAddress := nats_discovery.NewAddressNested(contactAddress, string(establisher.myIdentity))
+	dialogAddress := nats_discovery.NewAddressNested(contactAddress, string(establisher.myIdentity.Id))
```
Waldz: Dont need to cast `string` -> `string`

communication/nats_discovery/address.go
```diff
@@ -15,8 +16,8 @@ func NewAddress(server string, port int, topic string) *NatsAddress {
 	}
 }
 
-func NewAddressForIdentity(identity dto_discovery.Identity) *NatsAddress {
-	return NewAddress("127.0.0.1", 4222, string(identity))
+func NewAddressForIdentity(identity id.Identity) *NatsAddress {
+	return NewAddress("127.0.0.1", 4222, string(identity.Id))
```
Waldz: Dont need to cast `string` -> `string`

identity/manager.go
```diff
@@ -41,21 +40,21 @@ func (idm *identityManager) CreateNewIdentity(passphrase string) (identity dto.I
...
 
 	return ids
 }
 
-func (idm *identityManager) GetIdentity(identityString string) (identity dto.Identity, err error) {
+func (idm *identityManager) GetIdentity(identityString string) (identity Identity, err error) {
 	identityString = strings.ToLower(identityString)
 	for _, identity := range idm.GetIdentities() {
-		if strings.ToLower(string(identity)) == identityString {
+		if strings.ToLower(string(identity.Id)) == identityString {
```
Waldz: Dont need to cast `string` -> `string`

identity/node_handler.go
```diff
@@ -50,10 +49,10 @@ func (ih *identityHandler) Create() (id dto.Identity, err error) {
 	return
 }
 
-func (ih *identityHandler) getIdentityFromCache() (identity dto.Identity, err error) {
+func (ih *identityHandler) getIdentityFromCache() (identity Identity, err error) {
 	identity, err = ih.cache.GetIdentity()
 
-	if err != nil || !ih.manager.HasIdentity(string(identity)) {
+	if err != nil || !ih.manager.HasIdentity(string(identity.Id)) {
```
Waldz: Dont need to cast `string` -> `string`

PR #65, opened 2017-12-19
MYST-102 client connection manager

tadovas:
Last portion of migrating client to daemon world and exposing local api


client_connection/manager_test.go
```diff
@@ -0,0 +1,141 @@
...
+		tc.connManager.Connect(identity.FromAddress("identity-1"), "vpn-node-1")
+	}()
+	//wait for go function actually start, to avoid race condition, when we query Status before Connect call even begins.
+	wg.Wait()
+	assert.Equal(tc.T(), ConnectionStatus{NEGOTIATING, "", nil}, tc.connManager.Status())
+	tc.fakeOpenVpn.resumeStart()
+}
+
+func (tc *test_context) TearDownTest() {
+
```
ignasbernotas: Is this needed?

cmd/mysterium_monitor/command_run/command.go
```diff
@@ -43,22 +43,19 @@ func (cmd *CommandRun) Run(options CommandOptions) error {
 		return errors.New("Failed to get original IP: " + err.Error())
 	}
 
-	cmd.clientCommand = command_client.NewCommand(
-		state_client.NewMiddleware(cmd.checkClientIpWhenConnected),
-	)
+	cmd.clientCommand, err = command_client.NewCommand(command_client.CommandOptions{
+		DirectoryRuntime: options.DirectoryRuntime,
+	})
+
```
ignasbernotas: Unneeded newline before error handling

tadovas:
Will fix those two with next feature

cmd/mysterium_client/command_run/command.go
```diff
@@ -5,90 +5,94 @@ import (
...
+		mysteriumClient,
+		identityManager,
+		vpnManager,
+		httpApiServer,
+	}, nil
+}
+
+func (cmd *CommandRun) Run() {
+	router := tequilapi.NewApiRouter()
+	endpoints.RegisterIdentitiesEndpoint(router, cmd.identityManager)
```
Waldz: Should not this already be factories and injected?

cmd/mysterium_client/command_run/command.go
```diff
@@ -5,90 +5,94 @@ import (
...
 	"time"
 )
 
 type CommandRun struct {
+	//TODO this must disappear or become a private field
 	MysteriumClient server.Client
 
-	DialogEstablisherFactory func(identity identity.Identity) communication.DialogEstablisher
-	dialog                   communication.Dialog
+	identityManager identity.IdentityManagerInterface
```
Waldz: Very strange to inject all those variables, just to use them for reinjecting in `Run`

cmd/mysterium_client/command_run/command.go
```diff
@@ -5,90 +5,94 @@ import (
...
 	return cmd.httpApiServer.Wait()
 }
 
 func (cmd *CommandRun) Kill() {
-	cmd.dialog.Close()
-	cmd.vpnClient.Stop()
 	cmd.httpApiServer.Stop()
 }
+
+func configureVpnClientFactory(mysteriumApiClient server.Client, vpnClientRuntimeDirectory string) client_connection.VpnClientFactory {
```
Waldz: Looks like Openvpn specific logic, so I expect it to be next to `vpnConnectionManager`

cmd/mysterium_client/command_run/command.go
```diff
@@ -5,90 +5,94 @@ import (
...
+	return func(vpnSession *session.VpnSession, session *dto.Session) (openvpn.Client, error) {
```
Waldz: - Very strange to inject all `*session.VpnSession` but need just `sessionConfig`
- Very strange to inject all `*dto.Session`, then You need just `sessionId`

Extra tradeoff:
- Then you try to pass by reference, which leads to unsafe `nil` passing

client_connection/manager.go
```diff
@@ -1,55 +1,99 @@
...
+		return err
+	}
+
+	vpnSession, err := vpn_session.RequestSessionCreate(manager.dialog, proposal.Id)
+	if err != nil {
+		manager.status = statusError(err)
+		return err
+	}
+
+	manager.vpnClient, err = manager.vpnClientFactory(vpnSession, &session)
```
Waldz: Strange passing by reference, is it mutated inside?

client_connection/manager.go
```diff
@@ -1,55 +1,99 @@
...
-func (nm *fakeManager) Connect(identity identity.Identity, NodeKey string) error {
+func (manager *connectionManager) Disconnect() error {
+	manager.dialog.Close()
+	manager.vpnClient.Stop()
 	return nil
 }
 
-func (nm *fakeManager) Status() ConnectionStatus {
-	return ConnectionStatus{NOT_CONNECTED, ""}
+func (manager *connectionManager) Wait() error {
```
Waldz: We dont need `Wait()` anymore, correct?

client_connection/interface.go
```diff
@@ -0,0 +1,31 @@
...
+	"errors"
+	"github.com/mysterium/node/identity"
+)
+
+type State string
+
+const (
+	NOT_CONNECTED = State("NOT_CONNECTED")
+	NEGOTIATING   = State("NEGOTIATING")
+	CONNECTED     = State("CONNECTED")
```
donce: Why are we using ALL_CAPS_CASE for constants as in other programming languages? It should be `NotConnected`, `Negotiating`, `Connected` in Go.

client_connection/manager_test.go
```diff
@@ -0,0 +1,141 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/suite"
+	"sync"
+	"testing"
+)
+
+type test_context struct {
+	suite.Suite
+	connManager         *connectionManager
+	fakeDiscoveryClient server.Client
```
Waldz: Specify more exact type, if You know that it is fake

client_connection/interface.go
ignasbernotas: I agree, but this convention is throughout the project:)

client_connection/manager.go
```diff
@@ -1,55 +1,99 @@
...
 }
 
-func NewManager() *fakeManager {
-	return &fakeManager{make(chan error)}
+func (manager *connectionManager) Status() ConnectionStatus {
+	return manager.status
 }
 
-func (nm *fakeManager) Connect(identity identity.Identity, NodeKey string) error {
+func (manager *connectionManager) Disconnect() error {
```
donce: This method never returns an `error` - why do we need it as a return param?

client_connection/manager_test.go
Waldz: Should not You destroy some resources gracefully?
- `connManager.Stop()`
- `fakeOpenVpn.Stop()`

client_connection/manager_test.go
```diff
@@ -0,0 +1,141 @@
...
+
+func (foc *fake_openvpn_client) Wait() error {
+	return nil
+}
+
+func (foc *fake_openvpn_client) Stop() error {
+	return nil
+}
+
+type fake_dialog struct {
```
Waldz: I would add this Fake next to `communication` package, so that others could reuse it.
Or we can do it on next case

cmd/mysterium_client/mysterium_client.go
```diff
@@ -7,19 +7,21 @@ import (
...
 	}
 
-	if err := cmd.Run(options); err != nil {
+	cmd, err := command_run.NewCommand(options)
+	if err != nil {
 		fmt.Fprintln(os.Stderr, err)
 		os.Exit(1)
 	}
 
+	cmd.Run()
```
Waldz: Run without errors?

cmd/mysterium_fake/mysterium_fake.go
```diff
@@ -58,20 +66,13 @@ func runServer(serverCommand *command_server.CommandRun, waiter *sync.WaitGroup)
...
 func runClient(clientCommand *command_client.CommandRun, waiter *sync.WaitGroup) {
-	err := clientCommand.Run(command_client.CommandOptions{
-		NodeKey:          NODE_KEY,
-		DirectoryRuntime: CLIENT_DIRECTORY_RUNTIME,
-	})
-	if err != nil {
-		fmt.Fprintln(os.Stderr, "Client starting error: ", err)
-		os.Exit(1)
-	}
+	clientCommand.Run()
```
Waldz: Drop unused constants

cmd/mysterium_monitor/command_run/command.go
```diff
@@ -43,22 +43,19 @@ func (cmd *CommandRun) Run(options CommandOptions) error {
 		return errors.New("Failed to get original IP: " + err.Error())
 	}
 
-	cmd.clientCommand = command_client.NewCommand(
-		state_client.NewMiddleware(cmd.checkClientIpWhenConnected),
-	)
+	cmd.clientCommand, err = command_client.NewCommand(command_client.CommandOptions{
```
Waldz: Would be good not to disconnect middleware `state_client.NewMiddleware(cmd.checkClientIpWhenConnected)`

openvpn/client.go
```diff
@@ -2,25 +2,31 @@ package openvpn
 
 import "sync"
 
-func NewClient(config *ClientConfig, directoryRuntime string, middlewares ...ManagementMiddleware) *Client {
+type Client interface {
```
Waldz: `interface.go`?

client_connection/interface.go
tadovas: Well we can agree to start using go way constants and this is place would be an example :) Will fix names

client_connection/manager.go
tadovas: No it's not mutated, that's why passing by reference avoids unnecessary copying

client_connection/manager.go
tadovas: I am thinking about adding a check if ( ! status.connected) return NotConnected error

client_connection/manager.go
tadovas: Well if we are shutting connection manager on client exiting, maybe we would like to wait until it stops gracefully

client_connection/manager_test.go
tadovas: Good point

client_connection/manager_test.go
tadovas: In my opinion, reusing these "fake" or "stub" implementations is not very practical, as these implementations are usually test case specific anyway and will differ from component to component testing needs.

cmd/mysterium_client/command_run/command.go
tadovas: So your idea is that we should construct everything in factory method, and leave Run method to start serving tequilapi only (or do any other startup actions needed) ?

cmd/mysterium_client/command_run/command.go
tadovas: As per comment above - this can be moved to creation method

cmd/mysterium_client/command_run/command.go
tadovas: It's a result of breaking a much bigger function and trying to isolate creating vpn client from dialog passed configuration. But a few moments - dto.Session should disappear at all, as vpnSession should contain all config and ids needed to construct vpn client. Secondly our only supported transport carrier is VPN, but this factory would be a good place to create a specific transport connection negotiated between client and node, i.e. WebRTC connection or anything else. VpnSession is bad naming as it's esentially an agreement between client and node - what kind of service will be used, it could be not only VPN.
As there is single place that calls this method - caller guarantees that nils will not be passed. Again - why to make extra copy of those two structures?

cmd/mysterium_client/mysterium_client.go
tadovas: cm.Run() does not return any errors, all errors are catched on command creation.

cmd/mysterium_monitor/command_run/command.go
tadovas: I think this client api checking needs reworking anyway, as it's not practical to pass low level VPN spefic middlewares to client just for testing purposes. Since a lot has changed, how client setups VPN connection. Maybe we should write a simple functionality which setups vpn connection by using tequilapi and then checks client's outgoing ip address?

cmd/mysterium_client/command_run/command.go
tadovas: It can be moved

openvpn/client.go
tadovas: Why not

client_connection/manager_test.go
tadovas: Actually I cannot do that, as server.NewClientFake() returns Client interface already.

cmd/mysterium_client/command_run/command.go
tadovas: Refactored.

cmd/mysterium_fake/mysterium_fake.go
tadovas: All constants are still in use. Even NodeKey is still used by the node. Refactored constant names to camelcase anyway.

client_connection/manager.go
tadovas: Deciced: pass by value

client_connection/manager_test.go
tadovas: Decided: NewClientFake() should return not interface but concrete structure.

client_connection/interface.go
```diff
@@ -0,0 +1,31 @@
...
+type State string
+
+const (
+	NotConnected = State("NotConnected")
+	Connecting   = State("Connecting")
+	Connected    = State("Connected")
+)
+
+var (
+	AlreadyConnectedError = errors.New("already connected")
```
donce: This seems to be unused.

cmd/mysterium_fake/mysterium_fake.go
tadovas: Done.

cmd/mysterium_monitor/command_run/command.go
tadovas: This will be fixed later with mysterium_monitor

openvpn/client.go
tadovas: Good idea but with next

client_connection/interface.go
tadovas: Removed

tadovas:
@Waldz  as per discussion, fixed essential places

PR #69, opened 2017-12-22
MYST-181 Use proposals endpoint instead of create-session

donce:
For client, proposals are fetched from a dedicated endpoint instead `/client-create-session`.

This should only be merged after updating discovery api DEV env from `release/v1`.

### DIFF
--- a/client_connection/manager.go
+++ b/client_connection/manager.go
@@ -1,6 +1,7 @@
 package client_connection
 
 import (
+	"errors"
 	"github.com/mysterium/node/bytescount_client"
 	"github.com/mysterium/node/communication"
 	"github.com/mysterium/node/identity"
@@ -38,15 +39,27 @@ func NewManager(mysteriumClient server.Client, dialogEstablisherFactory DialogEs
 	}
 }
 
-func (manager *connectionManager) Connect(identity identity.Identity, NodeKey string) error {
+func (manager *connectionManager) Connect(identity identity.Identity, nodeKey string) error {
 	manager.status = statusConnecting()
-	clientSession, err := manager.mysteriumClient.SessionCreate(NodeKey)
+
+	proposals, err := manager.mysteriumClient.Proposals(nodeKey)
 	if err != nil {
 		manager.status = statusError(err)
 		return err
 	}
+	if len(proposals) == 0 {
+		err = errors.New("node has no service proposals")
+		manager.status = statusError(err)
+		return err
+	}
+	proposal := proposals[0]
 
-	proposal := clientSession.ServiceProposal
+	// TODO: pass in selected proposal to session creation
+	clientSession, err := manager.mysteriumClient.SessionCreate(nodeKey)
+	if err != nil {
+		manager.status = statusError(err)
+		return err
+	}
 
 	dialogEstablisher := manager.dialogEstablisherFactory(identity)
 	manager.dialog, err = dialogEstablisher.CreateDialog(proposal.ProviderContacts[0])
--- a/client_connection/manager.go
+++ b/client_connection/manager.go
@@ -42,7 +42,7 @@ func NewManager(mysteriumClient server.Client, dialogEstablisherFactory DialogEs
 func (manager *connectionManager) Connect(identity identity.Identity, nodeKey string) error {
 	manager.status = statusConnecting()
 
-	proposals, err := manager.mysteriumClient.Proposals(nodeKey)
+	proposals, err := manager.mysteriumClient.FindProposals(nodeKey)
 	if err != nil {
 		manager.status = statusError(err)
 		return err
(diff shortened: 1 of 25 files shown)


client_connection/manager.go
```diff
@@ -38,15 +39,27 @@ func NewManager(mysteriumClient server.Client, dialogEstablisherFactory DialogEs
...
+	if len(proposals) == 0 {
+		err = errors.New("node has no service proposals")
+		manager.status = statusError(err)
+		return err
+	}
+	proposal := proposals[0]
 
-	proposal := clientSession.ServiceProposal
+	// TODO: pass in selected proposal to session creation
+	clientSession, err := manager.mysteriumClient.SessionCreate(nodeKey)
```
tadovas: This is not needed and should go away. Client gets session id from node with last dialog message (VpnSession at the moment)

server/client_rest.go
```diff
@@ -70,6 +71,27 @@ func (client *clientRest) NodeSendStats(nodeKey string, sessionList []dto.Sessio
 	return nil
 }
 
+func (client *clientRest) Proposals(nodeKey string) (proposals []dto_discovery.ServiceProposal, err error) {
+	response, err := client.doRequest("GET", "proposals", dto.ProposalsRequest{
```
tadovas: Shouldn't it be like "/proposals?byNodeKey=<ourpassedkey>" ? Get cannot have body

server/client_rest.go
```diff
@@ -70,6 +71,27 @@ func (client *clientRest) NodeSendStats(nodeKey string, sessionList []dto.Sessio
...
+		NodeKey: nodeKey,
+	})
+
+	if err == nil {
```
shroomist: Let's do the following way:
```if err != nil {
    return
}
noErrorLogic()```

File: client_connection/manager.go (hunk 1 above)
Comment (tadovas) [inline client_connection/manager.go]: This is not needed and should go away. Client gets session id from node with last dialog message (VpnSession at the moment)

File: server/client_fake.go [hunk 4]
```diff
@@ -38,19 +37,23 @@ func (client *ClientFake) NodeSendStats(nodeKey string, sessionStats []dto.Sessi
...
+	log.Info(MYSTERIUM_API_LOG_PREFIX, "Fake proposals requested for node_key: ", nodeKey)
 	if proposal, ok := client.proposalsByProvider[nodeKey]; ok {
-		session = dto.Session{
-			Id:              nodeKey + "-session",
-			ServiceProposal: proposal,
-		}
-		return
+		proposals = []dto_discovery.ServiceProposal{proposal}
+	} else {
+		proposals = []dto_discovery.ServiceProposal{}
```
shroomist: shouldn't we return an error here instead of new empty proposal?

tadovas: If client by contract returns zero or more proposals then it's ok to return zero propsals, caller will decide if it is error (connection manager) or its normal( /proposals tequilapi endpoint)

server/client_rest.go
donce: Yes, will fix.

server/client_rest.go
donce: This looked strange for me as well, but there are a lot such cases in this file!
This fix this instance, but other instances should be refactored in a separate PR not to explode this one.

client_connection/manager.go
```diff
@@ -38,15 +39,27 @@ func NewManager(mysteriumClient server.Client, dialogEstablisherFactory DialogEs
 	}
 }
 
-func (manager *connectionManager) Connect(identity identity.Identity, NodeKey string) error {
+func (manager *connectionManager) Connect(identity identity.Identity, nodeKey string) error {
 	manager.status = statusConnecting()
-	clientSession, err := manager.mysteriumClient.SessionCreate(NodeKey)
+
+	proposals, err := manager.mysteriumClient.Proposals(nodeKey)
```
Waldz: `Proposals()` -> `FindProposals()`

client_connection/manager.go
Waldz: he can do it in separate request

donce: Done (although separate request would have been better..)

server/client_fake.go
Waldz: he decided to push this error to upper logic level

client_connection/manager.go
donce: Done.

server/client_rest.go
donce: Fixed.

server/client_rest.go
donce: Done.

server/client_fake.go
donce: Yeah - `/v1/proposals` endpoint works this way:
- By default, it returns all proposals.
- It allows filtering by some values. If filtering doesn't return any matches, it doesn't mean that it's an error - it means that there are no objects by given filters.

Waldz: +1

Waldz:
Looks Good To Me

PR #71, opened 2017-12-26
MYST-143 tequila disconnect

shroomist:


tequilapi/endpoints/connection_test.go
```diff
@@ -40,6 +40,29 @@ func (fm *fakeManager) Wait() error {
...
+	assert.Nil(t, err)
+	resp := httptest.NewRecorder()
+
+	connEndpoint.Status(resp, req, httprouter.Params{})
+
+	assert.Equal(t, http.StatusOK, resp.Code)
+	assert.JSONEq(
+		t,
+		`{
+            "status" : "Disconnecting"
```
tadovas: Formatting again :) mixing spaces with tabs perhaps?

client_connection/manager_test.go
```diff
@@ -18,9 +18,10 @@ import (
 
 type test_context struct {
 	suite.Suite
-	connManager         *connectionManager
-	fakeDiscoveryClient *server.ClientFake
-	fakeOpenVpn         *fake_openvpn_client
+	connManager                *connectionManager
+	fakeDiscoveryClient        *server.ClientFake
+	fakeOpenVpn                *fake_openvpn_client
+	fakeDialogResumeDisconnect chan int
```
tadovas: Why can't we move this channel to a field inside fake_dialog struct ?
And define two specials operations on fake_dialog itself ?

client_connection/manager.go
```diff
@@ -100,6 +102,10 @@ func statusNotConnected() ConnectionStatus {
 	return ConnectionStatus{NotConnected, "", nil}
 }
 
+func statusDisconnecting() ConnectionStatus {
+	return ConnectionStatus{Disconnecting, "", nil}
```
donce: Why are we using empty strings and not `nil` in session ids?

Waldz: That's go conventions. But on DTO layer I suggest to have mapping to `null` in JSON value

client_connection/manager.go
```diff
@@ -100,6 +102,10 @@ func statusNotConnected() ConnectionStatus {
 	return ConnectionStatus{NotConnected, "", nil}
 }
 
+func statusDisconnecting() ConnectionStatus {
```
Waldz: Strange to nullify sessions ID, while session still exists

client_connection/manager.go
tadovas: Json encoder can automatically handle empty strings by not rendering fields at all. It's already done by the way.

client_connection/manager.go
tadovas: Does session have any meaning on disconnecting state?

client_connection/manager_test.go
shroomist: because fake_dialog is not accessible from this context and is private to connection manager.

client_connection/manager.go
Waldz: Probably we dont have real case

Waldz:
Looks good to me

PR #72, opened 2017-12-27
MYST-179 Create Docker container with minimal Alpine image

Waldz:


CONTRIBUTING.md
```diff
@@ -32,7 +32,7 @@ Running
 ------------
 ``` bash
 # Start communication gateway
-docker-compose up gateway
+docker-compose up broker
```
ignasbernotas: Are we considering to officially name it broker? :)

bin/server_docker/ubuntu/Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM ubuntu:16.04
+MAINTAINER Valdas Petrulis <petrulis.valdas@gmail.com>
```
ignasbernotas: Personal email intended?:) Just in case you missed it.

.env_example
```diff
@@ -1 +1,2 @@
 MYSTERIUM_API_URL=<url>
+NATS_SERVER_IP=127.0.0.1
```
tadovas: Why do we still need nats server ip at compile time instead of passing it at least on runtime command args?

CONTRIBUTING.md
Waldz: Yes, such recommendation was in the air

.env_example
Waldz: Currently compile args is the only injection engine.
Not sure if passing args is what we want, as we need some defaults.

bin/server_docker/ubuntu/Dockerfile
Waldz: Yes, I dont care

PR #74, opened 2017-12-28
MYST-114 / Message signing and verification

ignasbernotas:


identity/keystore_interface.go
```diff
@@ -5,4 +5,7 @@ import "github.com/ethereum/go-ethereum/accounts"
 type keystoreInterface interface {
 	Accounts() []accounts.Account
 	NewAccount(passphrase string) (accounts.Account, error)
+	Find(a accounts.Account) (accounts.Account, error)
```
Waldz: Could You start using `keystore.Find()` in `manager.GetIdentity()` as-well?

identity/manager.go
```diff
@@ -51,6 +54,37 @@ func (idm *identityManager) GetIdentities() []Identity {
 	return ids
 }
 
+func (idm *identityManager) Unlock(identity Identity, passphrase string) error {
+	keystoreManager := idm.keystoreManager
+	accountExisting := identityToAccount(identity)
+
+	account, err := keystoreManager.Find(accountExisting)
```
Waldz: What if I already have "found" identity which I retrieved with `manager.GetIdentity()`?

identity/manager.go
```diff
@@ -51,6 +54,37 @@ func (idm *identityManager) GetIdentities() []Identity {
...
+	err = keystoreManager.Unlock(account, passphrase)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (idm *identityManager) IsUnlocked(identity Identity) bool {
+	_, err := idm.keystoreManager.SignHash(identityToAccount(identity), messageHash([]byte("1")))
```
Waldz: - Strange that readonly function does write action
- I would skip implementing `IsUnlocked` flag without real usecase

identity/manager.go
ignasbernotas: What if you didn't? :D

identity/signer_test.go
```diff
@@ -0,0 +1,23 @@
...
+	"encoding/hex"
+)
+
+func TestSigning(t *testing.T) {
+	ks := getIdentityTestKeystore()
+	identity := FromAddress("0x1e35193c8cadAA15b43B05ae3D882C91F49BB0Aa")
+	manager := NewIdentityManager(ks)
+	manager.Unlock(identity, "test_passphrase")
+
+	signer := manager.GetSigner(identity)
```
Waldz: You dont need full manager to test only `signer.Sign()`

identity/manager.go
ignasbernotas: I agree, it just seemed like we will need something like this and this will be more friendly than err != keystore.ErrLocked. Plus we won't need to use keystore package externally for these checks:)

identity/signer_test.go
ignasbernotas: Accounts still need to be unlocked. How would you approach this?

identity/keystore_interface.go
ignasbernotas: Yeah, good idea, cheers!

identity/manager.go
Waldz: Seems like we have 2 structures:
- `IdentityAddress` or `IdentityId` - just unique string `0x4e878562376372CCa10397245CBc2950d42246e1`, could point to another node
- `Identity` or `IdentityWallet`- which is my controlled identity from my owned keystore. With Lock/Unlock functions

identity/signer.go
```diff
@@ -0,0 +1,44 @@
...
+	"github.com/ethereum/go-ethereum/crypto"
+	"fmt"
+)
+
+type Signer interface {
+	Sign(message []byte) ([]byte, error)
+}
+
+type keystoreSigner struct {
+	keystoreManager keystoreInterface
```
Waldz: Just noticed. `Signer` reminds very much structure `keystoreWallet` from `go-ethereum/accounts/accounts`

identity/manager.go
Waldz: And if finally we do have `isLocked` flag, we should read it from Ethereum library see `keystoreWallet.Status()`

identity/manager_test.go
```diff
@@ -3,16 +3,16 @@ package identity
 import (
 	"errors"
 	"testing"
-
 	"github.com/ethereum/go-ethereum/accounts"
 	"github.com/stretchr/testify/assert"
+	"github.com/ethereum/go-ethereum/accounts/keystore"
```
Waldz: This unit was using mocked keystore, we should not need Ethereum library and real keystore in filesystem.

Waldz: If do, would be good good to separate such tests to `integration_test.go`

ignasbernotas: We talked about this with @tadovas. It seemed logical to have some sort of checking whether the underlying ethereum keystore works. What if we update the eth keystore package and something breaks? We won't find out about it with mocks.

identity/manager_test.go
```diff
@@ -89,3 +89,66 @@ func Test_HasIdentity(t *testing.T) {
...
+
+func Test_UnlockError(t *testing.T) {
+	ks := getIdentityTestKeystore()
+	testData := struct {
+		Address    string
+		Passphrase string
+	}{
+		"0x53a835143c0eF3bBCBFa796D7EB738CA7dd28f68",
+		"123",
+	}
```
donce: Is it a common practice to define test data struct even in a single case test?

ignasbernotas: Probably unnecessary:)

identity/signer.go
donce: If it satisfies `keystoreInterface` interface (and not `keystoreManagerInterface`), maybe `Manager` in `keystoreManager` variable name is redundant?

identity/verifier.go
```diff
@@ -0,0 +1,27 @@
...
+}
+
+func (ev *ethereumVerifier) Verify(message []byte, signature []byte) bool {
+	hash := messageHash(message)
+	pk, err := crypto.Ecrecover(hash, signature)
+	if err != nil {
+		return false
+	}
+	recoveredAddress := crypto.PubkeyToAddress(*crypto.ToECDSAPub(pk)).Hex()
+	return FromAddress(recoveredAddress) == ev.peerIdentity
```
Waldz: Just notice. Would be possible to drop attribute `peerIdentity`,
by changing interface:
`Verify(message []byte, signature []byte) bool`
->
`GetSigner(message []byte, signature []byte) Identity`

identity/signer.go
```diff
@@ -0,0 +1,44 @@
...
+		return nil, err
+	}
+
+	return signature, nil
+}
+
+// messageHash is a helper function that calculates a hash for the given message that can be
+// safely used to calculate a signature from.
+//
+// The hash is calulcated as
```
donce: *calculated*

identity/signer.go
Waldz: Yeah `keystore` is enough

identity/signer_test.go
```diff
@@ -0,0 +1,23 @@
...
+	sig, err := signer.Sign([]byte("Boop!"))
+	assert.NoError(t, err)
+	assert.Equal(
+		t,
+		hex.EncodeToString(sig),
+		"51e8c02f544c20a5b5b92894ffdd4dad90a71d994cad608cb3157b9ed7757de2758b9c0fc51bfaf079a4d60e81cc83c14cf6900c9bc031ba3f44cb119b82a5f000",
+	)
```
donce: Second argument should be `expected`, and the third `actual,` not in reverse (https://godoc.org/github.com/stretchr/testify/assert#Equal):
```go
func Equal(t TestingT, expected, actual interface{}, msgAndArgs ...interface{}) bool
```

identity/manager_test.go
Waldz: Yes, agree on such integration test:
- 1 integration with happy path
- 10 units with rare edge cases

identity/signer_test.go
Waldz: Usualy tests skip regular factory `New..`.
And loads just dependencies which are tested:
```
keystore := getIdentityTestKeystore()
keystore.Unlock("0x1e35193c8cadAA15b43B05ae3D882C91F49BB0Aa")

signer := keystoreSigner{keystore: keystore}
```

identity/verifier_test.go
```diff
@@ -0,0 +1,42 @@
...
+	identity := toIdentity(privateKey)
+	assert.True(t, NewVerifier(identity).Verify(message, signature))
+}
+
+func TestVerifyFunctionReturnsFalseWhenSignatureIsIncorrect(t *testing.T) {
+	message := []byte("I am message!")
+	privateKey, err := crypto.GenerateKey()
+	assert.NoError(t, err)
+
+	signature, err := crypto.Sign(crypto.Keccak256(message), privateKey)
```
donce: If I'm not mistaken, we could re-use `messageHash` here instead of `crypto.Keccak256`, as in `TestVerifyFunctionReturnsTrueWhenSignatureIsCorrect`. That it would be easier to see the difference between these two tests - only different parts will be different.

identity/signer.go
```diff
@@ -0,0 +1,44 @@
...
+
+func NewSigner(keystore keystoreInterface, identity Identity) Signer {
+	return &keystoreSigner{
+		keystoreManager: keystore,
+		identity:        identity,
+	}
+}
+
+func (ksSigner *keystoreSigner) Sign(message []byte) ([]byte, error) {
+	signature, err := ksSigner.keystoreManager.SignHash(identityToAccount(ksSigner.identity), messageHash(message))
```
Waldz: If I am signing 100 messages, same `identityToAccount` is called several times for variable `self.identity` is is always same

identity/keystore_interface.go
ignasbernotas: should fakeKeystore.Find implement the actual find through fake []Accounts? Otherwise the tests fail, but seems redundant to kind of do it.

identity/manager.go
tadovas: If this IsUnlocked functionality already in wallet then we are good to go.
Use case:
1. Consumer gets list of identities from tequila endoint
2. Consumer needs to decide some how if he needs to unlock identity he wants to use for connection. Calling unlock everytime when he wants to use it - sound a bit overhead? lock-status subresource with locked : true| false would be a nice indication in tequila endpoint

identity/verifier.go
tadovas: I thought about this approach, but then verification from caller side would be a bit clumsy. Also comparing extracted identity with given identity would require a) have a reference to expected identity b) compare it everywhere verification is needed. I think bool return value is cleaner indication that verification succeeded or not

identity/verifier_test.go
tadovas: Missed that. Good point.

identity/manager.go
```diff
@@ -51,6 +54,37 @@ func (idm *identityManager) GetIdentities() []Identity {
...
+	if err != nil {
+		return err
+	}
+
+	err = keystoreManager.Unlock(account, passphrase)
+	if err != nil {
+		return err
+	}
+
+	return nil
```
donce: Instead of these 5 lines of code, we could just return result of `Unlock`:
```go
return keystoreManager.Unlock(account, passphrase)
```
Unless you think it's good to keep `err != nil` check here for consistency?

ignasbernotas: If we decide to get rid of IsLocked method then maybe we could abstract some of the errors with our own? So that packages using this manager wouldn't have to include ethereum's keystore package just for error constants.

identity/manager.go
ignasbernotas: keystoreWallet is private to the keystore package as far as I can see. Have you seen it exposed through some function?

identity/keystore_interface.go
ignasbernotas: Fixed.

identity/signer_test.go
ignasbernotas: Fixed

identity/signer_test.go
ignasbernotas: Fixed

identity/signer.go
ignasbernotas: Yep, but keystoreWallet is private.

ignasbernotas: Fixed now.

identity/signer.go
ignasbernotas: Fixed.

ignasbernotas:
Test message

identity/cache_test.go
```diff
@@ -18,7 +18,7 @@ func TestIdentityCache_StoreIdentity(t *testing.T) {
 	assert.Nil(t, err)
 }
 
-func Test_IdentityCacheGetIdentity(t *testing.T) {
+func TestIdentityCache_GetIdentity(t *testing.T) {
```
ignasbernotas: test line comment

PR #75, opened 2017-12-28
HOTFIX Node creates identity on each run

Waldz:


cmd/mysterium_server/command_run/identity_handler.go
```diff
@@ -0,0 +1,71 @@
...
+	if err != nil || !ih.manager.HasIdentity(identity.Address) {
+		return identity, errors.New("identity not found in cache")
+	}
+
+	return identity, nil
+}
+
+func (ih *identityHandler) launchNew() (id identity.Identity, err error) {
+	// if all fails, create a new one
+	id, err = ih.manager.CreateNewIdentity("")
```
shroomist: Perhaps it's not the best place to hardcode default empty passphrase.
maybe a getter function, to take it from config, context or elsewhere?

cmd/mysterium_server/command_run/factory.go
```diff
@@ -1,34 +1,74 @@
...
+		dialogWaiterFactory: func(identity identity.Identity) (communication.DialogWaiter, dto_discovery.Contact) {
 			address := nats_discovery.NewAddressForIdentity(identity)
 			return nats_dialog.NewDialogWaiter(address), address.GetContact()
 		},
-		SessionManager: &session.Manager{
-			Generator: &session.Generator{},
+		sessionManagerFactory: func(vpnServerIp string) session.ManagerInterface {
+			return openvpn_session.NewManager(openvpn.NewClientConfig(
+				vpnServerIp,
+				options.DirectoryConfig+"/ca.crt",
```
tadovas: Don't hardcode slashes - the code is not portable between OSes. https://golang.org/pkg/path/filepath/#Join is your friend here.

cmd/mysterium_server/command_run/identity_handler.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+
+	if id, err = ih.launchLast(); err == nil {
+		return id, err
+	}
+
+	return ih.launchNew()
+}
+
+func (ih *identityHandler) launchExisting(address string) (id identity.Identity, err error) {
```
ignasbernotas: Launch sounds strange. I'd prefer "use" instead:)

Waldz: Fixed

cmd/mysterium_server/command_run/identity_handler.go
Waldz: Moved to constant `nodeIdentityPassword`

cmd/mysterium_server/command_run/factory.go
Waldz: Fixed in all codebase

Waldz:
Fixes made, @tadovas, please check

PR #76, opened 2017-12-29
myst-177 proposal list

shroomist:


tequilapi/endpoints/proposals.go
```diff
@@ -22,8 +22,8 @@ func NewProposalsEndpoint(mc server.Client) *proposalsEndpoint {
 
 func (pe *proposalsEndpoint) List(resp http.ResponseWriter, req *http.Request, params httprouter.Params) {
 
-	nodeKey := params.ByName("nodeKey")
-	proposals, err := pe.mysteriumClient.FindProposals(nodeKey)
+	nodeId := req.URL.Query().Get("nodeid")
```
donce: `node_id`?

server/client_fake.go
```diff
@@ -39,13 +39,18 @@ func (client *ClientFake) NodeSendStats(nodeKey string, sessionStats []dto.Sessi
...
 func (client *ClientFake) FindProposals(nodeKey string) (proposals []dto_discovery.ServiceProposal, err error) {
 	log.Info(MYSTERIUM_API_LOG_PREFIX, "Fake proposals requested for node_key: ", nodeKey)
-	if proposal, ok := client.proposalsByProvider[nodeKey]; ok {
-		proposals = []dto_discovery.ServiceProposal{proposal}
-	} else {
-		proposals = []dto_discovery.ServiceProposal{}
+
+	for _, proposal := range client.proposalsMock {
+		var filterMatched = true
+		if nodeKey != "" {
```
donce: This line means that if we create a node with empty nodeKey and search for proposals with nodeKey="A", we would get nodes with nodeKey="" as well. Do we really need this line?

tequilapi/endpoints/proposals.go
```diff
@@ -0,0 +1,38 @@
...
+	"net/http"
+)
+
+type proposalsEndpoint struct {
+	mysteriumClient server.Client
+}
+
+type proposalsDto struct {
+	Proposals []dto_discovery.ServiceProposal `json:"proposals"`
+}
```
donce: This struct is duplicate with `server.dto.ProposalsResponse`, maybe we can keep only one of those?

shroomist: your are right, but as of Waldz reqiurement, we want to decouple tequila endpoint structures from internals of app and from discovery.

tequilapi/endpoints/proposals.go
shroomist: perhaps yes

server/client_fake.go
tadovas: What about moving this check outside loop and making like this:
if (nodeKey == "") return client.proposalsMock

tequilapi/endpoints/proposals.go
tadovas: Two different apis should use their own internal serialization/deserialization structures as these are absolutely independent ones - no matter that data is the same at the moment. Agree with @shroomist

tequilapi/endpoints/proposals.go
```diff
@@ -0,0 +1,38 @@
...
+
+func (pe *proposalsEndpoint) List(resp http.ResponseWriter, req *http.Request, params httprouter.Params) {
+
+	nodeId := req.URL.Query().Get("nodeid")
+	proposals, err := pe.mysteriumClient.FindProposals(nodeId)
+	if err != nil {
+		utils.SendError(resp, err, http.StatusInternalServerError)
+		return
+	}
+	proposalsDto := proposalsDto{proposals}
```
tadovas: I don't think it's a good idea to pass json structure as is from discovery service to tequila endpoint response - any change inside discovery service response will automatically break tequila api consumers. What about creating our own ProposalDto and putting only required data? For example node id, proposal number ant some human readable/ decision important info like price etc.

tequilapi/endpoints/proposals.go
```diff
@@ -0,0 +1,38 @@
...
+	Proposals []dto_discovery.ServiceProposal `json:"proposals"`
+}
+
+func NewProposalsEndpoint(mc server.Client) *proposalsEndpoint {
+	return &proposalsEndpoint{mc}
+}
+
+func (pe *proposalsEndpoint) List(resp http.ResponseWriter, req *http.Request, params httprouter.Params) {
+
+	nodeId := req.URL.Query().Get("nodeid")
```
Waldz: Lets call it `provider_id`

tequilapi/endpoints/proposals_test.go
```diff
@@ -0,0 +1,111 @@
...
+	req.URL.RawQuery = query.Encode()
+
+	resp := httptest.NewRecorder()
+	handlerFunc := NewProposalsEndpoint(discoveryApi).List
+	handlerFunc(resp, req, nil)
+
+	assert.JSONEq(
+		t,
+		`{
+            "proposals": [`+string(proposalBytes)+`]
```
Waldz: formatting

tequilapi/endpoints/proposals_test.go
```diff
@@ -0,0 +1,111 @@
...
+        }`,
+		resp.Body.String(),
+	)
+}
+
+func TestProposalsEndpoint_List(t *testing.T) {
+	proposalsSerializable := proposalsDto{proposals}
```
Waldz: Dont use internal dto of tested function.

tequilapi/endpoints/proposals_test.go
```diff
@@ -0,0 +1,111 @@
...
+	)
+	assert.Nil(t, err)
+
+	resp := httptest.NewRecorder()
+	handlerFunc := NewProposalsEndpoint(discoveryApi).List
+	handlerFunc(resp, req, nil)
+
+	assert.JSONEq(
+		t,
+		string(proposalsBytes),
```
Waldz:
```
{
    "proposals": [
        0: serializeProposal(proposals[0]),
        1: serializeProposal(proposals[1])
    ]
}
```

tequilapi/endpoints/proposals.go
donce: Totally agree - I had this idea in mind, just wanted it to be clarified :D

server/client_fake.go
Waldz: I did not wanted conditional code with&without filter

tequilapi/endpoints/proposals_test.go
Waldz: Or jsut freeze wanted static string of tested endpoint

server/client_fake.go
tadovas: And now you have conditional code with and without filter inside a loop

tequilapi/endpoints/proposals_test.go
shroomist: fixed

tequilapi/endpoints/proposals_test.go
shroomist: fixed

tequilapi/endpoints/proposals.go
shroomist: done

server/client_fake.go
shroomist: @tadovas this is client mock, do you think its important to optimize here?

tadovas: Well it's not about optimization, more about readability. And yes you are right - since it's just for testing - no point to do some optimizations here, we can focus more on readability.

tequilapi/endpoints/proposals_test.go
```diff
@@ -0,0 +1,135 @@
...
+                        }
+                    }
+                }
+            ]
+        }`,
+        resp.Body.String(),
+    )
+}
+
+func TestProposalsEndpoint_List(t *testing.T) {
```
ignasbernotas: Why do we need an underscore in the name? I think we should stick to camelCase convention.

server/client_fake.go
donce: Plus, it's extendable - if we add more filters, it will be easy to extend this method by adding more ifs inside loop :)

server/client_fake.go
```diff
@@ -39,13 +39,18 @@ func (client *ClientFake) NodeSendStats(nodeKey string, sessionStats []dto.Sessi
 
 func (client *ClientFake) FindProposals(nodeKey string) (proposals []dto_discovery.ServiceProposal, err error) {
 	log.Info(MYSTERIUM_API_LOG_PREFIX, "Fake proposals requested for node_key: ", nodeKey)
-	if proposal, ok := client.proposalsByProvider[nodeKey]; ok {
-		proposals = []dto_discovery.ServiceProposal{proposal}
+
+	if nodeKey == "" {
```
Waldz: Conditionals are evil, because they work differently

tequilapi/endpoints/proposals.go
```diff
@@ -0,0 +1,74 @@
...
+    dto_discovery "github.com/mysterium/node/service_discovery/dto"
+    "github.com/mysterium/node/tequilapi/utils"
+    "net/http"
+)
+
+type proposalsDto struct {
+    Proposals []proposalTeqDto `json:"proposals"`
+}
+
+type locationTeqDto struct {
```
Waldz: Dont split so much tequila in all places

tequilapi/endpoints/proposals.go
```diff
@@ -0,0 +1,74 @@
...
+    Id                int                     `json:"id"`
+    ProviderId        string                  `json:"providerId"`
+    ServiceDefinition serviceDefinitionTeqDto `json:"serviceDefinition"`
+}
+
+func proposalToDto(p dto_discovery.ServiceProposal) proposalTeqDto {
+    return proposalTeqDto{
+        Id:         p.Id,
+        ProviderId: p.ProviderId,
+        ServiceDefinition: serviceDefinitionTeqDto{
```
Waldz: Add `ServiceType` also

tequilapi/endpoints/proposals.go
```diff
@@ -0,0 +1,74 @@
...
+            LocationOriginate: locationTeqDto{
+                ASN: p.ServiceDefinition.GetLocation().ASN,
```
Waldz: Where is country and city?

tequilapi/endpoints/proposals_test.go
```diff
@@ -0,0 +1,135 @@
...
+type TestPaymentMethod struct{}
+
+func (method TestPaymentMethod) GetPrice() money.Money {
+	return money.Money{}
+}
+
+var proposals = []dto_discovery.ServiceProposal{
+	{
+		Id:                1,
+		Format:            "service-proposal/v1",
```
Waldz: Remove fields, unnecessary for your tests: `Format`, `PaymentMethod`, `PaymentMethodType`, `ProviderContacts`

tequilapi/endpoints/proposals.go
```diff
@@ -0,0 +1,74 @@
...
+    mysteriumClient server.Client
+}
+
+func NewProposalsEndpoint(mc server.Client) *proposalsEndpoint {
+    return &proposalsEndpoint{mc}
+}
+
+func (pe *proposalsEndpoint) List(resp http.ResponseWriter, req *http.Request, params httprouter.Params) {
+
+    nodeId := req.URL.Query().Get("provider_id")
```
Waldz: Your started using camelCase naming in Tequilapi, it's OK, just be consistent

Waldz: It's not nodeId anymore ;)

shroomist:
@Waldz your comments fixed, can you guys approve this once again? @tadovas

PR #77, opened 2017-12-29
MYST-182 Discovery. Use new stats endpoint

donce:
Should be merged only after https://github.com/MysteriumNetwork/api/pull/11.

### DIFF
--- a/server/client_rest.go
+++ b/server/client_rest.go
@@ -95,7 +95,7 @@ func (client *clientRest) FindProposals(nodeKey string) (proposals []dto_discove
 	return
 }
 
-func (client *clientRest) CreateSessionStats(sessionId string, sessionStats dto.SessionStats) (err error) {
+func (client *clientRest) SendSessionStats(sessionId string, sessionStats dto.SessionStats) (err error) {
 	path := fmt.Sprintf("sessions/%s/stats", sessionId)
 	response, err := client.doPostRequest(path, sessionStats)
 	if err == nil {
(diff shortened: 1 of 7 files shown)


server/client_rest.go
```diff
@@ -95,8 +95,9 @@ func (client *clientRest) FindProposals(nodeKey string) (proposals []dto_discove
 	return
 }
 
-func (client *clientRest) SessionSendStats(sessionId string, sessionStats dto.SessionStats) (err error) {
-	response, err := client.doPostRequest("client_send_stats", sessionStats)
+func (client *clientRest) CreateSessionStats(sessionId string, sessionStats dto.SessionStats) (err error) {
+	path := fmt.Sprintf("sessions/%s/stats", sessionId)
+	response, err := client.doPostRequest(path, sessionStats)
```
shroomist: function name misleading - this doesn't create but does send the stats

tadovas: Agreed.

shroomist:
Please use conventional PR Titles.
MYST-182 Discovery. Use new stats endpoint

server/client_rest.go
donce: Done! 🍭

server/dto/node_stats_request.go
```diff
@@ -1,6 +1,6 @@
 package dto
 
 type NodeStatsRequest struct {
-	NodeKey  string         `json:"node_key"`
-	Sessions []SessionStats `json:"sessions"`
```
Waldz: You can remove `NodeStatsRequest.Sessions` fields at all

donce: It's currently being used and is required by api. It could be removed from everywhere, since it's empty array at the moment, but it will require changing api as well, and might be worth to do this in https://mysteriumnetwork.atlassian.net/browse/MYST-92

Issue #42, opened 2017-10-24
GoLint?


jiminoc:
Whats the project's stance on golint and comments? 

So far from a casual glance I'm not seeing any comments or golint compliant source files. These things are much easier to add up front than backfilling and opens up the source to more devs who can follow along in the codebase. 

thx

goktugyil:
Any updates?

Waldz:
Kudos to @jiminoc for pointing out ;)

Waldz:
@goktugyil we setuped helper command `bin/lint`.
Project still has lots of linter errors, and we are actively clearing them package by package.

PR #84, opened 2018-01-04
MYST-163 Add `golint` package

donce:
`./bin/lint` prints linter warnings.
  
Solves https://github.com/MysteriumNetwork/node/issues/42.

### DIFF
--- a/bin/lint
+++ b/bin/lint
@@ -1,0 +1,1 @@
+go list ./... | xargs go run vendor/github.com/golang/lint/golint/*.go
--- a/glide.yaml
+++ b/glide.yaml
@@ -19,3 +19,6 @@ testImport:
   subpackages:
   - assert
   - suite
+- package: github.com/golang/lint
+  subpackages:
+  - golint
--- a/glide.lock
+++ b/glide.lock
@@ -1,5 +1,5 @@
-hash: 84070bc56a0d892ef493515c34a6f29ef647d73755911a975f5c0b41d305e11b
-updated: 2017-12-14T21:45:17.235275+02:00
+hash: fcb0014fc6c0c66c464c53652f7f1479735043de32692a98f0f19aa568f97669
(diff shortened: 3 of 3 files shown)


glide.yaml
```diff
@@ -14,6 +14,9 @@ import:
   version: ^1.1.0
 - package: github.com/mitchellh/go-homedir
   version: b8bc1bf767474819792c23f32d8286a45736f1c6
+- package: github.com/golang/lint
```
Waldz: Should not it be under `testImport` - just for DEV environment

Waldz: I tried to run `go run vendor/github.com/golang/lint/lint.go` and it fails, because child dependencies are not installed

bin/lint
```diff
@@ -0,0 +1 @@
+go list ./... | xargs golint
```
Waldz: `go run vendor/github.com/golang/lint/lint.go`
1. Would it be more efficient to call `lint` only once, instead per-package
2. Seems like we check style on `vendor` also, see how `bin/test` is made

Waldz: 3. Would be nice to actually use installed dependency in helper `bin/lint`
e.g `go run vendor/github.com/golang/lint/lint.go`
What google says about best practices about that?

donce: >3. Would be nice to actually use installed dependency in helper bin/lint
e.g go run vendor/github.com/golang/lint/golint/*.go nat

Good point, updated.

glide.yaml
donce: Yes, fixed.

donce:
PR is ready to be reviewed 💰

shroomist:
perhaps we shouldn't remove unused packages and bump versions in yaml.lock with this ticket.

Waldz:
+100 We have a Linter!
Lets start clearing errors ;)

PR #85, opened 2018-01-05
Update glide.lock

donce:


Waldz:
Did not You recalculate all dependencies in PR #84?

PR #86, opened 2018-01-08
Feature/myst 27 broker on public domain

zolia:


communication/nats_discovery/contact.go
```diff
@@ -5,4 +5,6 @@ const CONTACT_NATS_V1 = "nats/v1"
 type ContactNATSV1 struct {
 	// Topic on which client is getting message
 	Topic string
+	// NATS servers used by node and should be contacted via
+	BrokerAddresses []string
```
Waldz: 1. Add JSON annotations
2 Freeze JSON serialisation with tests

communication/nats_discovery/address_test.go
```diff
@@ -40,6 +40,7 @@ func TestNewAddressForContact(t *testing.T) {
 		Type: "nats/v1",
 		Definition: ContactNATSV1{
 			Topic: "123456",
+			BrokerAddresses: []string{"nats://" + natsServerIp + ":4222"},
```
Waldz: Dont depend on live variables.
Just pass mocked variable like `"nats://far-server:1234"` and assert that it was parsed correctly

communication/nats_discovery/address.go
```diff
@@ -19,6 +19,13 @@ func NewAddress(server string, port int, topic string) *NatsAddress {
 	}
 }
 
+func NewBrokerAddresses(contact ContactNATSV1) *NatsAddress {
```
Waldz: We already had such factory `NewAddressForContact`, lets improve same function.
And lets improve `NewAddress(server, port, topic)` -> `NewAddress(topic, address...)`

communication/nats_discovery/address.go
```diff
@@ -19,6 +19,13 @@ func NewAddress(server string, port int, topic string) *NatsAddress {
...
+	return &NatsAddress{
+		servers: contact.BrokerAddresses,
+		topic: contact.Topic,
+	}
+}
+
 func NewAddressForIdentity(identity identity.Identity) *NatsAddress {
```
Waldz: Lets rename this to `NewAddressGenerate()` - it announces node's address

communication/nats_dialog/dialog_establisher.go
```diff
@@ -2,7 +2,7 @@ package nats_dialog
 
 import (
 	"fmt"
-	"github.com/mgutz/logxi/v1"
```
donce: Why did we have `logxi` in the first place?
Maybe it makes sense to remove it from `glide.lock? It will safe us in the future from using logger which is not working.

nat/service_iptables.go
```diff
@@ -102,6 +100,6 @@ func (service *serviceIpTables) disableRules() error {
 	return nil
 }
 
-func (service *serviceIpTables) clearStaleRules() error {
-	return service.disableRules()
+func (service *serviceIpTables) clearStaleRules() {
+	service.disableRules()
```
donce: If deletion of forwarding rules fails, maybe it's worth at least to add some warning log? I believe that this failure of deletion can mean multiple things, and ignoring that might be dangerous.

PR #88, opened 2018-01-08
Feature/myst-114-sign-mysterium-apimessages

tadovas:


server/json_client.go
```diff
@@ -0,0 +1,105 @@
...
+const (
+	mysteriumAgentName       = "goclient-v0.1"
+	authenticationHeaderName = "Authorization"
+	authenticationSchemaName = "Signature"
+)
+
+var noSignature = identity.SignatureBytes(nil)
+
+type HttpClient interface {
+	DoGet(path string, values url.Values) (*http.Response, error)
```
ignasbernotas: Is it worth prefixing the methods with "Do"?

server/mysterium_api.go
```diff
@@ -0,0 +1,122 @@
+package server
```
ignasbernotas: I know this is probably the best naming we can come up with, but maybe worth naming it discovery_api? mysterium_api is pretty generic:)

server/json_client.go
tadovas: Well obj.Get usually says get something from object. I see a better intent expression here. as GET POST etc. are actions and DoAction provides better readability. Ofcourse if it's confusing it can be renamed

server/mysterium_api.go
tadovas: Well from node/client perspective this api provides 3 distinct functionalities:
1. Register identities
2. Register/provide proposals (actual discovery)
3. Collect statistics
We call remote service "discovery service" as we don't have a better naming yet. But from client/node perspective - its not discovery but something more generic - i.e. mysterium_centralized_api

server/json_client.go
donce: Agree with @ignasbernotas - `Get` is a verb itself and since this is not Java, getters like `GetAmounts` shouldn't be very common to confuse with.

identity/signature.go
```diff
@@ -19,10 +23,14 @@ func (signature *Signature) Hex() string {
...
+func (signature *Signature) Base64() string {
+	return base64.StdEncoding.EncodeToString(signature.Bytes())
 }
 
 func (signature *Signature) Bytes() []byte {
 	return signature.raw
 }
+
+func (signature Signature) EqualsTo(other Signature) bool {
+	return bytes.Equal(signature.raw, other.raw)
```
donce: Is it safe to ignore cases when `signature` or `other` is `nil`?

server/mysterium_api.go
```diff
@@ -0,0 +1,122 @@
...
+		NodeKey: nodeKey,
+		// TODO Refactor Node statistics with new `SessionStats` DTO
+		Sessions: []dto.SessionStats{},
+	})
+	if err == nil {
+		defer response.Body.Close()
+		log.Info(mysteriumApiLogPrefix, "Node stats sent: ", nodeKey)
+	}
+
+	return nil
```
donce: Shouldn't this be `return`, instead `return nil`?
It seems that we are always returning no error, even when error happens.

identity/signature.go
tadovas: Since receiver and other are values but not pointers they cannot be nil. If you try to pass nil here you will get compiler error

server/json_client.go
tadovas: Will rename it to Get Post and SignedPost

server/mysterium_api.go
```diff
@@ -0,0 +1,122 @@
...
+
+var mysteriumApiUrl string
+
+type mysteriumApiClient struct {
+	http HttpClient
+}
+
+func NewClient() Client {
+	return &mysteriumApiClient{
+		http: NewJsonClient(
```
Waldz: Looks quite complicated `http`, `transport` and `httpTransport`

server/json_client.go
```diff
@@ -0,0 +1,105 @@
...
+
+const (
+	mysteriumAgentName       = "goclient-v0.1"
+	authenticationHeaderName = "Authorization"
+	authenticationSchemaName = "Signature"
+)
+
+var noSignature = identity.SignatureBytes(nil)
+
+type HttpClient interface {
```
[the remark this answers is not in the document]
Waldz: It could be event simpler interface: `httpClient.Do(request)` which is compatible with `http.Client`

server/json_client.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+}
+
+type jsonHttpClient struct {
+	baseApiUrl string
+	transport  HttpTransport
+}
+
+func (jhc *jsonHttpClient) DoGet(path string, values url.Values) (*http.Response, error) {
+	pathWithQuery := fmt.Sprintf("%v?%v", path, values.Encode())
```
Waldz: Looks like not building "/path", then dont have query params

server/mysterium_api.go
Waldz: Yeah, I dont se case then error is returned

server/mysterium_api.go
```diff
@@ -0,0 +1,122 @@
...
+	response, err := client.http.DoSignedPost(path, sessionStats, signer)
+	if err == nil {
+		defer response.Body.Close()
+		log.Info(mysteriumApiLogPrefix, "Session stats sent: ", sessionId)
+	}
+
+	return nil
+}
+
+func parseResponseJson(response *http.Response, dto interface{}) error {
```
Waldz: Marshar|Unmarshar now come in different classes, should not they come together?

server/mysterium_api.go
donce: Same applies for some of other endpoints.
This isn't related to this PR, so adding a TODO or Jira ticket seems enough.

client_connection/manager.go
```diff
@@ -111,8 +113,8 @@ func statusDisconnecting() ConnectionStatus {
 	return ConnectionStatus{Disconnecting, "", nil}
 }
 
-func ConfigureVpnClientFactory(mysteriumApiClient server.Client, vpnClientRuntimeDirectory string) VpnClientFactory {
-	return func(vpnSession session.SessionDto) (openvpn.Client, error) {
+func ConfigureVpnClientFactory(mysteriumApiClient server.Client, vpnClientRuntimeDirectory string, signerFactory SignerFactory) VpnClientFactory {
+	return func(vpnSession session.SessionDto, identity identity.Identity) (openvpn.Client, error) {
```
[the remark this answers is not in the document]
tadovas: I also thought about first idea, but then it became that connection manager only creates a signer and passes it to vpn client factory, also signer is too low level and communication specific (sign outogoing requests). So i decided that it has to be open vpn connection factory dependency on configuration phase.
As for option two - middlewares are very openvpn specific and should stay that way - inside openvpnfactory and on openvpn client construction, too much low level info moved outside.

server/json_client.go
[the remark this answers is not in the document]
tadovas: native httpClient.Post get etc hides constructed request objects so there is no way insert custom headers, access bodies etc. httpClient.Do is too low level - operates on req created elsewere. The point of this interface was to provide more flexible http actions to uper level component (mysterium_api_client) and to have reusable bit of functionality - http client with request bodies serialized to json and auto signed if needed. It could disappear after we fully migrate mysterium api to signed requests.

server/json_client.go
tadovas: Please explain. The idea is path or endpoint concated it with params. I.e. find proposals:
client.DoGet("proposals" , {map with params to find proposal with node id})

server/json_client.go
```diff
@@ -0,0 +1,105 @@
...
+func (jhc *jsonHttpClient) DoSignedPost(path string, body interface{}, signer identity.Signer) (*http.Response, error) {
+	return jhc.doSignedPayloadRequest(http.MethodPost, path, body, signer)
+}
+
+func (jhc *jsonHttpClient) doSignedPayloadRequest(method string, path string, body interface{}, signer identity.Signer) (*http.Response, error) {
+	payloadJson, err := json.Marshal(body)
+	if err != nil {
+		return nil, err
+	}
+	signature, err := signer.Sign(payloadJson)
```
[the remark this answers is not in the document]
tadovas: we cannot have a signedHttpClient as per our discussion - in that case signer will become a state and new httpClient will be needed to construct for each signer

server/json_client.go
```diff
@@ -0,0 +1,105 @@
...
+	fullPath := fmt.Sprintf("%v/%v", jhc.baseApiUrl, path)
+	req, err := http.NewRequest(method, fullPath, bytes.NewBuffer(body))
+	if err != nil {
+		return nil, err
+	}
+	req.Header.Set("User-Agent", mysteriumAgentName)
+	req.Header.Set("Content-Type", "application/json")
+	req.Header.Set("Accept", "application/json")
+
+	if !signature.EqualsTo(noSignature) {
```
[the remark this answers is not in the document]
tadovas: In this case it's not - currently signature is optional for posts and exists only if signer returns real signature instead of no signature constant. It will disappear after all posts become signed

server/mysterium_api.go
tadovas: It must unless jsonClient can get a respose struct factory somehow to know which types of concrete structures to unmarshal and return as response.

server/mysterium_api.go
tadovas: A better naming might help. Will fix that.

bytescount_client/stats_sender.go
```diff
@@ -1,19 +1,23 @@
...
-		})
-	})
+	return func(bytesSent, bytesReceived int) error {
+		return mysteriumClient.SendSessionStats(
+			sessionIdString,
+			dto.SessionStats{
+				BytesSent:     bytesSent,
+				BytesReceived: bytesReceived,
+			},
+			signer)
```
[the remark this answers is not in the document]
tadovas: Fixed

server/mysterium_api.go
tadovas: Will fix this just as example for other endpoints.

server/mysterium_api.go
tadovas: it became mysteriumApi {
     json : NewJsonClient()
}

server/json_client.go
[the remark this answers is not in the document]
Waldz: Signing actually has much simpler flow:
```
request := client.createPOSTRequest("v1/node_register", requestDto{})
client.signRequest(&request, signer)

response, err := client.httpClient.Do(request)

err = parseResponseJson(response, &dto.ProposalsResponse{})

```

I think most complication comes, because we mixed *request creating* + *request transporting*:
- `doGetRequest()`
- `doPostRequest()`
versus:
- `createGetRequest()`
- `signRequest(request)`
- `doRequest(request)` wich already exist in `http.Client.Do()`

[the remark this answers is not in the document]
tadovas: In that case JsonClient becomes requestFactory?

tadovas:
@Waldz Refactored mysterium api client internals (requests are created independently)

server/mysterium_api.go
```diff
@@ -17,105 +15,111 @@ const (
...
-	response, err := mApi.json.Post("node_register", dto.NodeRegisterRequest{
+func (mApi *mysteriumApi) NodeRegister(proposal dto_discovery.ServiceProposal) error {
+	req, err := newPostRequest("node_register", dto.NodeRegisterRequest{
 		ServiceProposal: proposal,
 	})
+	if err != nil {
+		return err
+	}
 
+	resp, err := mApi.http.Do(req)
```
Waldz: This block could be DRYied:
```
resp, err := mApi.http.Do(req)
if err == nil {
    defer resp.Body.Close()
}
```
Similar places was refactored in `ipify.doRequest(reqest, responseDto)`

server/request.go
```diff
@@ -0,0 +1,70 @@
...
+func newPostRequest(path string, requestBody interface{}) (*http.Request, error) {
+	bodyBytes, err := encodeToJson(requestBody)
+	if err != nil {
+		return nil, err
+	}
+	return newRequest(http.MethodPost, path, bodyBytes)
+}
+
+func newSignedPostRequest(path string, requestBody interface{}, signer identity.Signer) (*http.Request, error) {
+	bodyBytes, err := encodeToJson(requestBody)
```
Waldz: I would just reuse `newPostRequest()` + add header

client_connection/manager.go
```diff
@@ -14,7 +14,9 @@ import (
 
 type DialogEstablisherFactory func(identity identity.Identity) communication.DialogEstablisher
 
-type VpnClientFactory func(vpnSession session.SessionDto) (openvpn.Client, error)
+type VpnClientFactory func(vpnSession session.SessionDto, identity identity.Identity) (openvpn.Client, error)
+
+type SignerFactory func(identity identity.Identity) identity.Signer
```
Waldz: `identity` is package name, lets not use it for variables.
Use `id`, `providerId`, `customerId`, `myId`, `signerId`

client_connection/manager.go
```diff
@@ -66,7 +68,7 @@ func (manager *connectionManager) Connect(identity identity.Identity, nodeKey st
 		return err
 	}
 
-	manager.vpnClient, err = manager.vpnClientFactory(*vpnSession)
+	manager.vpnClient, err = manager.vpnClientFactory(*vpnSession, identity)
```
Waldz: `identity` is package name

client_connection/manager.go
```diff
@@ -121,7 +123,7 @@ func ConfigureVpnClientFactory(mysteriumApiClient server.Client, vpnClientRuntim
 			return nil, err
 		}
 
-		statsSender := bytescount_client.NewSessionStatsSender(mysteriumApiClient, vpnSession.Id)
+		statsSender := bytescount_client.NewSessionStatsSender(mysteriumApiClient, vpnSession.Id, signerFactory(identity))
```
Waldz: `identity` is package name

cmd/mysterium_client/command_run/command.go
```diff
@@ -36,7 +36,11 @@ func NewCommandWith(
 		return nats_dialog.NewDialogEstablisher(identity)
 	}
 
-	vpnClientFactory := client_connection.ConfigureVpnClientFactory(mysteriumClient, options.DirectoryRuntime)
+	signerFactory := func(id identity.Identity) identity.Signer {
```
Waldz: +1

server/json_client.go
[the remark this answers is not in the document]
Waldz: Anyway this function does not single responseibility, it does: requestCreation + requestAuthentification + requestTransport.
I like new refactoring better

server/json_client.go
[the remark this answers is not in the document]
Waldz: I did not get idea of `JsonClient` much. I am just advocating for function with single responsibility

server/json_client.go
Waldz: I am talking about case, then query params are empty.
`DoGet('v1/path', url.Values{})`
IMHO should be asserted to become:
`/path`

client_connection/manager.go
tadovas: Will fix that.

server/mysterium_api.go
tadovas: I will check it

server/request.go
tadovas: That's the problem - newPostRequest serializes body inside and inserts into request - no clean way to access body bytes later and create signature.

tadovas:
@Waldz all your mentiond issues addressed :)

PR #89, opened 2018-01-09
MYST-224 Refactor `identityHandler`, add tests

donce:
I'm going to work around `identityHandler` for `MYST-224`, so it should have at least basic tests.

### DIFF
--- a/cmd/mysterium_server/command_run/factory.go
+++ b/cmd/mysterium_server/command_run/factory.go
@@ -75,15 +75,3 @@ func NewCommandWith(
 		},
 	}
 }
-
-func SelectIdentity(identityHandler *identityHandler, keyOption string) (id identity.Identity, err error) {
-	if len(keyOption) > 0 {
-		return identityHandler.UseExisting(keyOption)
-	}
-
-	if id, err = identityHandler.UseLast(); err == nil {
-		return id, err
-	}
-
-	return identityHandler.UseNew()
-}
--- a/cmd/mysterium_server/command_run/factory.go
+++ b/cmd/mysterium_server/command_run/factory.go
@@ -33,15 +33,16 @@ func NewCommandWith(
 ) *CommandRun {
 
 	keystoreInstance := keystore.NewKeyStore(options.DirectoryKeystore, keystore.StandardScryptN, keystore.StandardScryptP)
+	cache := identity.NewIdentityCache(options.DirectoryKeystore, "remember.json")
 	identityHandler := NewNodeIdentityHandler(
 		identity.NewIdentityManager(keystoreInstance),
 		mysteriumClient,
-		options.DirectoryKeystore,
+		cache,
 	)
 
 	return &CommandRun{
 		identitySelector: func() (identity.Identity, error) {
-			return identityHandler.Select(options.NodeKey)
+			return SelectIdentity(identityHandler, options.NodeKey)
 		},
 		ipifyClient:     ipifyClient,
 		mysteriumClient: mysteriumClient,
@@ -73,4 +74,16 @@ func NewCommandWith(
 			return openvpn.NewServer(vpnServerConfig, options.DirectoryRuntime)
 		},
 	}
+}
+
+func SelectIdentity(identityHandler *identityHandler, keyOption string) (id identity.Identity, err error) {
+	if len(keyOption) > 0 {
+		return identityHandler.UseExisting(keyOption)
+	}
+
+	if id, err = identityHandler.UseLast(); err == nil {
+		return id, err
+	}
+
+	return identityHandler.UseNew()
 }
--- a/cmd/mysterium_server/command_run/identity_handler_test.go
+++ b/cmd/mysterium_server/command_run/identity_handler_test.go
@@ -18,6 +18,7 @@ func Test_identityHandler_UseExisting(t *testing.T) {
 	id, err := handler.UseExisting("address")
 	assert.Equal(t, identityManager.FakeIdentity1, id)
 	assert.Nil(t, err)
+	assert.Equal(t, "", client.RegisteredIdentity.Address)
 }
 
 func Test_identityHandler_UseLast(t *testing.T) {
@@ -33,6 +34,7 @@ func Test_identityHandler_UseLast(t *testing.T) {
 	id, err := handler.UseLast()
 	assert.Equal(t, fakeIdentity, id)
 	assert.Nil(t, err)
+	assert.Equal(t, "", client.RegisteredIdentity.Address)
 }
 
 func Test_identityHandler_UseNew(t *testing.T) {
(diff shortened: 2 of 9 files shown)


cmd/mysterium_server/command_run/factory.go
```diff
@@ -73,3 +75,15 @@ func NewCommandWith(
 		},
 	}
 }
+
+func SelectIdentity(identityHandler *identityHandler, keyOption string) (id identity.Identity, err error) {
```
Waldz: Idea of handler was, to have thiner code in command.
This why, I moved it to `identity_handler.go` next to `command.go`.

cmd/mysterium_server/command_run/identity_handler_test.go
```diff
@@ -0,0 +1,48 @@
...
+
+import (
+	"testing"
+
+	"github.com/mysterium/node/identity"
+	"github.com/mysterium/node/server"
+	"github.com/stretchr/testify/assert"
+)
+
+func Test_identityHandler_UseExisting(t *testing.T) {
```
Waldz: +100

cmd/mysterium_server/command_run/identity_handler_test.go
```diff
@@ -0,0 +1,48 @@
...
+	identityManager := identity.NewIdentityManagerFake()
+	client := server.NewClientFake()
+	cache := identity.NewIdentityCacheFake()
+
+	handler := NewNodeIdentityHandler(identityManager, client, cache)
+
+	id, err := handler.UseExisting("address")
```
Waldz: But such value "address" is not mocked in `identityManager`. is not it?

cmd/mysterium_server/command_run/identity_handler_test.go
```diff
@@ -0,0 +1,48 @@
...
+
+func Test_identityHandler_UseNew(t *testing.T) {
+	identityManager := identity.NewIdentityManagerFake()
+	client := server.NewClientFake()
+	cache := identity.NewIdentityCacheFake()
+
+	handler := NewNodeIdentityHandler(identityManager, client, cache)
+
+	id, err := handler.UseNew()
+	assert.Equal(t, identityManager.FakeIdentity2, id)
```
Waldz: You should assert if it was registered via `client`

identity/cache.go
```diff
@@ -15,23 +16,24 @@ type IdentityCache struct {
...
 }
 
 func (ic *IdentityCache) GetIdentity() (identity Identity, err error) {
-	if ic.cacheExists() {
-		cache, err := ic.readCache()
-		if err != nil {
-			return identity, err
-		}
+	if !ic.cacheExists() {
+		err = errors.New("cache file does not exist")
```
Waldz: It's not an error if file is missing. It's same as "identity non found in cache"

cmd/mysterium_server/command_run/factory.go
donce: well, `SelectIdentity` is independent function - maybe it's worth to move it to new file to make less code in this file?

cmd/mysterium_server/command_run/identity_handler_test.go
donce: This address is ignored in fake identity manager:
```go
func (fakeIdm *idmFake) GetIdentity(_ string) (Identity, error) {
	return fakeIdm.FakeIdentity1, nil
}
```

identity/cache.go
donce: 1. This method is named `GetIdentity` and it seems a common sense to get an identity or an error.

2. Looking in `identity_handler.go`:
```go
	identity, err = ih.cache.GetIdentity()
	if err != nil || !ih.manager.HasIdentity(identity.Address) {
		return identity, errors.New("identity not found in cache")
	}
```
If cache file does not exist, this results in "identity not found in cache" error either way - with or without this change.
What this change is about is not only making method more predictable, but also avoiding calling
`ih.manager.HasIdentity("")` for empty identity, which is returned when cache does not exist. If we know that cache file does not exist, we should not check for an empty identity in identity manager.

cmd/mysterium_server/command_run/identity_handler_test.go
donce: Done.

cmd/mysterium_server/command_run/identity_handler_test.go
Waldz: But how can we use *existing* "address" if it dont exist actually. Could could double check, plz

cmd/mysterium_server/command_run/factory.go
Waldz: Clear, decoupled from from handler struct.
This separate function could still be in `node_handler.go`

cmd/mysterium_server/command_run/identity_handler_test.go
```diff
@@ -0,0 +1,49 @@
...
+	client := server.NewClientFake()
+	cache := identity.NewIdentityCacheFake()
+
+	fakeIdentity := identity.FromAddress("abc")
+	cache.StoreIdentity(fakeIdentity)
+
+	handler := NewNodeIdentityHandler(identityManager, client, cache)
+
+	id, err := handler.UseLast()
+	assert.Equal(t, fakeIdentity, id)
```
Waldz: Could your assert that no identity was registered?

cmd/mysterium_server/command_run/identity_handler_test.go
```diff
@@ -0,0 +1,49 @@
...
+
+func Test_identityHandler_UseExisting(t *testing.T) {
+	identityManager := identity.NewIdentityManagerFake()
+	client := server.NewClientFake()
+	cache := identity.NewIdentityCacheFake()
+
+	handler := NewNodeIdentityHandler(identityManager, client, cache)
+
+	id, err := handler.UseExisting("address")
+	assert.Equal(t, identityManager.FakeIdentity1, id)
```
Waldz: Could your assert that no identity was registered?

cmd/mysterium_server/command_run/identity_handler_test.go
donce: Double checked - everything seems ok, not sure what's confusing you.

cmd/mysterium_server/command_run/factory.go
donce: Moved it back to `node_handler` as a function.

cmd/mysterium_server/command_run/identity_handler_test.go
donce: Done.

cmd/mysterium_server/command_run/identity_handler_test.go
donce: Done.

donce:
All comments done.

PR #90, opened 2018-01-10
MYST-228 Permit empty identity password when creating identity

donce:


tequilapi/endpoints/identities.go
```diff
@@ -123,7 +123,7 @@ func validateRegistrationRequest(regReq identityRegistrationDto) (err error) {
 
 func validateCreationRequest(createReq *identityCreationDto) (errors *validation.FieldErrorMap) {
 	errors = validation.NewErrorMap()
-	if len(createReq.Password) == 0 {
+	if createReq.Password == nil {
```
tadovas: what if { "password" : "" } maybe it's better to check if len(createReq.Password) ==0 ? In this case we will handle both (not present) and (present but empty) cases.

donce: It's a bit dangerous to allow making empty `POST /identities`, which returns you an account with empty password - api user might not realise that empty password was used!

Requiring to pass in `{"password": ""}` ensures that api user is aware that he is creating an account with empty password.

PR #93, opened 2018-01-11
MYST-114 sign all mysterium api rest calls

tadovas:


server/mysterium_api.go
```diff
@@ -48,10 +48,10 @@ func (mApi *mysteriumApi) RegisterIdentity(identity identity.Identity, signer id
 	return err
 }
 
-func (mApi *mysteriumApi) NodeRegister(proposal dto_discovery.ServiceProposal) error {
-	req, err := newPostRequest("node_register", dto.NodeRegisterRequest{
+func (mApi *mysteriumApi) RegisterProposal(proposal dto_discovery.ServiceProposal, signer identity.Signer) error {
+	req, err := newSignedPostRequest("node_register", dto.NodeRegisterRequest{
```
donce: This renaming makes sense! Could you rename the url of discovery-api as well? Calling same endpoint 2 different names looks very confusing.

server/interface.go
```diff
@@ -8,9 +8,10 @@ import (
 
 //Client interface for mysterium centralized api - will be removed in the future
 type Client interface {
-	RegisterIdentity(identity identity.Identity, signer identity.Signer) (err error)
-	NodeRegister(proposal dto_discovery.ServiceProposal) (err error)
-	NodeSendStats(nodeKey string) (err error)
 	FindProposals(nodeKey string) (proposals []dto_discovery.ServiceProposal, err error)
+	//these functions have side effects - therefore signer required
```
donce: Signer is not required only because these functions have side-effects:
In the future we might have `GET` endpoints without side-effects, which requires authorization.
What about this?
```
//these functions are signed because they require authorization
```

tadovas: Looks better. Fixed

PR #95, opened 2018-01-11
Add cors headers for tequilapi service

tadovas:


tequilapi/http_middlewares.go
```diff
@@ -0,0 +1,18 @@
+package tequilapi
+
+import "net/http"
+
+type corsHandler struct {
+	originalHandler http.Handler
+}
+
+func (wrapper corsHandler) ServeHTTP(resp http.ResponseWriter, req *http.Request) {
+	resp.Header().Set("Access-Control-Allow-Origin", "*")
```
shroomist: so now basically anybody could drink my tequila?
currently electron dev-env runs on `http://localhost:9080`

tadovas: not anybod. Any browser :D

PR #96, opened 2018-01-11
hex encoded signature changed to base64

zolia:


identity/signature.go
```diff
@@ -6,31 +6,39 @@ import (
...
-	return hex.EncodeToString(signature.raw)
+// Base64Encode encodes signature into Base64 format
+func (signature *Signature) Base64Encode() string {
+	return base64.StdEncoding.EncodeToString(signature.Bytes())
 }
 
-func (signature *Signature) Base64() string {
-	return base64.StdEncoding.EncodeToString(signature.Bytes())
+// SignatureBase64Decode decodes base64 string signature into raw bytes format
+func SignatureBase64Decode(signature string) Signature {
```
Waldz: It's factory, so name should be `SignatureBase64`

server/request.go
```diff
@@ -46,7 +46,7 @@ func newSignedPostRequest(path string, requestBody interface{}, signer identity.
 		return nil, err
 	}
 
-	req.Header.Add(authenticationHeaderName, authenticationSchemaName+" "+signature.Base64())
+	req.Header.Add(authenticationHeaderName, authenticationSchemaName+" "+signature.Base64Encode())
```
Waldz: Should header name be repeated 2 times?

identity/signature.go
```diff
@@ -6,31 +6,39 @@ import (
...
+	signatureBytes, _ := base64.StdEncoding.DecodeString(signature)
+	return Signature{signatureBytes}
 }
 
+// Bytes returns signature in raw bytes format
 func (signature *Signature) Bytes() []byte {
 	return signature.raw
 }
 
+// EqualsTo compares current signature with a given one
```
Waldz: You connect this function to `VerifierFake`

server/request.go
Waldz: My mistake

PR #97, opened 2018-01-11
MYST-114 Fix linter rules in "communication" package

Waldz:


bin/helpers/functions.sh
```diff
@@ -4,6 +4,6 @@
 function get_linker_ldflags {
     echo "
         -X 'github.com/mysterium/node/server.mysteriumApiUrl=${MYSTERIUM_API_URL}'
-        -X 'github.com/mysterium/node/communication/nats_discovery.natsServerIp=${NATS_SERVER_IP}'
+        -X 'github.com/mysterium/node/communication/natsdiscovery.natsServerIp=${NATS_SERVER_IP}'
```
donce: From effective Go (https://golang.org/doc/effective_go.html#names):
> By convention, packages are given lower case, single-word names; there should be no need for underscores or mixedCaps.

Can we name it just `communucation/discovery`? `natsdiscovery` is very unfriendly and is against effective go.

bin/lint
```diff
@@ -1 +1,16 @@
...
+#
+# Usage example:
+#> bin/lint
+#> bin/lint ./...
+#> bin/lint ./communication
+#> bin/lint ./communication/...
+
+ARGUMENTS=$@
+if [ -z "$ARGUMENTS" ]; then
+    ARGUMENTS=`go list ./... | grep -v /vendor/`
```
donce: grepping here is not necessary - `go list ./...` already excludes vendor packages. And since we're planning to do compilation using docker - we can choose not to use old `Go` version.

bin/lint
```diff
@@ -1 +1,16 @@
...
+fi
+
+go run vendor/github.com/golang/lint/golint/*.go -min_confidence=1 -set_exit_status ${ARGUMENTS}
```
donce: What errors are we hiding by increasing `min_confidence`?

communication/request.go
```diff
@@ -1,15 +1,18 @@
 package communication
 
-type RequestType string
+// RequestEndpoint is ppecial type that describes unique message endpoint
```
donce: `ppecial`

Waldz: Fixed

PR #98, opened 2018-01-15
feature/MYST-226 no pointers in consumers providers

zolia:


zolia:
too complicated to know required type of an interface for casting. Leave as it is now.

PR #104, opened 2018-01-21
Fix linter errors in "command_run.identity" and "location"

donce:


location/detection.go
```diff
@@ -9,10 +9,12 @@ import (
 // Database downloaded from http://dev.maxmind.com/geoip/geoip2/geolite2/
 const Database = "data/GeoLite2-Country.mmdb"
```
Waldz: Code source has nothing common with environment there compiled application is executed

- This config should be configured from CLI argument `--ip-database`
OR
- This config should be loader from `ConfigDirectory + "/GeoLite2-Country.mmdb"`

donce: Yes - I am working on that in a separate PR, because this PR is not related to this issue.

PR #107, opened 2018-01-21
MYST-148 gracefully close on Interrupt, sigterm and sighup

shroomist:
is this enough for graceful termination, or do we need a more elaborate system?

### DIFF
--- a/cmd/mysterium_client/run/command.go
+++ b/cmd/mysterium_client/run/command.go
@@ -12,9 +12,6 @@ import (
 	"github.com/mysterium/node/server"
 	"github.com/mysterium/node/tequilapi"
 	tequilapi_endpoints "github.com/mysterium/node/tequilapi/endpoints"
-	"os"
-	"os/signal"
-	"syscall"
 )
 
 //NewCommand function created new client command with options passed from commandline
@@ -70,10 +67,6 @@ type CommandRun struct {
 
 //Run starts Tequilapi service - does not block
 func (cmd *CommandRun) Run() error {
-	sigterm := make(chan os.Signal, 2)
-	signal.Notify(sigterm, os.Interrupt, syscall.SIGTERM, syscall.SIGHUP)
-	go sigtermListener(sigterm, cmd)
-
 	err := cmd.httpApiServer.StartServing()
 	if err != nil {
 		return err
@@ -104,14 +97,4 @@ func (cmd *CommandRun) Kill() error {
 	fmt.Printf("Api stopped\n")
 
 	return nil
-}
-
-func sigtermListener(terminator chan os.Signal, cmd *CommandRun) {
-	<-terminator
-	err := cmd.Kill()
-	if err != nil {
-		fmt.Printf("Unable to kill one of subroutines %q\n", err.Error())
-	}
-	fmt.Println("Good bye")
-	os.Exit(1)
 }
--- a/cmd/mysterium_client/run/command.go
+++ b/cmd/mysterium_client/run/command.go
@@ -12,6 +12,9 @@ import (
 	"github.com/mysterium/node/server"
 	"github.com/mysterium/node/tequilapi"
 	tequilapi_endpoints "github.com/mysterium/node/tequilapi/endpoints"
+	"os"
+	"os/signal"
+	"syscall"
 )
 
 //NewCommand function created new client command with options passed from commandline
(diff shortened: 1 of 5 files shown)


cmd/mysterium_client/run/command.go
```diff
@@ -52,11 +55,25 @@ func NewCommandWith(
...
 	httpApiServer := tequilapi.NewServer(options.TequilapiAddress, options.TequilapiPort, router)
-
-	return &CommandRun{
+	cmd := &CommandRun{
 		connectionManager,
 		httpApiServer,
 	}
+	sigterm := make(chan os.Signal, 2)
+	signal.Notify(sigterm, os.Interrupt, syscall.SIGTERM, syscall.SIGHUP)
+	go sarahConnor(sigterm, cmd)
```
Waldz: Factory should be responsible 1 responsibility - for factoring, but not running

shroomist: alright, this is now in Run()

cmd/mysterium_client/run/command.go
```diff
@@ -67,6 +70,10 @@ type CommandRun struct {
 
 //Run starts Tequilapi service - does not block
 func (cmd *CommandRun) Run() error {
+	sigterm := make(chan os.Signal, 2)
```
donce: Why `2`, isn't `1` enough?

cmd/mysterium_client/run/command.go
```diff
@@ -98,3 +105,13 @@ func (cmd *CommandRun) Kill() error {
...
+
+func sigtermListener(terminator chan os.Signal, cmd *CommandRun) {
+	<-terminator
+	err := cmd.Kill()
+	if err != nil {
+		fmt.Printf("Unable to kill one of subroutines %q\n", err.Error())
+	}
+	fmt.Println("Good bye")
+	os.Exit(1)
+}
```
donce: Closing subroutines and exiting is already implemented in `mysterium.client.go/newStopHandler` - can we re-use that somehow?

cmd/mysterium_client/run/command.go
```diff
@@ -98,3 +105,13 @@ func (cmd *CommandRun) Kill() error {
 
 	return nil
 }
+
+func sigtermListener(terminator chan os.Signal, cmd *CommandRun) {
```
donce: Function should be named as a verb, not as a noun - maybe `listenSigterm` suites better?

cmd/mysterium_client/run/command.go
```diff
@@ -98,3 +105,13 @@ func (cmd *CommandRun) Kill() error {
...
+	<-terminator
+	err := cmd.Kill()
+	if err != nil {
+		fmt.Printf("Unable to kill one of subroutines %q\n", err.Error())
```
donce: `os.Stderr` should be re-used for consistency, just like this:
```go
fmt.Fprintln(os.Stderr, err)
```

cmd/mysterium_client/run/command.go
```diff
@@ -98,3 +105,13 @@ func (cmd *CommandRun) Kill() error {
...
+	}
+	fmt.Println("Good bye")
+	os.Exit(1)
```
Waldz: `Exit(1)` mean program ending with errorNo=1, success graceful stop  should end with errorNo=0

cmd/mysterium_client/run/command.go
donce: It's hooked up differently, but the action it does on exit is duplicated:
```go
err := cmd.Kill()
if err != nil {
	fmt.Printf("Unable to kill one of subroutines %q\n", err.Error())
}
fmt.Println("Good bye")
os.Exit(1)
```

```go
if err := cmdRun.Kill(); err != nil {
	return err
}

os.Exit(0)
return nil
```

What bothers me is that these two duplications already differs - one returns exit `1` and prints outs "Good bye", another returns `0` without text.

We can easily DRY these - to move them to a separate method on `CommandRun` or to extend `CommandRun.Kill` with these duplications.

cmd/mysterium_client/run/command.go
donce: agree - we are already doing `os.Exit(0)` in `mysterium_client.go`

cmd/mysterium_server/command_run/command.go
```diff
@@ -110,8 +110,15 @@ func (cmd *CommandRun) Wait() error {
 	return cmd.vpnServer.Wait()
 }
 
-func (cmd *CommandRun) Kill() {
+func (cmd *CommandRun) Kill() (err error) {
```
donce: You're not really using named result variable - you can just leave `error`.

cmd/terminator.go
```diff
@@ -0,0 +1,32 @@
...
+}
+
+func NewTerminator(command Command) {
+	sigterm := make(chan os.Signal)
+	signal.Notify(sigterm, os.Interrupt, syscall.SIGTERM, syscall.SIGHUP)
+
+	go waitTerminatorSignals(sigterm, command)
+}
+
+func waitTerminatorSignals(terminator chan os.Signal, command Command) {
```
shroomist: for those who don't know -- this function is Sarah Connor

shroomist:
@donce, is it better now?

PR #109, opened 2018-01-22
Change "password" term with "passphrase" to unify naming

donce:


cmd/mysterium_client/cli/command.go
```diff
@@ -126,7 +126,7 @@ func (c *Command) connect(line string) {
 	consumerId, providerId := identities[0], identities[1]
 
 	if consumerId == "new" {
-		id, err := c.tequilapi.NewIdentity(identityDefaultPassword)
```
Waldz: But unlock passphrase should be taken from `Options.IdentityPassphrase`.
And Golang will default this option to `""`, You dont need constant for that

donce: There is no `Options` here - it's CLI. It should be passed in a command `identity new mypassphrase` - I've already created a subtask for that https://mysteriumnetwork.atlassian.net/browse/MYST-265. This subtask is only about renaming.

Waldz: clear

PR #110, opened 2018-01-22
Cors handling middleware improved to handle preflight cors requests

tadovas:


tequilapi/http_middlewares.go
```diff
@@ -7,12 +7,32 @@ type corsHandler struct {
...
+}
+
+func isPreflightCorsRequest(req *http.Request) bool {
+	isOptionsMethod := req.Method == http.MethodOptions
+	containsAccessControlRequestMethod := req.Header.Get("Access-Control-Request-Method") != ""
+	containsAccessControlRequestHeader := req.Header.Get("Access-Control-Request-Headers") != ""
+	return isOptionsMethod && containsAccessControlRequestHeader && containsAccessControlRequestMethod
+}
+
+func applyPreflightCorsResponse(resp http.ResponseWriter) {
```
Waldz: Why conditional check is done?
As i understand, later both `applyPreflightCorsResponse()` and `allowAllCorsActions()` does the same

tequilapi/http_middlewares.go
```diff
@@ -7,12 +7,32 @@ type corsHandler struct {
 }
 
 func (wrapper corsHandler) ServeHTTP(resp http.ResponseWriter, req *http.Request) {
-	resp.Header().Set("Access-Control-Allow-Origin", "*")
-	resp.Header().Set("Access-Control-Allow-Methods", "POST, GET, OPTIONS, PUT, DELETE")
+	if isPreflightCorsRequest(req) {
```
Waldz: Idea - would not it be easier to connect middleware for each route, we are registering?

tadovas: It would make sense if we really gonna use CORS for more grained control checking in our api. At the moment - it is simplier to handle all cors (allow all actions) in single place which naturaly is single http handling function.

tequilapi/http_middlewares.go
tadovas: Actually there are two cases:
1. If it's preflight request - add headers and return EMPTY response
2. Any other case - simply add cors headers.
I wanted to show intent that it's separate case - return empty response with headers. Maybe renaming function to generatePreflightCorsResponse will express intetion better.

shroomist:
I'm still getting `Request header field Content-Type is not allowed by Access-Control-Allow-Headers in preflight response`

donce:
For those who review and doesn't know about preflight requests (like me):
https://developer.mozilla.org/en-US/docs/Glossary/Preflight_request

tequilapi/http_middlewares.go
```diff
@@ -7,12 +7,33 @@ type corsHandler struct {
...
 
 //ApplyCors wraps original handler by adding cors headers to response BEFORE original ServeHTTP method is called
 func ApplyCors(original http.Handler) http.Handler {
 	return corsHandler{original}
 }
+
+func allowAllCorsActions(resp http.ResponseWriter) {
+	resp.Header().Set("Access-Control-Allow-Origin", "*")
+	resp.Header().Set("Access-Control-Allow-Methods", "POST, GET, OPTIONS, PUT, DELETE")
+	resp.Header().Set("Access-Control-Allow-Headers", "Content-type")
```
shroomist: `Access-Control-Allow-Headers", "Content-type"` fixes preflights for me.

tequilapi/http_middlewares.go
donce: What about this?
```go
allowAllCorsActions(resp)
if isPreflightCorsRequest(req) {
  return
}
wrapper.originalHandler.ServeHTTP(resp, req)
```
This shows that same headers are applied always and that empty response is returned in case it's preflight request.

tequilapi/http_middlewares_test.go
```diff
@@ -22,6 +22,25 @@ func TestCorsHeadersAreAppliedToResponse(t *testing.T) {
 	assert.True(t, mock.wasCalled)
 }
 
+func TestPreflightCorsCheckIsHandled(t *testing.T) {
+	req, err := http.NewRequest(http.MethodOptions, "/not-important", nil)
+	assert.NoError(t, err)
+
+	req.Header.Add("Access-Control-Request-Method", "DELETE")
+	req.Header.Add("Access-Control-Request-Headers", "origin, x-requested-with")
```
donce: Can we add content-type header here, to check that it's allowed? (which was why it was failing before).

tadovas:
Refactored to allow any headers specified by request
@donce It works now, and I think your idea was good, but not relevant anymore. Please recheck

tequilapi/http_middlewares_test.go
tadovas: Refactored handler now simply allows all headers defined in request

PR #114, opened 2018-01-23
Fix runtime error when aborting program before http server started

donce:


tequilapi/http_api_server_test.go
```diff
@@ -17,3 +17,8 @@ func TestLocalAPIServerPortIsAsExpected(t *testing.T) {
 	server.Stop()
 	server.Wait()
 }
+
+func TestStopBeforeStartingListening(t *testing.T) {
```
tadovas: Maybe a better test name would express why this test is needed like "StoppingServerBeforeListenning should not cause segfault" (this was the problem right?). But it's minor

PR #115, opened 2018-01-23
Fix command run error handling, which sometimes results in infinite loop

donce:
I've experienced this when port `4050` was already in use. In this case, node was running in loop without any output.


donce:
@tadovas I believe no unit-tests for this class explains that :D

PR #117, opened 2018-01-24
Fix country database file to be reachable in docker

donce:


cmd/mysterium_server/command_run/command.go
```diff
@@ -92,14 +93,14 @@ func (cmd *CommandRun) Run() (err error) {
 	return nil
 }
 
-func detectCountry() (string, error) {
+func detectCountry(detector location.Detector) (string, error) {
 	ipifyClient := ipify.NewClient()
```
Waldz: `ipifyClient` is already injected into command

donce: Right! Updated to re-use `ipifyClient`

PR #118, opened 2018-01-24
MYST-268 Add endpoint and CLI command for current IP

donce:


tequilapi/client/client.go
```diff
@@ -110,6 +110,20 @@ func (client *Client) Status() (status StatusDto, err error) {
 	return status, err
 }
 
+func (client *Client) Ip() (string, error) {
```
Waldz: `GetIP()`

tequilapi/client/client.go
```diff
@@ -110,6 +110,20 @@ func (client *Client) Status() (status StatusDto, err error) {
...
+	response, err := client.http.Get("ip", url.Values{})
+	if err != nil {
+		return "", err
+	}
+	defer response.Body.Close()
+
+	var ipData struct {
+		Ip string `json:"ip"`
```
Waldz: `IP`

tequilapi/endpoints/current_ip.go
```diff
@@ -0,0 +1,35 @@
+package endpoints
+
+import (
+	"github.com/julienschmidt/httprouter"
+	"github.com/mysterium/node/tequilapi/utils"
+	"net/http"
+)
+
+type currentIPEndpoint struct {
+	ipResolver ipResolver
```
Waldz: All `ipifyClient` could be this dependency. And finally it could renamed `ipifyClient` -> `IPResolver`

donce: Not quite sure I understood the suggestion - are you suggesting to refactor `ipify.Client` interface with all its implementations into functions (instead of a classes), i.e.:
`func ipify.NewClient (ipify.Client, error)` -> `func ipify.NewIPResolver (IPResolver, error)` ?

Waldz: No, I suggest to inject full ipifyClient here `ipResolver ipify.Client`

tequilapi/http_routes.go
```diff
@@ -2,16 +2,21 @@ package tequilapi
...
 
+// NewAPIRouter returns new api router with status endpoints
 func NewAPIRouter() *httprouter.Router {
 	router := httprouter.New()
 	router.HandleMethodNotAllowed = true
 
 	router.GET("/healthcheck", endpoints.HealthCheckEndpointFactory(time.Now, os.Getpid).HealthCheck)
 
+	ipifyClient := ipify.NewClient()
+	router.GET("/ip", endpoints.NewCurrentIPEndpoint(ipifyClient.GetPublicIP).CurrentIP)
```
tadovas: Maybe we can make a more generic endpoint and call it /network which can return more network information, not only ip?

Waldz: It should go somewhere under `/v1..`

tequilapi/http_routes.go
```diff
@@ -2,16 +2,21 @@ package tequilapi
...
 )
 
+// NewAPIRouter returns new api router with status endpoints
 func NewAPIRouter() *httprouter.Router {
 	router := httprouter.New()
 	router.HandleMethodNotAllowed = true
 
 	router.GET("/healthcheck", endpoints.HealthCheckEndpointFactory(time.Now, os.Getpid).HealthCheck)
 
+	ipifyClient := ipify.NewClient()
```
Waldz: Dont factory on the fly, pass dependency from upper layer

Waldz: But I just noticed that `NewAPIRouter()` factory is not right place to connect features

tequilapi/http_routes.go
Waldz: Maybe `GET /v1/connection/ip` is right place

donce: We don't have `/v1`, but it makes sense to move it inside `connection`.

tequilapi/http_routes.go
donce: Moved to connection endpoints

tequilapi/http_routes.go
donce: moved to `connection`

donce: It's not hard to rename later on if we have to add more fields - for now, all it does is return IP, because that's the user story.

tequilapi/http_routes.go
donce: Also, added dependency injection.

donce:
@Waldz fixed all issues, ready for review :)

ip/rest_resolver.go
```diff
@@ -11,15 +11,15 @@ import (
 	"time"
 )
 
-const IPIFY_API_URL = "https://api.ipify.org/"
-const IPIFY_API_CLIENT = "goclient-v0.1"
-const IPIFY_API_LOG_PREFIX = "[ipify.api] "
+const IpifyAPIURL = "https://api.ipify.org/"
+const IpifiAPICLient = "goclient-v0.1"
```
Waldz: `Ipifi` -> `Ipify..`
+ Could be private constants

ip/resolver.go
```diff
@@ -1,6 +1,6 @@
-package ipify
+package ip
 
-type Client interface {
+type Resolver interface {
 	GetPublicIP() (string, error)
```
Waldz: Could be 2 implementations with same `ip. Resolver` interface e.g. `NewPublicResolver()`, `NewOutboundResolver()`

ip/rest_resolver.go
donce: Done.

PR #123, opened 2018-01-29
MYST-274 Session duration

donce:


tequilapi/endpoints/connection.go
```diff
@@ -87,19 +88,29 @@ func (ce *connectionEndpoint) GetIP(writer http.ResponseWriter, request *http.Re
 // GetStatistics returns statistics about current connection
 func (ce *connectionEndpoint) GetStatistics(writer http.ResponseWriter, request *http.Request, params httprouter.Params) {
 	stats := ce.statsKeeper.Retrieve()
+
+	duration, err := ce.statsKeeper.GetSessionDuration()
+	if err != nil {
+		newErr := fmt.Errorf("unable to retrieve session duration: %v", err.Error())
+		utils.SendError(writer, newErr, http.StatusUnprocessableEntity)
```
tadovas: Unprocessable entity error is meant to use for http request body validation and is not allowed for GET requests, not to report some internal state problems. Maybe duration can be simply 0 if session is not started? Do we really need to handle that "special" case?

tequilapi/endpoints/connection_test.go
```diff
@@ -76,7 +84,11 @@ func TestAddRoutesForConnectionAddsRoutes(t *testing.T) {
 		},
 		{
 			http.MethodGet, "/connection/statistics", "",
-			http.StatusOK, `{"bytesSent": 0, "bytesReceived": 0}`,
+			http.StatusOK, `{
+				"bytesSent": 0,
+				"bytesReceived": 0,
+				"durationSeconds": 60
```
tadovas: what about just "sessionDuration"? It can be just convention - sent,received are reported as bytes, duration is reported as seconds. Api consumer when can reformat to any precision needed.

openvpn/middlewares/client/bytescount/stats_keeper.go
```diff
@@ -1,16 +1,53 @@
...
+
+// MarkSessionStart marks current time as session start time for statistics
+func (keeper *sessionStatsKeeper) MarkSessionStart() {
+	time := keeper.timeGetter()
+	keeper.sessionStart = &time
+}
+
+// GetSessionDuration returns elapsed time from marked session start
+func (keeper *sessionStatsKeeper) GetSessionDuration() (time.Duration, error) {
+	if keeper.sessionStart == nil {
```
tadovas: Do we really need to throw errors here and force any caller to handle them? It's just for statistics. Maybe we can simply return zero in case session is not started.

tequilapi/endpoints/connection_test.go
donce: `bytesSent` and `bytesReceived` specifies that it's `bytes` in the name, while `duration` does not. At first I implemented it in `duration` myself, but having raw number without explanation about value seemed confusing - also` durationSeconds` is more consistent with `bytesSent`/`bytesReceived`.

tadovas: In that case: maybe sentBytes, receivedBytes and sessionSeconds? :) or vice versa. More consistent

openvpn/middlewares/client/bytescount/stats_keeper.go
donce: Simplified to return zero in this case.

tequilapi/endpoints/connection.go
donce: Simplified to return zero in this case - no custom http status left.

PR #124, opened 2018-01-30
MYST-272 CLI for connection statistics

donce:
screen shot 2018-01-30 at 11 44 25

### DIFF
--- a/cmd/mysterium_client/cli/command.go
+++ b/cmd/mysterium_client/cli/command.go
@@ -213,7 +213,7 @@ func (c *Command) status() {
 		if err != nil {
 			warn(err)
 		} else {
-			info(fmt.Sprintf("Connection duration: %ds", statistics.DurationSeconds))
+			info(fmt.Sprintf("Connection duration: %ds", statistics.Duration))
 			info("Bytes sent:", statistics.BytesSent)
 			info("Bytes received:", statistics.BytesReceived)
 		}
--- a/cmd/mysterium_client/cli/command.go
+++ b/cmd/mysterium_client/cli/command.go
@@ -202,11 +202,19 @@ func (c *Command) status() {
 	status, err := c.tequilapi.Status()
 	if err != nil {
 		warn(err)
-		return
+	} else {
+		info("Status:", status.Status)
+		info("SID:", status.SessionId)
 	}
 
-	info("Status:", status.Status)
-	info("SID:", status.SessionId)
+	statistics, err := c.tequilapi.ConnectionStatistics()
+	if err != nil {
+		warn(err)
+	} else {
+		info(fmt.Sprintf("Connection duration: %ds", statistics.DurationSeconds))
+		info("Bytes sent:", statistics.BytesSent)
+		info("Bytes received:", statistics.BytesReceived)
+	}
 }
 
 func (c *Command) ip() {
--- a/cmd/mysterium_client/cli/command.go
+++ b/cmd/mysterium_client/cli/command.go
@@ -204,7 +204,7 @@ func (c *Command) status() {
 		warn(err)
 	} else {
 		info("Status:", status.Status)
-		info("SID:", status.SessionId)
+		info("SID:", status.SessionID)
 	}
 
 	statistics, err := c.tequilapi.ConnectionStatistics()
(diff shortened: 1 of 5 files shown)


cmd/mysterium_client/cli/command.go
```diff
@@ -202,11 +202,19 @@ func (c *Command) status() {
...
 		warn(err)
-		return
+	} else {
+		info("Status:", status.Status)
+		info("SID:", status.SessionID)
 	}
 
-	info("Status:", status.Status)
-	info("SID:", status.SessionId)
+	statistics, err := c.tequilapi.ConnectionStatistics()
```
tadovas: It would be nice if statistics are printed only if status == connected, otherwise - why to trash screen with not relevant information?

tequilapi/client/client.go
```diff
@@ -106,16 +106,28 @@ func (client *Client) Disconnect() (err error) {
 	return nil
 }
 
+// ConnectionStatistics returns statistics about current connection
+func (client *Client) ConnectionStatistics() (statistics StatisticsDTO, err error) {
```
Waldz: Should not it be getter?

tequilapi/client/client.go
```diff
@@ -106,16 +106,28 @@ func (client *Client) Disconnect() (err error) {
...
+func (client *Client) Status() (status StatusDTO, err error) {
 	response, err := client.http.Get("connection", url.Values{})
 	if err != nil {
 		return
 	}
 	defer response.Body.Close()
 
 	err = parseResponseJson(response, &status)
-	return status, err
+	return
```
Waldz: Is it bad to have clearer declaration?

tequilapi/client/dto.go
```diff
@@ -1,17 +1,24 @@
...
 	Status    string `json:"status"`
-	SessionId string `json:"sessionId"`
+	SessionID string `json:"sessionId"`
 }
 
-// IdentityDto holds identity address
-type IdentityDto struct {
+// StatisticsDTO holds statistics about connection
+type StatisticsDTO struct {
+	BytesSent       int `json:"bytesSent"`
```
Waldz: For external users, I would name it `uploadBytes` and `downloadBytes`

tequilapi/client/dto.go
```diff
@@ -1,17 +1,24 @@
...
+	BytesReceived   int `json:"bytesReceived"`
+	DurationSeconds int `json:"durationSeconds"`
```
Waldz: Should we add `..seconds` to attribute names?

donce: It specifies unit in which these values are returned. i.e.:
```
{
"sent": 23,
"received": 54,
"duration": 3
}
```
This would not be easy to understand, because it's unclear what units these values are in.

tequilapi/client/client.go
donce: Since result parameters are named, I don't see much value added by rewriting the same named parameters here which is the default of `return`

tequilapi/client/dto.go
donce: `"uploadAmount": 5` is a bit confusing - it doesn't specify units. But I like `upload` - what do you think about `uploadedBytes` and `downloadedBytes` ?

tequilapi/client/client.go
donce: It is a getter in a way, but it doesn't need `Get` prefix:
From https://golang.org/doc/effective_go.html#Getters:
> it's neither idiomatic nor necessary to put Get into the getter's name. If you have a field called owner (lower case, unexported), the getter method should be called Owner (upper case, exported), not GetOwner

cmd/mysterium_client/cli/command.go
donce: Updated

tequilapi/client/client.go
donce: Removed named parameters to reduce discussions on this.

tadovas:
LGTM

tequilapi/client/dto.go
donce: Applied naming we agreed on - renamed `durationSeconds` to `duration`, left byte fields unotuched.

tequilapi/client/client.go
Waldz: Cool.
Would be good to consistent in one object, I remember something like  `GetIP` and `GetStatus`

PR #135, opened 2018-02-01
HOTFIX One step towards single node

Waldz:


Waldz:
Before merged, but not to master :)

PR #136, opened 2018-02-02
execute script to update mac system dns servers after openvpn connect…

zolia:
3 executables needs to be shipped in neighbourhood:
- `mysterium_client`
- `openvpn`
- `update-resolv-conf`

### DIFF
--- a/openvpn/config.go
+++ b/openvpn/config.go
@@ -1,6 +1,7 @@
 package openvpn
 
 import (
+	"path/filepath"
 	"strconv"
 )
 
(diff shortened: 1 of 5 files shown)


openvpn/config.go
```diff
@@ -39,6 +40,11 @@ func (c *Config) SetManagementSocket(socketAddress string) {
 	c.setFlag("management-client")
 }
 
+func (c *Config) SetScripts(runtimeDirectory string) {
+	c.setParam("up", filepath.Join(runtimeDirectory, "update-resolv-conf"))
```
Waldz: Call this script relativelly as `update-resolv-conf`:
- Will be searched in neighbourhood
- Will be searched in PATH

PR #140, opened 2018-02-05
Remove mysterium monitor command

donce:
This is not used anymore and it's broken. Since we're not supporting it, we can remove it for now - this can be reverted when (if) we actually need it.


shroomist:
trust you on this, I don't mind.

PR #142, opened 2018-02-06
Remove mysterium fake command

donce:
This functionality is being replaced with `docker` :)


nat/service_fake.go
```diff
@@ -1,20 +0,0 @@
-package nat
-
-func NewServiceFake() *serviceFake {
```
Waldz: Dont remove, this stub is useful

Waldz:
Suggest to remove helper scripts from `bin`

PR #158, opened 2018-02-13
New cli command - stop, stops mysterium client

tadovas:


tadovas:
Well technically they are independent :) stop - stops client, exit - exits cli itself. But good point

PR #162, opened 2018-02-14
HOTFIX. Freeze Openvpn dependency

Waldz:


Waldz:
We use arguments and crypto features, which are supported only in newest version

PR #165, opened 2018-02-15
HOTFIX Dont omit location database on --location.database=""

Waldz:


cmd/commands/server/factory.go
```diff
@@ -53,10 +53,8 @@ func NewCommandWith(
 	var locationDetector location.Detector
 	if options.LocationCountry != "" {
 		locationDetector = location.NewDetectorFake(options.LocationCountry)
-	} else if options.LocationDatabase != "" {
-		locationDetector = location.NewDetector(filepath.Join(options.DirectoryConfig, options.LocationDatabase))
 	} else {
-		locationDetector = location.NewDetectorFake("")
+		locationDetector = location.NewDetector(filepath.Join(options.DirectoryConfig, options.LocationDatabase))
```
donce: Mutating `options.LocationDatabase` (in options.go) is bad - this makes it hard to understand in later code, why `options.LocationDatabase` is assumed to be present, even though this option is optional.
To solve this, we can use default location detabase here:
```go
...
else {
  var locationDatabase string
  if options.DirectoryConfig != "" {
    locationDatabase = options.DirectoryConfig
  } else {
    locationDatabase = defaultLocationDatabase
  }
  locationDetector = location.NewDetector(filepath.Join(locationDatabase, options.LocationDatabase))
}
```

Waldz: Fixed

PR #166, opened 2018-02-19
MYST-324 Return "Service Unavailable" when ipify is unavailable

donce:


tequilapi/endpoints/connection.go
```diff
@@ -61,6 +64,7 @@ func (ce *connectionEndpoint) Create(resp http.ResponseWriter, req *http.Request
 		case connection.ErrAlreadyExists:
 			utils.SendError(resp, err, http.StatusConflict)
 		default:
+			log.Error(connectionLogPrefix, err)
```
Waldz: Should not logs be handled in connection manager, and it it could have more detailed logs on intermediate steps

Waldz: At least, we applied this approach in other places

donce: There is no convenient way to log all connect errors in connection manager - take a look at it's implementation:
```go
func (manager *connectionManager) Connect(consumerID, providerID identity.Identity) error {
	if manager.status.State != NotConnected {
		return ErrAlreadyExists
	}

	proposal, err := manager.findProposalByProviderID(providerID)
	if err != nil {
		return err
	}

	dialogEstablisher := manager.newDialogEstablisher(consumerID)
	manager.dialog, err = dialogEstablisher.EstablishDialog(providerID, proposal.ProviderContacts[0])
	if err != nil {
		return err
	}

	vpnSession, err := session.RequestSessionCreate(manager.dialog, proposal.ID)
	if err != nil {
		return err
	}
	manager.sessionID = vpnSession.ID

	manager.openvpnClient, err = manager.newVpnClient(*vpnSession, consumerID, providerID, manager.onVpnStatusUpdate)
	if err != nil {
		manager.dialog.Close()
		return err
	}

	if err := manager.openvpnClient.Start(); err != nil {
		manager.dialog.Close()
		return err
	}
	return nil
}
```
There are 6 places where error can occur - I want to log all of them, but adding 6 log statements just because this class did not have any logs before doesn't make sense to me.

PR #167, opened 2018-02-20
MYST 355: limit openvpn client reconnects

zolia:
currently, if node identity changes, client auth simply fails and client hangs in reconnecting state. this PR ensures, that if auth failed, do not attempt to reconnect. Myst client now reports "not connected".


bin/server_package/config/ca.crt
```diff
@@ -1,19 +0,0 @@
------BEGIN CERTIFICATE-----
```
donce: Why are these removed?

zolia: these are generated now, leftovers from the old auth model.

location/detector_test.go
```diff
@@ -16,6 +16,7 @@ func TestDetectorDetectCountry(t *testing.T) {
 		{"95.85.39.36", "NL", ""},
 		{"127.0.0.1", "", ""},
 		{"8.8.8.8.8", "", "failed to parse IP"},
+		{"185.243.112.225", "", ""},
```
donce: If we return error when we are unable to found country in database, using country detector would be much easier - if error was not returned, that means country was returned :)

openvpn/config.go
```diff
@@ -60,6 +60,13 @@ func (c *Config) SetTLSCrypt(cryptFile string) {
 	c.AddOptions(OptionFile("tls-crypt", cryptFile))
 }
 
+func (c *Config) SetReconnectLimits() {
+	c.setParam("connect-retry-max", "2")
+	c.setParam("remap-usr1", "SIGTERM")
+	c.setFlag("single-session")
+	c.setFlag("tls-exit")
```
donce: Some of params doesn't seem to be used for reconnection limit, which is the purpose of this method. I.e. `tls-exit` and `remap-usr1` seems to be used for allowing to exit vpn in some graceful way.

openvpn/factory.go
```diff
@@ -40,6 +40,7 @@ func NewClientConfig(
 	config.SetClientMode(remote, 1194)
 	config.SetTLSCACertificate(caCertPath)
 	config.SetTLSCrypt(tlsCryptKeyPath)
+	config.SetReconnectLimits()
```
donce: If number `2` were a parameter to this method, we could see what that limit is in here - also, that would allow us to change it in factory without changing config class itself.

openvpn/config.go
zolia: i use limit in broader sense, not a number, but limitation, not to reconnect.

zolia: ant these are parameters that enforces that limitation..

donce: > i use limit in broader sense, not a number, but limitation, not to reconnect.

I don't quite understand - anyways, documentation for this method is missing - maybe documenting this would explain what broader sense you have?

openvpn/factory.go
shroomist: intuitively I want to pass parameters to a function that has a `set` prefix.
perhaps `RestrictConnectRetryMax(2)`

bin/server_package/config/ca.crt
donce: If it was in a separate commit with an explanatory message, it would cause less confusion :)

location/detector_test.go
Waldz: return errors.New("unkown country for IP: x.x.x")

openvpn/config.go
```diff
@@ -60,6 +60,13 @@ func (c *Config) SetTLSCrypt(cryptFile string) {
 	c.AddOptions(OptionFile("tls-crypt", cryptFile))
 }
 
+func (c *Config) SetReconnectLimits() {
+	c.setParam("connect-retry-max", "2")
```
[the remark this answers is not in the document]
zolia: I believe no. At least not in here. This is just openvpn related. True connect / reconnect attempts should go from tequila api. Failure to connect should probably mean either real connectivity issues or failed auth due to state / outdated auth procedure in dialog stage.

openvpn/factory.go
zolia: Restrict prob better term here. Will rename.

PR #172, opened 2018-02-21
MYST 361: add -discovery-address and -broker-address command line options

zolia:
PS adds -discovery-address and -broker-address command line options to override compiled-in options

### DIFF
--- a/bin/client_package/installation/default
+++ b/bin/client_package/installation/default
@@ -6,5 +6,5 @@ MYSTERIUM_CLIENT_TEQUILAPI_ADDRESS="localhost"
 # Port for listening incoming api requests
 MYSTERIUM_CLIENT_TEQUILAPI_PORT="4050"
 
-# Mysterium discovery address (default: http://testnet.mysterium.network:5000/v1)
+# Mysterium discovery address (eg.: http://testnet.mysterium.network:5000/v1)
 MYSTERIUM_DISCOVERY_ADDRESS: ""
--- a/bin/client_package/installation/default
+++ b/bin/client_package/installation/default
@@ -5,3 +5,6 @@ MYSTERIUM_CLIENT_TEQUILAPI_ADDRESS="localhost"
 
 # Port for listening incoming api requests
 MYSTERIUM_CLIENT_TEQUILAPI_PORT="4050"
+
+# Mysterium discovery address (default: http://testnet.mysterium.network:5000/v1)
+MYSTERIUM_DISCOVERY_ADDRESS: ""
--- a/bin/client_run
+++ b/bin/client_run
@@ -1,3 +1,10 @@
 #!/bin/bash
 
-sudo ./build/client/mysterium_client --config-dir=bin/client_package/config --runtime-dir=build/client $@
+source bin/helpers/load_run_environment.sh
+
+sudo ./build/client/mysterium_client \
+   --config-dir=bin/client_package/config \
+   --runtime-dir=build/client \
+   $DISCOVERY \
+   $BROKER \
+   $@
--- a/bin/client_run
+++ b/bin/client_run
@@ -6,5 +6,4 @@ sudo ./build/client/mysterium_client \
    --config-dir=bin/client_package/config \
    --runtime-dir=build/client \
    $DISCOVERY \
-   $BROKER \
    $@
--- a/bin/client_run
+++ b/bin/client_run
@@ -5,5 +5,5 @@ source bin/helpers/load_run_environment.sh
 sudo ./build/client/mysterium_client \
    --config-dir=bin/client_package/config \
    --runtime-dir=build/client \
-   $DISCOVERY \
+   $DISCOVERY_OPTION\
    $@
--- a/bin/helpers/functions.sh
+++ b/bin/helpers/functions.sh
@@ -3,8 +3,7 @@
 # Map environment variables to flags for Golang linker's -ldflags usage
 function get_linker_ldflags {
     echo "
-        -X 'github.com/mysterium/node/server.mysteriumApiUrl=${MYSTERIUM_API_URL}'
-        -X 'github.com/mysterium/node/server.natsServerIP=${NATS_SERVER_IP}'
-        -X 'github.com/mysterium/node/client.mysteriumApiUrl=${MYSTERIUM_API_URL}'
+        -X 'github.com/mysterium/node/cmd/commands/server.MysteriumApiUrl=${MYSTERIUM_API_URL}'
+        -X 'github.com/mysterium/node/cmd/commands/server.natsServerIP=${NATS_SERVER_IP}'
     "
 }
--- a/bin/helpers/functions.sh
+++ b/bin/helpers/functions.sh
@@ -3,7 +3,7 @@
 # Map environment variables to flags for Golang linker's -ldflags usage
 function get_linker_ldflags {
     echo "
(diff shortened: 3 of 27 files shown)


bin/helpers/functions.sh
```diff
@@ -4,6 +4,7 @@
 function get_linker_ldflags {
     echo "
         -X 'github.com/mysterium/node/server.mysteriumApiUrl=${MYSTERIUM_API_URL}'
-        -X 'github.com/mysterium/node/communication/nats/discovery.natsServerIP=${NATS_SERVER_IP}'
+        -X 'github.com/mysterium/node/server.natsServerIP=${NATS_SERVER_IP}'
+        -X 'github.com/mysterium/node/client.mysteriumApiUrl=${MYSTERIUM_API_URL}'
```
Waldz: Why `mysteriumApiUrl ` several times

zolia: its in different packages. to remove duplication I could make it public..

server/request_test.go
```diff
@@ -17,6 +17,8 @@ type mockedSigner struct {
 	signatureToReturn identity.Signature
 }
 
+var testMysteriumApiUrl = "http://testUrl"
```
donce: We shouldn't depend that url is `mysterium` - it can be just `testRequestURL`.
Also, `URL` has to be upper-case.

cmd/commands/server/options.go
```diff
@@ -17,10 +17,16 @@ type CommandOptions struct {
...
 	LocationCountry  string
 	LocationDatabase string
+
+	DiscoveryAPIAddress string
+	BrokerAddress       string
 }
 
 const defaultLocationDatabase = "GeoLite2-Country.mmdb"
 
+var MysteriumApiUrl string
```
donce: Undocumented.

bin/helpers/load_run_environment.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+if [ -f .env ]; then
+   source .env
+
+   [ -n "$MYSTERIUM_API_URL" ] && DISCOVERY="-discovery-address=$MYSTERIUM_API_URL"
+   [ -n "$NATS_SERVER_IP" ] && BROKER="-broker-address=$NATS_SERVER_IP"
```
donce: Isn't too dashes common practice for key-value params?
--discovery-address=FOO

cmd/commands/client/options.go
```diff
@@ -60,6 +64,13 @@ func ParseArguments(args []string) (options CommandOptions, err error) {
 		"Run an interactive CLI based Mysterium UI",
 	)
 
+	flags.StringVar(
+		&options.DiscoveryAPIAddress,
+		"discovery-address",
+		server.MysteriumApiUrl,
```
donce: `client` shouldn't rely on `server` for options - they should be independant.
Can we extract this param into common class in `cmd` package, i.e. `cmd.CommonOptions`?

tadovas: This is gotcha :) and example of bad naming. "server" package has nothing to do with server functionality, it's simply part of discovery api functionality.

cmd/commands/server/options.go
donce: `MysteriumAPIURL`

bin/helpers/load_run_environment.sh
zolia: I agree, but its true for all options. Legacy.

server/request.go
```diff
@@ -15,22 +15,20 @@ const (
 	authenticationSchemaName = "Signature"
 )
 
-var mysteriumApiUrl string
-
-func newGetRequest(path string, params url.Values) (*http.Request, error) {
+func newGetRequest(apiURI, path string, params url.Values) (*http.Request, error) {
```
donce: Since `mApi.discoveryAPIAddress` seems to be used  as `apiURL` param value in all calls, does it make sense to parametrize it?

cmd/commands/client/options.go
zolia: true, moved to cmd package.

bin/client_run
```diff
@@ -1,3 +1,9 @@
 #!/bin/bash
 
-sudo ./build/client/mysterium_client --config-dir=bin/client_package/config --runtime-dir=build/client $@
+source bin/helpers/load_run_environment.sh
+
+sudo ./build/client/mysterium_client \
+   --config-dir=bin/client_package/config \
+   --runtime-dir=build/client \
+   $DISCOVERY \
```
donce: Shouldn't this param be passed in in docker-entrypoints as well?

server/request.go
zolia: not sure about this place, since its functions and not member methods we should not depend on internal state. Should we hardcoded in apiURI, can it be considered a state or never changing constant..

bin/client_run
zolia: it depends from environment we want to build for.. we should compile in for testnet and only for development to show our own discovery.

requestor/request_test.go
```diff
@@ -17,6 +17,8 @@ type mockedSigner struct {
 	signatureToReturn identity.Signature
 }
 
+var testRequestApiUrl = "http://testUrl"
```
donce: `testRequestAPIURL`

requestor/request.go
```diff
@@ -1,4 +1,4 @@
-package server
+package requestor
```
donce: `requestor` name is confusing, since there is no object `requestor` here - it's just a bunch of functions.
What about package `request` or `requests`?
I.e. `requests.NewGetRequest` looks quite good.

tadovas: requests.Get looks even better :)

bin/client_run
zolia: is is passed now

bin/client_package/installation/default
```diff
@@ -5,3 +5,6 @@ MYSTERIUM_CLIENT_TEQUILAPI_ADDRESS="localhost"
 
 # Port for listening incoming api requests
 MYSTERIUM_CLIENT_TEQUILAPI_PORT="4050"
+
+# Mysterium discovery address (default: http://testnet.mysterium.network:5000/v1)
```
donce: How `http://testnet.mysterium.network:5000/v1` is default? Isn't value in `.env` the default?

bin/helpers/load_run_environment.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+if [ -f .env ]; then
+   source .env
+
+   [ -n "$MYSTERIUM_API_URL" ] && DISCOVERY="--discovery-address=$MYSTERIUM_API_URL"
+   [ -n "$NATS_SERVER_IP" ] && BROKER="--broker-address=$NATS_SERVER_IP"
```
donce: `DISCOVERY` and `BROKER` are too common - perhaps `DISCOVERY_OPTION` and `BROKER_OPTION` are more explicit?

cmd/commands/server/options.go
```diff
@@ -17,10 +17,15 @@ type CommandOptions struct {
...
 	LocationCountry  string
 	LocationDatabase string
+
+	DiscoveryAPIAddress string
+	BrokerAddress       string
 }
 
 const defaultLocationDatabase = "GeoLite2-Country.mmdb"
 
+var natsServerIP string
```
donce: `brokerIP`?

PR #173, opened 2018-02-22
feature/MYST-149 command line option to specify openvpn executable

tadovas:


cmd/commands/client/options.go
```diff
@@ -39,6 +40,12 @@ func ParseArguments(args []string) (options CommandOptions, err error) {
 		filepath.Join(cmd.GetDataDirectory(), "run"),
 		"Runtime writable directory for temp files",
 	)
+	flags.StringVar(
+		&options.OpenvpnBinary,
+		"openvpn.binary",
+		"openvpn", //search in $PATH by default,
+		"openvpn binary to use for Open VPN connections",
+	)
```
ignasbernotas: Would be nice to output an error and kill the client if openvpn isn't found, otherwise you'll only find that out during connect.

tadovas: That was my idea for next PR - on client/server startup execute openvpn.binary --version - print it if successfull, fail with error otherwise :)

PR #178, opened 2018-02-23
Add simple integration test for client with CLI

donce:
Test script starts-up docker containers and executes "status" command in CLI. "Status: NotConnected" is expected in output.

Sample `integration_tests/test.sh` output:
```
Creating mysterium-broker ... done
Creating mysterium-server-ubuntu ... done
Creating mysterium-client ... done
Creating mysterium-server-ubuntu ...
Creating mysterium-client ...
Full CLI output:
[INFO] Status: NotConnected
[INFO] SID:
Good bye

Status line found:
[INFO] Status: NotConnected
Tests passed
```


integration_tests/test.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+
+docker-compose up -d
+docker cp integration_tests/data/ mysterium-client:/usr/lib/integration_tests
```
Waldz: Does it have to be copied into container?

tadovas: That's strange to copy something into running container at all

donce: Yes - tequilapi binds on localhost, so it is not reachable from host machine.

Waldz: Yes, it is reachable:
- map 4050 to wanted host port 
- in wort case, You can bind on other unix socket (see container environment variables https://github.com/MysteriumNetwork/node/blob/master/bin/client_package/installation/default)

integration_tests/data/test.sh
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
+
+output=`/usr/bin/mysterium_client --cli < /usr/lib/integration_tests/cli-input`
```
Waldz: Nicest way to to fetch Tequila endpoint with `curl -X POST ..:4050/...`

integration_tests/test.sh
tadovas: It's ok to change it to "0.0.0.0" for testing purposes. At least that's why this parameter --tequilapi.address "0.0.0.0" was intended for ;)

donce: Testing environment should as as close to real as possible, but I can't find any better solution to keep that, so we'll have to make it public in tests :/

integration_tests/test.sh
```diff
@@ -0,0 +1,13 @@
...
+result=`curl localhost:4050/connection`
+
+if [ "$result" != '{"status":"NotConnected"}' ]
+then
+    echo "Unexpected status response: $result"
+    exit 1
+fi
+
+echo "Tests passed"
+exit 0
```
tadovas: And where is docker-compose down ? I expect tests to cleanup after themselves

PR #179, opened 2018-02-26
MYST-362 prevent server premature shutdown on missing discovery ping

zolia:


cmd/commands/server/command_server.go
```diff
@@ -100,9 +100,8 @@ func (cmd *Command) Start() (err error) {
 			case <-time.After(1 * time.Minute):
 				err := cmd.mysteriumClient.PingProposal(proposal, signer)
 				if err != nil {
-					//TODO failed to refresh proposal. Stop everything?
 					log.Error("Failed to ping proposal", err)
-					cmd.Kill()
+					// do not stop server on missing ping to discovery. More on this in MYST-362 and MYST-370
```
tadovas: Maybe 'failed' instead of 'missing'. Also do we really need to write ticket numbers here?

PR #180, opened 2018-02-26
MYST-323 Increase stats update frequency

donce:


client/connection/openvpn.go
```diff
@@ -39,8 +39,11 @@ func ConfigureVpnClientFactory(
 			go statsSender(stats)
 			return nil
 		}
-
-		statsHandler := bytescount.NewCompositeStatsHandler(statsSaver, asyncStatsSender)
+		asyncIntervalStatsSender, err := bytescount.NewIntervalStatsHandler(asyncStatsSender, time.Now, 60)
```
tadovas: Maybe asyncIntervalStatsSender -> just intervalStatsSender?

openvpn/middlewares/client/bytescount/selective_stats_handler.go
```diff
@@ -0,0 +1,22 @@
...
+	"github.com/pkg/errors"
+	"time"
+)
+
+// NewIntervalStatsHandler creates and returns composite handler, which invokes internal handler at given interval
+func NewIntervalStatsHandler(handler SessionStatsHandler, clock func() time.Time, interval time.Duration) (SessionStatsHandler, error) {
+	if interval < 0 {
+		return nil, errors.New("Invalid 'interval' parameter")
+	}
+	var lastTime *time.Time = nil
```
tadovas: var lastTime = &time.Time{}. No need to check for nil, also it would be initialized to zero (unix style) and will trigger first call to handler. Maybe even pointer is not needed here.

openvpn/middlewares/client/bytescount/selective_stats_handler.go
```diff
@@ -0,0 +1,22 @@
...
+	return func(sessionStats SessionStats) error {
+		now := clock()
+		if lastTime == nil || (now.Sub(*lastTime)) >= interval {
```
tadovas: Maybe lastTime.Add(interval) < now ?

openvpn/middlewares/client/bytescount/selective_stats_handler.go
```diff
@@ -0,0 +1,22 @@
+package bytescount
+
+import (
+	"github.com/pkg/errors"
+	"time"
+)
+
+// NewIntervalStatsHandler creates and returns composite handler, which invokes internal handler at given interval
+func NewIntervalStatsHandler(handler SessionStatsHandler, clock func() time.Time, interval time.Duration) (SessionStatsHandler, error) {
```
tadovas: clock -> currentTime ? It would look nicer now:=currentTime()

openvpn/middlewares/client/bytescount/selective_stats_handler.go
donce: Not sure if this is better, especially since you've made a mistake in your example: `>=` should be used instead of `<` :D

tadovas: No I didn't.

openvpn/middlewares/client/bytescount/selective_stats_handler.go
donce: Done.

openvpn/middlewares/client/bytescount/selective_stats_handler.go
donce: Yeah, you didn't - it's more confusing than I initially thought :D
Comparing passed time with interval time seems much easier.

openvpn/middlewares/client/bytescount/selective_stats_handler.go
donce: Added boolean var

client/connection/openvpn.go
donce: Done.

PR #187, opened 2018-02-27
Fail fast if could not locate openvpn binary on both client and node

tadovas:


cmd/commands/server/command_server.go
```diff
@@ -34,11 +34,17 @@ type Command struct {
 	vpnServerFactory func(sessionManager session.Manager, serviceLocation dto_discovery.Location,
 		providerID identity.Identity, callback state.Callback) *openvpn.Server
 
-	vpnServer *openvpn.Server
+	vpnServer          *openvpn.Server
+	openvpnBinaryCheck func() error
```
donce: Why not `checkOpenvpn` as in client command?

openvpn/version.go
```diff
@@ -0,0 +1,38 @@
+package openvpn
+
+import (
+	"os/exec"
+	"syscall"
+)
+
+//Version function checks that openvpn is available, given path to openvpn binary
+func Version(openvpnBinary string) error {
```
donce: Function name declares implementation, not the purpose of this function. `CheckOpenVPNBinary` ?

openvpn/version.go
```diff
@@ -0,0 +1,38 @@
...
+	process := NewProcess(openvpnBinary, "[openvpn binary check]")
+	if err := process.Start([]string{"--version"}); err != nil {
+		return err
+	}
+	cmdResult := process.Wait()
+
+	exitCode, err := extractExitCodeFromCmdResult(cmdResult)
+	//openvpn returns exit code 1 in case of --version paramter, if anything else is returned - treat as error
+	if err != nil || exitCode != 1 {
+		return err
```
donce: In case `err == nil` and `exitCode == 0`, you'll return `nil`, which means that this function passes. A separate error should be returned in that case, i.e.
```go
if err != nil {
  return err
}
if exitCode != 1 {
  return errors.New("Oh shit!")
}
```

Maybe this indicates a need for unit-test? ;)

openvpn/version.go
```diff
@@ -0,0 +1,38 @@
...
+func Version(openvpnBinary string) error {
+
+	process := NewProcess(openvpnBinary, "[openvpn binary check]")
+	if err := process.Start([]string{"--version"}); err != nil {
+		return err
+	}
+	cmdResult := process.Wait()
+
+	exitCode, err := extractExitCodeFromCmdResult(cmdResult)
+	//openvpn returns exit code 1 in case of --version paramter, if anything else is returned - treat as error
```
zolia: It will rarely will be something else on error than '1', in theory we should parse and look for version.. but prob, kad skip this for now.. :/

cmd/commands/server/command_server.go
tadovas: Saw this after PR was submitted. Will make the names the same

openvpn/version.go
tadovas: Yes and yes :) it could be unittested

openvpn/version.go
tadovas: That's the idea of this PR - to check that path to openvpn was correct. Exact openvpn version maching and even freeze is out of scope for this PR and I not sure that its needed.
This exitCode == 1 test is simply to disable Error messages that executable exited with status non-zero which does openvpn given version parameter

openvpn/version.go
tadovas: The idea was to make a function whchi returns (string, error) with string as openvpn version and error if exec somehow fails. But for the moment yes - it can be named CheckOpenVpnBinary

tadovas:
All comments have been addressed

openvpn/check.go
```diff
@@ -0,0 +1,45 @@
...
+
+//given error value from cmd.Wait() extract exit code if possible, otherwise returns error itself
+//this is ugly but there is no standart and os independent way to extract exit status in golang
+func extractExitCodeFromCmdResult(cmdResult error) (int, error) {
+	exitError, ok := cmdResult.(*exec.ExitError)
+	if !ok {
+		return 0, cmdResult
+	}
+
+	exitStatus, ok := exitError.Sys().(syscall.WaitStatus)
```
zolia: Question is if it will behave the same for Windows..

tadovas: Confirmed by guys in openstack. Works on windows and linux. And I just tried on MacOS :)

openvpn/testdata/exit-with-one.sh
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
+
+echo "Some output"
+exit 1
```
donce: Newline

openvpn/testdata/exit-with-zero.sh
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
+echo "Some output"
+exit 0
```
donce: Newline.

openvpn/check_test.go
```diff
@@ -0,0 +1,18 @@
...
+	assert.Error(t, CheckOpenvpnBinary("non-existent-binary"))
+}
+
+func TestErrorIsReturnedOnExitCodeZero(t *testing.T) {
+	assert.Error(t, CheckOpenvpnBinary("testdata/exit-with-zero.sh"))
+}
+
+func TestNoErrorIsReturnedOnExitCodeOne(t *testing.T) {
+	assert.NoError(t, CheckOpenvpnBinary("testdata/exit-with-one.sh"))
+}
```
donce: Nice ;)

openvpn/check.go
```diff
@@ -0,0 +1,45 @@
+package openvpn
+
+import (
+	"errors"
+	"os/exec"
+	"strconv"
+	"syscall"
+)
+
+//CheckOpenvpnBinary function checks that openvpn is available, given path to openvpn binary
```
donce: Space after `//`

openvpn/check.go
```diff
@@ -0,0 +1,45 @@
...
+func CheckOpenvpnBinary(openvpnBinary string) error {
```
donce: Shouldn't we use abbrevation naming conventions for `vpn`, i.e. `CheckOpenVPNBinary`?

tadovas: Well its not an abrevation but more well known trademark "openvpn".

openvpn/testdata/exit-with-zero.sh
tadovas: Fixed.

openvpn/testdata/exit-with-one.sh
tadovas: Fixed.

openvpn/check_test.go
tadovas: Fixed.

openvpn/check.go
tadovas: Fixed.

PR #191, opened 2018-03-02
Bugfix/MYST-393

tadovas:


openvpn/middlewares/server/auth/middleware.go
```diff
@@ -37,143 +60,104 @@ func (m *middleware) Stop(commandWriter management.Connection) error {
...
-		if consumed {
-			return m.authenticateClient()
+func (m *middleware) handleClientEvent(event clientEvent) {
+	switch event.eventType {
+	case connect, reauth:
+		username := event.env["username"]
+		password := event.env["password"]
+		err := m.authenticateClient(event.clientID, event.clientKey, username, password)
+		if err != nil {
+			log.Critical("Unable to authenticate client. Error: ", err)
```
zolia: I woudn't say its critical. Critical usually means that application cannot proceed. In this case just 1 client..

PR #192, opened 2018-03-03
build info in struct /healthcheck

shroomist:


version/version.go
```diff
@@ -4,6 +4,12 @@ package version
 
 import "fmt"
 
+type Info struct {
```
donce: Missing doc.

version/version.go
```diff
@@ -17,3 +23,11 @@ var BuildNumber = "dev-build"
 func AsString() string {
 	return fmt.Sprintf("Branch: %s. Build id: %s. Commit: %s", GitBranch, BuildNumber, GitCommit)
 }
+
+func GetInfo() *Info {
```
donce: Missing doc.

tequilapi/endpoints/health_check.go
```diff
@@ -3,28 +3,35 @@ package endpoints
...
+	Branch      string `json:"branch"`
+	BuildNumber string `json:"buildNumber"`
 }
 
 type healthCheckEndpoint struct {
 	startTime       time.Time
 	currentTimeFunc func() time.Time
 	processNumber   int
-	buildVersion    string
+	buildVersion    *version.Info
```
donce: maybe `versionInfo` is more accurate name?

tequilapi/endpoints/health_check.go
```diff
@@ -3,28 +3,35 @@ package endpoints
...
-	Version string `json:"version"`
+	Uptime  string    `json:"uptime"`
+	Process int       `json:"process"`
+	Version buildInfo `json:"version"`
+}
+
+type buildInfo struct {
+	Commit      string `json:"commit"`
+	Branch      string `json:"branch"`
+	BuildNumber string `json:"buildNumber"`
```
donce: `buildInfo` is identical to `version.Info`, except `json` names. Is this intentional to make this module less coupled to `version` package..?

version/version.go
donce: Since we're exposing this method to get version info, can `GitCommit`, `GitBranch` and `BuildNumber` be private? Or do they have to be public because we set them in bash script?

bin/helpers/load_environment.sh
```diff
@@ -4,4 +4,7 @@ if [ ! -f .env ]; then
     printf "\e[0;31m%s\e[0m\n" "Environment file must be set!"
     exit 1
 fi
-source .env
\ No newline at end of file
+source .env
+
+TRAVIS_COMMIT="$(git rev-parse --short HEAD 2> /dev/null | sed "s/\(.*\)/\1/")"
+TRAVIS_BRANCH="$(git symbolic-ref --short -q HEAD)"
```
tadovas: Travis CI has it's own target branch exposing logic. Why are you are overriding this?
At least put both variables in conditional logc if its not set - set to default value

version/version.go
tadovas: The logic should be very simple. Version package provides a few exported variables and a function which concats them to string. Why do we need separate GetInfo function with variables put in structure? Just define structure where you need it - i.e. in healthcheck endpoint

tequilapi/endpoints/health_check.go
shroomist: yes

bin/helpers/functions.sh
```diff
@@ -4,8 +4,8 @@
 function get_linker_ldflags {
     echo -n "-X 'github.com/mysterium/node/cmd.MysteriumAPIURL=${MYSTERIUM_API_URL}' "
     echo -n "-X 'github.com/mysterium/node/cmd/commands/server.natsServerIP=${NATS_SERVER_IP}' "
-    [ -n "$TRAVIS_BRANCH" ] && echo -n "-X 'github.com/mysterium/node/version.GitBranch=${TRAVIS_BRANCH}' "
-    [ -n "$TRAVIS_COMMIT" ] && echo -n "-X 'github.com/mysterium/node/version.GitCommit=${TRAVIS_COMMIT}' "
-    [ -n "$TRAVIS_JOB_NUMBER" ] && echo -n "-X 'github.com/mysterium/node/version.BuildNumber=${TRAVIS_JOB_NUMBER}' "
+    [ -n "$BRANCH" ] && echo -n "-X 'github.com/mysterium/node/version.gitBranch=${BRANCH}' "
+    [ -n "$COMMIT" ] && echo -n "-X 'github.com/mysterium/node/version.gitCommit=${COMMIT}' "
+    [ -n "$TRAVIS_JOB_NUMBER" ] && echo -n "-X 'github.com/mysterium/node/version.buildNumber=${TRAVIS_JOB_NUMBER}' "
```
donce: shouldn't `TRAVIS_JOB_NUMBER` be generalised as well, i.e. to `BUILD_NUMBER`?

tequilapi/endpoints/health_check.go
```diff
@@ -3,28 +3,35 @@ package endpoints
...
-	buildVersion    string
+	versionInfo     *version.Info
 }
 
 /*
 HealthCheckEndpointFactory creates a structure with single HealthCheck method for healthcheck serving as http,
 currentTimeFunc is injected for easier testing
 */
-func HealthCheckEndpointFactory(currentTimeFunc func() time.Time, procID func() int, buildVersion string) *healthCheckEndpoint {
+func HealthCheckEndpointFactory(currentTimeFunc func() time.Time, procID func() int, buildVersion *version.Info) *healthCheckEndpoint {
```
donce: `buildVersion` -> `versionInfo`

version/version.go
```diff
@@ -4,16 +4,32 @@ package version
...
+var gitCommit = "<unknown>"
 
-// GitBranch comes from TRAVIS_BRANCH env variable - if it's github release, this variable will contain release tag name
-var GitBranch = "<unknown>"
+// GitBranch comes from BRANCH env variable - if it's github release, this variable will contain release tag name
+var gitBranch = "<unknown>"
 
 // BuildNumber comes from TRAVIS_JOB_NUMBER env variable
-var BuildNumber = "dev-build"
+var buildNumber = "dev-build"
```
donce: comments of GitCommit, GitBranch and BuildNumber should be lowercased as well.

PR #195, opened 2018-03-08
Feature/enable-e2e-tests-on-pull-requests

tadovas:


e2e/Dockerfile.node
```diff
@@ -0,0 +1,28 @@
+FROM golang:1.9.2-alpine AS builder
```
donce: Why do we need separate `Dockerfile` for e2e?

e2e/connection_test.go
```diff
@@ -0,0 +1,21 @@
...
+	"flag"
+	"github.com/mysterium/node/tequilapi/client"
+	"github.com/stretchr/testify/assert"
+	"testing"
+)
+
+var host = flag.String("tequila.host", "localhost", "Specify tequila host for e2e tests")
+var port = flag.Int("tequila.port", 4050, "Specify tequila port for e2e tests")
+
+func TestClientConnectsToNode(t *testing.T) {
```
donce: Method name is misleading - it doesn't test that.

.travis.yml
```diff
@@ -32,10 +35,15 @@ install:
 
 script:
   - bin/test
-  - bin/client_build
-  - bin/server_build
+  - if [ "$TRAVIS_OS_NAME" == 'linux' ] && [ "$TRAVIS_PULL_REQUEST" != "false" ]; then
```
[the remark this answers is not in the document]
tadovas: From Travis documentatin about $TRAVIS_PULL_REQUEST:
"TRAVIS_PULL_REQUEST: The pull request number if the current job is a pull request, “false” if it’s not a pull request."
So if it's pull request this variable will not contain true, but the PR number

e2e/Dockerfile.node
tadovas: Because we don't have lightweight docker image for building both node/client and ability to start them up.

e2e/connection_test.go
tadovas: Renamed.

e2e/Dockerfile.node
Waldz: I agree, I see this `Dockerfile.node` as duplication of our official image.

e2e/Dockerfile.node
```diff
@@ -0,0 +1,28 @@
...
+# Copy node stuff
+COPY --from=builder /go/src/github.com/mysterium/node/build/server/mysterium_server .
+COPY --from=builder /go/src/github.com/mysterium/node/bin/server_package/config server_config
+
+#Copy client stuff
+COPY --from=builder /go/src/github.com/mysterium/node/build/client/mysterium_client .
+COPY --from=builder /go/src/github.com/mysterium/node/bin/client_package/config client_config
+
+
+ENTRYPOINT ["/mysterium/mysterium_server"]
```
Waldz: You compile 2 executables, but use only `mysterium_server`. Do you have a case then You need run both executables *in same machine*?

e2e/Dockerfile.node
tadovas: Which one? I don't see any official images of building client + node on alpine

e2e/docker-compose.yml
```diff
@@ -0,0 +1,42 @@
...
+      - broker
+      - node
+    privileged: true
+    cap_add:
+      - MKNOD
+      - NET_ADMIN
+    ports:
+      - 4052:4050
+    entrypoint:
+      "/mysterium/mysterium_client"
```
Waldz: We have Dockerfile for client also, just nobody is testing them. Would be nice integration covers existing tools. Because they are breaking from time to time: Dockerfile of server, Dockerfile of client

bin/test
```diff
@@ -11,7 +11,7 @@
 
 ARGUMENTS=$@
 if [ -z "$ARGUMENTS" ]; then
-    ARGUMENTS=`go list ./...`
+    ARGUMENTS=`go list ./... | sed '/e2e/d'` #skip e2e package - integration tests by default
```
Waldz: Go test has some tags feature, would be good to investigate that

e2e/docker-compose.yml
```diff
@@ -0,0 +1,42 @@
...
+      dockerfile: e2e/Dockerfile.node
+    depends_on:
+      - broker
+      - node
+    privileged: true
+    cap_add:
+      - MKNOD
+      - NET_ADMIN
+    ports:
+      - 4052:4050
```
Waldz: Do we really have to take hosts port then running tests? What if tests are run in docker-compose's neighbour container?

.travis.yml
```diff
@@ -32,10 +35,15 @@ install:
...
+      bin/travis_scripts/check_docker.sh;
+      bin/travis_scripts/check_docker_compose.sh;
+      bin/run_e2e_tests;
```
Waldz: `bin/test` and `bin/test_e2e`

bin/run_e2e_tests
```diff
@@ -29,16 +32,16 @@ setup () {
...
 
 setup
 
-result=`curl localhost:4050/connection`
-if [ "$result" != '{"status":"NotConnected"}' ]
+go test -v ./e2e/... -args --tequila.host=localhost --tequila.port=4052
+if [ ! $? -eq 0 ]
 then
-    print_error "Unexpected status response: $result"
+    print_error "Tests failed"
```
Waldz: Why not cleaning containers if tests succeeds?

.travis.yml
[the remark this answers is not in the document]
donce: wow, that is confusing :D

bin/run_e2e_tests
donce: It is being cleanup, look where `cleanup` function is invoked.

e2e/Dockerfile.node
tadovas: I am using the same image to start both server and node container - in that case container is built once, with two execs ready. See docker-compose node and client definitions.

e2e/docker-compose.yml
tadovas: Let's make a clean separation between integration testing and packaging testing

bin/test
tadovas: Tags have some limitations when it comes for full package of tests switching of. Tried that - does not help

e2e/docker-compose.yml
tadovas: Possible, but a bit complicated - at the moment is simplest way to run tests agains full started up system with only *single* port exposed

bin/run_e2e_tests
tadovas: it is:
print_success "Tests passed"
cleanup

bin/travis_scripts/check_docker.sh
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
+set -e
```
zolia: What is the point to use set -e if there is only 1 command?

tadovas: Reserved for future usage :)

e2e/Dockerfile.node
Waldz: `mysteriumnetwork/mysterium-node:alpine`
`mysteriumnetwork/mysterium-client:alpine`

bin/test
Waldz: Anyway would be nice to filter by full package name

bin/travis_scripts/check_docker.sh
tadovas: Fixed

bin/test
tadovas: Yea will need to investigate further - maybe some other time.

e2e/Dockerfile.node
tadovas: Ha ha, as they are commited one hour ago :) I will use them - thanks

bin/helpers/load_environment.sh
```diff
@@ -5,13 +5,3 @@ if [ ! -f .env ]; then
     exit 1
 fi
 source .env
-
-COMMIT="$TRAVIS_COMMIT"
```
Waldz: Should we move these to .evn_example maybe?

e2e/Dockerfile.node
Waldz: This `Dockerfile.node` is unused now

tadovas: Removed

bin/helpers/load_environment.sh
tadovas: Well they are optional and travis specific anyway. Should we trash that .env file even more?

PR #197, opened 2018-03-12
MYST-413 Fix broken Alpine image

Waldz:


bin/client_docker/alpine/Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+FROM alpine:3.6
+MAINTAINER Valdas Petrulis <petrulis.valdas@gmail.com>
+
+# Install packages
+RUN apk add --update --no-cache bash iptables ca-certificates openvpn \
+    && rm -rf /var/cache/apk/*
+
+# Install application
+COPY bin/client_docker/docker-entrypoint.sh /usr/local/bin/docker-entrypoint.sh
+ENTRYPOINT ["/usr/local/bin/docker-entrypoint.sh"]
```
tadovas: From docker perspective it's not very efficient to put executable AND it's args into entry point, as it will be very hard to override params later. It should be like this:
ENTRYPOINT ["your exec and args which must be always present and never change"]
CMD["default additional args which can be customized when using docker run <image> <args> style"]
those could be tequila.port , address , discovery address etc.
Do we really even need docker-entrypoint.sh ? it hides important stuff, also uses env vars internally and it's hard to track what is going on

docker-compose.yml
```diff
@@ -10,6 +10,25 @@ services:
...
+    tty: true
+    build:
+      context: .
+      dockerfile: bin/server_docker/alpine/Dockerfile
+    depends_on:
+      - broker
+    cap_add:
+      - MKNOD
+      - NET_ADMIN
+    networks:
```
tadovas: You don't need to force use default network. Each docker compose creates and maintains it's own network automatically

bin/client_docker/alpine/Dockerfile
Waldz: Good ideas, will depreciate Docker environments variables in upcoming PRs

bin/client_build
```diff
@@ -13,7 +13,7 @@
 #> GOOS=windows GOARCH=amd64 bin/client_build
 #
 # Check if program has dynamic libraries:
-#> readelf -d /build/client/mysterium_client
+#> readelf -d build/client/mysterium_client
```
zolia: There are no readelf on osx by default. We should add help notice how to get it.

docker-compose.yml
Waldz: Removed network injection

bin/client_build
Waldz: Fixed

PR #202, opened 2018-03-12
MYST-413 Optimize dockers to have more cached layers

Waldz:


tadovas:
What about alpine image building? Needs to be migrated to our builder too?

Waldz:
Had same idea, but alpine currently compiled in alpine, and builder is Ubuntu image

bin/client_package/config/update-resolv-conf
```diff
@@ -20,7 +17,11 @@ NMSRVRS=()
 SRCHS=()
 adapters=()
 
+# TODO Investigate DNS on Linux environment
 NETWORKSETUP=/usr/sbin/networksetup
+if [ ! -x "$NETWORKSETUP" ]; then
```
tadovas: NETWORKSETUP=(which networksetup)
if [ - "$NETWORKSETUP" ]; then //not OSX
don't rely on hardcoded path - maybe we decide to bundle our own

Waldz: Removed my logic, waiting for your fix

PR #203, opened 2018-03-13
Hotfix/enable slack notifications

tadovas:
Currently I didn't find a clean way to trigger notifications on master branch only.
The behaviour is the following:
1. On failing builds - always notify
2. On success builds - only notify when status changes i.e. from success to fail and vice versa
3. Less commits with failing builds - less spam :)


Waldz:
14 commits :)

PR #204, opened 2018-03-14
Feature/implement e2e client connects to node test

tadovas:


.gitignore
```diff
@@ -3,3 +3,4 @@
 /vendor
 /.env
 testdataoutput
+e2e_tests.log
```
donce: newline! :D

tadovas: Not again :) fixed.

bin/client_package/config/update-resolv-conf
```diff
@@ -13,6 +13,9 @@
 # copy update-resolv-conf to myst client current directory
 # build/client/ (by default)
 
+#This check helps to avoid running this script in linux (win :)) based environments
+[[ $OSTYPE = "darwin"* ]] || exit 0
```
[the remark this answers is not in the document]
tadovas: It's a special case when expanding os type and making string match (wildchar at the end) so yes this expression is valid.

bin/client_docker/docker-entrypoint.sh
```diff
@@ -5,10 +5,13 @@ OS_DIR_CONFIG="/etc/mysterium-client"
 OS_DIR_DATA="/var/lib/mysterium-client"
 OS_DIR_RUN="/var/run/mysterium-client"
 
+mkdir -p $OS_DIR_RUN
```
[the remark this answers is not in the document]
tadovas: Done

e2e/connection_test.go
```diff
@@ -2,20 +2,76 @@ package e2e
...
 	status, err := tequilApi.Status()
 	assert.NoError(t, err)
 	assert.Equal(t, "NotConnected", status.Status)
+
+	identities, err := tequilApi.GetIdentities()
+	assert.NoError(t, err)
+
+	var identity client.IdentityDTO
+	if len(identities) < 1 {
+		identity, err = tequilApi.NewIdentity("")
```
Waldz: It's 2nd green path. Should not we have 2 separate tests for that?

tadovas: It's not green path. It's ensurement that we have available identity to use for connection - test cannot proceed without that.

bin/run_e2e_tests
```diff
@@ -17,17 +17,33 @@ print_error () {
...
+    if [ ! $? -eq 0 ]; then
+        print_error "Db startup failed"
+        cleanup
+        exit 1
+    fi
+
     ${dockerComposeCmd} build
     if [ ! $? -eq 0 ]; then
         print_error "Building docker images failed"
         exit 1
```
donce: Cleanup missing.

bin/run_e2e_tests
```diff
@@ -17,17 +17,33 @@ print_error () {
...
 
 setup () {
+    ${dockerComposeCmd} up -d db # start database first - it takes about 10 sec untils db startsup, and otherwise db migration fails
+    if [ ! $? -eq 0 ]; then
+        print_error "Db startup failed"
+        cleanup
+        exit 1
+    fi
+
     ${dockerComposeCmd} build
```
donce: What if images are already built, won't this finish instantly while database is not set-up yet?

bin/run_e2e_tests
```diff
@@ -17,17 +17,33 @@ print_error () {
...
     if [ ! $? -eq 0 ]; then
         print_error "Building docker images failed"
         exit 1
     fi
 
+    ${dockerComposeCmd} run --entrypoint bin/db-upgrade discovery
+    if [ ! $? -eq 0 ]; then
+        print_error "Db migration failed"
+        cleanup
+        exit 1
```
donce: These 3 lines repeats on each step - we can make a function `fatal(error_message)` to DRY this.

cmd/commands/cli/command_cli.go
```diff
@@ -226,10 +226,8 @@ func (c *Command) proposals() {
 	for _, proposal := range proposals {
 		country := proposal.ServiceDefinition.LocationOriginate.Country
 		var countryString string
-		if country != nil {
-			countryString = *country
-		} else {
-			countryString = "Unknown"
+		if len(country) == 0 {
```
donce: I'm confusied - Isn't `countryString` left empty if `len(country)` != 0?

tadovas: Fixed

bin/run_e2e_tests
tadovas: It's bash - next time

tequilapi/client/dto.go
```diff
@@ -25,14 +27,18 @@ type ProposalDTO struct {
 	ServiceDefinition ServiceDefinitionDTO `json:"serviceDefinition"`
 }
 
+func (p ProposalDTO) String() string {
+	return fmt.Sprintf("Id: %d , Provider: %s, Country: %s", p.ID, p.ProviderID, p.ServiceDefinition.LocationOriginate.Country)
```
donce:
```go
p.ServiceDefinition.LocationOriginate.Country
```
This is very long call chain and it seems to be repeated everywhere we need country. Perhaps service proposal or service definition could have `country` shortcut method?

PR #206, opened 2018-03-14
MYST-420 / manager.Connect now sets itself into Connecting state

ignasbernotas:


tadovas:
What about a test case -> trying to connect, errors occurs - state reverts to not connected ?

client/connection/manager.go
```diff
@@ -43,11 +43,18 @@ func NewManager(mysteriumClient server.Client, dialogEstablisherCreator DialogEs
...
-func (manager *connectionManager) Connect(consumerID, providerID identity.Identity) error {
+func (manager *connectionManager) Connect(consumerID, providerID identity.Identity) (err error) {
 	if manager.status.State != NotConnected {
 		return ErrAlreadyExists
 	}
 
+	manager.status = statusConnecting()
+	defer func() {
+		if err != nil {
+			manager.status = statusNotConnected()
```
Waldz: Would be nice this assert this status in tests

ignasbernotas: It was already covered, but I cleaned up the tests

ignasbernotas: @tadovas @Waldz

client/connection/manager.go
```diff
@@ -43,11 +43,18 @@ func NewManager(mysteriumClient server.Client, dialogEstablisherCreator DialogEs
 	}
 }
 
-func (manager *connectionManager) Connect(consumerID, providerID identity.Identity) error {
+func (manager *connectionManager) Connect(consumerID, providerID identity.Identity) (err error) {
```
donce: It seems this method always returns error explicily (i.e. `return err`). Why do we need this var here?

ignasbernotas: Due to the deferred closure, it needs to know whether there was an error

client/connection/manager.go
Waldz: Anyway missing test about status Connecting

tadovas: Added test case. @Waldz  up to you now.

PR #208, opened 2018-03-16
new server option to specify protocol

zolia:


nat/service_iptables.go
```diff
@@ -43,6 +45,17 @@ func (service *serviceIPTables) Stop() error {
 }
 
 func (service *serviceIPTables) enableIPForwarding() (err error) {
+	out, err := exec.Command("sysctl", "-n", "net.ipv4.ip_forward").CombinedOutput()
```
Waldz: Would be nice to that `enableIPForwarding` would have 1 responsibility.
Suggest to move this logic to another function e.g. `isIPForwardingEnabled()`

bin/server_package/installation/default
```diff
@@ -17,3 +17,6 @@ MYSTERIUM_DISCOVERY_ADDRESS: ""
 
 # Mysterium broker address (eg.: testnet.mysterium.network)
 MYSTERIUM_BROKER_ADDRESS: ""
+
+# Mysterium server protocol (options: {udp, tcp})
+MYSTERIUM_SERVER_PROTOCOL: ""
```
Waldz: Wrong syntax here `VAR: "value"` should be `VAR="value"`

Waldz: But @tadovas is dropping container ENV variables, dont spent time on introducing new ones

cmd/commands/server/options.go
```diff
@@ -104,6 +106,13 @@ func ParseArguments(args []string) (options CommandOptions, err error) {
 		"Address (URL form) of ipify service",
 	)
 
+	flags.StringVar(
+		&options.Protocol,
+		"protocol",
```
Waldz: I would go with configuring style like this:
```
mysterium_server 
  --identity=..
  --openvpn.binary=openvpn
  --openvpn.proto=tcp
```

tadovas: Absolutely, lets look at options as variables. "protocol" is global and doesn't say anything about context it is used. openvpn.protocol cleary states that this option is for openvpn and has something to do with openvpn connect type

zolia: We might need to have it global, bus since we don't have other transports yet, will do as you suggest.

nat/service_iptables.go
```diff
@@ -42,7 +44,32 @@ func (service *serviceIPTables) Stop() error {
...
 
+
+func (service *serviceIPTables) isIPForwardingEnabled() (enabled bool, err error) {
+	out, err := exec.Command("sysctl", "-n", "net.ipv4.ip_forward").CombinedOutput()
+	if err != nil {
+		return false, fmt.Errorf("Failed to check IP forwarding status: %s", err)
+	}
+
+	if strings.TrimSpace(string(out)) == "1" {
+		service.forward = true
```
Waldz: Getter function `IsEnabled()` should only read, but not modify state

PR #210, opened 2018-03-19
removed privileged and mknod cap; docker entrypoints reworked

zolia:


tadovas:
Are you sure? 80% of the changes reverts alpine images and env variables. What was the intention?

bin/client_docker/alpine/Dockerfile
```diff
@@ -19,17 +19,12 @@ RUN apk add --update --no-cache iptables ca-certificates openvpn bash \
...
 COPY bin/client_package/config /etc/mysterium-client
 COPY --from=builder /go/src/github.com/mysterium/node/build/client/mysterium_client /usr/bin/mysterium_client
 
 WORKDIR /var/run/mysterium-client
-
-ENTRYPOINT [ "/usr/bin/mysterium_client", \
-    "--config-dir=/etc/mysterium-client", \
-    "--data-dir=/var/lib/mysterium-client", \
-    "--runtime-dir=/var/run/mysterium-client", \
-    "--tequilapi.address=0.0.0.0" ]
```
donce: Why are these params not needed anymore?

zolia:
We cannot move device creation into image, as container reuses hosts's /dev tree and does not have its own. Leaving as is.

tadovas:
Ok so basically what I found on the internet is:
1. docker containers have mknod cap by default
2. without priviledged flag, docker /dev mount goes through cgroups protection and therefore host device list is not accessible
So now it clear - we need to setup /dev/net/tun on entrypoint

bin/client_docker/prepare-run-env.sh
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
```
Waldz: Looks like same script duplicated twice

bin/server_package_docker
```diff
@@ -21,7 +21,6 @@ DOCKER_IMAGE="mysteriumnetwork/mysterium-node"
 printf "Building Alpine image..\n" \
     && docker build \
         --file bin/server_docker/alpine/Dockerfile \
-        --build-arg PACKAGE_VERSION=$VERSION \
```
Waldz: This is build arguments, ant it's still required by Ubuntu's Dockerfile

PR #212, opened 2018-03-29
Feature/MYST-431-make-connect-method-cancelable

tadovas:


utils/cancelable.go
```diff
@@ -0,0 +1,114 @@
...
+// ErrRequestCancelled error is returned by CancelableAction.Call() method when Cancel() method was called
+var ErrRequestCancelled = errors.New("request was cancelled")
+
+// ErrUndefinedRequest error is returned by CancelableAction.Call() by default when BlockingOperation was not defined
+// ( CancelableAction.Request(BlockingOperation) was not called )
+var ErrUndefinedRequest = errors.New("undefined request called")
+
+// SkipOnError decorator takes callback function with single val as parameter and returns CleanupCallback function, which calls
+// given callback only if no error was returned by blocking operation (i.e. good when cleanup is need only when blocking operation succeded but was Cancelled before)
+func SkipOnError(callback func(interface{})) CleanupCallback {
```
donce: First though when reading hame is that this is empty cleanup callback - it simple does nothing (skips callback).
Maybe `InvokeOnSuccess`?

utils/cancelable.go
```diff
@@ -0,0 +1,114 @@
...
+		requestAction: undefinedRequest,
+		cleanupAction: noCleanup,
+		cancelAction: CallOnce(func() {
+			close(channel)
+		}),
+	}
+}
+
+// Request method takes long running operation function as parameter and returns copy of CancelableAction as a result
+func (c CancelableAction) Request(method BlockingOperation) CancelableAction {
```
donce: `Request` and `Cleanup` names seem like they perform request and cleanup - when in fact they don't, they just set the values. What about `SetRequestAction`or `OnRequest`? Or maybe builder pattern is better here, extracting these methods to a separate class which is responsible for building an object will `Request` and `Cleanup` already set?

utils/cancelable.go
```diff
@@ -0,0 +1,114 @@
...
+}
+
+type actionResult struct {
+	val interface{}
+	err error
+}
+
+type actionResultChannel chan actionResult
+
+func callBlockingAction(blockingOp BlockingOperation, cleanup CleanupCallback, cancel CancelChannel) (interface{}, error) {
```
donce: Maybe `cancel` should be named `cancelled`, since channel on this side is used for checking state, not as an action?

client/connection/manager.go
```diff
@@ -55,33 +68,72 @@ func (manager *connectionManager) Connect(consumerID, providerID identity.Identi
...
+		log.Info(managerLogPrefix, "Closing active connection")
+		manager.status = statusDisconnecting()
+		cancelable.Cancel()
+	})
+
+	val, err := cancelable.
+		Request(func() (interface{}, error) {
+			return manager.findProposalByProviderID(providerID)
+		}).
+		Call()
```
donce: Looks very unusual to keep `.` in the previous line - is it a common practice in Go?
I wasn't able to find these usages because I was looking for `.Call()` :D

client/connection/manager.go
```diff
@@ -55,33 +68,72 @@ func (manager *connectionManager) Connect(consumerID, providerID identity.Identi
 		}
 	}()
 
-	proposal, err := manager.findProposalByProviderID(providerID)
+	cancelable := utils.NewCancelable()
+	manager.closeAction = utils.CallOnce(func() {
```
donce: `manager.closeAction` should be set before changing `manager.status = statusConnecting()` - that way we'll prevent a gap when we are in 'connecting' state but `closeAction` is not yet created, which will result in `warnOnClose`

client/connection/manager.go
```diff
@@ -2,44 +2,57 @@ package connection
...
-	status        ConnectionStatus
-	dialog        communication.Dialog
-	openvpnClient openvpn.Client
-	sessionID     session.SessionID
+	status      ConnectionStatus
+	closeAction func()
+}
+
+func warnOnClose() {
+	log.Warn(managerLogPrefix, "WARNING! Trying to close when there is nothing to close. Possible bug or race condition")
```
donce: We don't need to start each warning with "WARNING!" - `log.Warn` is responsible to show that this is a warning :D

client/connection/manager_test.go
```diff
@@ -78,52 +93,61 @@ func (tc *testContext) TestWithUnknownProviderConnectionIsNotMade() {
...
+		errorChannel <- tc.connManager.Connect(myID, activeProviderID)
+	}()
+
+	waitABit()
+	select {
+	case err := <-errorChannel:
+		assert.Fail(tc.T(), "Connect error not expected: ", err)
+	default:
+		assert.Equal(tc.T(), statusConnecting(), tc.connManager.Status())
+	}
```
donce: We can use boolean for that - getting boolean value from variable and setting it are atomic operations AFAIK.
```go
func (tc *testContext) TestStatusReportsConnectingWhenConnectionIsInProgress() {
	tc.fakeOpenVpn.onStartReportStates = []openvpn.State{}

	connected := false
	go func() {
		tc.connManager.Connect(myID, activeProviderID)
		connected = true
	}()

	waitABit()
	assert.Equal(tc.T(), statusConnecting(), tc.connManager.Status())
	assert.False(tc.T(), connected)
}
```

donce:
```go
func (tc *testContext) TestStatusReportsConnectingWhenConnectionIsInProgress() {
	tc.fakeOpenVpn.onStartReportStates = []openvpn.State{}

	go func() {
		tc.connManager.Connect(myID, activeProviderID)
		assert.Fail(tc.T(), "failed")
	}()

	waitABit()
	assert.Equal(tc.T(), statusConnecting(), tc.connManager.Status())
}
```

client/connection/manager_test.go
```diff
@@ -78,52 +93,61 @@ func (tc *testContext) TestWithUnknownProviderConnectionIsNotMade() {
...
+	errorChannel := make(chan error, 1)
+
+	go func() {
+		errorChannel <- tc.connManager.Connect(myID, activeProviderID)
+	}()
+
+	waitABit()
+	select {
+	case err := <-errorChannel:
+		assert.Fail(tc.T(), "Connect error not expected: ", err)
```
donce: Invalid params - Fail does not accept `error` params.

cmd/commands/client/command_client.go
```diff
@@ -42,8 +43,9 @@ func NewCommandWith(
 
 	identityManager := identity.NewIdentityManager(keystoreInstance)
 
-	dialogEstablisherFactory := func(myID identity.Identity) communication.DialogEstablisher {
-		return nats_dialog.NewDialogEstablisher(myID, identity.NewSigner(keystoreInstance, myID))
+	dialogFactory := func(consumerID, providerID identity.Identity, contact dto.Contact) (communication.Dialog, error) {
+		dialogEstablisher := nats_dialog.NewDialogEstablisher(consumerID, identity.NewSigner(keystoreInstance, consumerID))
```
Waldz: Why new `dialogEstablisher` struct is created for each dialog?

client/connection/manager.go
```diff
@@ -2,44 +2,57 @@ package connection
...
-		newDialogEstablisher: dialogEstablisherCreator,
-		newVpnClient:         vpnClientCreator,
-		statsKeeper:          statsKeeper,
-		status:               statusNotConnected(),
+		mysteriumClient: mysteriumClient,
+		newDialog:       dialogCreator,
+		newVpnClient:    vpnClientCreator,
+		statsKeeper:     statsKeeper,
+		status:          statusNotConnected(),
+		closeAction:     warnOnClose,
```
Waldz: What about `connectionTeardown`, `cleanConnection`?

client/connection/manager.go
```diff
@@ -55,33 +68,72 @@ func (manager *connectionManager) Connect(consumerID, providerID identity.Identi
...
 	}
+	proposal := val.(*dto.ServiceProposal)
 
-	dialogEstablisher := manager.newDialogEstablisher(consumerID)
-	manager.dialog, err = dialogEstablisher.EstablishDialog(providerID, proposal.ProviderContacts[0])
+	val, err = cancelable.
+		Request(func() (interface{}, error) {
+			return manager.newDialog(consumerID, providerID, proposal.ProviderContacts[0])
+		}).
+		Cleanup(utils.SkipOnError(func(val interface{}) {
```
Waldz: This error skipping looks complicated:
```
Cleanup(utils.SkipOnError(...))
```

`cancelable.Done()` - executes on success only
`cancelable.Always()` - executes on errors also

client/connection/manager.go
```diff
@@ -2,44 +2,57 @@ package connection
...
 	//these are populated by Connect at runtime
-	status        ConnectionStatus
-	dialog        communication.Dialog
-	openvpnClient openvpn.Client
-	sessionID     session.SessionID
+	status      ConnectionStatus
+	closeAction func()
+}
+
+func warnOnClose() {
```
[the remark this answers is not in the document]
tadovas: Moved to the bottom

utils/cancelable.go
tadovas: Nice naming. Fixed

utils/cancelable.go
tadovas: Good point. Renamed

client/connection/manager.go
tadovas: Yes that go syntax style. Otherwise - syntax error :)

client/connection/manager.go
tadovas: yes there is a tiny possibility. But mutexes and locking would be better. Lets leave it to other PR

client/connection/manager.go
```diff
@@ -123,3 +161,83 @@ func (manager *connectionManager) findProposalByProviderID(providerID identity.I
...
+	<-cancelRequest
+	log.Debug(managerLogPrefix, "Stopping openvpn client")
+	err := openvpnClient.Stop()
+	if err != nil {
+		log.Error(managerLogPrefix, "Failed to stop openvpn client: ", err)
+	}
+}
+
+func openvpnClientWaiter(openvpnClient openvpn.Client, dialog communication.Dialog) {
+	defer dialog.Close()
```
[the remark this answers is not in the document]
tadovas: Fixed

client/connection/manager.go
```diff
@@ -123,3 +161,83 @@ func (manager *connectionManager) findProposalByProviderID(providerID identity.I
...
+}
+
+func (manager *connectionManager) consumeOpenvpnStates(stateChannel <-chan openvpn.State, sessionID session.SessionID) {
+	func() {
+		for state := range stateChannel {
+			manager.onStateChanged(state, sessionID)
+		}
+		manager.status = statusNotConnected()
+		log.Debug(managerLogPrefix, "State updater stopped")
+	}()
```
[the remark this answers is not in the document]
tadovas: Removed

client/connection/manager.go
```diff
@@ -55,33 +68,72 @@ func (manager *connectionManager) Connect(consumerID, providerID identity.Identi
...
-	if err = manager.openvpnClient.Start(); err != nil {
-		manager.dialog.Close()
+	err = manager.waitForConnectedState(stateChannel, vpnSession.ID, cancelable.Cancelled)
+	if err != nil {
+		dialog.Close()
+		openvpnClient.Stop()
 		return err
 	}
+
+	go openvpnClientStopper(openvpnClient, cancelable.Cancelled)
```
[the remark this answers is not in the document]
tadovas: Refactored. However - discussable. Do we have a better readability?

client/connection/manager.go
tadovas: Removed

client/connection/manager_test.go
tadovas: Removed

client/connection/manager.go
tadovas: Now it's Cleanup(utils.InvokeOnSuccess(...)) because it's not a part of cancelable request, but a function decorator which invoces callback only if error == nil

cmd/commands/client/command_client.go
```diff
@@ -42,8 +43,9 @@ func NewCommandWith(
 
 	identityManager := identity.NewIdentityManager(keystoreInstance)
 
-	dialogEstablisherFactory := func(myID identity.Identity) communication.DialogEstablisher {
-		return nats_dialog.NewDialogEstablisher(myID, identity.NewSigner(keystoreInstance, myID))
+	dialogFactory := func(consumerID, providerID identity.Identity, contact dto.Contact) (communication.Dialog, error) {
```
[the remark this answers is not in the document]
tadovas: This dialogCreator is called from client connection manager, where consumer and provider is always know. Simply do one function call to create dialog instead of two inside connection manager

cmd/commands/client/command_client.go
tadovas: That's because when connecting we can provide different identities

client/connection/manager.go
tadovas: I like cleanConnection more maybe. Renamed

openvpn/testdata/100-milisec-process.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+
+echo "Starting"
+sleep 0.2
```
donce: why is `100-milisec-process` doing 200ms sleep? :D

openvpn/process.go
```diff
@@ -9,16 +9,18 @@ import (
...
 		logPrefix:          logPrefix,
 		openvpnBinary:      openvpnBinary,
-		cmdExitError:       make(chan error),
+		cmdExitError:       make(chan error, 1), //channel should have capacity to hold single process exit error
 		cmdShutdownStarted: make(chan bool),
 		cmdShutdownWaiter:  sync.WaitGroup{},
 	}
 }
 
+// Process struct define process wrapper which handles clean shutdown, tracks executable exit errors, etc.
```
donce: defines*

openvpn/check.go
```diff
@@ -10,7 +10,7 @@ import (
 // CheckOpenvpnBinary function checks that openvpn is available, given path to openvpn binary
 func CheckOpenvpnBinary(openvpnBinary string) error {
 
-	process := NewProcess(openvpnBinary, "[openvpn binary check]")
+	process := NewProcess(openvpnBinary, "[openvpn binary check] ")
```
donce: This is out of scope of this PR, but I hate that we're doing this prefix formatting of `"[%s] "` in so many places - it should be done in one place :/

tadovas: Well [] it's just nice surround. I don't see any problem not to use [ ] at all.

openvpn/process.go
tadovas: Fixed

openvpn/testdata/100-milisec-process.sh
tadovas: Nice catch :) fixed

openvpn/check.go
donce: Using it is good, it increases readability - but I think this formatting should be done in one place so that we could easily change that. Not this formatting is duplicated across whole system.

client/connection/manager.go
Waldz: I know pattern is applied, but still complicated

client/connection/manager_test.go
```diff
@@ -189,22 +190,27 @@ func TestConnectionManagerSuite(t *testing.T) {
...
 }
 
 func (foc *fakeOpenvpnClient) Stop() error {
+	close(foc.stopProcess)
 	return nil
 }
 
 func (foc *fakeOpenvpnClient) reportState(state openvpn.State) {
 	foc.stateCallback(state)
+	time.Sleep(time.Millisecond)
```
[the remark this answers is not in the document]
tadovas: Removed - not needed anyway. Every test itself now does waiting for state changes as long as its needed.

client/connection/manager.go
```diff
@@ -123,3 +170,71 @@ func (manager *connectionManager) findProposalByProviderID(providerID identity.I
...
+	}
+
+	return openvpnClient, nil
+}
+
+func (manager *connectionManager) waitForConnectedState(stateChannel <-chan openvpn.State, sessionID session.SessionID, cancelRequest utils.CancelChannel) error {
+
+	for {
+		select {
+		case state := <-stateChannel:
```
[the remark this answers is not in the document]
tadovas: Fixed. read state and also check if channel was closed explicitly

client/connection/manager.go
```diff
@@ -55,33 +64,81 @@ func (manager *connectionManager) Connect(consumerID, providerID identity.Identi
...
 		return err
 	}
+
+	manager.cleanConnection = func() {
+		log.Info(managerLogPrefix, "Closing active connection")
+		manager.status = statusDisconnecting()
+		if err := openvpnClient.Stop(); err != nil {
+			log.Warn(managerLogPrefix, "Openvpn process stopped with error: ", err)
+		} else {
+			log.Info(managerLogPrefix, "Openvpn client stopped")
```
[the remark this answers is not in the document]
tadovas: Fixed

client/connection/manager.go
tadovas: either if err == nil then do cleanup or function decorator. I would stay with decorator

PR #213, opened 2018-03-30
add UnregisterProposal

antaveiv:
more unit tests in progress


cmd/commands/server/command_server.go
```diff
@@ -132,10 +133,24 @@ func (cmd *Command) Wait() error {
...
-	err := cmd.dialogWaiter.Stop()
+
+	providerId, err := cmd.identityLoader()
+	if err != nil {
+		return err
+	}
+
+	signer := cmd.createSigner(providerId)
+
+	err = cmd.mysteriumClient.UnregisterProposal(providerId.Address, signer)
```
tadovas: You don't need to do anything in this place. Kill command signals openvpn process to exit, openvpn process reports exiting state, and then pinger is stopped and proposal unregistered.

cmd/commands/server/command_server.go
```diff
@@ -93,15 +95,14 @@ func (cmd *Command) Start() (err error) {
 		case openvpn.ExitingState:
 			log.Info("Open vpn service exiting")
 			close(stopPinger)
+			cmd.mysteriumClient.UnregisterProposal(providerID.Address, signer)
```
tadovas: Don't swallow possibly returned error - at least log it.

server/dto/node_unregister_request.go
```diff
@@ -0,0 +1,6 @@
+package dto
+
+type NodeUnregisterRequest struct {
```
tadovas: ProposalUnregisterRequest ?

Waldz: `NodeUnregisterRequest` is ok also, it's more consistent with remote endpoint naming

server/interface.go
```diff
@@ -12,6 +12,7 @@ type Client interface {
 
 	FindProposals(providerID string) (proposals []dto_discovery.ServiceProposal, err error)
 	RegisterProposal(proposal dto_discovery.ServiceProposal, signer identity.Signer) (err error)
+	UnregisterProposal(providerId string, signer identity.Signer) (err error)
```
Waldz: Be consistent with another method in interface `PingProposal(proposal)`:
- pass full `proposal` structure
- real adapted takes that he needs for unregistration DTO

cmd/commands/server/command_server.go
Waldz: Also this unregistration+logging is duplicated. I would like same cleanup function here e.g. `cmd.Kill()`

cmd/commands/server/command_server.go
```diff
@@ -132,10 +133,24 @@ func (cmd *Command) Wait() error {
 // Kill stops server
 func (cmd *Command) Kill() error {
 	cmd.vpnServer.Stop()
-	err := cmd.dialogWaiter.Stop()
+
+	providerId, err := cmd.identityLoader()
```
Waldz: Identity is loaded in startup, loading 2nd time looks too excess

cmd/commands/server/command_server.go
Waldz: Should we depend unregistration after openvpn exit? I think unregistration can be ran separately even before openvpn exit

server/mysterium_api_fake.go
```diff
@@ -29,6 +29,28 @@ func (client *ClientFake) RegisterProposal(proposal dto_discovery.ServiceProposa
 	return nil
 }
 
+//UnregisterProposal unregisters a service proposal when client disconnects
+func (client *ClientFake) UnregisterProposal(providerID string, signer identity.Signer) error {
+	matchingProposals, err := client.FindProposals(providerID)
```
Waldz: I suggest not to couple on function `FindProposals`, just traverse `client.proposalsMock` and delete matched values

cmd/commands/server/command_server.go
Waldz: But mostly I miss tests which imitates all those cases we talk here

tadovas: Actually we should call unregister AFTER pinger is stopped. So basically the best place to put Unregister is on command_server.go:117, where stopPinger channel case is handled.

Waldz: So it is:
```
stop service (vpnServer, dialogWaiter, natService) -> stop pinger -> unregister proposal
```

server/dto/node_unregister_request.go
tadovas: Fine by me , until this structure not leaked anywhere else out of mysterium api scope

cmd/commands/server/command_server_test.go
```diff
@@ -0,0 +1,38 @@
...
+	myID                  = identity.FromAddress("identity-1")
+	activeProviderID      = identity.FromAddress("vpn-node-1")
+	activeProviderContact = dto_discovery.Contact{}
+	activeProposal        = dto_discovery.ServiceProposal{
+		ProviderID:       activeProviderID.Address,
+		ProviderContacts: []dto_discovery.Contact{activeProviderContact},
+	}
+)
+
+func (tc *testContext) TestProposalUnregisteredWhenPingerClosed() {
```
tadovas: Simple TestXXX(t * testing.T) is enough here.
testContext structure is needed when you want to use Setup/Teardown callbacks before each test or test suite.

cmd/commands/server/command_server_test.go
```diff
@@ -0,0 +1,38 @@
...
+
+import (
+	"github.com/stretchr/testify/suite"
+	"github.com/mysterium/node/server"
+	"github.com/mysterium/node/identity"
+	dto_discovery "github.com/mysterium/node/service_discovery/dto"
+	"github.com/stretchr/testify/assert"
+)
+
+type testContext struct {
```
tadovas: This structure is not needed

cmd/commands/server/command_server.go
```diff
@@ -132,11 +119,34 @@ func (cmd *Command) Wait() error {
...
 		return err
 	}
+
 	err = cmd.natService.Stop()
 
 	return err
 }
+
+// Ping proposal until stopped, then unregister proposal
+func PingProposalLoop(proposal dto_discovery.ServiceProposal, mysteriumClient  server.Client, signer identity.Signer, stopPinger <- chan int) {
```
tadovas: This function is used inside package only - it can be private

Waldz: It's internal function, no point to make it public

cmd/commands/server/command_server_test.go
```diff
@@ -0,0 +1,38 @@
...
+	fakeDiscoveryClient.RegisterProposal(activeProposal, nil)
+
+	go PingProposalLoop(activeProposal, fakeDiscoveryClient, nil, stopPinger)
+
+	close(stopPinger) //causes proposal to be unregistered
+
+	proposals, err := fakeDiscoveryClient.FindProposals(activeProposal.ProviderID)
+
+	assert.NoError(tc.T(), err)
+	assert.Empty(tc.T(), proposals)
```
Waldz: `Empty` checks if object is empty, I would rather check length of proposals array

cmd/commands/server/command_server_test.go
```diff
@@ -0,0 +1,38 @@
...
+
+type testContext struct {
+	suite.Suite
+}
+
+var (
+	myID                  = identity.FromAddress("identity-1")
+	activeProviderID      = identity.FromAddress("vpn-node-1")
+	activeProviderContact = dto_discovery.Contact{}
+	activeProposal        = dto_discovery.ServiceProposal{
```
Waldz: You dont need so rich proposal, having `dto_discovery.ServiceProposal{}` would be enought

tadovas:
Closed as irrelevant

Waldz:
Comments fixed in #232

PR #217, opened 2018-04-09
Disable caching on all tequilapi responses

tadovas:


tequilapi/http_middlewares.go
```diff
@@ -40,3 +42,26 @@ func generatePreflightResponse(req *http.Request, resp http.ResponseWriter) {
...
 }
+
+const cacheControlHeader = "Cache-control"
+const noStore = "no-store"
+const noCache = "no-cache"
+const mustRevalidate = "must-revalidate"
+
+type cacheControl struct {
+	originalHandler http.Handler
+}
```
donce: This is duplicated with `corsHandler` - only `ServeHTTP` differs.

We can DRY this by having
```go
type handlerWrapper struct {
  originalHandler http.Handler
  serveHttp func(resp http.ResponseWriter, req *http.Request)
}

func DisableCaching(original http.Handler) http.Handler {
  return &handlerWrapper{original, serveDisabledCaching}
}

func serveDisabledCaching...
```
and same applies for `ApplyCors`.

tadovas: Possible. But only two structures at the moment. Maybe when count reaches 3 :)

tequilapi/http_api_server.go
```diff
@@ -23,33 +24,38 @@ type apiServer struct {
...
 
+// Stop method stops underlying http server
 func (server *apiServer) Stop() {
 	if server.listener == nil {
 		return
 	}
 	server.listener.Close()
 }
 
+// Wait method waits for http server to finish handling requets (i.e. when Stop() was called)
```
zolia: 'requets' -> 'requests' typo

tequilapi/http_api_server.go
```diff
@@ -23,33 +24,38 @@ type apiServer struct {
 	listener      net.Listener
 }
 
+// NewServer creates http api server for given addres port and http handler
 func NewServer(address string, port int, handler http.Handler) APIServer {
 	server := apiServer{
 		make(chan error, 1),
-		ApplyCors(handler),
+		DisableCaching(ApplyCors(handler)),
```
zolia: It would be nicer to seen headers being applied sequentially, not wrapped like this. How this would look if we would need to add 4x different sets of headers?

tadovas: Nice idea. It could be improvement with composer which takes request, a list of decorators and applies them in given order like this:
Decorate(originRequest , ApplyCors, DisableCaching ... etc... )

PR #219, opened 2018-04-09
Stop tequila after current request is finished, instead of delay

tadovas:


tequilapi/endpoints/stop.go
```diff
@@ -4,28 +4,25 @@ import (
...
 		response.WriteHeader(http.StatusAccepted)
 	}
 }
-
-func delayExecution(function func(), duration time.Duration) {
-	go func() {
-		time.Sleep(duration)
-		function()
-	}()
+func doWhenNotified(notify <-chan struct{}, stopApplication ApplicationStopper) {
```
donce: Function name is generic `doWhenNotified`, but param is concrete - `stopApplication`. Can this function perform other things when notified?

tequilapi/endpoints/stop.go
```diff
@@ -4,28 +4,25 @@ import (
...
 
-func newStopHandler(stop ApplicationStopper, delay time.Duration) httprouter.Handle {
-	return func(response http.ResponseWriter, _ *http.Request, _ httprouter.Params) {
-		log.Info("Client is stopping")
-		delayExecution(stop, delay)
+func newStopHandler(stop ApplicationStopper) httprouter.Handle {
+	return func(response http.ResponseWriter, req *http.Request, _ httprouter.Params) {
+		log.Info("Application stop requested")
+
+		go doWhenNotified(req.Context().Done(), stop)
```
donce: Nice improvement 👍

tequilapi/endpoints/stop.go
tadovas: Well it's to private and two-linear to do other things. Maybe callStopWhenNotified?

donce: >to private and two-linear
you mean "too private and too linear"? :D
For me this looks like a generic thing - this was the original intention. Of course it's private, so you can't re-use it now, but in case you had to, it's possible to extract it - re-usable components are healthy even when not re-using :)

tadovas: I mean "too much internal" and two-(2 lines)-linear.  Yea I will stick with renaming. There is an old saying - don't DRY too much, you will suffer from dehidratation :)

tequilapi/endpoints/stop.go
```diff
@@ -4,28 +4,25 @@ import (
...
 		response.WriteHeader(http.StatusAccepted)
 	}
 }
-
-func delayExecution(function func(), duration time.Duration) {
-	go func() {
-		time.Sleep(duration)
-		function()
-	}()
+func callStopWhenNotified(notify <-chan struct{}, stopApplication ApplicationStopper) {
```
donce: why not simply `stopWhenNotified`?

PR #222, opened 2018-04-11
MYST-498 Cancel connection returns 499

donce:


client/connection/manager.go
```diff
@@ -62,9 +62,17 @@ func (manager *connectionManager) Connect(consumerID, providerID identity.Identi
 		}
 	}()
 
+	err = manager.startConnection(consumerID, providerID)
+	if err == utils.ErrRequestCancelled {
+		return ErrConnectionCancelled
```
tadovas: Do we really need separate error in manager if utils.ErrRequestCancelled is the only error which indicates cancelation ?

donce: Yes - we specify more concrete custom error message, which we use in our api response body.

donce: I was wondering, maybe this mapping can be done in api level - if manager.Connect returns `ErrRequestCancelled`, than we return custom error with response body we want.

donce: @tadovas pushed new commit of what I've meant.

client/connection/manager.go
```diff
@@ -20,8 +20,6 @@ var (
 	ErrNoConnection = errors.New("no connection exists")
 	// ErrAlreadyExists error indicates that aciton applieto to manager expects no active connection (i.e. connect)
 	ErrAlreadyExists = errors.New("connection already exists")
-	// ErrConnectionCancelled indicates that connection in progress was cancelled by request of api user
-	ErrConnectionCancelled = errors.New("connection was cancelled")
```
Waldz: What was wrong with more concrete error?

client/connection/manager.go
```diff
@@ -214,7 +216,7 @@ func (manager *connectionManager) waitForConnectedState(stateChannel <-chan open
 				manager.onStateChanged(state, sessionID)
 			}
 		case <-cancelRequest:
-			return ErrConnectionCancelled
+			return utils.ErrRequestCancelled
```
Waldz: Dont like that internally used library (`cancellable`) is exposed to external world

PR #233, opened 2018-05-02
Add notify on explicit exit openvpn flag

tadovas:
Add these flags to add a chance for graceful openvpn connection shutdown from both server and client sides (instead of waiting and timing out on TLS reauth)

### DIFF
--- a/openvpn/factory.go
+++ b/openvpn/factory.go
@@ -30,6 +30,7 @@ func NewServerConfig(
 	config.SetPingTimerRemote()
 	config.SetPersistTun()
 	config.SetPersistKey()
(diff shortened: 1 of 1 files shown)


openvpn/factory.go
```diff
@@ -30,6 +30,7 @@ func NewServerConfig(
 	config.SetPingTimerRemote()
 	config.SetPersistTun()
 	config.SetPersistKey()
+	config.setFlag("explicit-exit-notify")
```
Waldz: Would be nice to have comment that problem it fixes (command, commit, anythere)

tadovas: Good point for the future

PR #235, opened 2018-05-08
MYST 532 Tequilapi new Location Endpoint

interro:


client/connection/openvpn.go
```diff
@@ -38,19 +37,12 @@ func ConfigureVpnClientFactory(
 
 		statsSaver := bytescount.NewSessionStatsSaver(statsKeeper)
 
-		consumerCountry, err := locationDetector.DetectCountry()
```
Waldz: Why not detecting on connection start?

cmd/commands/client/command_client.go
```diff
@@ -61,14 +61,22 @@ func NewCommandWith(
 		filepath.Join(options.DirectoryConfig, options.LocationDatabase),
 	)
 
+	originalLocation, err := locationDetector.DetectLocation()
```
Waldz: Factory should just construct, not actual jobs

tequilapi/endpoints/connection.go
```diff
@@ -113,6 +119,24 @@ func (ce *ConnectionEndpoint) GetIP(writer http.ResponseWriter, request *http.Re
...
+// GetLocation responds with original and current countries
+func (ce *ConnectionEndpoint) GetLocation(writer http.ResponseWriter, request *http.Request, params httprouter.Params) {
+	currentLocation, err := ce.locationDetector.DetectLocation()
+	if err != nil {
+		utils.SendError(writer, err, http.StatusServiceUnavailable)
+		return
+	}
+
+	response := struct {
+		Original location.Location `json:"original"`
```
Waldz: Internal `Location` object should be mapped to Tequilapi response DTO

client/connection/openvpn.go
interro: fixed

cmd/commands/client/command_client.go
```diff
@@ -118,6 +123,13 @@ func (cmd *Command) Start() error {
 	}
 	log.Infof("Api started on: %d", port)
 
+	originalLocation, err := cmd.locationCache.RefreshAndGet()
```
Waldz: Why are't You using/injecting variable `originalLocation` anythere?

client/connection/openvpn.go
Waldz: Does it mean that we are always getting startup location?

tequilapi/endpoints/location.go
```diff
@@ -0,0 +1,52 @@
...
+func NewLocationEndpoint(locationDetector location.Detector, locationCache location.Cache) *LocationEndpoint {
+	return &LocationEndpoint{
+		locationDetector: locationDetector,
+		locationCache: locationCache,
+	}
+}
+
+// GetLocation responds with original and current countries
+func (ce *LocationEndpoint) GetLocation(writer http.ResponseWriter, request *http.Request, params httprouter.Params) {
+	originalLocation, err := ce.locationCache.Get()
```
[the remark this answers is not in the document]
zolia: I would make it '(le *LocationEndpoint)', not '(ce *LocationEndpoint)'

tequilapi/endpoints/location.go
```diff
@@ -0,0 +1,52 @@
...
+	}
+
+	currentLocation, err := ce.locationDetector.DetectLocation()
+	if err != nil {
+		utils.SendError(writer, err, http.StatusServiceUnavailable)
+		return
+	}
+
+	response := struct {
+		Original location.Location `json:"original"`
```
donce: `original` looks confusing - it doesn't specify what this location actually means. Maybe `startup` would be more clear?

location/interface.go
```diff
@@ -7,5 +7,11 @@ type Resolver interface {
...
 // Detector allows detecting location by current ip
 type Detector interface {
-	DetectCountry() (string, error)
+	DetectLocation() (Location, error)
+}
+
+// Cache allows caching location
+type Cache interface {
+	Get() (Location, error)
+	RefreshAndGet() (Location, error)
```
donce: Most of uses of `RefreshAndGet` seems to be made only for `Refresh` part, result is ignored.
We can simplify this method to assigning single responsibility to it - just `Refresh()`.

cmd/commands/client/command_client.go
tadovas: By the way - I suggest moving log.xx line after location is fetched. In case of error it should display Api started... and then error

tequilapi/endpoints/location.go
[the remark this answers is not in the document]
donce: > If we failed to get location (original) on startup do we always display that error?

In case only one of original and current ip is not available, it makes sense to return what we have. I.e.
```js
{
  original: null,
  current: "1.1.1.1"
}
```
or
```js
{
  original: "1.1.1.1",
  current: null
}

File: client/connection/openvpn.go (hunk 1 above)
Comment (interro) [reply client/connection/openvpn.go]: Getting location on app start

File: tequilapi/endpoints/location.go (hunk 5 above)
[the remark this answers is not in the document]
Comment (interro) [reply tequilapi/endpoints/location.go]: @tadovas changed logic where locationCache.Get() no longer returns error so the answer is no, we display error on startup only

File: tequilapi/endpoints/location.go (hunk 6 above)
Comment (Waldz) [reply tequilapi/endpoints/location.go]: `original` is just perfect

File: location/interface.go (hunk 7 above)
Comment (Waldz) [reply location/interface.go]: I agree

File: tequilapi/endpoints/location.go (hunk 6 above)
Comment (donce) [reply tequilapi/endpoints/location.go]: ok, when you put it like this, it does make sense - but at first look, it didn't

File: cmd/commands/client/command_client.go [hunk 8]
```diff
@@ -97,16 +101,24 @@ type Command struct {
...
 	err := cmd.checkOpenvpn()
 	if err != nil {
 		return err
 	}
 
+	originalLocation, err := cmd.originalLocationCache.RefreshAndGet()
+	if err != nil {
+		log.Warn("Failed to detect country", err)
+	} else {
+		log.Info("Country detected: ", originalLocation.Country)
```
donce: Logs of original and vpn country could differ - i.e. "Original country detected: Lithuania", "VPN country detected: China" - it would be easier to debug.

location/cache.go
```diff
@@ -0,0 +1,25 @@
...
+	}
+}
+
+// Gets location from cache
+func (lc *cache) Get() (Location) {
+	return lc.location
+}
+
+// Stores location to cache
+func (lc *cache) RefreshAndGet() (Location, error)  {
```
donce: `Refresh()`?

cmd/commands/client/command_client.go
interro: added "original" to log message

PR #237, opened 2018-05-11
Add go vet to check for dead code

donce:
Dead code is no good. Also, [go report card](https://goreportcard.com/report/github.com/MysteriumNetwork/node#go_vet) runs go vet, so not having dead code increases our score as well :)

`go vet` supports multiple checks - not sure if others are needed, but we can add more rules if we find them useful :)

### DIFF
--- a/bin/test_commit
+++ b/bin/test_commit
@@ -1,0 +1,1 @@
+bin/test && bin/vet && bin/lint_git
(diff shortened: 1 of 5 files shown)


donce:
@zolia added lint_git script

bin/test_commit
```diff
@@ -0,0 +1 @@
+bin/test && bin/vet && bin/lint_git
```
Waldz: What is purpose of `bin/test_commit` helper?
Example usage documentation missing.

donce: @Waldz added docs.

PR #238, opened 2018-05-11
Hotfix/unregistered provider print fix

zolia:


e2e/docker-compose.yml
```diff
@@ -53,7 +53,7 @@ services:
       MYSQL_PASSWORD: myst_api
 
   discovery:
-    image: mysteriumnetwork/mysterium-api:0.1.7
+    image: mysteriumnetwork/mysterium-api:latest
```
tadovas: Do we really need latest here? I would like to have hardcoded API version against which we are running e2e tests - otherwise tests will become non-deterministic - in case of api version bump, they will start to fail.
For future - maybe we can define two api containers - stable version and latest and run tests against both of them.

Waldz: Does not integrations test have a purpose to identify any service that broke after changes?
it means:
- it should be in separate repo
- any component (node, api) should trigger an integration check

tadovas: Normally yes. In this case - it's node integration testing not the whole system and we are testing node behaviour against specific api version

PR #241, opened 2018-05-17
Added win/osx/linux cross-compilation via xgo

ignasbernotas:
cd $GOPATH/src/github.com/mysterium/node/
docker build -t mysteriumnetwork/mysterium-client ./bin/util_docker/builder
./bin/package_all 0.1.1

It creates osx/linux/windows binaries, and packages debian files, but fails at creating docker images.

### DIFF
--- a/bin/package_all
+++ b/bin/package_all
@@ -9,6 +9,9 @@
 # Usage example:
 #> bin/package_all 0.0.1
 
+source bin/helpers/functions.sh
+source bin/helpers/load_environment.sh
+
 VERSION=$1
 if [ -z "$VERSION" ]; then
     printf "\e[0;31m%s\e[0m\n" "Missing version!"
@@ -16,6 +19,9 @@ if [ -z "$VERSION" ]; then
 fi
 
 DIR_BUILD="build/package"
+DOCKER_IMAGE="client_builder"
+DOCKER_LOCAL_DIR="/go/src/github.com/mysterium/node"
+PLATFORMS="darwin-10.11/amd64,windows-10.0/amd64,linux/amd64"
 
 printf "Creating build directory '$DIR_BUILD' ..\n" \
     && rm -rf $DIR_BUILD \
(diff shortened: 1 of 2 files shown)


bin/util_docker/builder/Dockerfile
```diff
@@ -9,7 +9,8 @@ RUN apt-get update \
 
 # Install Debber
 RUN go get github.com/debber/debber-v0.3/cmd/debber
+RUN go get github.com/oschwald/geoip2-golang
 
-WORKDIR /go/src/github.com/mysterium/node
+WORKDIR /
```
Waldz: This path was mounted in `builder_run`

bin/package_all
```diff
@@ -9,40 +9,69 @@
...
     && printf "\n\n"
 
-for ARCH in amd64 i386 armhf
-do
-    bin/builder_run "bin/server_build; bin/server_package_debian $VERSION $ARCH" \
-      && mv build/server/mysterium_server $DIR_BUILD/mysterium_server_linux_$ARCH \
-      && printf "\n\n"
-done
+# build client
+docker run --rm -v $(pwd):$DOCKER_LOCAL_DIR $DOCKER_IMAGE -c "
```
Waldz: You can assume that script is already inside docker, and call this script with:
`bin/builder_run bin/package_all`.
Script ecpects environment to be ready and calls `xgo` directly without proxying to docker

bin/package_all
```diff
@@ -9,40 +9,69 @@
...
+    xgo --targets='$PLATFORMS'            \
+        --ldflags='$(get_linker_ldflags)' \
+        --out='mysterium_client'          \
```
Waldz: --out=$DIR_BUILD

bin/package_all
ignasbernotas: This is so confusing, we're separating files into different responsibilities, but not reusing them out of package_all script.. I would prefer single larger file than many one liners

PR #242, opened 2018-05-17
Add license headers

donce:


Waldz:
Need Travis check for future generations. See ideas in MYST-487

client/connection/interface.go
```diff
@@ -1,3 +1,20 @@
+/*
+ * Copyright (C) 2018 The Mysterium Network Authors
```
Waldz: Should text same as in `LICENCE` file

tadovas:
Also sign-off check in github?

donce:
@Waldz @tadovas you're talking about other subtasks of MYST-487. This is just one of subtasks, MYST-242.

donce:
I'll add instructions with screenshots on how to setup GoLand for this after we've merged this PR :)

tadovas:
@Waldz got it. LGTM

PR #248, opened 2018-05-18
Feature/myst 497 improve session authorisation

zolia:
- prohibit second client with the same credentials
- cleanup session after client disconnects
- fixed bogus tests
- added new test


identity/extractor_test.go
```diff
@@ -37,6 +37,16 @@ func TestAuthenticate_WhenSignatureIsCorrect(t *testing.T) {
...
 }
 
+func TestAuthenticate_WhenBase64MessageSignatureIsCorrect(t *testing.T) {
+	message := []byte("MystVpnSessionId:Boop!")
+	signature := SignatureBase64("V6ifmvLuAT+hbtLBX/0xm3C0afywxTIdw1HqLmA4onpwmibHbxVhl50Gr3aRUZMqw1WxkfSIVdhpbCluHGBKsgE=")
+
+	extractor := &extractor{}
+	signerID, err := extractor.Extract(message, signature)
+	assert.NoError(t, err)
+	assert.Exactly(t, originalSignerID, signerID, "Original signer should be extracted")
```
tadovas: You are comparing two signerIDs here, error message should be "Signers should match"

openvpn/session/manager.go
```diff
@@ -0,0 +1,72 @@
...
+func (manager *manager) FindSession(id session.SessionID) (session.Session, bool) {
+	return manager.sessionManager.FindSession(id)
+}
+
+// FindSession finds session and sets clientID if it is not set yet, returns false on clientID conflict
+func (manager *manager) FindUpdateSession(clientID int, id session.SessionID) (session.Session, bool) {
+	// start enumerating clients from '1', since non-existing key, might return '0' as clientID value
+	clientID++
+	sessionInstance, found := manager.FindSession(id)
+	activeClientID := manager.sessionClientIDMap[id]
```
tadovas: instead of expecting clientID as 0 if not found, use second return value from map return:
clientId, found := map[id] if !found do ...

openvpn/session/manager.go
```diff
@@ -0,0 +1,72 @@
...
+package session
+
+import (
+	"github.com/mysterium/node/identity"
+	"sync"
+	"github.com/mysterium/node/session"
+)
+
+// NewManager returns session manager which maintains a map of session id -> session
+func NewManager(m session.Manager) *manager {
```
tadovas: What is the purpose of this "session" manager inside openvpn package? we already have one.

openvpn/session/manager.go
```diff
@@ -0,0 +1,72 @@
...
+	return manager.sessionManager.Create(peerID)
+}
+
+// FindSession returns session instance by given session id
+func (manager *manager) FindSession(id session.SessionID) (session.Session, bool) {
+	return manager.sessionManager.FindSession(id)
+}
+
+// FindSession finds session and sets clientID if it is not set yet, returns false on clientID conflict
+func (manager *manager) FindUpdateSession(clientID int, id session.SessionID) (session.Session, bool) {
```
tadovas: Finds and Updates? Maybe this method does too much and needs to be split.

openvpn/middlewares/server/auth/middleware.go
```diff
@@ -23,21 +23,26 @@ import (
...
 )
 
-// CredentialsChecker callback checks given auth primitives (i.e. customer identity signature / node's sessionId)
-type CredentialsChecker func(username, password string) (bool, error)
-
 type middleware struct {
-	checkCredentials CredentialsChecker
-	commandWriter    management.Connection
-	currentEvent     clientEvent
+	credentialsValidator CredentialsChecker
```
tadovas: I think that instead of having more and more depenencies on each use case (validating, cleaning) maybe an event channel could be a more generic solution, and middleware will have well defined purpose - generating events from incoming client messages through management connection and passing them up

openvpn/session/manager.go
```diff
@@ -0,0 +1,72 @@
...
+		sessionManager:     m,
+		sessionClientIDMap: make(map[session.SessionID]int),
+		creationLock:   	sync.Mutex{},
+	}
+}
+
+type manager struct {
+	sessionManager session.Manager
+	sessionClientIDMap     map[session.SessionID]int
+	creationLock   sync.Mutex
```
tadovas: Is this used anywhere? By the way this manager (map) is not thread-safe

openvpn/session/manager.go
```diff
@@ -0,0 +1,72 @@
...
+
+	if activeClientID != clientID {
+		return session.Session{}, false
+	}
+	return sessionInstance, found
+}
+
+// RemoveSession removes given session from underlying session managers
+func (manager *manager) RemoveSession(id session.SessionID) {
+	manager.sessionManager.RemoveSession(id)
```
tadovas: You need mutex locking here

openvpn/session/validator.go
```diff
@@ -0,0 +1,70 @@
...
+		signature := identity.SignatureBase64(signatureString)
+		extractedIdentity, err := v.identityExtractor.Extract([]byte(SignaturePrefix+sessionString), signature)
+		if err != nil {
+			return false, err
+		}
+		return currentSession.ConsumerID == extractedIdentity, nil
+}
+
+// Cleanup removes session from underlying session managers
+func (v *Validator) Cleanup(sessionString string) error {
```
tadovas: Cleanup is just callback from underlyging middleware, what caller has to do with returned error? Error handling (logging) can be performed here

openvpn/session/manager.go
zolia: Doc updated, this manager is kind of decorator for underlying session manager to account for openvpn clientID

identity/extractor_test.go
donce: To keep the detailed explanation and to specify failure, it could be something like this: `Extracted signer should match original signer`

openvpn/middlewares/server/auth/middleware.go
zolia: I got the idea, but maybe its better to split this one into separate case.

openvpn/session/validator.go
zolia: Cleanup is called as callback from middleware.handleClientEvent, error is handled there..

openvpn/session/manager.go
zolia: set a lock in underlying RemoveSession function.

zolia:
remarks fixed..

openvpn/session/validator.go
tadovas: That's my poiont. Why middleware which is lower level should handle cleanup errors? It doesn't know what to do with them anyway - except to log.

zolia: I think its important to be able to test for error returned by Cleanup callback, which can be nicely tested for in tests as in TestCleanupReturnsErrorIfSessionNotExists()

openvpn/middlewares/server/auth/middleware.go
```diff
@@ -23,21 +23,27 @@ import (
...
-type CredentialsChecker func(username, password string) (bool, error)
-
 type middleware struct {
-	checkCredentials CredentialsChecker
-	commandWriter    management.Connection
-	currentEvent     clientEvent
+	// TODO: consider implementing event channel to communicate required callbacks
+	credentialsValidator CredentialsValidator
+	sessionCleaner       SessionCleaner
+	commandWriter        management.Connection
```
donce: Not related to this PR, but this variable and class naming difference bugs me - why not call interface `management.CommandWriter` instead of `management.Connection`? Isn't `Connection` just a detail of implementation?

openvpn/middlewares/server/auth/middleware.go
```diff
@@ -23,21 +23,27 @@ import (
...
 
+// CredentialsValidator callback checks given auth primitives (i.e. customer identity signature / node's sessionId)
+type CredentialsValidator func(clientID int, username, password string) (bool, error)
+
+// SessionCleaner callback cleans up session after client disconnects
+type SessionCleaner func(username string) error
+
 // NewMiddleware creates server user_auth challenge authentication middleware
-func NewMiddleware(credentialsChecker CredentialsChecker) *middleware {
+func NewMiddleware(credentialsChecker CredentialsValidator, cleaner SessionCleaner) *middleware {
```
donce: `credentialsChecker` -> `credentialsValidator`

openvpn/session/credentials/provider.go
```diff
@@ -0,0 +1,33 @@
...
+import (
+	"github.com/mysterium/node/session"
+	ovpnsession "github.com/mysterium/node/openvpn/session"
+	"github.com/mysterium/node/identity"
+	"github.com/mysterium/node/openvpn/middlewares/client/auth"
+)
+
+// SignatureCredentialsProvider returns session id as username and id signed with given signer as password
+func SignatureCredentialsProvider(id session.SessionID, signer identity.Signer) auth.CredentialsProvider {
+	return func() (string, string, error) {
```
donce: Returns list is hard to understand, especially with 2 unnamed strings. You have to just to `auth.CredentialsProvider` to understand what this function returns.

Naming return values solves this:
```
	return func() (username string, password string, err error) {
		signature, err := signer.Sign([]byte(ovpnsession.SignaturePrefix + id))
		return string(id), signature.Base64(), err
	}
```

openvpn/session/manager.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+type manager struct {
+	sessionManager     session.Manager
+	sessionClientIDMap map[session.SessionID]int
+	sessionMapLock     sync.Mutex
+}
+
+// Create delegates session creation to underlying session.manager
+func (manager *manager) Create(peerID identity.Identity) (sessionInstance session.Session, err error) {
```
donce: No need to name return values here, they are pretty obvious.

openvpn/session/validator.go
```diff
@@ -0,0 +1,76 @@
...
+	sessionManager *manager
+	identityExtractor identity.Extractor
+}
+
+// NewValidator return Validator instance
+func NewValidator(m *manager, extractor identity.Extractor) (*Validator) {
+	return &Validator{
+				sessionManager: m,
+				identityExtractor: extractor,
+			}
```
donce: Very strange indentation.

openvpn/session/manager.go
```diff
@@ -0,0 +1,83 @@
...
+
+// FindSession returns OpenVPN session instance by given session id
+func (manager *manager) FindSession(clientID int, id session.SessionID) (session.Session, bool, error) {
+	sessionInstance, foundSession := manager.sessionManager.FindSession(id)
+
+	if !foundSession {
+		return session.Session{}, false, errors.New("no underlying session exists, possible break-in attempt")
+	}
+
+	activeClientID, foundClientID := manager.sessionClientIDMap[id]
```
donce: Naming is very confusing - `activeClientID` and `foundClientID` sounds like 2 differents client IDs, one which is active and another one which was found.
Maybe `sessionClientID` and `clientIDFound` is more readable?

openvpn/session/manager.go
```diff
@@ -0,0 +1,83 @@
...
+
+	if foundClientID && clientID != activeClientID {
```
donce: This condition is not very readable - it looks like 3 different client IDs are here. Another comment about `foundClientID` would fix this.

openvpn/session/manager.go
```diff
@@ -0,0 +1,83 @@
...
+
+	return sessionInstance, foundClientID, nil
+}
+
+// UpdateSession updates OpenVPN session with clientID, returns false on clientID conflict
+func (manager *manager) UpdateSession(clientID int, id session.SessionID) bool {
+	manager.sessionMapLock.Lock()
+	defer manager.sessionMapLock.Unlock()
+
+	_, foundClientID := manager.sessionClientIDMap[id]
```
donce: Do we really need to have `Map` in variable name? `sessionClientIDs[id]` looks quite readable as well for me.
I.e. we don't name arrays with `Array` prefix, as in `SessionArray`

openvpn/session/validator_mocks.go
```diff
@@ -0,0 +1,67 @@
...
+
+type mockedConfigProvider func() string
+
+func (mcp mockedConfigProvider) ProvideServiceConfig() (session.ServiceConfiguration, error) {
+	return mcp(), nil
+}
+
+func provideMockedVPNConfig() string {
+	return mockedVPNConfig
+}
```
donce: Why do we need this function, can't we take `const` directly?

zolia: We cannot use plain string here since 'session.NewManager()' requires function as an argument where it is used in TestValidateReturnsFalseWhenNoSessionFound()

PR #249, opened 2018-05-22
Add go fmt check to CI

donce:
screen shot 2018-05-21 at 15 00 23
screen shot 2018-05-21 at 15 00 43


bin/check_go_fmt
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+# Checks if all project files are formatted using go fmt
+#
+# Usage example:
+#> bin/check_gofmt
```
zolia: bin/check_go_fmt typo

PR #250, opened 2018-05-22
OpenVPN management network address

ignasbernotas:
Our current openvpn management connection is using unix sockets, this isn't supported in windows, therefore we need to switch to TCP.

### DIFF
--- a/openvpn/client.go
+++ b/openvpn/client.go
@@ -39,13 +39,17 @@ type openVpnClient struct {
 
 // NewClient creates openvpn client with given config params
 func NewClient(openvpnBinary string, config *ClientConfig, directoryRuntime string, middlewares ...management.Middleware) *openVpnClient {
+	managementAddress, err := management.GetAvailableAddress()
+	if err != nil {
+		return nil
+	}
+
 	// Add the management interface socketAddress to the config
-	socketAddress := tempFilename(directoryRuntime, "openvpn-management-", ".sock")
-	config.SetManagementSocket(socketAddress)
+	config.SetManagementSocket(managementAddress.IP, managementAddress.Port)
 
 	return &openVpnClient{
 		config:     config,
-		management: management.NewManagement(socketAddress, "[client-management] ", middlewares...),
+		management: management.NewManagement(managementAddress.String(), "[client-management] ", middlewares...),
 		process:    NewProcess(openvpnBinary, "[client-openvpn] "),
 	}
 }
--- a/openvpn/client.go
+++ b/openvpn/client.go
@@ -40,7 +40,7 @@ type openVpnClient struct {
 // NewClient creates openvpn client with given config params
 func NewClient(openvpnBinary string, config *ClientConfig, managementAddress *management.Address, middlewares ...management.Middleware) *openVpnClient {
 	// Add the management interface socketAddress to the config
-	config.SetManagementSocket(managementAddress.IP, managementAddress.Port)
+	config.SetManagementAddress(managementAddress.IP, managementAddress.Port)
 
 	return &openVpnClient{
 		config:     config,
--- a/openvpn/client.go
+++ b/openvpn/client.go
@@ -38,12 +38,7 @@ type openVpnClient struct {
 }
 
 // NewClient creates openvpn client with given config params
-func NewClient(openvpnBinary string, config *ClientConfig, directoryRuntime string, middlewares ...management.Middleware) *openVpnClient {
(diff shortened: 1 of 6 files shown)


openvpn/server.go
```diff
@@ -19,18 +19,21 @@ package openvpn
...
-
 	"github.com/mysterium/node/openvpn/tls"
 	"sync"
 )
 
 // NewServer constructs new openvpn server instance
 func NewServer(openvpnBinary string, generateConfig ServerConfigGenerator, directoryRuntime string, middlewares ...management.Middleware) *Server {
 	// Add the management interface socketAddress to the config
-	socketAddress := tempFilename(directoryRuntime, "openvpn-management-", ".sock")
+	managementAddress, err := management.GetAvailableAddress()
```
ignasbernotas: Wasn't sure how to handle it @tadovas

openvpn/management/address.go
```diff
@@ -0,0 +1,40 @@
...
+
+func GetAddressFromString(address string) Address {
+	str := strings.Split(address, ":")
+	port, _ := strconv.Atoi(str[1])
+
+	return Address{
+		IP:   str[0],
+		Port: port,
+	}
+}
```
donce: What happens if I call `GetAddressFromString("1.1.1.1")`, `GetAddressFromString("::")` or `GetAddressFromString("127.0.0.1:ragai")`?

openvpn/management/address.go
```diff
@@ -0,0 +1,40 @@
...
+type Address struct {
+	IP   string
+	Port int
+}
+
+func (ma *Address) String() string {
+	return fmt.Sprintf("%s:%d", ma.IP, ma.Port)
+}
+
+func GetAvailableAddress() (*Address, error) {
```
donce: Not quite sure what this function does?

openvpn/management/address.go
```diff
@@ -0,0 +1,40 @@
+package management
+
+import (
+	"net"
+	"fmt"
+	"strings"
+	"strconv"
+)
+
+type Address struct {
```
donce: Please document public objects.

openvpn/client.go
```diff
@@ -39,13 +39,17 @@ type openVpnClient struct {
 
 // NewClient creates openvpn client with given config params
 func NewClient(openvpnBinary string, config *ClientConfig, directoryRuntime string, middlewares ...management.Middleware) *openVpnClient {
+	managementAddress, err := management.GetAvailableAddress()
+	if err != nil {
+		return nil
```
zolia: How this 'nil' should be handled in caller? Since its factory, all logic should be performed outside and result passed as argument, such as: 
func NewClient(managementAddress *Address ... ) *openVpnClient

openvpn/management/address.go
ignasbernotas: Fixed, added a docblock

openvpn/management/address.go
ignasbernotas: fixed

openvpn/management/address.go
ignasbernotas: fixed

openvpn/client.go
ignasbernotas: Good point! I'm on it:) Thanks!

openvpn/config.go
```diff
@@ -58,8 +59,10 @@ func (c *Config) setFlag(name string) {
 }
 
 // SetManagementSocket creates unix socket style socket option for communication with openvpn process
-func (c *Config) SetManagementSocket(socketAddress string) {
-	c.setParam("management", socketAddress+" unix")
+func (c *Config) SetManagementSocket(address string, port int) {
```
zolia: I would rename it to reflect more information, SetManagementAddress

openvpn/management/address.go
```diff
@@ -0,0 +1,57 @@
+package management
+
+import (
+	"net"
+	"fmt"
+	"strings"
+	"strconv"
+	"github.com/kataras/iris/core/errors"
```
zolia: whats this package about?

ignasbernotas: already removed:)

ignasbernotas:
@tadovas you commented on an older version of the file, yet your points are valid. The problem I encountered was that the the config was set before the management.Start is called and since we can only get the port number from the live connection, in the current order of things we couldn't use this approach without too many changes. I didn't want to refactor and break things and took the easy way :/

PR #251, opened 2018-05-22
generic request timeout

zolia:
this timeout covers all request phases

### DIFF
--- a/server/mysterium_api.go
+++ b/server/mysterium_api.go
@@ -39,13 +39,13 @@ type HTTPTransport interface {
 	Do(*http.Request) (*http.Response, error)
 }
 
-func newHTTPTransport(responseTimeout time.Duration) HTTPTransport {
+func newHTTPTransport(requestTimeout time.Duration) HTTPTransport {
 	return &http.Client{
 		Transport: &http.Transport{
 			//Don't reuse tcp connections for request - see ip/rest_resolver.go for details
-			DisableKeepAlives:     true,
-			ResponseHeaderTimeout: responseTimeout,
+			DisableKeepAlives: true,
 		},
+		Timeout: requestTimeout,
 	}
 }
 


server/mysterium_api.go
```diff
@@ -39,13 +39,13 @@ type HTTPTransport interface {
...
-func newHTTPTransport(responseTimeout time.Duration) HTTPTransport {
+func newHTTPTransport(requestTimeout time.Duration) HTTPTransport {
 	return &http.Client{
 		Transport: &http.Transport{
 			//Don't reuse tcp connections for request - see ip/rest_resolver.go for details
-			DisableKeepAlives:     true,
-			ResponseHeaderTimeout: responseTimeout,
+			DisableKeepAlives: true,
 		},
+		Timeout: requestTimeout,
```
donce: So the difference is that `Timeout` includes all processes, while `ResponseHeaderTimeout` is just one of process (getting headers)?

`ResponseHeaderTimeout`:

> // ResponseHeaderTimeout, if non-zero, specifies the amount of
        // time to wait for a server's response headers after fully
        // writing the request (including its body, if any). This
        // time does not include the time to read the response body.

`Timeout`:

> // Timeout specifies a time limit for requests made by this
        // Client. The timeout includes connection time, any
        // redirects, and reading the response body. The timer remains
        // running after Get, Head, Post, or Do return and will
        // interrupt reading of the Response.Body.
        //
        // A Timeout of zero means no timeout.
        //
        // The Client cancels requests to the underlying Transport
        // using the Request.Cancel mechanism. Requests passed
        // to Client.Do may still set Request.Cancel; both will
        // cancel the request.
        //
        // For compatibility, the Client will also use the deprecated
        // CancelRequest method on Transport if found. New
        // RoundTripper implementations should use Request.Cancel
        // instead of implementing CancelRequest.

zolia: exactly.

tadovas: 💯 tokens

PR #252, opened 2018-05-23
Moved openvpn management from unix socket to TCP to support windows

ignasbernotas:


openvpn/client.go
```diff
@@ -86,6 +85,9 @@ func (client *openVpnClient) Start() error {
 		return err
 	}
 
+	addr := client.management.ActiveSocketAddress()
```
tadovas: We don't need directoryRuntime anymore as a parameter? In that case it can be removed

openvpn/management/listener.go
```diff
@@ -76,12 +81,17 @@ func (management *Management) Stop() {
...
 	management.listenerShutdownWaiter.Wait()
+
 	log.Info(management.logPrefix, "Shutdown finished")
 }
 
-// SocketAddress returns management socket address
-func (management *Management) SocketAddress() string {
-	return management.socketAddress
+// ActiveSocketAddress returns management socket address
+func (management *Management) ActiveSocketAddress() *net.TCPAddr {
```
tadovas: Maybe net.Addr is enough? net.TCPAddr is too specific.

openvpn/server.go
```diff
@@ -63,13 +63,14 @@ type Server struct {
...
 	config := server.generateConfig()
 
-	config.SetManagementSocket(server.management.SocketAddress())
-
 	// Start the management interface (if it isnt already started)
 	if err := server.management.Start(); err != nil {
 		return err
 	}
 
+	addr := server.management.ActiveSocketAddress()
```
tadovas: directoryRuntime parameter can be removed?

openvpn/client.go
```diff
@@ -86,6 +85,9 @@ func (client *openVpnClient) Start() error {
...
+	client.config.SetManagementAddress(addr.IP.String(), addr.Port)
```
zolia: Why not just pass addr? More generic and less verbose.

PR #257, opened 2018-05-31
MYST-495 Show license on startup with subcommands for more info

donce:


cmd/commands/client/options.go
```diff
@@ -91,6 +93,20 @@ func ParseArguments(args []string) (options CommandOptions, err error) {
 		"Run an interactive CLI based Mysterium UI",
 	)
 
+	flags.BoolVar(
+		&options.Warranty,
+		"warranty",
```
Waldz: `licence.warranty`

cmd/commands/cli/command_cli.go
```diff
@@ -53,6 +53,8 @@ const statusConnected = "Connected"
 
 // Run runs CLI interface synchronously, in the same thread while blocking it
 func (c *Command) Run() (err error) {
+	license := cmd.GetStartupLicense("type `show w'", "type `show c'")
```
Waldz: Looks like `show w` is wrong recomendation

zolia: GetStartupLicense() - I don't see it defined. I would guess it should return error (file not found.. etc).

cmd/commands/client/options.go
donce: Updated.

cmd/commands/cli/command_cli.go
donce: Fixed

donce: It was defined in `license.go`, but it was too long file to be displayed in PR, so it was collapsed.

Split that file so that code would be separate from long texts.

cmd/commands/cli/command_cli.go
```diff
@@ -392,5 +405,7 @@ func newAutocompleter(tequilapi *tequilapi_client.Client, proposals []tequilapi_
 				getIdentityOptionList(tequilapi),
 			),
 		),
+		readline.PcItem("warranty"),
```
Waldz: What about licence prefix here, to be consistent with other places

donce: Added nesting, which makes more sense in CLI.

cmd/commands/client/command_client.go
```diff
@@ -151,6 +153,14 @@ func (cmd *Command) Start() error {
 	return nil
 }
 
+func printLicense() {
+	startupLicense := license.GetStartupLicense(
+		"run program with '-warranty' option",
```
Waldz: Still not consistent

cmd/commands/client/options.go
```diff
@@ -33,7 +33,9 @@ type CommandOptions struct {
 	TequilapiAddress string
 	TequilapiPort    int
 
-	CLI bool
+	CLI        bool
+	Warranty   bool
+	Conditions bool
```
Waldz: Still not consistent

donce:
@Waldz fixed

PR #258, opened 2018-06-04
MYST-546/Refactor spaces in option values handling

tadovas:
Actually args passed to openvpn doesn't need any escaping. It was a problem on our side - how we serialized params and options into array of strings as args. We did splitting by space our selfes - ignoring the fact that arg itself can contain spaces.

### DIFF
--- a/openvpn/factory.go
+++ b/openvpn/factory.go
@@ -18,6 +18,7 @@
 package openvpn
 
 import (
+	"fmt"
 	"github.com/mysterium/node/openvpn/tls"
 	"io/ioutil"
 	"path/filepath"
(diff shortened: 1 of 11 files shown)


openvpn/factory.go
```diff
@@ -106,8 +107,8 @@ func NewClientConfigFromSession(vpnConfig *VPNConfig, configDir string, configFi
 	config := ClientConfig{NewConfig(configDir)}
 	config.AddOptions(OptionFile("config", configAsString, configFile))
 
-	config.setParam("up", filepath.Join(configDir, "update-resolv-conf"))
-	config.setParam("down", filepath.Join(configDir, "update-resolv-conf"))
+	config.setParam("up", fmt.Sprintf("\"%s\"", filepath.Join(configDir, "update-resolv-conf")))
```
Waldz: Not clear intention why this hack is done

Waldz: Could also be ``"%s"``

PR #264, opened 2018-06-09
MYST-578 Start using XGO in CI flow

Waldz:


bin/server_package_standalone
```diff
@@ -0,0 +1,22 @@
...
+#> bin/server_package_standalone
+
+set -e
+
+BINARY=${BINARY:="build/server/mysterium_server"}
+DIR_BUILD="build/package"
+PACKAGE_FILE="${DIR_BUILD}/`basename -s .exe ${BINARY}`.tar.gz"
+
+# Put all files to package directory (to avoid preserving directory structure)
+DIR_TEMP=`mktemp -d ${DIR_BUILD}/${tempname}.XXXXXX`
```
ignasbernotas: what's the XXXXXX for?

bin/server_package_standalone
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
```
ignasbernotas: This is identical to bin/client_package_standalone except for binary name. Couldn't we use a single script for both of these?

ignasbernotas: If not, then could we at least start to tidy up the file structure in bin/ separate the node from client scripts?

bin/server_package_standalone
Waldz: It's template for tmp directory generation

bin/server_package_standalone
Waldz: Soon we will migrate to single binary `mysterium_node`, so I care about it much ATM

bin/server_package_standalone
```diff
@@ -0,0 +1,22 @@
...
+
+###########################################################################
+# Packaging script which creates standalone package.
+#
+# Usage:
+#> bin/server_package_standalone
+
+set -e
+
+BINARY=${BINARY:="build/server/mysterium_server"}
```
donce: What does this syntax mean?

bin/server_package_standalone
donce: @ignasbernotas more about it here: https://www.mktemp.org/manual.html

donce: Where does this `tempname` come from? :O

bin/server_package_standalone
```diff
@@ -0,0 +1,22 @@
...
+DIR_BUILD="build/package"
+PACKAGE_FILE="${DIR_BUILD}/`basename -s .exe ${BINARY}`.tar.gz"
+
+# Put all files to package directory (to avoid preserving directory structure)
```
donce: *to avoid preserving* - did you mean *to preserve*?

bin/server_package_standalone
Waldz: Variable can be passed to script via environment variables, but has default value

bin/server_package_standalone
Waldz: `mkdir` templating has some predefined variables

bin/server_package_standalone
Waldz: No. By default, `mkdir` preserves directory structure of global paths

PR #262, opened 2018-06-08
MYST-549 Terms & conditions

donce:
- terms flag for node
- summary terms text
- link to TERMS.md on github
- agreed-terms-and-conditions flag by default on docker_compose scripts

PR #266, opened 2018-06-13
Remove docker-compose.yml

donce:
@Waldz is suggesting to drop support for docker-compose to avoid using ENV variables in https://github.com/MysteriumNetwork/node/pull/262. If we want to go in that direction, we should be clear with that - we shouldn't have `docker-compose.yml` which we no longer can support.


Waldz:
It's documentation how to run a network locally, it works, lets keeps it

PR #267, opened 2018-06-18
MYST-559 - fix forever hanging openvpn client if process exits too early

tadovas:


openvpn/check.go
```diff
@@ -18,31 +18,50 @@
...
+	outputBuffer, cmdResult := command.Output()
 	exitCode, err := extractExitCodeFromCmdResult(cmdResult)
 	if err != nil {
 		return err
 	}
-
 	//openvpn returns exit code 1 in case of --version parameter, if anything else is returned - treat as error
 	if exitCode != 1 {
-		return errors.New("unexpected openvpn code: " + strconv.Itoa(exitCode))
+		log.Error("[Openvpn check] ", "Check failed. Output of executed command: ", string(outputBuffer))
```
zolia: "Openvpn check" we could move this to separate prefix.

openvpn/check.go
```diff
@@ -18,31 +18,50 @@
...
+		}
+		log.Info("[Openvpn check] ", str)
+	}
+
+	//optional custom tag
+	str, err := stringReader.ReadLine()
+	if err == nil {
+		log.Info("[Openvpn check] ", "Custom tag: ", str)
+	} else if err != io.EOF {
+		//EOF is expected and doesn't fail check
```
zolia: language here does not make much sense

openvpn/cmd_wrapper.go
```diff
@@ -0,0 +1,122 @@
...
+	"io"
+	"syscall"
+)
+
+// NewCmdWrapper returns process wrapper for given executable
+func NewCmdWrapper(executablePath, logPrefix string) *CmdWrapper {
+	return &CmdWrapper{
+		logPrefix:          logPrefix,
+		executablePath:     executablePath,
+		CmdExitError:       make(chan error, 1), //channel should have capacity to hold single process exit error
```
zolia: isn't it '1' by default?

openvpn/openvpn_process.go
```diff
@@ -0,0 +1,104 @@
...
+	"github.com/mysterium/node/openvpn/management"
+	"sync"
+	"time"
+)
+
+// Process defines openvpn process interface with basic controls
+type Process interface {
+	Start() error
+	Wait() error
+	Stop() error
```
zolia: error can be safely removed here, it should not be used.

zolia:
LGTM

openvpn/check.go
zolia: move it to constant

openvpn/check.go
tadovas: Improved comment

openvpn/check.go
tadovas: Moved

PR #271, opened 2018-06-21
Document all API endpoints

interro:


tequilapi/endpoints/identities.go
```diff
@@ -87,17 +103,17 @@ func (endpoint *identitiesAPI) List(resp http.ResponseWriter, request *http.Requ
...
 // swagger:operation POST /identities Identity createIdentity
 // ---
-// summary: Create new identity
+// summary: Creates new identity
 // description: Creates identity and stores in keystore encrypted with passphrase
 // parameters:
 //   - in: body
 //     name: body
-//     description: Parameters for creating new identity
+//     description: Parameter in body (passphrase) required for creating new identity
```
donce: Do we need to specify "in body"? This is transport, we should not couple description to transport.

interro: In swagger template this message looks more informative

tequilapi/endpoints/connection.go
```diff
@@ -75,19 +95,33 @@ func NewConnectionEndpoint(manager connection.Manager, ipResolver ip.Resolver, s
...
+//   500:
+//     description: Internal server error
+//     schema:
+//       "$ref": "#/definitions/errorMessage"
 func (ce *ConnectionEndpoint) Status(resp http.ResponseWriter, _ *http.Request, _ httprouter.Params) {
 	statusResponse := toStatusResponse(ce.manager.Status())
 	utils.WriteAsJSON(statusResponse, resp)
 }
 
+// Creates starts new connection
```
zolia: should be "Create", shouldn't it be caught by linter? Did you run it?

tequilapi/endpoints/health_check.go
```diff
@@ -25,16 +25,27 @@ import (
...
+
+	// example: 0.0.6
 	Version   string    `json:"version"`
 	BuildInfo buildInfo `json:"buildInfo"`
 }
 
 type buildInfo struct {
-	Commit      string `json:"commit"`
-	Branch      string `json:"branch"`
+	// example: <unknown>
```
zolia: Why this struct doesn't have "// swagger:model"  heading?

tequilapi/endpoints/connection.go
interro: fixed, linter configured to have some tolerance

tequilapi/endpoints/health_check.go
interro: added missing meta tags

bin/check_lint
```diff
@@ -41,7 +41,7 @@ MESSAGES_ERROR=`golint --set_exit_status --min_confidence=1 ${ARGUMENTS} 2>/dev/
 MESSAGES_RECONFIGURE=()
 
 check_uncleaned_package "github.com/mysterium/node/identity" 9
-check_uncleaned_package "github.com/mysterium/node/tequilapi" 12
+check_uncleaned_package "github.com/mysterium/node/tequilapi" 11
```
Waldz: +1

tequilapi/endpoints/connection.go
```diff
@@ -56,6 +56,26 @@ type statusResponse struct {
 	SessionID string `json:"sessionId,omitempty"`
 }
 
+// swagger:model
+type ipResponse struct {
```
Waldz: Very strange that structures are private.
Lets make them public, because it is so.

PR #272, opened 2018-06-27
MYST-569 get rid of build time env variables

tadovas:
…ined networks with posibility to override values one by one


cmd/commands/client/command_client.go
```diff
@@ -176,3 +177,15 @@ func (cmd *Command) Kill() error {
 
 	return nil
 }
+
+// TODO this function can be aligned with server function when client and server options will merge into
+func getNetworkDefinition(options CommandOptions) metadata.NetworkDefinition {
+	network := metadata.TestNetworkDefinition
```
zolia: I would use 'TestNetDefinition', since we talk about specific network name 'TestNet'

zolia:
other stuff looks good.

metadata/network.go
```diff
@@ -15,7 +15,16 @@
...
-// MysteriumAPIURL stores MYSTERIUM_API_URL env variable that is complied in
-var MysteriumAPIURL string
+// NetworkDefinition structure holds all parameters which describe particular network
+type NetworkDefinition struct {
+	DiscoveryAPIAddress string
+	BrokerAddress       string
+}
+
+// TestNetworkDefinition defines parameters for test network (currently default network)
+var TestNetworkDefinition = NetworkDefinition{
```
Waldz: TestNetwork.. -< Testnet..

metadata/network.go
```diff
@@ -15,7 +15,16 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-package cmd
+package metadata
 
-// MysteriumAPIURL stores MYSTERIUM_API_URL env variable that is complied in
-var MysteriumAPIURL string
+// NetworkDefinition structure holds all parameters which describe particular network
+type NetworkDefinition struct {
```
Waldz: +1 kudos

cmd/commands/server/options.go
```diff
@@ -125,13 +122,14 @@ func ParseArguments(args []string) (options CommandOptions, err error) {
 	flags.StringVar(
 		&options.DiscoveryAPIAddress,
 		"discovery-address",
-		cmd.MysteriumAPIURL,
+		"",
```
Waldz: Would be nice to print current defaults as was before

tadovas: There are no defaults anymore

cmd/commands/client/command_client.go
tadovas: Fixed

cmd/commands/server/factory.go
```diff
@@ -91,7 +95,7 @@ func NewCommandWith(
 		natService:       natService,
 		dialogWaiterFactory: func(myID identity.Identity) communication.DialogWaiter {
 			return nats_dialog.NewDialogWaiter(
-				nats_discovery.NewAddressGenerate(options.BrokerAddress, myID),
+				nats_discovery.NewAddressGenerate(networkDefintion.BrokerAddress, myID),
```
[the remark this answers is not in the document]
tadovas: fixed

metadata/network.go
tadovas: fixed

cmd/commands/server/options.go
Waldz: There are, DEVNET is default, I want to show it clearly

tadovas: Testnet is default as defined in network.go there are now default override values.

cmd/commands/client/command_client.go
```diff
@@ -176,3 +177,20 @@ func (cmd *Command) Kill() error {
 
 	return nil
 }
+
+// TODO this function can be aligned with server function when client and server options will merge into
+func getNetworkDefinition(options CommandOptions) metadata.NetworkDefinition {
+	network := metadata.TestNetDefinition
+
+	switch {
+	case options.LocalNet:
```
Waldz: Why not to have `default` in switch statement

tadovas: you will need bigger scope variable anyway, so either it can be left empty and initialized in default: of switch or it can be initialized right away and let the switch to do modification if needed. I chose the second approach

cmd/commands/server/options.go
Waldz: We already started to use terminology Testnet/Devnet or TESTNET/DEVNET in communication, event You in previous sentence

tadovas: Ok I will stick to "Testnet" "Devnet" naming

PR #274, opened 2018-07-03
MYST-564 Begin v0.2 release cycle

Waldz:


Waldz:
- [ ] We can't start release without Terms and Conditions feature

Waldz:
We're not updating code anymore

PR #277, opened 2018-07-03
MYST-590 terms for deb install

zolia:
debconf based popup to accept link to terms

### DIFF
--- a/bin/server_package/installation/post-uninstall.sh
+++ b/bin/server_package/installation/post-uninstall.sh
@@ -1,10 +1,14 @@
 #!/bin/bash
 
-if [ "$1" = "purge" -a -e /usr/share/debconf/confmodule ]; then
-    # Source debconf library.
-    . /usr/share/debconf/confmodule
-    # Remove my changes to the db.
-    db_purge || true
+if [ "$1" = "purge" ]; then
+    if [[ -e /usr/share/debconf/confmodule ]]; then
+        # Source debconf library.
+        . /usr/share/debconf/confmodule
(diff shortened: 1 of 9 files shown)


bin/server_package/installation/post-uninstall.sh
```diff
@@ -1,15 +1,27 @@
 #!/bin/bash
 
+if [ "$1" = "purge" -a -e /usr/share/debconf/confmodule ]; then
```
Waldz: Such silently fails are very strange

bin/server_package/installation/post-uninstall.sh
```diff
@@ -1,15 +1,27 @@
...
+    # Source debconf library.
+    . /usr/share/debconf/confmodule
+    # Remove my changes to the db.
+    db_purge || true
```
Waldz: Same here

bin/server_package/installation/systemd.service
```diff
@@ -11,6 +11,7 @@ Group=mysterium-node
 
 RuntimeDirectory=mysterium-node
 RuntimeDirectoryMode=0750
+LogsDirectory=mysterium-client
```
Waldz: It's `server_package...`, whould be `mysterium-client`?

bin/server_package/installation/post-uninstall.sh
zolia: -a -e /usr/share/debconf/confmodule checks for existence of debconf package. And purge does not and must not fail.

PR #278, opened 2018-07-03
feature/MYST-596 DNS update for windows

tadovas:


openvpn/default_client_config.go
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package openvpn
+
+import "github.com/mysterium/node/openvpn/config"
+
+func defaultClientConfig(runtimeDir string, scriptSearchPath string) *ClientConfig {
```
Waldz: Would be nice to leave this function in `openvpn/client_config.go` to make Git history more searchable

tadovas: Good point 👍

PR #285, opened 2018-07-09
MYST-577 Fix missing version during "dev" builds

Waldz:


.travis.yml
```diff
@@ -39,18 +38,25 @@ script:
 
 before_deploy:
   - bin/travis_scripts/ensure_xgo.sh
-  - bin/package_all $VERSION
+  - bin/package_all $(cat VERSION)
```
zolia: do we need '$' before (cat VERSION) ?

zolia: should `cat VERSION` work?

Waldz: Approach now is different

metadata/version.go
```diff
@@ -42,6 +29,21 @@ const versionSummaryFormat = `Mysterium Node
 %s
 %s`
 
+// VersionAsString returns all defined version constants as single string
+func VersionAsString() string {
+	if Version != "" {
+		return Version
+	}
+
+	version := "source"
```
zolia: maybe space after 'source'?

Waldz: Fixed

PR #287, opened 2018-07-11
MYST-643 Make CI parallel building

tadovas:
Split checks into parallel builds on test phase
Possibly run e2e and packaging in compile phase (or integrity checking)
Make deployment phase with parallel deployments of the same binaries (provided in previous phase)


.travis.yml
```diff
@@ -17,25 +17,42 @@ env:
...
+    #test stage
+    - stage: test
+      script: bin/check_vet
+    - script: bin/builder_run bin/check_lint
+    - script: bin/check_fmt
+    - script: bin/check_license
+    - script: bin/test
+    #e2e stage
+    - stage: e2e-test
+      script: bin/travis_scripts/check_docker.sh && bin/travis_scripts/check_docker_compose.sh && bin/test_e2e
```
Waldz: Not much value of those checks, drop them maybe?
Could they've be in some `before_*` phase?

.travis.yml
```diff
@@ -17,25 +17,42 @@ env:
 cache:
   directories:
     - "$CACHE_ROOT_DIR"
+    - vendor
```
Waldz: Why we cache it, but then delete it?

soffokl: Vendor in the cache is used here to pass it between stages, to avoid multiple `glide install` calls. Deleting it everytime is not required.

.travis.yml
```diff
@@ -13,69 +13,140 @@ env:
...
+    - script:
+      - bin/server_package_docker_ubuntu $BUILD_VERSION
+      - docker save mysterium-node:ubuntu | gzip > mysterium-node_ubuntu.tgz
+      - bin/s3 cp mysterium-node_ubuntu.tgz s3://docker-images
+      name: "Ubuntu docker image for node"
+
+    - script:
+      - bin/client_package_docker_ubuntu $BUILD_VERSION
+      - docker save mysterium-client:ubuntu | gzip > mysterium-client_ubuntu.tgz
+      - bin/s3 cp mysterium-client_ubuntu.tgz s3://docker-images
```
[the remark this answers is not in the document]
soffokl: The idea was to avoid pushing other images if at least one failed since it could be incompatible. So push to docker registry is done on the next stage.

.travis.yml
```diff
@@ -14,68 +14,142 @@ env:
...
+      name: "Vet check"
+
+    - script: bin/builder_run bin/check_lint
+      name: "Linter"
+
+    - script: bin/check_fmt
+      name: "Formating check"
+
+    - script: bin/check_license
+      name: "License check"
```
tadovas: These sanity checkers like license, linter, vet are very small jobs (few sec), is there a point to paralelize them - overhead of single job on travis is around 30 secs, and there is limited number of parallel jobs.

soffokl: I did a test, the first test was with license+vet+fmt in one job, and the second test with all steps in separate jobs. Got the same results for the whole stage: 1:54 for both cases.
If we are starting all jobs in parallel, the time of the stage will be equal to the longest job. 
We still need to wait for other jobs to finish, like unit tests. 
And, I think, separated jobs looks more clear in the Travis CI UI.

tadovas: 💯  karma points for making tests and measurements. And yes - I was thinking between clear separations and a bit of optimisation. Looks good to me now

.travis.yml
```diff
@@ -14,68 +14,142 @@ env:
...
+
+    - script:
+      - BINARY=build/server/mysterium_server_windows_amd64.exe bin/server_package_standalone windows
+      - BINARY=build/client/mysterium_client_windows_amd64.exe bin/client_package_standalone windows
+      - bin/s3 sync build/package s3://build-artifacts
+      name: "Windows packages"
+
+    - script:
+      - bin/server_package_docker
+      - docker save mysterium-node:alpine | gzip > mysterium-node_alpine.tgz
```
zolia: you pipe here to gzip only, thus extension should be .gz only, not .tgz.

soffokl:
@zolia @Waldz Could you please review this PR once again. I have moved E2E-tests to the early stage to finish it faster.

PR #292, opened 2018-07-12
Improvement/naive detection of nat

tadovas:


cmd/commands/server/command_server.go
```diff
@@ -105,7 +110,16 @@ func (cmd *Command) Start() (err error) {
 		return err
 	}
 
-	sessionManager := cmd.sessionManagerFactory(primitives, vpnServerIP)
+	if outboundIP != publicIP {
+		log.Infof(
```
Waldz: Should be higher level `Error()`

cmd/commands/server/command_server.go
```diff
@@ -75,15 +75,20 @@ func (cmd *Command) Start() (err error) {
...
 	providerContact, err := cmd.dialogWaiter.Start()
 
+	publicIP, err := cmd.ipResolver.GetPublicIP()
+	if err != nil {
+		return err
+	}
+
 	// if for some reason we will need truly external IP, use GetPublicIP()
-	vpnServerIP, err := cmd.ipResolver.GetOutboundIP()
+	outboundIP, err := cmd.ipResolver.GetOutboundIP()
```
Waldz: Aren't we doing option `openvpn.ip` with this?
For those who wants has different public IP and wats to map it in router

cmd/commands/server/command_server.go
tadovas: It's just a warning - not an error

cmd/commands/server/command_server.go
tadovas: There is not such option at the moment. And this is a quick detection with posibility for user to take some action.

cmd/commands/server/factory.go
```diff
@@ -137,6 +139,31 @@ func NewCommandWith(
...
+			outboundIP, err := ipResolver.GetOutboundIP()
+			if err != nil {
+				return "", err
+			}
+
+			if publicIP != outboundIP {
+				forwardInfo := fmt.Sprintf("%s:%v -> %s:%v", publicIP, options.OpenvpnPort, outboundIP, options.OpenvpnPort)
+				log.Warnf(
+					`WARNING: It seems that publicaly visible ip: [%s] does not match your local machines ip: [%s]. 
+You should probaly need to do port forwarding on your router: %s.`,
```
Waldz: `..on your router: PUBLIC_IP:1194 -> MACHINE_IP:1194 e.g. %s:1194->%s:1194`
and you can put example values from detected values

cmd/commands/server/command_server.go
```diff
@@ -76,14 +77,14 @@ func (cmd *Command) Start() (err error) {
 	providerContact, err := cmd.dialogWaiter.Start()
 
 	// if for some reason we will need truly external IP, use GetPublicIP()
-	vpnServerIP, err := cmd.ipResolver.GetOutboundIP()
+	outboundIP, err := cmd.ipResolver.GetOutboundIP()
```
Waldz: Sad that we do 2-3 requests. Is it possible to avoid it?

cmd/commands/server/factory.go
tadovas: Hmm ok now it looks like this:
2018-07-12T11:45:11.643023557 [Warn] WARNING: It seems that publicaly visible ip: [46.255.210.150] does not match your local machines ip: [192.168.0.131]. 
You should probaly need to do port forwarding on your router: 46.255.210.150:1194 -> 192.168.0.131:1194.

Are you suggesting to expand message even more like "... on your router: PUBLIC_IP:1194 -> MACHINE_IP:1194 e.g. 46.255.210.150:1194 -> 192.168.0.131:1194." ?

cmd/commands/server/command_server.go
tadovas: Without too much refactoring? Yes - by passing outboundIP to openvpnServiceAddress function as parameter instead of looking up it again inside. I can live with this approach :)

cmd/commands/server/factory.go
Waldz: Sorry, did not see that it's already like that

tadovas: 👍

cmd/commands/server/command_server.go
tadovas: Fixed :) reduced ipify query to 1 request

PR #293, opened 2018-07-12
copy client configs travis way

zolia:


bin/client_package_standalone
```diff
@@ -28,9 +28,12 @@ DIR_BUILD="build/package"
 # Put all files to package directory (to avoid preserving directory structure)
 printf "Building Standalone package for OS '$OS' ..\n"
 DIR_TEMP=`mktemp -d ${DIR_BUILD}/${tempname}.XXXXXX`
-cp -p ${BINARY} ${DIR_TEMP}/mysterium_client
-cp -rp "bin/common_package/" ${DIR_TEMP}/config
-cp -rp "bin/client_package/config/${OS}/" ${DIR_TEMP}/config 2>/dev/null || true
+cp -vp ${BINARY} ${DIR_TEMP}/mysterium_client
+cp -vrp "bin/common_package/" ${DIR_TEMP}/config
+
+if [[ -d "bin/client_package/config/${OS}/" ]]; then
```
Waldz: Same code is duplicated in `bin/client_build` and would be nice to fix both places.
Duplication is also an issue

PR #296, opened 2018-07-12
HOTFIX: copy proper config files

Waldz:
Picked same fixed for release


bin/helpers/functions.sh
```diff
@@ -7,3 +7,12 @@ function get_linker_ldflags {
     [ -n "$BUILD_NUMBER" ] && echo -n "-X 'github.com/mysterium/node/metadata.BuildNumber=${BUILD_NUMBER}' "
     [ -n "$BUILD_VERSION" ] && echo -n "-X 'github.com/mysterium/node/metadata.Version=${BUILD_VERSION}' "
 }
+
+function copy_client_config {
```
tadovas: Can this function be generlized even more? I.e. copy_config with parameters <client or server> $OS $DIR_TEMP ?

PR #298, opened 2018-07-16
darwin nat implementation

zolia:
Client still has to be on different machine. 
This is due to 10.8.0.0/24 subnet clash. That is server wants to tunnel it through utun1, but client would need to tunnel it through utun2, but cannot do it since route already exists.

### DIFF
--- a/nat/service_pfctl.go
+++ b/nat/service_pfctl.go
@@ -28,8 +28,8 @@ import (
 )
 
 type servicePFCtl struct {
-	rules   []RuleForwarding
-	forward bool
+	rules     []RuleForwarding
+	ipForward serviceIPForward
 }
 
 func (service *servicePFCtl) Add(rule RuleForwarding) {
@@ -37,11 +37,12 @@ func (service *servicePFCtl) Add(rule RuleForwarding) {
 }
 
 func (service *servicePFCtl) Start() error {
-	service.clearStaleRules()
-	err := service.enableIPForwarding()
+	err := service.ipForward.Enable()
 	if err != nil {
 		return err
 	}
+
+	service.clearStaleRules()
 	err = service.enableRules()
 	if err != nil {
 		return err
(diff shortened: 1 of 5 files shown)


nat/service_pfctl.go
```diff
@@ -0,0 +1,164 @@
...
+}
+
+func (service *servicePFCtl) enableRules() error {
+	for _, rule := range service.rules {
+		iface, err := ifaceByAddress(rule.TargetIP)
+		if err != nil {
+			return err
+		}
+		natRule := fmt.Sprintf("nat on %v inet from %v to any -> %v", iface, rule.SourceAddress, rule.TargetIP)
+		arguments := fmt.Sprintf("echo \"%v\" | /sbin/pfctl -vEf -", natRule)
```
Waldz: Use qoute ` to avoid \\" escaping

nat/service_pfctl.go
```diff
@@ -0,0 +1,164 @@
...
+			}
+		}
+
+		log.Info(natLogPrefix, "NAT from '", rule.SourceAddress, "' to IP: ", rule.TargetIP)
+	}
+	return nil
+}
+
+func (service *servicePFCtl) disableRules() {
+	arguments := fmt.Sprintf("/sbin/pfctl -F nat")
```
Waldz: `Sprintf ` is redundant, because there aren't variables

nat/service_pfctl.go
```diff
@@ -0,0 +1,164 @@
...
+	cmd := exec.Command(
+		"sh",
+		"-c",
+		arguments,
```
Waldz: Looks like sub-command quoting is missing, is not it?

nat/service_pfctl.go
```diff
@@ -0,0 +1,164 @@
...
+		if err != nil {
+			return "", err
+		}
+		for _, address := range addresses {
+			if address.(*net.IPNet).IP.String() == ipAddress {
+				return ifi.Name, nil
+			}
+		}
+	}
+	return "undefined", nil
```
Waldz: Is not it error?

nat/service_pfctl.go
```diff
@@ -0,0 +1,164 @@
...
+	log.Info(natLogPrefix, "IP forwarding enabled")
+	return nil
+}
+
+func (service *servicePFCtl) disableIPForwarding() {
+	if service.forward {
+		return
+	}
+
+	cmd := utils.SplitCommand("/usr/sbin/sysctl", "-w net.inet.ip.forwarding=0")
```
Waldz: Looks like duplicated code, could be DRY'ied

nat/service_pfctl.go
```diff
@@ -0,0 +1,164 @@
...
+		cmd := exec.Command(
+			"sh",
+			"-c",
+			arguments,
+		)
+
+		if output, err := cmd.CombinedOutput(); err != nil {
+			if !strings.Contains(string(output), natRule) {
+				log.Warn("Failed to create pfctl rule: ", cmd.Args, " Returned exit error: ", err.Error(), " Cmd output: ", string(output))
+				log.Flush()
```
Waldz: What is so special about your logs, that you flush it?

nat/service_pfctl.go
Waldz: Could could be just `/sbin/pfctl -vE -f %s`?

nat/service_pfctl.go
zolia: its not, different path of command and different arguments.

nat/service_pfctl.go
zolia: no, its input from stdin (-)

nat/service_pfctl.go
tadovas: There is no need to flush logs on every step. Flushing is done on main package just before program exiting

PR #302, opened 2018-07-20
Better errors handling

soffokl:
Handling of some errors was ignored in several places.


Waldz:
What tool do You use to identify unhandled errors?

soffokl:
@Waldz it just [errcheck](https://github.com/kisielk/errcheck), but it's much easy to use it with tools like [golangci-lint](https://github.com/golangci/golangci-lint) or [gometalinter](https://github.com/alecthomas/gometalinter). 
Both of them could be easily integrated to most of editors or IDE.

PR #303, opened 2018-07-24
Improvement/Merge client and server network options into single structure

tadovas:
1. New testnet option introduced (overrides any previously defined localnet)
2. Client and server network options now share common structure
3. Fuction to parse options from flags
4. Function to construct network definition from provided options


cmd/network.go
```diff
@@ -0,0 +1,87 @@
...
+		&options.Localnet,
+		"localnet",
+		false,
+		"Defines network configuration which expects localy deployed broker and discovery services",
+	)
+
+	flags.BoolVar(
+		&options.Testnet,
+		"testnet",
+		false,
```
Waldz: But `testnet` is true by default

PR #304, opened 2018-07-24
MYST-648 Pack binaries with .exe for Windows

interro:


bin/client_package_standalone
```diff
@@ -29,7 +29,13 @@ DIR_BUILD="build/package"
 # Put all files to package directory (to avoid preserving directory structure)
 printf "Building Standalone package for OS '$OS' ..\n"
 DIR_TEMP=`mktemp -d ${DIR_BUILD}/${tempname}.XXXXXX`
-cp -vp ${BINARY} ${DIR_TEMP}/mysterium_client
+
+if [ "$OS" == "windows" ]; then
```
tadovas: I suggest simply appending .exe to target filename instead of duplicating copy command. I.e.: 
target_name = mysterium_client
if os == windows {
target_name = target_name + .exe
}
cp ..... $target_name

Waldz:
- [ ] Rebase branch onto `release/0.2`

bin/client_package_standalone
interro: Thanks, suggestion is nice to have, however I think we should not optimise here, because these are OS specific things and thus script is more readable

tadovas: OK, fair enough

PR #312, opened 2018-08-13
Updated go-homedir

ignasbernotas:
`mysterium_client` failed to run in launchd context due to bug in go-homedir. This PR updates package version which includes a needed fix.

https://github.com/mitchellh/go-homedir/pull/24

### DIFF
--- a/glide.yaml
+++ b/glide.yaml
@@ -15,7 +15,7 @@ import:
 - package: github.com/julienschmidt/httprouter
   version: 1.1.0
 - package: github.com/mitchellh/go-homedir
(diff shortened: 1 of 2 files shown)


glide.yaml
```diff
@@ -15,7 +15,7 @@ import:
 - package: github.com/julienschmidt/httprouter
   version: 1.1.0
 - package: github.com/mitchellh/go-homedir
-  version: b8bc1bf767474819792c23f32d8286a45736f1c6
```
Waldz: Should update lock file also

ignasbernotas: It's already updated

PR #314, opened 2018-08-20
MYST-681 Start having command "mysterium_node"

Waldz:


bin/run
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
```
soffokl: `/bin/run` and `bin/client_run` files looks 100% equal. Do we need to keep both?

cmd/mysterium_node/mysterium_node.go
```diff
@@ -0,0 +1,89 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package main
+
+import (
+	"fmt"
```
soffokl: What do you think about start using [goimorts](http://goinbigdata.com/goimports-vs-gofmt/) for formatting a file? It groups imports into 3 parts: 1 - standard library; 2 - external dependencies; 3 - internal dependencies, which makes much easy to read a list of imports.

bin/build
```diff
@@ -0,0 +1,36 @@
...
+. ./bin/helpers/functions.sh
+
+export GOOS=${GOOS:-`go env GOHOSTOS`}
+export GOARCH=${GOARCH:-`go env GOHOSTARCH`}
+export GOBIN=`pwd`/build/node
+echo "Compiling 'mysterium_node' for '$GOOS/$GOARCH'.."
+
+go install -ldflags="$(get_linker_ldflags)" cmd/mysterium_node/mysterium_node.go
+if [ $? -ne 0 ]; then
+    printf "\e[0;31m%s\e[0m\n" "Compile failed!"
```
soffokl: There is a `bin/helpers/output.sh` with a `print_error` function, that already added for printing colored output.

cmd/directories.go
```diff
@@ -44,7 +44,7 @@ func ParseFromCmdArgs(flags *flag.FlagSet, options *DirectoryOptions) error {
 		return err
 	}
 
-	currentDir, err := os.Getwd()
+	currentDir, err := os.Executable()
```
tadovas: Why Getwd doesn't fit ?

cmd/mysterium_node/mysterium_node.go
tadovas: We are using goimports on daily basis ( at least some of us). There is a task to migrate full code base and add checks to CI

bin/run
Waldz: We will drop `bin/client_*` and `bin/server_` then single executable will be ready and stable

cmd/mysterium_node/mysterium_node.go
Waldz: Fixed imports for my changed files

bin/build
Waldz: Improved

cmd/directories.go
Waldz: I made command runnable from any location in filesystem:
```
./build/client/mysterium_client
```

tadovas: so basically current dir is a path to executable ?

Waldz: Yes, exactly

PR #319, opened 2018-08-24
Migration from glide to dep

soffokl:


Gopkg.toml
```diff
@@ -0,0 +1,70 @@
...
+  name = "github.com/stretchr/testify"
+  version = "1.2.2"
+
+[prune]
+  go-tests = true
+  unused-packages = true
+
+  [[prune.project]]
+    name = "github.com/ethereum/go-ethereum"
+    unused-packages = false
```
tadovas: New line expected

Gopkg.toml
```diff
@@ -0,0 +1,70 @@
...
+
+[[constraint]]
+  name = "github.com/stretchr/testify"
+  version = "1.2.2"
+
+[prune]
+  go-tests = true
+  unused-packages = true
+
+  [[prune.project]]
```
tadovas: Global unused-packages is a little evil I agree - so no we can specificy this flag per dependency? It would be a bit tricky to track if we need this flag for specific depedency or not - especially if it has transitive dependencies. and this flag is very important if package uses cgo - as any .c .h files will be pruned if this flag is true. That's a bit of consideration

Gopkg.toml
soffokl: Added

Gopkg.toml
soffokl: Not sure when it was introduced. 
Found the only bug when it was broken: https://github.com/golang/dep/issues/1561 )

.travis.yml
```diff
@@ -38,8 +34,8 @@ jobs:
     - stage: dep-cache
       name: "Vendor update"
       script:
-        - source bin/travis_scripts/ensure_glide.sh $BUILD_TOOLS_PATH "v0.13.1"
-        - glide "-home" $GLIDE_HOME install
+        - curl https://raw.githubusercontent.com/golang/dep/master/install.sh | sh
```
Waldz: What about `go get ..` way?

CONTRIBUTING.md
```diff
@@ -6,7 +6,7 @@
 * **Step 1.** Get project dependencies
 ```bash
 brew install go
-brew install glide
+brew install dep
```
Waldz: We collect our development dependencies into `node-builder` image here `bin/builder_docker/Dockerfile`

.travis.yml
soffokl: > It is strongly [recommended](https://github.com/golang/dep#installation) that you use a released version.

And `go get` requires to get sources and build it, so it probably a bit slower than download just binary.

soffokl:
@Waldz, @tadovas, @zolia could you please review it again, I have rebased it to master.

PR #321, opened 2018-08-27
MYST-178. Fixed import path in packages

soffokl:


tadovas:
@Waldz  we discussed about this long time ago. I cannot remember what approach or conclusion we made.

.travis.yml
```diff
@@ -2,7 +2,7 @@ sudo: true
 os: linux
 language: go
 go: [1.9.2]
-go_import_path: github.com/mysterium/node
+go_import_path: github.com/MysteriumNetwork/node
```
zolia: I am not sure about upper cased MysteriumNetwork, apart that it looks bad its again common practice and "not all file systems are case sensitive." (source: https://talks.golang.org/2014/names.slide#16)

soffokl: If the file system is case insensitive it will support both paths for import, the upper case will break nothing here. 
And [here](https://github.com/golang/go/issues/6430#issuecomment-66085787) is a comment, that describes that path of the URL is case sensitive since go get use it, it should support it.

Waldz: A. I would go for `github.com/mysterium-network/node`
or
B. A. Rename our organization to `github.com/mysterium-network/node`
or
C. Event rename our repo to `github.com/mysterium-network/go-node`

zolia: Working with such repo in case-insensitive env might produce code that will be interpreted in other OS'es as lower case thus braking work flow per-se.

soffokl: For my any of **A, B** or **C** is good.
I just don't like the way when we need to rename cloned repo to start a build.

soffokl: My vote is for `github.com/mysterium-network/node`.

Waldz: I like github.com/mysteriumnetwork/go-node

Waldz: I renamed organization. Go ahead with name `github.com/mysteriumnetwork/node`

soffokl:
@Waldz @tadovas @zolia this PR is ready for review.

.travis.yml
```diff
@@ -2,7 +2,7 @@ sudo: true
 os: linux
 language: go
 go: [1.9.2]
-go_import_path: github.com/mysterium/node
+go_import_path: github.com/mysteriumnetwork/node
```
tadovas: I guess this is not needed on travis anymore

CONTRIBUTING.md
```diff
@@ -10,8 +10,8 @@ brew install dep
 brew install openvpn
 
 export GOPATH=~/workspace/go
-git clone git@github.com:MysteriumNetwork/node.git $GOPATH/src/github.com/mysterium/node
-cd $GOPATH/src/github.com/mysterium/node
+git clone git@github.com:MysteriumNetwork/node.git $GOPATH/src/github.com/mysteriumnetwork/node
```
tadovas: What about 'git clone ..MysteriumNetwork..' ? seems like leftover?

.travis.yml
soffokl: Looks like this, removed it.

CONTRIBUTING.md
soffokl: 👍 right, fixed too.

PR #322, opened 2018-08-27
MYST-304. Enabled data race detector for unit tests

soffokl:
Some tests were fixed to allow passing with enabled data race detector.

### DIFF
--- a/client/connection/manager.go
+++ b/client/connection/manager.go
@@ -19,6 +19,7 @@ package connection
 
 import (
 	"errors"
+	"sync"
 
 	log "github.com/cihub/seelog"
 	"github.com/mysterium/node/communication"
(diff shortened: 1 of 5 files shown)


client/connection/manager.go
```diff
@@ -151,23 +161,31 @@ func (manager *connectionManager) startConnection(consumerID, providerID identit
...
 	}
+	manager.mutex.Unlock()
 
 	go openvpnClientWaiter(openvpnClient, dialog)
 	go manager.consumeOpenvpnStates(stateChannel, vpnSession.ID)
 	return nil
 }
 
 func (manager *connectionManager) Status() ConnectionStatus {
+	manager.mutex.RLock()
```
Waldz: Is there a point to block simultaneous status checks?

soffokl: This is only read-lock, multiple reads can be done at the same time. 
The point for this lock was a complaint by the data race detector.

PR #327, opened 2018-08-31
CI builds for all artifacts with a single client+server binary

soffokl:


bin/package_debian
```diff
@@ -0,0 +1,89 @@
+#!/bin/bash
```
Waldz: Would be nice to move `bin/server_*` scripts in some manner that we dont loose Git history

PR #328, opened 2018-09-05
Fixed of building docker images for Ubuntu

soffokl:


soffokl:
@tadovas looks like it's faster to build every time, then pass binaries via the cache, with these staged builds.

PR #331, opened 2018-09-09
MYST-726 Make openvpn tunnel setup mandatory for process

Waldz:


openvpn/process_factory.go
```diff
@@ -25,7 +25,11 @@ import (
 )
 
 // CreateNewProcess creates new openvpn process with given config params
-func CreateNewProcess(openvpnBinary string, config *config.GenericConfig, middlewares ...management.Middleware) Process {
+func CreateNewProcess(
```
soffokl: It feels like one-line function declaration was more readable for me.

openvpn/process_factory_linux.go
```diff
@@ -22,9 +22,21 @@ package openvpn
...
 	"github.com/mysteriumnetwork/node/openvpn/config"
 	"github.com/mysteriumnetwork/node/openvpn/management"
+	"github.com/mysteriumnetwork/node/openvpn/tunnel"
 )
 
-// CreateNewProcess function creates new linux process and overrides default function in linux environment
-func CreateNewProcess(openvpnBinary string, config *config.GenericConfig, middlewares ...management.Middleware) Process {
-	return NewLinuxProcess(openvpnBinary, config, middlewares...)
+// CreateNewProcess function creates Linux OS customized openvpn process
+func CreateNewProcess(
```
soffokl: The same here:
> It feels like one-line function declaration was more readable for me.

openvpn/tunnel/setup_noop.go
```diff
@@ -0,0 +1,36 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package tunnel
+
+import "github.com/mysteriumnetwork/node/openvpn/config"
+
+// NewNoopSetup creates no-op tunnel setup
+func NewNoopSetup() *noopSetup {
```
soffokl: Maybe just make the struct public, if we are doing nothing in the creation function?

openvpn/process.go
```diff
@@ -24,32 +24,42 @@ import (
...
-type Process interface {
-	Start() error
-	Wait() error
-	Stop()
+type tunnelSetup interface {
+	Setup(config *config.GenericConfig) error
+	Teardown()
 }
 
 type openvpnProcess struct {
```
tadovas: As far as I remember OpenvpnProcess is part of public api ( it's returned from package to outside world) and can be safely made public. As general approach - structures returned by New.. functions should always be exported ones. Even linter mentions that.

openvpn/process.go
```diff
@@ -108,5 +122,11 @@ func (openvpn *openvpnProcess) Stop() {
 		openvpn.management.Stop()
 	}()
 
+	waiter.Add(1)
+	go func() {
+		defer waiter.Done()
+		openvpn.tunnelSetup.Teardown()
```
tadovas: This tunnel tear down must be called ONLY AFTER openvnp process is completed, to avoid funny openvpn or tunnel setup fail errors. It cannot be go routined.

openvpn/process.go
zolia: 👍 I think the same

openvpn/process_factory.go
zolia: it is better shorter line to comply with 80 symbols length style.

zolia: I would just call it 'NewProcess', Create and New its two of the same.

openvpn/tunnel/setup_noop.go
zolia: 👍 I also think we should move to such style. Especially for not over complicated initialisation of structs.

openvpn/tunnel/setup_tun_interface.go
```diff
@@ -32,88 +33,107 @@ const tunLogPrefix = "[linux tun service] "
...
+	}
+
+	service.device = device
+	config.SetPersistTun()
+	config.SetDevice(device.Name)
+	return nil
 }
 
-func (service *serviceLinuxTun) Stop() {
+func (service *serviceLinuxTun) Teardown() {
```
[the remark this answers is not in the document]
tadovas: I don't know why we are calling it tun service at all, it's simple device manager with CreateNewDevice, RemoveDevice

openvpn/process_factory.go
soffokl: I think 80 symbols lines is not a limit anymore:
Here is a mention about it in the [Effective Go](https://golang.org/doc/effective_go.html):
> Line length
> Go has no line length limit. Don't worry about overflowing a punched card.

tadovas: But humans do have line length limit, it's all about readability I guess

soffokl: Yep, that is why I mentioned that single line is looking more readable for me.

And another link to the line length in golang: https://github.com/golang/go/wiki/CodeReviewComments#line-length

openvpn/tunnel/setup_tun_interface.go
```diff
@@ -15,15 +15,16 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-package linux
+package tunnel
```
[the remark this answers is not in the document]
vkuznecovas: I don't think this single file deserves to be a library on its own, especially since we're moving openvpn away from the node repo. Can we leave it here for now and if needed split at a later date?

openvpn/process_factory_linux.go
```diff
@@ -22,9 +22,17 @@ package openvpn
...
 )
 
-// CreateNewProcess function creates new linux process and overrides default function in linux environment
-func CreateNewProcess(openvpnBinary string, config *config.GenericConfig, middlewares ...management.Middleware) Process {
-	return NewLinuxProcess(openvpnBinary, config, middlewares...)
+// CreateNewProcess function creates Linux OS customized openvpn process
+func CreateNewProcess(openvpnBinary string, configuration *config.GenericConfig, middlewares ...management.Middleware) *OpenvpnProcess {
+	configuration.SetScriptParam("iproute", config.SimplePath("nonpriv-ip"))
+
+	tunnelSetup := tunnel.NewTunInterfaceSetup(
```
Waldz: +1

openvpn/tunnel/setup_tun_interface.go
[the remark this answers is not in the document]
Waldz: Both ideas is good, but lets do it then we need to reuse

Waldz:
LGTM

openvpn/tunnel/setup_tun.go
```diff
@@ -0,0 +1,25 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package tunnel
+
+// GetTunnelSetup sets the required configuration options for the tunnel and returns a Setup for it
+func GetTunnelSetup() Setup {
```
Waldz: We tend to call factories `New*()` or `new*()`

openvpn/process.go
```diff
@@ -24,24 +24,20 @@ import (
 
 	"github.com/mysteriumnetwork/node/openvpn/config"
 	"github.com/mysteriumnetwork/node/openvpn/management"
+	"github.com/mysteriumnetwork/node/openvpn/tunnel"
 )
 
-type tunnelSetup interface {
```
Waldz: You reverted 2 improvements
- Interface should be on caller side. Not in implementation
- Implementation factory should return not interface but exact structs `new() return *SetupTunInterface`

Waldz: @tadovas, @sofokkl we kind of agreed to continue like this. But now we have 2 conventions and that missleads

tadovas: In this case the only option is to return interface, if we want to choose os specific implementation on compile time, and hide those details from process (generic tun setup on mac, win) (specific tun on linux), 
On the other hand, we can have two functions with the same name and conditionaliy compiled, but with different return values, i.e. NewTunnelSetup() * LinuxTunnel NewTunnelSetup() * GenericTunnel, but I think it will hurt readability as the same function will return different result on different OS. I would vote for interface declaration on implementor side in this particular case.

tadovas: Return structures depend on interfaces is generic golang approach and it's fine, but there are edge cases where you have to choose :) I think it's the case

Waldz: > ..if we want to choose os specific implementation on compile time.. 

To solve this i moved conditional factory to upper layer:
+ caller defines interface `TunnelSetup` https://github.com/mysteriumnetwork/node/commit/84b2b3b3d698ebae264f8bb041c6478faea7054a

+ caller decides implementation to go for Linux/ARM because he knows where he is compiling

+ Caller can construct `SetupTunInterface` in ARM. And now is made private

This was idea - to make `tunnel` package independent from our process so that someone can use just Tunnel seperately.

PR #362, opened 2018-09-13
Consumer issues new promise for the provider

soffokl:
This PR includes "Promise balance notifications": https://github.com/mysteriumnetwork/node/pull/381


core/connection/manager_test.go
```diff
@@ -91,6 +94,11 @@ func (tc *testContext) SetupTest() {
 		sync.RWMutex{},
 	}
 
+	promiseFactory := func(consumerID, providerID identity.Identity, amount money.Money) (*promise.SignedPromise, error) {
```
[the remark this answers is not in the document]
Waldz: I am making this rename in my branch

soffokl:
@tadovas @vkuznecovas @Waldz @zolia
the basic promise issuing ready for review

core/promise/producer.go
```diff
@@ -0,0 +1,55 @@
...
+)
+
+// Producer creates requests/responses of the "promise-create" events
+type Producer struct {
+	SignedPromise *SignedPromise
+}
+
+// Request structure represents message from service provider to receive new promise from consumer
+type Request struct {
+	SignedPromise *SignedPromise
```
zolia: Why to have SignedPromise inside Request? Its just 1 member, why not simply SignedPromise?

cmd/flags_node.go
```diff
@@ -34,6 +34,17 @@ var (
...
 	}
+
+	promiseCurrencyFlag = cli.StringFlag{
+		Name:  "promise.currency",
+		Usage: "Type of currency that will be used for issuing promises",
+		Value: "MYST",
+	}
+	promiseAmountFlag = cli.IntFlag{
+		Name:  "promise.amount",
+		Usage: "Amount of money that will be used for issuing a single promise",
```
zolia: We should specify format for a user.  I.e: Integer, 1000 == 1 MYST

core/promise/ballance_consumer.go
```diff
@@ -0,0 +1,44 @@
...
+
+// GetMessageEndpoint returns endpoint there to receive messages
+func (consumer *BalanceMessageConsumer) GetMessageEndpoint() communication.MessageEndpoint {
+	return balanceEndpoint
+}
+
+// NewMessage creates struct where message from endpoint will be serialized
+func (consumer *BalanceMessageConsumer) NewMessage() (messagePtr interface{}) {
+	var message BalanceMessage
+	return &message
```
zolia: we could just write `retrun &BalanceMessage{}`

core/promise/ballance_consumer.go
```diff
@@ -0,0 +1,44 @@
...
+
+package promise
+
+import (
+	"github.com/mysteriumnetwork/node/communication"
+)
+
+// BalanceMessageConsumer process balance notification from communication channel.
+type BalanceMessageConsumer struct {
+	Callback func(*BalanceMessage)
```
zolia: Can we name it ConsumeCallback ?

core/promise/ballance_consumer.go
```diff
@@ -0,0 +1,44 @@
...
+// NewMessage creates struct where message from endpoint will be serialized
+func (consumer *BalanceMessageConsumer) NewMessage() (messagePtr interface{}) {
+	var message BalanceMessage
+	return &message
+}
+
+// Consume handles messages from endpoint
+func (consumer *BalanceMessageConsumer) Consume(messagePtr interface{}) error {
+	consumer.Callback(messagePtr.(*BalanceMessage))
+	return nil
```
zolia: wired to have error contract, but no errors produced..

cmd/flags_node.go
soffokl: I was thinking that it will be minimal allowed value, as a `wei`

core/promise/producer.go
soffokl: I think keeping `request/response` here, how it used in the communication dialog makes it more clear for understanding.

cmd/flags_node.go
Waldz: Do we need configurable promise at all at this stage.
We agreed to issue `promise` which has full amount from `serviceProposal`

core/connection/manager_test.go
```diff
@@ -369,15 +370,25 @@ func (fd *fakeDialog) Send(producer communication.MessageProducer) error {
...
+				Success: true,
+				Message: "Everything is great!",
 			},
-		},
-		nil
+			nil
+	case *session.CreateProducer:
+		return &session.CreateResponse{
+				Success: true,
+				Message: "Everything is great!",
```
Waldz: This field is used for errors only `Success: false`

core/node/node.go
```diff
@@ -54,8 +55,15 @@ func NewNode(
 		return dialogEstablisher.EstablishDialog(providerID, contact)
 	}
 
-	promiseIssuerFactory := func(dialog communication.Dialog) connection.PromiseIssuer {
-		return noop.NewPromiseIssuer(dialog)
+	promiseIssuerFactory := func(issuerID identity.Identity, dialog communication.Dialog) connection.PromiseIssuer {
+		return &noop.PromiseIssuer{
```
Waldz: Use factory here

core/node/node.go
```diff
@@ -54,8 +55,15 @@ func NewNode(
...
+			IssuerID: issuerID,
+			Dialog:   dialog,
+			Signer:   signerFactory(issuerID),
+			Amount: money.Money{
```
Waldz: You are getting `issuer.Start(proposal)` and that data should be enough to create promises, because it holds `primise.PaymentMethod`.
Probably we need something as `NoopPaymentMethod`

Later:
We agreed to create `PromisePolicy` interface , which answers to political questions. We can implement it with later PR.

core/promise/consumer.go
```diff
@@ -0,0 +1,48 @@
...
+func (c *consumer) Consume(requestPtr interface{}) (response interface{}, err error) {
+	// request := requestPtr.(*Request)
+
+	// TODO there should be some validation of the received proposal and storing it somewhere for the server needs.
+	// TODO signature validation of the promise should be here too.
+
+	// if request.SignedPromise.IssuerSignature == valid {
+	// 		return &Response{
+	// 			Success: false,
+	// 			Message: fmt.Sprintf("Bas signature: %s", request.SignedPromise.IssuerSignature),
```
Waldz: Do we need this unfinished code?

Waldz: Maybe You can add this validation with separate PR

core/promise/methods/noop/promise_issuer.go
```diff
@@ -23,21 +23,19 @@ import (
...
-		dialog: dialog,
-	}
-}
-
-// PromiseIssuer issues promises in such way, what no actual money is added to promise
+// PromiseIssuer issues promises in such way, that no actual money is added to promise
 type PromiseIssuer struct {
-	dialog communication.Dialog
+	IssuerID identity.Identity
+	Dialog   communication.Dialog
```
Waldz: Private attributes and factories. Caller will now know how to initiate all runtime variables

session/create_producer.go
```diff
@@ -24,27 +24,31 @@ import (
...
+// Produce creates requests for the endpoint
+func (producer *CreateProducer) Produce() (requestPtr interface{}) {
 	return &CreateRequest{
 		ProposalId: producer.ProposalID,
 	}
 }
 
 // RequestSessionCreate requests session creation and returns session DTO
 func RequestSessionCreate(sender communication.Sender, proposalID int, sessionPtr *Session) error {
-	responsePtr, err := sender.Request(&createProducer{
```
Waldz: Why to make public? Why to refactor session at all?

core/promise/promise.go
```diff
@@ -0,0 +1,69 @@
...
+	signature, err := issuerSigner.Sign(out)
+
+	return &SignedPromise{
+		Promise:         *promise,
+		IssuerSignature: Signature(signature.Base64()),
+	}, err
+}
+
+// Send sends signed promise via the communication channel
+func Send(signedPromise *SignedPromise, sender communication.Sender) (*Response, error) {
```
Waldz: Does it belong to DTO file?
Maybe should be similar approach as `RequestSessionCreate`- helper on `provider` side

core/promise/promise.go
```diff
@@ -0,0 +1,69 @@
...
+	responsePtr, err := sender.Request(&Producer{
+		SignedPromise: signedPromise,
+	})
+
+	response := responsePtr.(*Response)
+	if err != nil || !response.Success {
+		return nil, errors.New("Promise issuing failed: " + response.Message)
+	}
+
+	return response, nil
```
Waldz: Do we need response that promise was received? What data will be there?
Very  hard to see the case without `Send` being called in some actual place

core/promise/consumer.go
soffokl: it's already coming with a https://github.com/mysteriumnetwork/node/pull/404

core/promise/promise.go
soffokl: Removed response.

core/promise/promise.go
soffokl: I think Send is related only to `SignedPromise`, I have changed it to be a method of it to point it explicitly. It was added to the `promise.go` file since it related only to promises.

core/promise/methods/noop/promise_issuer.go
soffokl: Changed it to be a private.

core/node/node.go
soffokl: Changed it to getting `Amount` from the `promise`

core/connection/manager_test.go
soffokl: Removed `Message`.

core/node/node.go
soffokl: Changed it to use `NewPromiseIssuer`.

core/promise/consumer.go
soffokl: It was added as `TODO` example, next PR is adding real validation.

core/promise/producer.go
Waldz: Keep in mind, that will send SignedPromise structure to blockhain for clearing reasons

PR #380, opened 2018-09-18
Decouple session storage

Waldz:
updates #382


services/openvpn/session/client_map.go
```diff
@@ -25,14 +25,14 @@ import (
 )
 
 type clientMap struct {
-	sessionManager   session.Manager
+	sessions         *session.Storage
```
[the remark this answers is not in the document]
tadovas: clientMap ? what happened to agreement not to use terms "client" "server". Looks like ServiceConsumerMap

services/openvpn/session/validator.go
```diff
@@ -34,10 +34,10 @@ type Validator struct {
 }
 
 // NewValidator return Validator instance
-func NewValidator(sessionManager session.Manager, extractor identity.Extractor) *Validator {
+func NewValidator(sessionStorage *session.Storage, extractor identity.Extractor) *Validator {
```
tadovas: Depend on interfaces not on structures

services/openvpn/session/validator_mocks.go
```diff
@@ -22,10 +22,21 @@ import (
 	"github.com/mysteriumnetwork/node/session"
 )
 
-const mockedVPNConfig = "config_string"
+func mockStorage(sessions ...session.Session) *session.Storage {
```
tadovas: Doesn't look like mock storage to me, but a real prepopulated storage. Again - missing Storage interface ?

session/storage.go
```diff
@@ -0,0 +1,58 @@
...
+// NewStorage initiates new session storage
+func NewStorage() *Storage {
+	return &Storage{
+		sessionMap: make(map[SessionID]Session),
+		lock:       sync.Mutex{},
+	}
+}
+
+// Storage maintains a map of session id -> session
+type Storage struct {
```
tadovas: This can be called InMemoryStorage, and caller can define Storage interface with expected operations

services/openvpn/session/client_map.go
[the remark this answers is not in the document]
Waldz: This term `client` from from Openvpn management notifications.
And this map saves this `clientID -> sessions` metadata

services/openvpn/session/validator.go
Waldz: Refactored to 2 different interfaces on caller side

services/openvpn/session/validator_mocks.go
Waldz: Mocked storage

session/storage.go
Waldz: Renamed to `StorageMemory`

session/manager.go
```diff
@@ -27,46 +27,39 @@ import (
 type ServiceConfigProvider func() (ServiceConfiguration, error)
 
 // NewManager returns session manager which maintains a map of session id -> session
-func NewManager(serviceConfigProvider ServiceConfigProvider, idGenerator Generator) *manager {
+func NewManager(
```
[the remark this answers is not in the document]
Waldz: It's just 3 arguments. We make option map just >4 arguments, AFAIK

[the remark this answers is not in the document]
Waldz: Improved arguments to fit one line

Waldz:
All comments fixed

session/storage_memory.go
```diff
@@ -0,0 +1,59 @@
...
+}
+
+// StorageMemory maintains a map of session id -> session
+type StorageMemory struct {
+	sessionMap map[SessionID]Session
+	lock       sync.Mutex
+}
+
+// Add puts given session to storage. Multiple sessions per peerID is possible in case different services are used
+func (storage *StorageMemory) Add(sessionInstance Session) error {
```
soffokl: Looks like no error could be returned here, should we keep it in the function declaration?

session/manager.go
```diff
@@ -26,47 +26,41 @@ import (
 // ServiceConfigProvider defines configuration providing dependency
 type ServiceConfigProvider func() (ServiceConfiguration, error)
 
+// IDGenerator defines method for session id generation
+type IDGenerator func() SessionID
```
tadovas: Can we assume, that future implementations will always succeed and never posibliy return error? I.e. true crypto random generator? Maybe it's worth to define error as part of result - it's more relaxed definition.

session/storage_memory.go
Waldz: Fixed

bin/check_golint
```diff
@@ -42,7 +42,6 @@ MESSAGES_RECONFIGURE=()
 
 check_uncleaned_package "github.com/mysteriumnetwork/node/identity" 6
 check_uncleaned_package "github.com/mysteriumnetwork/node/tequilapi" 11
-check_uncleaned_package "github.com/mysteriumnetwork/node/session" 6
```
tadovas: Extra karma points for cleaning up packages

session/manager.go
Waldz: I just remove error from `SessionMap` just because current implementation `storageMemory` dont return error (because of feadback).
So.. it look that dont do assumtions on future, w;ll implement when needed.

bin/check_golint
Waldz: So where is my approval?

tadovas:
Fine by me, as long as @vkuznecovas happy :)

Issue #363, opened 2018-09-13
Release 0.3 version with Paid ID


Waldz:
Acceptance Criteria:
- make feature toggle `myst service --experiment-identity-check` and DI registers `IdentityRegisterNoop`
- Create branch `release/0.3` and tag release `0.3.0`
- Deploy WITH `--experiment-identity-check` to Devnet servers
- Deploy WITHOUT `--experiment-identity-check` to 1 Testnet server
- Test http://github.com/mysteriumnetwork/mysterium-vpn 0.2.* version with 1 new Testnet server
- Deploy WITHOUT `--experiment-identity-check` to all Testnet servers

soffokl:
Looks like everything done in the scope of the issue.

PR #383, opened 2018-09-19
Identity check disabled by default, added flag to enable it. #363

soffokl:


core/node/options_network.go
```diff
@@ -22,6 +22,8 @@ type OptionsNetwork struct {
 	Testnet  bool
 	Localnet bool
 
+	IdentityCheck bool
```
Waldz: I would like these options to have `experiment` prefix or suffix. Because we will be deleting them later

soffokl: Agree, fixed it.

tadovas: 👍 absolutely

PR #384, opened 2018-09-19
Fixed uploading release images

soffokl:


soffokl:
yep, 
`Error response from daemon: No such image: myst:alpine`
since the image was not imported from the previous stage.

This PR adds a missing step of downloading images.

PR #388, opened 2018-09-19
Fixed upload artifacts to release and set iptables rules

soffokl:


nat/service_iptables.go
```diff
@@ -34,17 +34,9 @@ func (service *serviceIPTables) Add(rule RuleForwarding) {
 }
 
 func (service *serviceIPTables) Start() error {
-	err := service.ipForward.Enable()
-	if err != nil {
```
Waldz: Maybe we should skipp this error here, in high level logic.
But lets have `ipForward.Enable` not swelling errors

soffokl: Reverted error returning

nat/service_pfctl.go
```diff
@@ -38,17 +38,9 @@ func (service *servicePFCtl) Add(rule RuleForwarding) {
 }
 
 func (service *servicePFCtl) Start() error {
-	err := service.ipForward.Enable()
-	if err != nil {
-		return err
-	}
-
+	service.ipForward.Enable()
```
Waldz: Warning logging whould be exactly here in high logic.
So that everybody understands why we swallow errors

soffokl: We have logging in the `Enable`, do you want to duplicate it or move it to `Start` only?

soffokl: Added extra log line.

nat/service_pfctl.go
```diff
@@ -40,15 +40,11 @@ func (service *servicePFCtl) Add(rule RuleForwarding) {
 func (service *servicePFCtl) Start() error {
 	err := service.ipForward.Enable()
 	if err != nil {
-		return err
+		log.Warn(natLogPrefix, "Failed to enable IP forwarding: ", err)
```
Waldz: Now it's double loging, is not it?

soffokl: yeah, it prints messages from both functions.

soffokl: I think we can leave it this way for the release.

PR #403, opened 2018-09-24
CI improvements

Waldz:


bin/builder_run
```diff
@@ -23,5 +23,5 @@ if [ -z "$COMMAND" ]; then
 fi
 
 docker run --rm -ti \
-    -v $(pwd):/go/src/github.com/mysteriumnetwork/node mysteriumnetwork/mysterium-node-builder \
-    -c "cd /go/src/github.com/mysteriumnetwork/node && ${COMMAND}"
+    -v $(pwd):/go/src/github.com/mysteriumnetwork/node \
```
soffokl: Default working directory is `/go/src/github.com/mysterium/node` which does not exist at the moment.
We can rebuild docker image or just change the working directory during the start time.

Waldz: Yes, I rebuilt `builder` image

soffokl: hm, but it failed on CI

Waldz: Failed to push. Fixed

PR #404, opened 2018-09-25
Added promises validation and storing for the service provider. #347

### LINKED ISSUE #347: Payments. Provider stores Consumer promises (state)

soffokl:
Closes #347


core/promise/consumer.go
```diff
@@ -18,10 +18,31 @@
...
+	ethereum "github.com/ethereum/go-ethereum"
+	"github.com/ethereum/go-ethereum/common"
 	"github.com/mysteriumnetwork/node/communication"
+	"github.com/mysteriumnetwork/node/core/promise/storage"
+	"github.com/mysteriumnetwork/node/identity"
+	"github.com/mysteriumnetwork/node/service_discovery/dto"
+)
+
+var (
+	errLowAmount          = fmt.Errorf("promise amount less then the service proposal price")
```
vkuznecovas: typo "then" -> "than"

core/promise/consumer.go
```diff
@@ -18,10 +18,31 @@
...
+	errLowBalance         = fmt.Errorf("issuer balance less then the promise amount")
```
vkuznecovas: typo "then" -> "than"

core/promise/consumer.go
```diff
@@ -32,17 +53,60 @@ func (c *consumer) NewRequest() (requestPtr interface{}) {
...
+	}
+
+	receivedPromise, err := json.Marshal(request.SignedPromise.Promise)
+	if err != nil {
+		return nil, err
+	}
+
+	signature := identity.SignatureBase64(string(request.SignedPromise.IssuerSignature))
+	issuer := identity.FromAddress(request.SignedPromise.Promise.IssuerID)
+	verify := identity.NewVerifierIdentity(issuer)
```
vkuznecovas: this could be better named as verifier, since the line below "verify.Verify" does seem a bit confusing

core/promise/consumer.go
```diff
@@ -32,17 +53,60 @@ func (c *consumer) NewRequest() (requestPtr interface{}) {
...
+			Request: request,
+		}, nil
+	}
 
-	// TODO there should be some validation of the received proposal and storing it somewhere for the server needs.
-	// TODO signature validation of the promise should be here too.
+	balance, err := c.etherClient.BalanceAt(context.Background(), common.HexToAddress(issuer.Address), nil)
+	if err != nil {
+		return nil, err
+	} else if balance.Uint64() < amount {
```
vkuznecovas: redundant else

core/promise/consumer.go
```diff
@@ -32,17 +53,60 @@ func (c *consumer) NewRequest() (requestPtr interface{}) {
...
+		return &Response{
```
vkuznecovas: this forming of responses seems to be a bit repetitive, maybe create a helper function like "ResponseFromError(err error, request)"

soffokl:
@vkuznecovas comments are fixed.

core/promise/consumer.go
```diff
@@ -18,10 +18,31 @@
...
+	"github.com/ethereum/go-ethereum/common"
 	"github.com/mysteriumnetwork/node/communication"
+	"github.com/mysteriumnetwork/node/core/promise/storage"
+	"github.com/mysteriumnetwork/node/identity"
+	"github.com/mysteriumnetwork/node/service_discovery/dto"
 )
 
-type consumer struct{}
+var (
+	errLowAmount          = fmt.Errorf("promise amount less than the service proposal price")
```
zolia: Why not errors.New()? You dont need formatting in this case.

soffokl: Well, both `errors.New()` and `fmt.Errorf()` doing the same under the hood. 
It's more style question. Do we have any recommendation which one to use in the project?

zolia: Most of the time we used errors.New(), sorta indicates that we are creating "New" message.

soffokl: ok, clear, changed it.

core/promise/consumer.go
```diff
@@ -36,17 +66,49 @@ func (c *Consumer) NewRequest() (requestPtr interface{}) {
 
 // Consume handles requests from endpoint and replies with response
 func (c *Consumer) Consume(requestPtr interface{}) (response interface{}, err error) {
-	// request := requestPtr.(*Request)
+	request, ok := requestPtr.(*Request)
+	if !ok {
+		return nil, errUnsupportedRequest
```
Waldz: Standardised errors is nice, but it's just for logging AFAIK. What about response to sender?
Did you look how responses are standardised at `session/create_dto.go`?

core/promise/consumer.go
```diff
@@ -36,17 +66,49 @@ func (c *Consumer) NewRequest() (requestPtr interface{}) {
...
+		return nil, err
+	}
 
-	// if request.SignedPromise.IssuerSignature == valid {
-	// 		return &Response{
-	// 			Success: false,
-	// 			Message: fmt.Sprintf("Bas signature: %s", request.SignedPromise.IssuerSignature),
-	// 		}, fmt.Errorf("Bad promise signature")
-	// }
+	signature := identity.SignatureBase64(string(request.SignedPromise.IssuerSignature))
```
Waldz: Are you sure not factory `identity.SignatureBytes()` from []bytes?

core/promise/consumer.go
Waldz: Actually not good that we write a lot of logic in dialog transport layer.
I've just refactored `session/create_consumer.go` to be thin and `consumer.SessionManager.Create()` is fat.
Maybe here some `type ValidatePromise func(signedPromise) error`?

core/promise/consumer.go
```diff
@@ -36,17 +66,49 @@ func (c *Consumer) NewRequest() (requestPtr interface{}) {
...
+	if err := c.storage.Store(issuer.Address, &request.SignedPromise.Promise); err != nil {
+		return nil, err
+	}
+
+	return &Response{Success: true, Message: "Promise accepted", Request: request}, nil
+}
 
-	return &Response{Success: true, Message: "Promise accepted"}, nil
+func failedResponse(err error, request *Request) *Response {
+	return &Response{Success: false, Message: err.Error(), Request: request}
```
Waldz: We dont need to leak text errors. Actually it should be system-only error codes (see `session/create_dto.go`)

core/promise/consumer.go
```diff
@@ -36,17 +66,49 @@ func (c *Consumer) NewRequest() (requestPtr interface{}) {
...
+		return failedResponse(errBadSignature, request), nil
+	}
+
+	benefiter := identity.FromAddress(request.SignedPromise.Promise.BenefiterID)
+	if benefiter.Address != c.proposal.ProviderID {
+		return failedResponse(errUnknownBenefiter, request), nil
+	}
+
+	price := c.proposal.PaymentMethod.GetPrice()
+	amount := request.SignedPromise.Promise.Amount.Amount
```
Waldz: `.Amount.Amount`?

Waldz: `amountPromised`, `feePromised`?

core/promise/consumer.go
```diff
@@ -36,17 +66,49 @@ func (c *Consumer) NewRequest() (requestPtr interface{}) {
...
+	}
+	if balance.Uint64() < amount {
+		return failedResponse(errLowBalance, request), nil
+	}
+
+	if err := c.storage.Store(issuer.Address, &request.SignedPromise.Promise); err != nil {
+		return nil, err
+	}
+
+	return &Response{Success: true, Message: "Promise accepted", Request: request}, nil
```
Waldz: We dont need to resend request back

core/promise/consumer_test.go
```diff
@@ -18,24 +18,157 @@
...
+
+func (fp fakePayment) GetPrice() money.Money {
+	return money.Money{Amount: fp.amount}
+}
+
+type fakeBlockchain struct {
+	balance int64
 }
+
+func (fb *fakeBlockchain) BalanceAt(ctx context.Context, account common.Address, blockNumber *big.Int) (*big.Int, error) {
```
Waldz: If would have smaller interface, You would not need to mock all other functions `StorageAt`, `NonceAt `

core/promise/methods/noop/promise_processor.go
```diff
@@ -23,8 +23,10 @@ import (
 	"time"
 
 	log "github.com/cihub/seelog"
+	ethereum "github.com/ethereum/go-ethereum"
```
Waldz: Lets decouple `PromiseProcessor` from Ethereum

core/promise/methods/noop/promise_processor.go
```diff
@@ -80,6 +87,7 @@ func (processor *PromiseProcessor) Start(proposal discovery_dto.ServiceProposal)
 
 // Stop stops processing promises
 func (processor *PromiseProcessor) Stop() error {
+	processor.storage.Close()
```
Waldz: Can we stop storage it `Node.Stop` because this dependency will be needed in more places

e2e/connection_test.go
```diff
@@ -97,6 +97,9 @@ func consumerPicksProposal(t *testing.T, tequilapi *tequilapi_client.Client) teq
 }
 
 func consumerConnectFlow(t *testing.T, tequilapi *tequilapi_client.Client, consumerID string, proposal tequilapi_client.ProposalDTO) {
+	err := topUpAccount(consumerID)
```
Waldz: +1 niiiice

core/promise/consumer.go
Waldz: I suggest static response with `const responseInsufficient = ..`

core/promise/consumer.go
```diff
@@ -36,17 +66,49 @@ func (c *Consumer) NewRequest() (requestPtr interface{}) {
...
+	amount := request.SignedPromise.Promise.Amount.Amount
+	if amount < price.Amount {
+		return failedResponse(errLowAmount, request), nil
+	}
+
+	balance, err := c.etherClient.BalanceAt(context.Background(), common.HexToAddress(issuer.Address), nil)
+	if err != nil {
+		return nil, err
+	}
+	if balance.Uint64() < amount {
```
Waldz: Can `BalanceRegistry` return already parsed `money`?

core/promise/consumer.go
soffokl: I believe that we need to send error details to the client. 

Errors like `Internal error` from the server says that something happened internally on the server and user should retry request later changing nothing. 

But if we are sending him a message like `your promise format invalid` or `you have too low balance` it says that user did something wrong and he needs to change it before next sending. 

It will be very hard to debug client problem without propper error responses if user uses external exit node. 

I think it's good practice to responce with as much details as we can.

core/promise/consumer.go
soffokl: Changed it to the `promise.Fee`.

core/promise/consumer.go
soffokl: Removed it.

core/promise/consumer_test.go
soffokl: Fixed.

core/promise/methods/noop/promise_processor.go
soffokl: Done.

core/promise/methods/noop/promise_processor.go
soffokl: Added storage stopping to the `di.Shutdown()`

core/promise/consumer.go
soffokl: For `money.Money` we need to know `Currency` to provide it. Left only balance, since we don't know currency at that moment.

core/promise/consumer.go
```diff
@@ -36,17 +62,22 @@ func (c *Consumer) NewRequest() (requestPtr interface{}) {
...
-	// request := requestPtr.(*Request)
+	request, ok := requestPtr.(*Request)
+	if !ok {
+		return failedResponse(errUnsupportedRequest), errUnsupportedRequest
+	}
 
-	// TODO there should be some validation of the received proposal and storing it somewhere for the server needs.
-	// TODO signature validation of the promise should be here too.
+	if err := request.SignedPromise.Validate(c.proposal, c.balanceRegistry); err != nil {
+		return failedResponse(err), nil
```
zolia: prob here we should return err also: `return failedResponse(err), err`

core/promise/consumer.go
```diff
@@ -36,17 +62,22 @@ func (c *Consumer) NewRequest() (requestPtr interface{}) {
...
+	}
 
-	// if request.SignedPromise.IssuerSignature == valid {
-	// 		return &Response{
-	// 			Success: false,
-	// 			Message: fmt.Sprintf("Bas signature: %s", request.SignedPromise.IssuerSignature),
-	// 		}, fmt.Errorf("Bad promise signature")
-	// }
+	if err := c.storage.Store(request.SignedPromise.Promise.IssuerID, &request.SignedPromise.Promise); err != nil {
+		return nil, err
```
zolia: `err` not returned too. We should be able to log all improper activity by client on service log...

core/promise/dto.go
```diff
@@ -27,10 +27,10 @@ type SignedPromise struct {
...
 // Promise represents payment promise between two parties
 type Promise struct {
-	SerialNumber int
-	IssuerID     string
-	BenefiterID  string
-	Amount       money.Money
+	SerialNumber int    `storm:"id"`
+	IssuerID     string `storm:"index"`
+	BenefiterID  string `storm:"index"`
+	Fee          money.Money
```
zolia: I would stick with 'Amount'. Promise usually is not actual 'Fee', but just promised 'Amount'.

core/promise/methods/noop/promise_processor.go
```diff
@@ -63,12 +69,13 @@ type PromiseProcessor struct {
 }
 
 // Start processing promises for given service proposal
-func (processor *PromiseProcessor) Start(proposal discovery_dto.ServiceProposal) error {
+func (processor *PromiseProcessor) Start(proposal dto.ServiceProposal) error {
 	processor.lastPromise = promise.Promise{
-		Amount: money.NewMoney(10, money.CURRENCY_MYST),
+		Fee: money.NewMoney(10, money.CURRENCY_MYST),
```
zolia: Why '10'? We prob should externalise this one.

cmd/di.go
```diff
@@ -154,14 +175,19 @@ func (di *Dependencies) BootstrapServiceComponents(nodeOptions node.Options, ser
...
+
 	di.ServiceManager = service.NewManager(
 		di.NetworkDefinition,
 		identityLoader,
 		di.SignerFactory,
 		di.IdentityRegistry,
 		openvpnServiceManager,
 		func(proposal dto_discovery.ServiceProposal, configProvider session.ConfigProvider) communication.DialogHandler {
-			sessionManagerFactory := newSessionManagerFactory(proposal, configProvider, sessionStorage)
+			promiseHandler := func(dialog communication.Dialog) *noop.PromiseProcessor {
```
zolia: why still noop? Its doing something already :)

identity/balance.go
```diff
@@ -0,0 +1,36 @@
...
+
+import (
+	"context"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/ethclient"
+)
+
+// BalanceRegistry provides amount of money that identity have on the balance in the blockchain
+type BalanceRegistry func(Identity) (uint64, error)
```
zolia: Since it is just 1 function, I would name it with verb, like BalanceCheck, it is always better readable to see some action if you getting some data.

core/promise/consumer.go
Waldz: We already had these discussions and made decisions:
- protocol should not too verbose with errors
- it can be attack vector
- we not building app here, but crystallising protocol messages
- there might be other nodes, which wont respond with real error codes, but will fake them. So you dont control infra and you cant depend on error codes anyway

core/promise/methods/noop/promise_processor.go
Waldz: I think will be improved in upcoming PRs. TODO needed

core/promise/consumer.go
Waldz: Keep in mind, that in protocol we are coding for standardised error-codes, instead of text messages.
Because if it's text messages, which language should it be? English, Esperanto? :)

https://github.com/mysteriumnetwork/node/blob/master/session/create_dto.go

identity/balance.go
soffokl: Changed it to `Balance` since "Effective go" [recomends](https://golang.org/doc/effective_go.html#Getters) such approach for getters.

core/promise/methods/noop/promise_processor.go
```diff
@@ -95,7 +102,7 @@ balanceLoop:
 
 		case <-time.After(processor.balanceInterval):
 			processor.balanceSend(
-				promise.BalanceMessage{1, true, processor.lastPromise.Amount},
+				promise.BalanceMessage{1, true, processor.lastPromise.Fee},
```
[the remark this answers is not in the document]
soffokl: Added TODO

core/storage/storage.go
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package storage
+
+// Storage stores persistent objects for future usage
+type Storage interface {
+	Store(issuer string, data interface{}) error
```
tadovas: Issuer has nice datat type which is called Identity. Can we reuse it here? string type is too generic

core/promise/promise.go
```diff
@@ -65,3 +74,41 @@ func (sp *SignedPromise) Send(sender communication.Sender) error {
 
 	return nil
 }
+
+// Validate check signed promise to be valid. It checks signature, benefiter address.
+// Also it compares the promised amount to be enough for the proposal.
+// And finally it checks that issuer have enough balance to issue the promice.
+func (sp *SignedPromise) Validate(proposal dto.ServiceProposal, balance identity.Balance) error {
```
tadovas: It feels like this validate does too much. Why it has to depend on proposal? Signed promise with valid signature itself is perfectly valid, as both parties agreed on amount and other parameters

core/promise/dto.go
```diff
@@ -27,9 +27,9 @@ type SignedPromise struct {
 
 // Promise represents payment promise between two parties
 type Promise struct {
-	SerialNumber int
-	IssuerID     string
-	BenefiterID  string
+	SerialNumber int    `storm:"id"`
```
tadovas: I really really recommend to start using Promise structures from payments lib, as these promises will land on contract anyway, and a lot of work will need to be done to align them with payment contract expected structure later

core/storage/boltdb/storage.go
```diff
@@ -0,0 +1,60 @@
...
+
+	"github.com/asdine/storm"
+	"github.com/mysteriumnetwork/node/core/storage"
+)
+
+type bolt struct {
+	db *storm.DB
+}
+
+// NewStorage creates a new BoltDB storage for service promises
```
interro: storage can be used for generic objects, for example for sessions too.
Functions comments can be replaced like this:
"for service promises" -> "for storing values"

core/promise/dto.go
soffokl: Looks like it's out of the scope of this PR. 
Created issue for changing it in the same milestone. https://github.com/mysteriumnetwork/node/issues/454

core/storage/boltdb/storage.go
soffokl: For now, it's used only for storing promises, what is the main aim of the PR. 
Let's replace the comment (and probably not comment only) when we will have something different.

PR #406, opened 2018-09-25
Removed building of client and server, only myst left

soffokl:


Makefile
```diff
@@ -25,14 +21,8 @@ dep:
 dep-ensure:
 	dep ensure
 
-server:
+myst:
 	./bin/server_build
```
vkuznecovas: aren't you removing this file?

bin/release_docker
```diff
@@ -37,14 +35,6 @@ while test $# -gt 0; do
     docker_release_image myst:alpine mysteriumnetwork/myst:${VERSION_TAG}
     docker_release_image myst:alpine mysteriumnetwork/myst:${VERSION_TAG}-alpine
     docker_release_image myst:ubuntu mysteriumnetwork/myst:${VERSION_TAG}-ubuntu
-
-    docker_release_image myst:alpine mysteriumnetwork/mysterium-node:${VERSION_TAG}
-    docker_release_image myst:alpine mysteriumnetwork/mysterium-node:${VERSION_TAG}-alpine
-    docker_release_image myst:ubuntu mysteriumnetwork/mysterium-node:${VERSION_TAG}-ubuntu
```
Waldz: Are we stopping support for `mysterium-node` image?
Because internet is full of instructions how to setup it like here https://github.com/mysteriumnetwork/node/wiki/Links

Maybe we should push dummy image which says "You should be migrate to `mysteriumnetwork/myst service --agreed-terms-and-conditions`"

bin/client_package/installation/post-install.sh
```diff
@@ -1,83 +0,0 @@
-#!/bin/bash
```
Waldz: Is it possible to `git mv` instead copy&delete. To keep Git history somehow

bin/aliases
```diff
@@ -1,15 +0,0 @@
...
-#
-#> ln -s `pwd`/bin/aliases ~/.zsh.after/mysterium.zsh && source ~/.zshrc
-
-alias myst-home="cd $GOPATH/src/github.com/mysteriumnetwork/node"
-alias myst="myst-home"
-
-alias myst-broker="myst-home && docker-compose up broker"
-alias myst-nats="myst-broker"
-
-alias myst-client="myst-home && bin/client_build && bin/client_run"
```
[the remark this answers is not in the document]
soffokl: It looks more like a personal stuff. Should we keep it in public repo?

bin/ip2asn
```diff
@@ -1,23 +0,0 @@
-#!/bin/bash
```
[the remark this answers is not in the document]
soffokl: I don't really like to keep things "just in case" in a public repo. 
It makes a repo looks littered.

bin/release_docker
soffokl: We will still have the building of `mysterium-node` images in the `release/0.2` branch, so images will be active. 
Placing a dummy image that recommends using newer version should be done only after we finally deprecate `0.2.*` versions and stop doing any fixes there. 
I think it should be done a bit latter.

Waldz: now You clarified

soffokl:
@Waldz @vkuznecovas @zolia I have rebased it to the master. Please take a look on this once again.

bin/aliases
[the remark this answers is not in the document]
zolia: yech, you can make personal aliases through bash..

bin/server_package/config/nonpriv-ip
```diff
@@ -1,2 +0,0 @@
-#!/bin/sh
```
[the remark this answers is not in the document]
soffokl: Fixed it.

Waldz:
Dont forget to remove from https://github.com/MysteriumNetwork/build-artifacts

bin/helpers/functions.sh
```diff
@@ -8,11 +8,11 @@ function get_linker_ldflags {
...
 }
 
-function copy_client_config {
+function copy_config {
     local OS_DIR=$1
     local DST_DIR=$2
-    cp -vrp "bin/common_package/" ${DST_DIR}/config
-    if [[ -d "bin/client_package/config/${OS_DIR}/" ]]; then
-        cp -vrp "bin/client_package/config/${OS_DIR}/." ${DST_DIR}/config
+    cp -vrp "bin/package/config/common/" ${DST_DIR}/config
```
Waldz: I registered related bug #415

PR #412, opened 2018-09-27
Added stale configuration for auto closing inactive issues.

### LINKED ISSUE #359: Setup automatic stale issue actions with comments
Example: https://github.com/ethereum/go-ethereum/tree/master/.github

soffokl:
We need to define the best time frames for auto-closing stale issues. 

Closes #359


.github/stale.yml
```diff
@@ -0,0 +1,17 @@
...
+  - security
+# Label to use when marking an issue as stale
+staleLabel: stale
+# Comment to post when marking an issue as stale. Set to `false` to disable
+markComment: >
+  This issue has been automatically marked as stale because it has not had
+  recent activity. It will be closed if no further activity occurs. Thank you
+  for your contributions.
+# Comment to post when closing a stale issue. Set to `false` to disable
+closeComment: false
```
zolia: would suggest 'true' here. It is always good to know why issue was closed.

soffokl: Looks like there should be not `true` but some comment like in `markComment`.
Anyway, I'll add something there.

soffokl: Added comment.

PR #413, opened 2018-09-28
Fixed service starting in the dep packages

### LINKED ISSUE #395: myst deb packages does not provide service after installation
After merging `mysterium-server` and `mysterium-client` to the single binary `myst` appeared a problem, that `myst` deb package does not start both services. 

- [ ] Add service for the node;
- [ ] Add service for the client; 
- [ ] Fix `INSTALL.md` instruction for using only myst binaries;

soffokl:
Closes #395


INSTALL.md
```diff
@@ -1,26 +1,27 @@
...
 
 ### Fetching and running docker image
 ```bash
-sudo docker run --cap-add NET_ADMIN --net host --name mysterium-node -d mysteriumnetwork/mysterium-node --agreed-terms-and-conditions
+sudo docker run --cap-add NET_ADMIN --net host --name myst -d mysteriumnetwork/myst service --agreed-terms-and-conditions
 ```
 
-Note: to run server, you will have to accept terms & conditions by adding '--agreed-terms-and-conditions' command line option.
-Note 2: it's mandatory to run docker container with --net host to correctly detect VPN service ip which needs to be published to clients, assuming that host on which node is running has external interface with public ip
+>**Note:** to run server, you will have to accept terms & conditions by adding '--agreed-terms-and-conditions' command line option.
```
Waldz: We sync Node WIKI with same running guide. Update it also ;)

INSTALL.md
```diff
@@ -34,91 +35,41 @@ apt-get update
...
-wget https://github.com/mysteriumnetwork/node/releases/download/{VERSION}/mysterium-node_linux_amd64.deb
-sudo dpkg --install --force-depends mysterium-node_linux_amd64.deb
+wget https://github.com/mysteriumnetwork/node/releases/download/{VERSION}/myst_linux_amd64.deb
+sudo dpkg --install --force-depends myst_linux_amd64.deb
 sudo apt-get install -y --fix-broken
 ```
+
 ### Running service
 ```bash
 sudo service mysterium-node start
```
Waldz: `service myst-node start` or `service myst start`

INSTALL.md
```diff
@@ -34,91 +35,41 @@ apt-get update
...
-
-## Mysterium VPN client (Debian && Ubuntu) - tested on Ubuntu 16.04
-### Download
- * https://github.com/mysteriumnetwork/node/releases/download/{VERSION}/mysterium-client_linux_amd64.deb
-
-### Add latest OpenVPN repository
-
-```bash
-apt-get update && apt-get install -y curl
-curl -s https://swupdate.openvpn.net/repos/repo-public.gpg | apt-key add && echo "deb http://build.openvpn.net/debian/openvpn/stable xenial main" > /etc/apt/sources.list.d/openvpn-aptrepo.list && rm -rf /var/cache/apt/* /var/lib/apt/lists/*
```
Waldz: Why to remove these instructions?

INSTALL.md
```diff
@@ -34,91 +35,41 @@ apt-get update
...
-sudo tail -f /var/log/mysterium-client/*
-```
 ### Debugging standalone
 ```bash
-sudo mysterium_client --data-dir=/var/lib/mysterium-client --config-dir=/etc/mysterium-client --runtime-dir=/tmp
+sudo myst --data-dir=/var/lib/mysterium-node --config-dir=/etc/mysterium-node --runtime-dir=/tmp --identity=0x123456..
 ```
 
-
 ## Mysterium VPN node and client standalone binaries (.tar.gz)
```
Waldz: Why to remove this?

bin/package/installation/initd.sh
```diff
@@ -0,0 +1,220 @@
...
+# Required-Start:    $all
+# Default-Start:     2 3 4 5
+# Default-Stop:      0 1 6
+# Short-Description: Start the Mysterium VPN node process
+### END INIT INFO
+
+# If you modify this, please make sure to also edit systemd.service
+
+OS_DIR_BIN="/usr/bin"
+OS_DIR_CONFIG="/etc/mysterium-node"
```
Waldz: Where should be config|logs|data directories: `/etc/myst` or `/etc/mysterium-node` ?

INSTALL.md
Waldz: Maybe we can have service aliases and support both names:
- mysterium_node
- myst -> mysterium_node

Waldz: But I imagine future like this:
- myst <- name of binary
- `mysterium_node` service which starts `myst service ..` service
- `mysterium_client` service which starts `myst connect ..` service

bin/package/installation/initd.sh
```diff
@@ -0,0 +1,220 @@
...
+        --make-pidfile \
+        --pidfile $DAEMON_PIDFILE \
+        --user $DAEMON_USER \
+        --group $DAEMON_GROUP \
+        --exec $DAEMON_BIN \
+        -- \
+        --config-dir=$OS_DIR_CONFIG \
+        --data-dir=$OS_DIR_DATA \
+        --runtime-dir=$OS_DIR_RUN \
+        service \
```
Waldz: Currently we have travis job `name: "Ubuntu docker image for node"` to test .deb installation on Ubuntu.
But it did not identified this error. 
Do You have ideas how to do E2E test of Debian service?

core/node/options_directory.go
```diff
@@ -55,7 +55,7 @@ func ensureOrCreateDir(dir string) error {
 	err := ensureDirExists(dir)
 	if os.IsNotExist(err) {
 		log.Info("[Directory config checker] ", "Directory: ", dir, " does not exit. Creating new one")
-		return os.MkdirAll(dir, 0600)
+		return os.MkdirAll(dir, 0700)
```
Waldz: Was is a debugged place??

INSTALL.md
soffokl: Removed section for client download and installation. Since there are no more dedicated client binary and installation process, only instruction for installing single binary left.

INSTALL.md
soffokl: Removed section for client download and installation. Since there are no more dedicated client binary and installation process, only instruction for installing single binary left.

core/node/options_directory.go
soffokl: Yeah. Actually, every directory should have `x` attribute to be workable.

bin/package/installation/initd.sh
soffokl: This job just builds a docker image based on Ubuntu. It installs dep package, but in this case installation of dep package was not broken, that is why we didn't find this issue. 
Docker uses it's own `docker-entrypoint.sh` for starting services. But broken was system service.
Not sure how to test it correctly with a systemd service.

INSTALL.md
soffokl: 1) I don't think that we need to add aliases, it will be just extra confusion for the users. We will need to have extra documentation to describe the reasons for aliases. I do not see any advantages having two names for the single service. 

2) In the current implementation, it works exactly the way you described. `myst` is a bynary, `mysterium_node` service that starts `myst service ...`

INSTALL.md
soffokl: I was planning to do it when this PR will be merged after review.

bin/package/installation/initd.sh
soffokl: Well, my idea was that this service is starting mysterium exit node, so all configs and logs should be in the `mysterium-node` directories and I left it as it was before.

bin/package/installation/initd.sh
soffokl: Created issue for the future improvement: #417

INSTALL.md
Waldz: Cool

PR #422, opened 2018-10-02
moved to gofrs/uuid instead of satori/uuid

### LINKED ISSUE #376: move away from satori/go.uuid
satori/go.uuid is unmaintained and contains a critical bug that results in non random uuids. See https://github.com/satori/go.uuid/issues/73

We should deprecate it. There's an actively maintained fork with the issues fixed here https://github.com/gofrs/uuid

vkuznecovas:
closes #376


session/generator.go
```diff
@@ -17,9 +17,9 @@
 
 package session
 
-import "github.com/satori/go.uuid"
+import "github.com/gofrs/uuid"
 
 // GenerateUUID method returns ID based on random UUID
 func GenerateUUID() ID {
-	return ID(uuid.NewV4().String())
+	return ID(uuid.Must(uuid.NewV4()).String())
```
Waldz: Lets not crash full app, but handle error

session/manager.go
```diff
@@ -102,7 +102,11 @@ func (manager *manager) Create(consumerID identity.Identity, proposalID int) (se
 }
 
 func (manager *manager) createSession(consumerID identity.Identity) (sessionInstance Session, err error) {
-	sessionInstance.ID = manager.generateID()
+	id, err := manager.generateID()
```
Waldz: Put directly to `sessionInstance.ID`, dont need temp variable here

vkuznecovas: fixd

PR #423, opened 2018-10-03
Push master to dev pre-release

### LINKED ISSUE #416: Push 'dev' releases to public storage
Currently we are pushing `dev` version to https://github.com/mysteriumnetwork/build-artifacts/tree/mysterium-node

Problem1. This is private
Problem2. Using Git branch as storage solution

Acceptance Criteria:
- Implement storage solution, which is public (S3 probably)

soffokl:
Closes #416


.travis.yml
```diff
@@ -18,14 +18,16 @@ cache:
     - vendor
 
 stages:
-  - dep-cache
-  - test
+  - name: dep-cache
+    if: tag != dev
```
Waldz: we dont need to cache for `dev`?

.travis.yml
```diff
@@ -18,14 +18,16 @@ cache:
...
+  - name: test
+    if: tag != dev
   - name: build
-    if: (type != pull_request AND branch = master AND tag IS blank) OR tag IS present
+    if: (type != pull_request AND branch = master AND tag IS blank) OR (tag IS present AND tag != dev)
```
Waldz: Should not we run `build` together with `release`. Would simplify variable hell

.travis.yml
```diff
@@ -95,14 +97,21 @@ jobs:
     - stage: release-dev
       script: bin/s3 sync s3://build-artifacts build/package
       name: "Update for build-artifacts"
+      before_deploy:
+        - git tag -f dev
```
Waldz: Dont You want to add this to script?

.travis.yml
soffokl: We are introducing `dev` tag in the `release-dev` stage. When the new tag appears Travis starts a new build for this tag. This condition just prevents this run the second time.

.travis.yml
soffokl: We are introducing `dev` tag in the `release-dev` stage. When the new tag appears Travis starts a new build for this tag. This condition just prevents this run the second time.

.travis.yml
soffokl: I think that it's more readable to keep it here. 
It's not too big to have an extra script for that, and it never used for the second time somewhere else.

.travis.yml
Waldz: Sounds very complicated, how to make readable about this side effect?

vkuznecovas:
I feel we also need to include commit hashes with the packages(either on each of them, or somehow incorporate that into the bucket itself). Here's a scenario of why I think this is needed:

As a user, I'll open up the dev-release page(the s3 bucket in this case) - and I'll see all the packages - issue here is that I won't know the commit they are coming from. I might end up running a version that I think is the newest, although the packages have not been updated in the bucket yet.

soffokl:
There will be a single option to download dev version, only the latest one, no other options.

.travis.yml
```diff
@@ -93,23 +93,10 @@ jobs:
 
     # Development bleeding-edge release (on master commits only)
     - stage: release-dev
```
Waldz: It's not release anymore. It's just single `bin/release_goreport` task

PR #430, opened 2018-10-09
Fixed build version for deb packages

soffokl:


.travis.yml
```diff
@@ -11,7 +11,7 @@ env:
     - BUILD_VERSION_MAJOR=0
 
 before_install:
-  - export BUILD_VERSION=${TRAVIS_TAG:-"$BUILD_VERSION_MAJOR-dev.$TRAVIS_BUILD_NUMBER"}
+  - if [ -z "${TRAVIS_TAG/dev/}" ]; then export BUILD_VERSION="$BUILD_VERSION_MAJOR-dev.$TRAVIS_BUILD_NUMBER"; fi
```
Waldz: Lets tag our prerelease with `0-dev`, `0.0.0-dev` or `0.4-dev` and we dont need this conditional anymore

soffokl: `0-dev` and `0.0.0-dev` looks weird for me. `dev` means development version. 
`0.4-dev` is nice, but it will require a manual change after every major release. 
Do you think that it's better than this condition?

Waldz: I dont know. But this condition brings problems, not 1st time

PR #442, opened 2018-10-12
Cherry pick of "Added back agreed-terms-and-conditions flag"

soffokl:


cmd/commands/service/command.go
```diff
@@ -18,10 +18,14 @@
 package service
 
 import (
+	"fmt"
+
```
Waldz: extra line

PR #446, opened 2018-10-14
Dummy connection which does Noop encryption

Waldz:
Updates #446

### DIFF
--- a/core/connection/manager.go
+++ b/core/connection/manager.go
@@ -49,11 +49,11 @@ var (
 
 type connectionManager struct {
 	//these are passed on creation
-	mysteriumClient    server.Client
-	newDialog          DialogCreator
-	newPromiseIssuer   PromiseIssuerCreator
-	connectionCreators *map[string]ConnectionCreator
-	statsKeeper        stats.SessionStatsKeeper
+	mysteriumClient   server.Client
+	newDialog         DialogCreator
+	newPromiseIssuer  PromiseIssuerCreator
+	connectionCreator ConnectionCreator
+	statsKeeper       stats.SessionStatsKeeper
 	//these are populated by Connect at runtime
 	ctx             context.Context
 	mutex           sync.RWMutex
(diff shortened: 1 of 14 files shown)


vkuznecovas:
misslick

services/openvpn/connection_factory_test.go
```diff
@@ -0,0 +1,78 @@
...
+}
+
+func (cf *cacheFake) Get() location.Location {
+	return cf.location
+}
+func (cf *cacheFake) RefreshAndGet() (location.Location, error) {
+	return cf.location, cf.err
+}
+
+var _ connection.ConnectionCreator = &ProcessBasedConnectionFactory{}
```
zolia: why we need this one?

Waldz: It's compilation time check if `ProcessBasedConnectionFactory ` complies interface `ConnectionCreator `

core/connection/manager.go
```diff
@@ -122,6 +129,11 @@ func (manager *connectionManager) startConnection(consumerID, providerID identit
 		return err
 	}
 
+	connectionCreator, exists := (*manager.connectionCreators)[proposal.ServiceType]
```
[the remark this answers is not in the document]
Waldz: Improved by suggestion

core/connection/registry.go
```diff
@@ -0,0 +1,45 @@
...
+}
+
+// Register new plugable connection
+func (registry *Registry) Register(serviceType string, creator ConnectionCreator) {
+	registry.creators[serviceType] = creator
+}
+
+// CreateConnection create plugable connection
+func (registry *Registry) CreateConnection(options ConnectOptions, stateChannel StateChannel) (Connection, error) {
+	creator, exists := registry.creators[options.Proposal.ServiceType]
```
zolia: Most likely we need to distinguish between OpenVPN UDP and OpenVPN TCP. If that would be only OpenVPN service type, then it would be ambiguous ConnectionCreator. Maybe we could have not OpenVPN service type, but OpenvpnTCP and OpenvpnUDP service types?

Waldz: Proposal defines service and it's details (protocol, speed, price, contact, etc.)
Consumer is interested if any of details change, not only udp

PR #281, opened 2018-07-04
Compile in geoip2 database and avoid external file reading

tadovas:
Idea is to reduce external file dependencies and to build-in as much data/info as possible. First candidate geoip database.
Short summary:
1. Geoip database is converted to few source files by generator.go
2. New option for builtin country resolving for both client/server - used by default
3. No need to package db for runtime anymore

PR #447, opened 2018-10-15
Initial mobile entrypoint for iOS/Swift

Waldz:
```
https://releases.mysterium.network/Carthage.json

import UIKit
import Mysterium

@UIApplicationMain
class AppDelegate: UIResponder, UIApplicationDelegate {
    
    func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]?) -> Bool {
        
        MysteriumNewNode()
        print("Node created!!!")
        
        return true
    }

}
```

Updates #447

Leftover:
- [x] Create Xcode project with Golang compilation (in `research-ios`)
- [ ] Xcode Golang compile should have checksum
- [x] Inject ConnectionManager to DI container #446
- [x] Create dummy transport for ConnectionManager #446 
- [ ] ConnectionManager.Start() starts callbacking NetworkExtension for read/write of packets
- [ ] ConnectionManager.Stop() stop read/write of packets in NetworkExtension
- [x] TODO Implement different check for in "mobile" package
- [x] TODO Embbed all config file to released artifacts
- [x] TODO Where to save runtime data
- [x] TODO Embbed location DB to released artifacts #281
- [x] TODO Return node startup/runtime errors to mobile
- [x] Release artefact for iOS `gomobile bind -target=ios/arm64 -v github.com/mysteriumnetwork/node/mobile` #466 
- [ ] Strip debug symbols from .framework artefact

### DIFF
--- a/bin/package/ios/Info.plist
+++ b/bin/package/ios/Info.plist
@@ -1,0 +1,53 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>BuildMachineOSBuild</key>
+	<string>17G65</string>
+	<key>CFBundleDevelopmentRegion</key>
+	<string>en</string>
+	<key>CFBundleExecutable</key>
+	<string>Mysterium</string>
+	<key>CFBundleIdentifier</key>
+	<string>network.mysterium.sdk.ios</string>
+	<key>CFBundleInfoDictionaryVersion</key>
+	<string>6.0</string>
+	<key>CFBundleName</key>
+	<string>Mysterium</string>
+	<key>CFBundlePackageType</key>
+	<string>FMWK</string>
+	<key>CFBundleShortVersionString</key>
+	<string>0.5.0-rc1</string>
+	<key>CFBundleSignature</key>
+	<string>????</string>
+	<key>CFBundleSupportedPlatforms</key>
+	<array>
+		<string>MacOSX</string>
+	</array>
+	<key>CFBundleVersion</key>
+	<string>70</string>
+	<key>DTCompiler</key>
+	<string>com.apple.compilers.llvm.clang.1_0</string>
+	<key>DTPlatformBuild</key>
+	<string>9F2000</string>
+	<key>DTPlatformVersion</key>
+	<string>GM</string>
+	<key>DTSDKBuild</key>
+	<string>17E189</string>
+	<key>DTSDKName</key>
+	<string>macosx10.13</string>
+	<key>DTXcode</key>
+	<string>0941</string>
+	<key>DTXcodeBuild</key>
+	<string>10A255</string>
+	<key>NSHumanReadableCopyright</key>
+	<string>Copyright (C) 2018 The "MysteriumNetwork/node" Authors.</string>
+	<key>UIDeviceFamily</key>
+	<array>
+		<integer>1</integer>
+		<integer>2</integer>
+		<integer>3</integer>
+		<integer>4</integer>
+	</array>
+</dict>
+</plist>
--- a/bin/package_ios
+++ b/bin/package_ios
@@ -1,0 +1,32 @@
+#!/bin/bash
+
+###########################################################################
+# Packaging script which creates iOS framework (Swift compatible).
+#
+# Usage:
+#> bin/package_ios <architecture>
+#
+# Package (specific architecture: arm64, amd64):
+#> bin/package_ios arm64
+
+set -e
+
+source bin/helpers/output.sh
+source bin/helpers/functions.sh
+
+ARCH=$1
+if [[ ! "$ARCH" =~ ^(arm64|amd64)$ ]]; then
+    print_error "Missing architecture! Should be: arm64, amd64"
+    exit 1
+fi
+
+DIR_BUILD="build/package"
+PACKAGE_FILE="${DIR_BUILD}/Mysterium.framework"
+
+gomobile bind -v \
+    -target=ios/${ARCH} \
+    -o $PACKAGE_FILE \
+    github.com/mysteriumnetwork/node/mobile/mysterium
+
+print_success "iOS package '$PACKAGE_FILE' complete!"
+exit 0
--- a/bin/package_ios
+++ b/bin/package_ios
@@ -14,21 +14,13 @@ set -e
 source bin/helpers/output.sh
 source bin/helpers/functions.sh
 
-ARCH=$1
-if [[ ! "$ARCH" =~ ^(arm64|amd64)$ ]]; then
-    print_error "Missing architecture! Should be: arm64, amd64"
-    exit 1
-fi
-
 DIR_BUILD="build/package"
+mkdir -p $DIR_BUILD
 PACKAGE_FILE="${DIR_BUILD}/Mysterium.framework.zip"
 DIR_TEMP=`mktemp -d ${DIR_BUILD}/${tempname}.XXXXXX`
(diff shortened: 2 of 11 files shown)


mobile/mysterium/entrypoint.go
```diff
@@ -0,0 +1,114 @@
...
+		panic(err)
+	}
+	dataDir := filepath.Join(userHomeDir, ".mysterium")
+	currentDir := userHomeDir
+
+	err = di.Bootstrap(node.Options{
+		node.OptionsDirectory{
+			dataDir,
+			filepath.Join(dataDir, "keystore"),
+			// TODO Embbed all config file to released artifacts
```
soffokl: `Embbed` -> `Embed`

mobile/mysterium/entrypoint.go
```diff
@@ -0,0 +1,114 @@
...
+	var di cmd.Dependencies
+
+	userHomeDir, err := homedir.Dir()
+	if err != nil {
+		panic(err)
+	}
+	dataDir := filepath.Join(userHomeDir, ".mysterium")
+	currentDir := userHomeDir
+
+	err = di.Bootstrap(node.Options{
```
soffokl: I suggest using files keys for structs. It's really hard to read without it.

mobile/mysterium/entrypoint.go
```diff
@@ -0,0 +1,114 @@
...
+			currentDir,
+		},
+
+		"127.0.0.1",
+		4050,
+
+		openvpn_core.NodeOptions{},
+		node.OptionsLocation{
+			"https://api.ipify.org/",
+			// TODO Embbed location DB to released artifacts
```
soffokl: `Embbed` -> `Embed`

mobile/mysterium/entrypoint.go
```diff
@@ -0,0 +1,114 @@
...
+			// TODO Embbed all config file to released artifacts
+			filepath.Join(currentDir, "config"),
+			// TODO Where to save runtime data
+			currentDir,
+		},
+
+		"127.0.0.1",
+		4050,
+
+		openvpn_core.NodeOptions{},
```
soffokl: With keyed fields, we can skip setting filed by the zero value.

cmd/di.go
```diff
@@ -88,13 +88,14 @@ func (di *Dependencies) Bootstrap(nodeOptions node.Options) error {
...
 	log.Infof("Starting Mysterium Node (%s)", metadata.VersionAsString())
 
-	if err := nodeOptions.Directories.Check(); err != nil {
-		return err
-	}
-
-	if err := nodeOptions.Openvpn.Check(); err != nil {
-		return err
-	}
+	// TODO Implement different check for in "mobile" package
```
soffokl: Maybe we can add check if it's mobile or not, to do checks for normal systems? 
These checks look important.

bin/package_ios
```diff
@@ -0,0 +1,32 @@
...
+# Package (specific architecture: arm64, arm64):
+#> bin/package_ios arm64
+
+set -e
+
+source bin/helpers/output.sh
+source bin/helpers/functions.sh
+
+ARCH=$1
+if [[ ! "$ARCH" =~ ^(arm64|arm64)$ ]]; then
```
soffokl: Is it `arm64` or `arm64` condition? Looks like the same value repeated twice.

mobile/mysterium/entrypoint.go
Waldz: Fixed

mobile/mysterium/entrypoint.go
Waldz: Used for some fields

bin/package_ios
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+
+###########################################################################
+# Packaging script which creates iOS framework (Swift compatible).
+#
+# Usage:
+#> bin/package_ios <architecture>
+#
+# Package (specific architecture: arm64, arm64):
```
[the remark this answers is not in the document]
Waldz: Fixed

bin/package_ios
```diff
@@ -0,0 +1,36 @@
...
+# Package (specific architecture: arm64, amd64):
+#> bin/package_ios arm64
+
+set -e
+
+source bin/helpers/output.sh
+source bin/helpers/functions.sh
+
+ARCH=$1
+if [[ ! "$ARCH" =~ ^(arm64|amd64)$ ]]; then
```
Arrnas: Is this necessary? AFAIK xcode strips out unnecessary architectures from fat binaries on it's own so I think we should include all supported architectures and allow users of the framework to decide which architectures they'll need. Default `gomobile bind -target=ios` contains `Architectures in the fat file: Mysterium.framework/Mysterium are: armv7 i386 x86_64 arm64`, note that `armv7` is needed to support older devices.

bin/package/ios/Info.plist
```diff
@@ -0,0 +1,53 @@
...
+	<string>Mysterium</string>
+	<key>CFBundlePackageType</key>
+	<string>FMWK</string>
+	<key>CFBundleShortVersionString</key>
+	<string>0.5.0-rc1</string>
+	<key>CFBundleSignature</key>
+	<string>????</string>
+	<key>CFBundleSupportedPlatforms</key>
+	<array>
+		<string>MacOSX</string>
```
Arrnas: this might cause issues when uploading to the app store: https://stackoverflow.com/questions/32726939/error-itms-90542-invalid-cfbundlesupportedplatforms-value https://github.com/NativeScript/ios-runtime/issues/689

bin/package_ios
Waldz: I thought on mobile size matters: 
- 24MB binary size with single architecture 
- 100MB binary size with all architectures

Arrnas: https://pspdfkit.com/guides/ios/current/troubleshooting/removing-architectures/ both `cocoapods` and `carthage` automatically remove desktop architectures from release builds. 
In our case desktop architectures are not important for ios because we're using the framework in the network extension which can only run on real devices, but we still need both `armv7` and `arm64` architectures inside the `.framework` file.

tadovas: why we need arm7 anyway? I don't even crosscompiled necessary libs for this arch. We are not planning to support pre ios 10.x devices anyway.

cmd/di.go
tadovas: This can be abstracted to some kind of Transport.Check() which can be conditionaly compiled on different oses.

bin/package_ios
tadovas: I used this compability matrix http://iossupportmatrix.com/ and assumed that we start supporting ARMv8 and upwards which is basically ios 11 and greater. So far it will be enough. For older arms support it's a question for POs - what segments to target. The same is for android - initially 64 bit archs. Older ones - later, if needed.

Arrnas: in that case yeah, we can just compile `arm64` for ios

mobile/mysterium/entrypoint.go
```diff
@@ -0,0 +1,120 @@
...
+	"github.com/mysteriumnetwork/node/identity"
+	"github.com/mysteriumnetwork/node/metadata"
+	service_noop "github.com/mysteriumnetwork/node/services/noop"
+)
+
+// NewNode function creates new Node
+func NewNode() {
+	var di cmd.Dependencies
+
+	userHomeDir, err := homedir.Dir()
```
zolia: this requires additional external dependency for non-Darwin os'es:
`cmd := exec.Command("getent", "passwd", strconv.Itoa(os.Getuid()))` 
not sure its proper way to do it.

tadovas:
@zolia @vkuznecovas @Waldz let's wait no more, and move on with this PR. Uncompleted tasks form list above can be completed with other issues (need to register them)

Issue #450, opened 2018-10-16
Goimports check do nothing since goimport is not installed for CI


soffokl:
We are missing `goimports` tool on the CI, but it succeeds somehow anyway. 

```
bin/check_goimports
bin/check_goimports: line 23: goimports: command not found
All files are compliant to goimports.
```

It should fail if required dependencies not found. 
Required dependencies should be installed before running it.

PR #451, opened 2018-10-16
redirect stderr to stdout for goimports #450

u5surf:
#450 
command not found is output as stderr so, it can be redirected to stdout only using `2>&1`

result :
https://travis-ci.com/u5surf/node/jobs/152053145

btw, but also installing goimports in travis, it’s failed the required dependencies test.
Is it another issue?


soffokl:
@u5surf please consider [signing](https://help.github.com/articles/signing-commits/) your commit to be able to merge it.

vkuznecovas:
not entirely sure where first 4 commits are coming from

PR #452, opened 2018-10-16
Run goimports with builder

### LINKED ISSUE #450: Goimports check do nothing since goimport is not installed for CI

tadovas:
Closes #450 
Run check_goimports inside preinstalled builder container


tadovas:
This looks really important, as I already found one go file which was not "goimported"

tadovas:
@mysteriumnetwork/core @Waldz @soffokl  check this as priority

discovery/factory.go
```diff
@@ -52,12 +52,12 @@ func NewService(
 	signerCreate identity.SignerFactory,
 ) *Discovery {
 	return &Discovery{
-		identityRegistry:            identityRegistry,
-		identityRegistration:        identityRegistration,
-		mysteriumClient:             mysteriumClient,
-		signerCreate:                signerCreate,
-		statusChan:                  make(chan Status),
-		status:                      StatusUndefined,
+		identityRegistry:     identityRegistry,
```
soffokl: This was formatted by goimports in my system. Maybe we have some old version of goimports in the build image?

.travis.yml
```diff
@@ -48,7 +48,7 @@ jobs:
     - script: bin/builder_run bin/check_golint
       name: "golint check"
 
-    - script: bin/check_goimports
+    - script: bin/builder_run bin/check_goimports
```
soffokl: I'd try to avoid using extra pre-build images for simple task. 
Maybe it's better to do `go get goimports` instead of running it in extra docker container?

soffokl: And looks like direct call works a bit faster on my system:
image

soffokl: This ^ was not counting time for image downloading.

discovery/factory.go
soffokl: Found it, I was using old version from https://github.com/bradfitz/goimports

soffokl: or not
sorry for these comments fall :(

.travis.yml
tadovas: Your points are absolutely valid, but:
1. It's not improvement or feature, its a hotfix - intention was to fix it asap with minimal effort as build functionality is currently broken
2. golint is also go getted inside builder and is run the same way, maybe we can get rid of this builder completely when migrating to go style CI

PR #457, opened 2018-10-17
Reduce noise when running `check_golint`

tcharding:
Currently `check_golint` script prints a bunch (5) of warning lines when there are _no_ messages to warn about, such as this (including the newlines):
```
[warning] 0 messages in 'github.com/mysteriumnetwork/node/identity': 


```
This is a off putting; if there are no issues there should be no output.  If 0 messages implies some error then we should warn about that explicitly (I don't know if this is the case so this patch takes the first route).

Only print warning message if there are more than 0 messages to show.


soffokl:
Hm, I see no `[warning] 0` messages. Maybe it's related to golint version or something else? 
image

tcharding:
This is linked to #458 it shows up because `golint` is failing for me so there are 0 warnings.

Issue #443, opened 2018-10-14
Tequilapi "/stop" endpoint does not stop `myst service`


Waldz:
**How to reproduce:**
1. Start service
```
bin/build && bin/run service --agreed-terms-and-conditions
```
2. Try to stop service thru Tequilapi 
```
curl -X POST http://localhost:4050/stop

2018-10-14T11:53:15.974989 [Info] Application stop requested
2018-10-14T11:53:15.975132 [Info] [discovery] stopping discovery loop..
2018-10-14T11:53:15.975173 [Info] [discovery] cancelled identity registration
2018-10-14T11:53:15.975235 [Info] [service-manager] Waiting for service to finish
2018-10-14T11:53:15.986791 [Info] [nat] NAT rules cleared
2018-10-14T11:53:15.994995 [Info] [nat] IP forwarding disabled
2018-10-14T11:53:15.995061 [Info] [client-management] Shutdown
2018-10-14T11:53:15.995242 [Info] [client-management] Shutdown finished
2018-10-14T11:53:15.995282 [Info] No active connection - proceeding
2018-10-14T11:53:15.995419 [Trace] [openvpn-process] Stdout: Sun Oct 14 14:53:15 2018 event_wait : Interrupted system call (code=4)
2018-10-14T11:53:15.995481 [Info] Api stopped
Good bye
```
3. **Result1 (random):** Process does not stop and After 3s or some time service outputs again
```
2018-10-14T11:53:18.001838 [Trace] [openvpn-process] Stdout: Sun Oct 14 14:53:18 2018 Closing TUN/TAP interface
2018-10-14T11:53:18.003133 [Debug] [client-management] Line received: >STATE:1539517998,EXITING,SIGINT,,,,,
2018-10-14T11:53:18.003184 [Warn] [client-management] Connection failed to read. EOF
2018-10-14T11:53:18.003212 [Trace] [client-management] Line delivering: >STATE:1539517998,EXITING,SIGINT,,,,,
2018-10-14T11:53:18.003374 [Info] [client-management] Event consumer is done
2018-10-14T11:53:18.003431 [Info] [service-openvpn] Openvpn service exited
2018-10-14T11:53:18.003527 [Warn] [client-management] Middleware stop error. connection is gone
2018-10-14T11:53:18.003653 [Trace] [openvpn-process] Stdout: Sun Oct 14 14:53:18 2018 SIGINT[hard,] received, process exiting
[...]

PR #463, opened 2018-10-18
Quit cli when receiving command `quit`

### LINKED ISSUE #462: Quiting the cli hangs instead of quitting
Running `bin/run cli` then entering command `quit` (or `exit`) outputs a goodbye message then prints the cli prompt again, the cli hangs at this stage (Ctl-C does not work).

tcharding:
Currently the cli does not quit when told to.  We can use the `HardKiller` instead of the `SoftKiller` to fix this.

Closes #462

### DIFF
--- a/cmd/commands/cli/command.go
+++ b/cmd/commands/cli/command.go
@@ -332,7 +332,7 @@ func (c *cliApp) help() {
 
 // quit stops cli and client commands and exits application
 func (c *cliApp) quit() {
-	stop := utils.SoftKiller(c.Kill)
+	stop := utils.HardKiller(c.Kill)
 	stop()
 }
 


cmd/commands/cli/command.go
```diff
@@ -332,7 +332,7 @@ func (c *cliApp) help() {
 
 // quit stops cli and client commands and exits application
 func (c *cliApp) quit() {
-	stop := utils.SoftKiller(c.Kill)
+	stop := utils.HardKiller(c.Kill)
```
soffokl: `HardKiller` doing `os.Exit()` without proper shutting down other dependencies. 
I think there should be a better approach for this.

Waldz:
I registered bug related to this issue: #443

cmd/commands/cli/command.go
soffokl: CLI is starting [here](https://github.com/mysteriumnetwork/node/blob/master/cmd/commands/cli/command.go#L57), but if we are doing `os.Exit()` in the CLI we are missing the [following](https://github.com/mysteriumnetwork/node/blob/master/cmd/commands/cli/command.go#L63-L65) block. 
It's not only about CLI, but it's also about upper components.

soffokl:
Looks like this is a bug in the package that we are using https://github.com/chzyer/readline/pull/136
It fixed in master, but we are using the latest version and we don't have the fix released. 

We can workaround it somehow, or just use the master branch in `dep`

soffokl:
Looks like this is a bug in the package that we are using https://github.com/chzyer/readline/pull/136
It fixed in master, but we are using the latest version and we don't have the fix released. 

We can workaround it somehow, or just use the master branch in `dep`

PR #471, opened 2018-10-22
Fixed stopping of service, daemon and CLI

### LINKED ISSUE #443: Tequilapi "/stop" endpoint does not stop `myst service`

soffokl:
Closes #443

### DIFF
--- a/cmd/commands/daemon/command.go
+++ b/cmd/commands/daemon/command.go
@@ -31,7 +31,7 @@ func NewCommand() *cli.Command {
 		Usage:     "Starts Mysterium Tequilapi service",
 		ArgsUsage: " ",
 		Action: func(ctx *cli.Context) error {
-			errorChannel := make(chan error, 1)
+			errorChannel := make(chan error, 2)
 			if err := di.Bootstrap(cmd.ParseFlagsNode(ctx)); err != nil {
 				return err
 			}
--- a/cmd/commands/daemon/command.go
+++ b/cmd/commands/daemon/command.go
@@ -19,7 +19,6 @@ package daemon
 
 import (
 	"github.com/mysteriumnetwork/node/cmd"
-	"github.com/mysteriumnetwork/node/utils"
 	"github.com/urfave/cli"
 )
 
@@ -32,11 +31,15 @@ func NewCommand() *cli.Command {
 		Usage:     "Starts Mysterium Tequilapi service",
 		ArgsUsage: " ",
 		Action: func(ctx *cli.Context) error {
+			errorChannel := make(chan error, 1)
 			if err := di.Bootstrap(cmd.ParseFlagsNode(ctx)); err != nil {
 				return err
 			}
-			cmd.RegisterSignalCallback(utils.SoftKiller(di.Shutdown))
-			return di.Node.Wait()
+			go func() { errorChannel <- di.Node.Wait() }()
+
+			cmd.RegisterSignalCallback(func() { errorChannel <- nil })
+
+			return <-errorChannel
 		},
 		After: func(ctx *cli.Context) error {
 			return di.Shutdown()
--- a/cmd/commands/service/command.go
+++ b/cmd/commands/service/command.go
@@ -25,7 +25,6 @@ import (
 	"github.com/mysteriumnetwork/node/cmd/commands/license"
 	"github.com/mysteriumnetwork/node/core/service"
 	"github.com/mysteriumnetwork/node/metadata"
-	"github.com/mysteriumnetwork/node/utils"
 	"github.com/urfave/cli"
 )
 
@@ -101,7 +100,7 @@ func NewCommand(licenseCommandName string) *cli.Command {
 				errorChannel <- di.ServiceManager.Wait()
 			}()
 
-			cmd.RegisterSignalCallback(utils.SoftKiller(di.Shutdown))
+			cmd.RegisterSignalCallback(func() { errorChannel <- nil })
 
 			err := <-errorChannel
 			switch err {
--- a/cmd/commands/service/command.go
+++ b/cmd/commands/service/command.go
@@ -78,7 +78,7 @@ func NewCommand(licenseCommandName string) *cli.Command {
 				os.Exit(2)
 			}
 
-			errorChannel := make(chan error, 1)
+			errorChannel := make(chan error, 3)
 			if err := di.Bootstrap(cmd.ParseFlagsNode(ctx)); err != nil {
 				return err
 			}


cmd/commands/service/command.go
```diff
@@ -101,7 +100,7 @@ func NewCommand(licenseCommandName string) *cli.Command {
 				errorChannel <- di.ServiceManager.Wait()
 			}()
 
-			cmd.RegisterSignalCallback(utils.SoftKiller(di.Shutdown))
+			cmd.RegisterSignalCallback(func() { errorChannel <- nil })
```
Waldz: Why do we need dummy stopper at all?

cmd/commands/cli/command.go
```diff
@@ -38,6 +39,7 @@ const cliCommandName = "cli"
 // NewCommand constructs CLI based Mysterium UI with possibility to control quiting
 func NewCommand() *cli.Command {
 	var di cmd.Dependencies
+	cmdCLI := &cliApp{}
```
Waldz: Dont need to init:
`var cmdCLI *cliApp`

cmd/commands/service/command.go
soffokl: It's not dummy stopper. 

We need to have a graceful shutdown for the command. 
This handler returns `nil` error to stop the function, since it's expected behavior for the signals.

cmd/commands/daemon/command.go
```diff
@@ -32,11 +31,15 @@ func NewCommand() *cli.Command {
...
 		Action: func(ctx *cli.Context) error {
+			errorChannel := make(chan error, 1)
 			if err := di.Bootstrap(cmd.ParseFlagsNode(ctx)); err != nil {
 				return err
 			}
-			cmd.RegisterSignalCallback(utils.SoftKiller(di.Shutdown))
-			return di.Node.Wait()
+			go func() { errorChannel <- di.Node.Wait() }()
+
+			cmd.RegisterSignalCallback(func() { errorChannel <- nil })
```
tadovas: I see a channel of size 1 here, and two writes (from Node.Wait goroutine, and signaling goroutine), only one read. Another thing - when signal is received and Action exits (return <-errorChannel), what happens with Node which is still running? is it stopped somewhere else? doesn't sound right.

soffokl: A node will be stopped by the `di.Shutdown()` in the `After` section, a bit lower.

soffokl: Buffered channel is used here to not block goroutine and release it without waiting of reading.

tadovas: My hint was maybe a channel with size 2 should be more appropriate? So each goroutine can put its result independenly whenever another party read previous value or not

cmd/commands/service/command.go
Waldz: clear now, but just after explanation :)

cmd/commands/cli/command.go
Waldz: I meant - You should initiate full variable `&cliApp{}`, not only several attributes of it

soffokl: Unfortunately, fields of the `cliApp` depend on the variable that we are getting only from `ctx` in `Action` block.

cmd/commands/daemon/command.go
soffokl: Make sense, especially not only in this case but in other, with more waiters. 
Changed it in all places.

cmd/commands/cli/command.go
```diff
@@ -38,29 +39,31 @@ const cliCommandName = "cli"
...
-				historyFile: filepath.Join(nodeOptions.Directories.Data, ".cli_history"),
-				tequilapi:   tequilapi_client.NewClient(nodeOptions.TequilapiAddress, nodeOptions.TequilapiPort),
-			}
+			cmdCLI.historyFile = filepath.Join(nodeOptions.Directories.Data, ".cli_history")
+			cmdCLI.tequilapi = tequilapi_client.NewClient(nodeOptions.TequilapiAddress, nodeOptions.TequilapiPort)
+
 			go func() { errorChannel <- cmdCLI.Run() }()
-			cmd.RegisterSignalCallback(utils.SoftKiller(cmdCLI.Kill))
-			cmd.RegisterSignalCallback(utils.SoftKiller(di.Shutdown))
+			cmd.RegisterSignalCallback(func() { errorChannel <- nil })
```
Waldz: Did You alligned change to this stop flow also:
`tequilapi_endpoints.AddRouteForStop(router, utils.SoftKiller(di.Shutdown))`

soffokl: Didn't get a question. 

In the [master](https://github.com/mysteriumnetwork/node/blob/master/core/node/node.go#L117) I see 
```
utils.SoftKiller(node.Kill)
```

So it will stop the `node` and service will be exited correctly, and `di.Shutdown()` will be applies from `After` section.

Issue #431, opened 2018-10-09
Service and CLI should utilize existing API if it exists. 


soffokl:
During the starting `service` or `cli` sub-commands, we are starting `taquilapi` instance too.

If it failed to start because the same instance already exists, we should continue to utilize existing instance.

Log records should contain all information to describe what was happened during the start process.

PR #475, opened 2018-10-23
myst cli does not start new daemon

### LINKED ISSUE #402: Make 'myst daemon' as separate command
```
myst default -> help
myst daemon # to call call Tequilapi endpoinds
myst service --identity=0x1 --agreed-terms-and-conditions

# future commands
myst connect --provider=0x1
myst bar | taverna | cli
```

Acceptance Criteria:
- Create command deamon (now it's `myst`)
- Make help as default command 
- Make clear WIKI docs how to run a node with `myst service`

soffokl:
Closes #402

### DIFF
--- a/cmd/commands/cli/command.go
+++ b/cmd/commands/cli/command.go
@@ -37,31 +37,18 @@ const cliCommandName = "cli"
 
 // NewCommand constructs CLI based Mysterium UI with possibility to control quiting
 func NewCommand() *cli.Command {
-	var di cmd.Dependencies
-
 	return &cli.Command{
 		Name:  cliCommandName,
 		Usage: "Starts a CLI client with a Tequilapi",
 		Action: func(ctx *cli.Context) error {
-			errorChannel := make(chan error)
-			if err := di.Bootstrap(cmd.ParseFlagsNode(ctx)); err != nil {
-				return err
-			}
-			go func() { errorChannel <- di.Node.Wait() }()
-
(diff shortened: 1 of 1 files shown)


soffokl:
@tadovas no, it's not related. Hanging problem described [here](https://github.com/mysteriumnetwork/node/issues/462#issuecomment-431739662).

cmd/commands/cli/command.go
```diff
@@ -37,31 +37,18 @@ const cliCommandName = "cli"
...
 // NewCommand constructs CLI based Mysterium UI with possibility to control quiting
 func NewCommand() *cli.Command {
-	var di cmd.Dependencies
-
 	return &cli.Command{
 		Name:  cliCommandName,
 		Usage: "Starts a CLI client with a Tequilapi",
 		Action: func(ctx *cli.Context) error {
-			errorChannel := make(chan error)
-			if err := di.Bootstrap(cmd.ParseFlagsNode(ctx)); err != nil {
```
Waldz: Is it possible to remove Node bootstrapping for `myst service` command also?

soffokl: But what will the service do without Node?
I think it will be not comfortable to start daemon and service separately.

And we have issue https://github.com/mysteriumnetwork/node/issues/431, that will allow using `service` with existing `daemon`.

Waldz: > But what will the service do without Node?

Well flow should be: `myst daemon --data-dir..` and `myst service`

PR #476, opened 2018-10-23
Add filter to cli command 'proposals'

tcharding:
Currently when using the cli and running command `proposals` we get all
the proposals back.  It would be nice to be able to narrow down the
search i.e using a filter.  We can filter based on the country and also
the provider ID.

Add filter term to command `proposals`.  No documentation added.  Needs
extending to include something in the output of `help` to show that this
feature exists.

With this applied
```
» proposals SG
2018-10-23T05:59:50.896303068 [Info] [Mysterium.api] Proposals fetched: 41
[INFO] Found 41 proposals (filter: 'SG')
[INFO] - provider id: 0x44ceda7262fcc48d591ee468c8a7319585d51aee, proposal id: 1, country: SG
[INFO] - provider id: 0xeed4ee266b2175a7f4063ecd59eef414b60f9d1d, proposal id: 1, country: SG
[INFO] - provider id: 0xf03f364719d09b37260872735c113baaee4f4f1a, proposal id: 1, country: SG
```

### DIFF
--- a/cmd/commands/cli/command.go
+++ b/cmd/commands/cli/command.go
@@ -121,7 +121,6 @@ func (c *cliApp) handleActions(line string) {
 		{"quit", c.quit},
 		{"help", c.help},
 		{"status", c.status},
-		{"proposals", c.proposals},
 		{"healthcheck", c.healthcheck},
 		{"ip", c.ip},
 		{"disconnect", c.disconnect},
@@ -138,6 +137,7 @@ func (c *cliApp) handleActions(line string) {
 		{command: "version", handler: c.version},
 		{command: "license", handler: c.license},
 		{command: "registration", handler: c.registration},
+		{command: "proposals", handler: c.proposals},
 	}
 
 	for _, cmd := range staticCmds {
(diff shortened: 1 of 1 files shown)


tcharding:
It would be nice to get this
```
» list
[INFO] Mysterium CLI tequilapi commands:
  connect 
  identities 
  ├── new 
  ├── list 
  status 
  healthcheck 
  proposals [filter-term]
...
```

cmd/commands/cli/command.go
```diff
@@ -291,18 +291,29 @@ func (c *cliApp) healthcheck() {
...
 		if country == "" {
 			country = "Unknown"
 		}
+
 		msg := fmt.Sprintf("- provider id: %v, proposal id: %v, country: %v", proposal.ProviderID, proposal.ID, country)
-		info(msg)
+
+		if filter == "" ||
+			strings.Contains(proposal.ProviderID, filter) ||
+			strings.Contains(country, filter) {
```
soffokl: If it contains the only country, maybe `filter` is not a best name? Why not call it `country` or something similar?

zolia: this filters not just for the country, but ProviderID itself, we might leave it 'filter'.

soffokl: true, missed it somehow, sorry.

tcharding:
Looks like this was a tooling issue again.  New version of `goimports` put the import list in a different order.  Let's see how travis likes it now.

cmd/commands/cli/command.go
tcharding: Other functions that accept an argument use `argString` for the argument identifier - I used 'filter' here to be more descriptive.

tcharding:
No love from Travis today.  Doesn't look like travis is running the build now and the very useful error message I can see is ` The Travis CI build could not complete du...`

Is there a limit on number of times a non-member  can push to a PR perhaps?

PR #479, opened 2018-10-24
Make broker flag configurable

### LINKED ISSUE #226: Refactor: BrokerPort should be configurable
In `communication/nats/discovery/config.go`, broker port is currently hard-coding as

```
BrokerPort             = 4222
```

It impossible to change if broker (NATS) use a different port.

tcharding:
Currently broker flag is hard coded as 4222.  We can make it configurable by adding an option `--broker-port`.

Make broker flag configurable.

Closes #226

### Testing
PR is untested.  With the following patch applied and running `bin/localnet/setup.sh` I can confirm port is open
```
sudo netstat -tlpn | grep 4333
tcp6       0      0 :::4333                 :::*                    LISTEN      20868/docker-pro
```

I cannot really work out ATM how to do further testing using the local net.  I'm a bit confused by `bin/run daemon` vs `bin/run service`.  (I've not got localnet running on master branch either.) 
```
diff --git a/bin/docker/docker-entrypoint.sh b/bin/docker/docker-entrypoint.sh
index 8be87972e9d9..80d67c28fe5d 100755
--- a/bin/docker/docker-entrypoint.sh
+++ b/bin/docker/docker-entrypoint.sh
@@ -12,4 +12,5 @@ exec /usr/bin/myst \
  --data-dir=$OS_DIR_DATA \
  --runtime-dir=$OS_DIR_RUN \
  --tequilapi.address=0.0.0.0 \
+ --broker-port=4333 \
  $@
diff --git a/bin/localnet/docker-compose.yml b/bin/localnet/docker-compose.yml
index 38f3a6373e52..9c669f195e27 100644
--- a/bin/localnet/docker-compose.yml
+++ b/bin/localnet/docker-compose.yml
@@ -4,7 +4,7 @@ services:
   broker:
     image: nats
     expose:
-      - 4222
+      - 4333
       - 8222
 
   #infrastructure - centralized api and db
diff --git a/bin/localnet/publish-ports.yml b/bin/localnet/publish-ports.yml
index f2c4867be384..ef0a41a74a57 100644
--- a/bin/localnet/publish-ports.yml
+++ b/bin/localnet/publish-ports.yml
@@ -3,7 +3,7 @@ services:
 
   broker:
     ports:
-      - 4222:4222
+      - 4333:4333
       - 8222:4222
 ```

### CODE REVIEW

**Comment by tcharding** [comment]
Yes @soffokl that was also suggested by @zyfdegh in issue #226.  @Waldz suggested a separate option.  My guess is this is to fit in with
```
bin/run -h
...
   --tequilapi.address value        IP address of interface to listen for incoming connections (default: "127.0.0.1")
   --tequilapi.port value           Port for listening incoming api requests (default: 4050)
...
```
but discovery address uses this method
```
run/bin -h
...
   --discovery-address value        Address (URL form) of discovery service (default: "https://testnet-api.mysterium.network/v1")
```
So we have an inconsistency here.  We should make it all uniform right?  Either the port is accepted as part of the address for _all_ addresses or for none.  My vote, for what its worth, goes with having the port as part of the address 'host:port' as is convention within the `net` package.

**Comment by tcharding** [comment]
What do you rekon @Waldz, I'll re-spin after you've made a call.

**Comment by Waldz** [comment]
I am for `contact.address=testnet-broker.mysterium.network:4222`

**Comment by tcharding** [comment]
Hey @Waldz could you explain your work flow to me please.  When you merge master into a pull request like done here what do you expect me to do when I go to update the PR?

1. Rebase and move the original patches to the top?
2. git reset hard and remove the merge then add new changes to the original patch?
3. Add patches on top (surely not :)
4. Other?

thanks man

**Comment by tcharding** [comment]
Updated this PR completely: 

- First patch fixes lint issues in `server/`
- Second patch fixes handling of `--broker-address`

Currently we do not handle command line option `--broker-address` correctly.  Firstly we hard code port number 4222 and append it to the string.  This means a user cannot configure NATS to use a different port.  We should check first if the string contains a port number and only use the default if required.  Secondly we should accept the protocol as part of the string.  We should support
```
--broker-address=[nats://]host[:port]
```
    
Fix handling of `--broker-address` to accept optional protocol and port number.  Keep defaults if not supplied.

File: communication/nats/discovery/address.go [hunk 1]
```diff
@@ -35,10 +37,22 @@ func NewAddress(topic string, addresses ...string) *AddressNATS {
...
-	address := fmt.Sprintf("nats://%s:%d", brokerIP, BrokerPort)
+func NewAddressGenerate(rawurl string, myID identity.Identity) *AddressNATS {
+	// url.Parse() functions incorrectly if rawurl doesn't start with protocol.
+	if !strings.HasPrefix(rawurl, "nats:") {
+		rawurl = fmt.Sprintf("nats://%s", rawurl)
+	}
+
+	url, err := url.Parse(rawurl)
+	if err != nil {
+		return nil
```
soffokl: We should not just hide an error. 
I'd change the function to return the error, and stop the application if the provided address was incorrect.

tcharding: I actually did that at first.  Then I thought that perhaps it was the callers responsibility to pass in a correct string.  Caller can still tell we failed by checking if returned string is "".  This imitates the way at least one function in the standard library works (I can't think which one off the top of my head though).  Doing it this way will mean we could verify the option strings when the `myst` command first starts.  This makes the rest of the code more simple since we _know_ we have valid option strings.  What do you think?

tadovas: Function naming. GenerateNewAddress ?

vkuznecovas: I'll agree with Dmitry - it's best to return the error here. Having to check for an empty string does not feel like a Go way of doing things. I'd stick with the error being returned and the good old if err != nil.

zolia:
```
--broker-address=[nats://]host[:port]
```
implies that you should use `nats` service. Idea for the future is possibly to have multiple types of brokers. Meaning we should have a generic URI of free form whose structure is not known a priori.

communication/nats/discovery/address.go
soffokl: > Function naming. GenerateNewAddress ?
What's about just `NewAddress`?

tcharding: Looks like we have consensus on this then.  Will re-do with an error return.  Thanks

tcharding:
> --broker-address=[nats://]host[:port]
> 
> implies that you should use nats service. Idea for the future is possibly to have multiple types of brokers. Meaning we should have a generic URI of free form whose structure is not known a priori.

This seems to add weight to the idea of doing some sort of verification on this option string (possibly when the command starts).  Did you want me to drop the use of 'nats' from the commit log @zolia?

> Function naming. GenerateNewAddress ?

@tadovas I don't understand this comment sorry.  Do you mean you don't like this function name?

communication/nats/discovery/address.go
tadovas: @tcharding replying to your question - yes I think GenerateNewAddress is a better name.

zolia:
> > --broker-address=[nats://]host[:port]
> > implies that you should use nats service. Idea for the future is possibly to have multiple types of brokers. Meaning we should have a generic URI of free form whose structure is not known a priori.
> 
> This seems to add weight to the idea of doing some sort of verification on this option string (possibly when the command starts). Did you want me to drop the use of 'nats' from the commit log @zolia?

from my understanding it should simply be: 
`--broker-address=URI`
Then, depending on what kind of broker was used, it should produce sensible error, describing format if needed.

communication/nats/discovery/address.go
tcharding: lol, my dislexia was playing up @tadovas I read your original comment a bunch of times before and didn't realize you had switched the word `Generate` from the back of the function name to the front.  Thanks for clarifying.

tcharding:
@tadovas I used function name `NewAddressFromHostAndID` and not your suggested `GenerateNewAddress` in order to fit in with the naming of other creator functions in the file.  Please feel free to improve upon the name.

Please review the changes to the help menu output since this is the first use of the back ticks.  Has there been discussion already on this usage?

thanks

tcharding:
@Waldz could you please remove WIP label (no longer 'Work in Progress') thanks.

soffokl:
@tadovas @zolia @vkuznecovas @Waldz Let's review this once again. Changes look reasonable, but it took too long without any actions.

soffokl:
I'll fix CI, so it's ready to review.

PR #505, opened 2018-10-25
bin: Add missing newline to end of file

tcharding:
A number of the shell scripts in bin/ do not end with a newline.  This is against UNIX common convention (and is bothersome because `cat $file` does not show the last line of the file).

Add newline to all scripts in bin/ that need it.  Done using:

	find -type f -print0 | xargs -0 -L1 bash -c 'test "$(tail -c 1 "$0")" && echo "" >> "$0"'

This PR seemingly goes against common commit format for this project; Is it OK to use a prefix for the git brief commit message?  If not I can re-spin without it.  For a new developer trying to gently ease his way into the project its nice to have the prefix so one can do simple PRs in a directory then iff merged do the same type of fix in other directories.  Having the prefix makes this a bit more explicit.  Thanks.

Signed-off-by: tcharding <me@tobin.cc>


bin/check_govet
```diff
@@ -22,4 +22,4 @@ fi
 go vet -composites=false ${ARGUMENTS}
 
 print_success "All files are compliant to go vet."
-exit 0
\ No newline at end of file
+exit 0
```
tadovas: I guess exit 0 can be removed here. As it's returned by default
Also add ```set -e flag``` at the beginning of the script - script will fail on any failing subcommand

bin/check_license
```diff
@@ -25,4 +25,4 @@ if [ ! -z "$bad_files" ]; then
 fi
 
 print_success "All files have copyright headers."
-exit 0
\ No newline at end of file
+exit 0
```
tadovas: Exit can safely be removed here

bin/localnet/deployer/keystore/UTC--2018-07-25T13-08-10.138246914Z--a754f0d31411d88e46aed455fa79b9fced122497
```diff
@@ -1 +1 @@
-{"address":"a754f0d31411d88e46aed455fa79b9fced122497","crypto":{"cipher":"aes-128-ctr","ciphertext":"56c3a9b187b9acfe64d9d9ab2f0236f990f533f99ddb05cc0d5be9340f7fb755","cipherparams":{"iv":"fa76b9c835859593db14cd4820078b88"},"kdf":"scrypt","kdfparams":{"dklen":32,"n":262144,"p":1,"r":8,"salt":"3dfdd74c4e7eb4d6ef3b449b836438f3082a34bd0c6dc82a85bde4f30421e1b9"},"mac":"97c1323a22cddb1dac9652a1b3b6b07706193a12cf056256dee963647686cde7"},"id":"965998e2-2a62-4aae-9b0d-b3a723b8d521","version":3}
\ No newline at end of file
+{"address":"a754f0d31411d88e46aed455fa79b9fced122497","crypto":{"cipher":"aes-128-ctr","ciphertext":"56c3a9b187b9acfe64d9d9ab2f0236f990f533f99ddb05cc0d5be9340f7fb755","cipherparams":{"iv":"fa76b9c835859593db14cd4820078b88"},"kdf":"scrypt","kdfparams":{"dklen":32,"n":262144,"p":1,"r":8,"salt":"3dfdd74c4e7eb4d6ef3b449b836438f3082a34bd0c6dc82a85bde4f30421e1b9"},"mac":"97c1323a22cddb1dac9652a1b3b6b07706193a12cf056256dee963647686cde7"},"id":"965998e2-2a62-4aae-9b0d-b3a723b8d521","version":3}
```
tadovas: These are autogenerated files by geth. Not sure if its meaningful to extra modify them

bin/check_govet
tcharding: > As it's returned by default

I disagree on that one @tadovas, Bash scripts exit with the status of the last command executed I believe.  You are right though if we use `set -e` then `exit 0` is redundant.  Sounds like we want project wide `set -e` in all project Bash scripts, no use of `exit 0`, and a newline terminating each file?

bin/localnet/deployer/keystore/UTC--2018-07-25T13-08-10.138246914Z--a754f0d31411d88e46aed455fa79b9fced122497
tcharding: Good catch, thanks.  Sloppy work from me.

bin/check_govet
tadovas: That default part I meant result of last command, thanks for correction. Well using set -e project wide is maybe too much, it depends on situation, but in this particular case I don't see a point to use explicit exit 0.

tcharding:
> If it happens, let's create CI job to prevent that bin/check_newlines

@Waldz I added a script to this PR `/bin/travis_scripts/ensure_nl.sh` with the intention that we could run it during CI.  Could you please review and see if that is what you meant.  I do not know how to plug it into travis.  thanks.

Waldz:
Direction looks good.
Just the same way, as `bin/check_license` script

PR #516, opened 2018-11-05
Fix build env

tadovas:


bin/package_ios
```diff
@@ -24,8 +24,13 @@ PACKAGE_FILE="${DIR_BUILD}/Mysterium.framework.zip"
...
 DIR_FRAMEWORK="${DIR_TEMP}/Mysterium.framework"
 
-xgo -image=mysteriumnetwork/xgo:1.11 -targets=ios/* -out Mysterium -dest $DIR_TEMP `pwd`/mobile/mysterium
+xgo \
+    -image=mysteriumnetwork/xgo:1.11 \
+    --ldflags="-w -s $(get_linker_ldflags)" \
+    -targets=ios/* -out Mysterium -dest $DIR_TEMP \
+    `pwd`/mobile/mysterium
 
+chown -R travis ${DIR_FRAMEWORK}
```
soffokl: This looks like a workaround for Travis CI only. What if I will run this script locally and I have no `travis` user?

tadovas: :(

tadovas: Well I guess we will find solution for local builds. Now fix that red master card

PR #521, opened 2018-11-06
mobile network options passed to entrypoint

zolia:


mobile/mysterium/entrypoint.go
```diff
@@ -88,8 +85,18 @@ func NewNode(appPath string) (*MobileNode, error) {
 	return &MobileNode{di}, nil
 }
 
+func GetTestnetMobileNetworkOptions() *MobileNetworkOptions {
```
tadovas: maybe DefaultNetworkOptions function would better show purpose

PR #523, opened 2018-11-06
Update session history

### LINKED ISSUE #433: Sessions endpoint
We need to keep track of past sessions. `GET /sessions` endpoint example:

```json
[
      {
        "id": "30f610a0-c096-11e8-b371-ebde26989839",
        "proposal": {
          "providerId": "0x3b03a513fba4bd4868edd340f77da0c920150f3e"
        },
        "location": {
          "country": "lt"
        },
        "statistics": {
          "dateStart": 1537787035230,
          "duration": 2100,
          "bytesSent": 1024,
          "bytesReceived": 6000
        }
      }
]
```

interro:
Closes #433
- This is second PR, which finishes basic scope of having sessions history
- [First part of PR](https://github.com/mysteriumnetwork/node/pull/498)
- Session history is updated on connection disconnect event
- Changes in stored session format, now saving TimeStarted and TimeUpdated instead of Duration.
- /sessions endpoint still returns Duration (TimeUpdated - TimeStarted) 
- Added session status


tequilapi/endpoints/sessions.go
```diff
@@ -102,21 +104,28 @@ func AddRoutesForSession(router *httprouter.Router, sessionStorage sessionStorag
...
 		ProviderID:      string(se.ProviderID.Address),
 		ServiceType:     se.ServiceType,
 		ProviderCountry: se.ProviderCountry,
 		DateStarted:     se.TimeStarted.Format("2018-10-29 16:22:05"),
-		BytesSent:       se.DataStats.BytesSent,
-		BytesReceived:   se.DataStats.BytesReceived,
-		Duration:        se.Duration,
+		BytesSent:       uint64(se.DataStats.BytesSent),
+		BytesReceived:   uint64(se.DataStats.BytesReceived),
+		Duration:        getDuration(se.TimeStarted, se.TimeUpdated),
```
vkuznecovas: move the getDuration method to connection.Session struct, that way you can just call se.GetDuration()

core/connection/session_storage.go
```diff
@@ -41,9 +43,9 @@ func (repo *SessionStorage) Save(se Session) error {
 }
 
 // Update updates specified fields of existing session by id
-func (repo *SessionStorage) Update(sessionID session.ID, duration int, dataStats stats.SessionStats) error {
+func (repo *SessionStorage) Update(sessionID session.ID, TimeUpdated time.Time, dataStats stats.SessionStats) error {
 	// update two fields by sessionID
-	se := Session{SessionID: sessionID, Duration: duration, DataStats: dataStats}
+	se := Session{SessionID: sessionID, TimeUpdated: TimeUpdated, DataStats: dataStats}
```
tadovas: Do we save session status - NEW, IN Progress, COMPLETED ? How can I tell which session is finished which is ongoing?

tequilapi/endpoints/sessions.go
interro: moved

core/connection/session_storage.go
interro: Is there a need for session status in history list? This can be added as feature later as we get feedback from using this endpoint

zolia: imho NEW and COMPLETED should be enough. New session will have TimeStarted set, completed session will have TimeUpdated set?

e2e/connection_test.go
```diff
@@ -127,6 +127,19 @@ func consumerConnectFlow(t *testing.T, tequilapi *tequilapi_client.Client, consu
 	assert.NoError(t, err)
 	seelog.Info("Shifted consumer IP: ", vpnIp)
```
zolia:
```suggestion
	seelog.Info("Changed consumer IP: ", vpnIp)
```

tequilapi/client/client.go
```diff
@@ -239,3 +239,16 @@ func (client *Client) Stop() error {
 
 	return nil
 }
+
+// GetSessions returns all sessions from history
+func (client *Client) GetSessions() (endpoints.SessionsDTO, error) {
```
zolia: `sessions := endpoints.SessionsDTO{}`
and using only `sessions` should be simpler.

interro: refactored

core/connection/session_storage.go
interro: added statuses New and Completed

core/connection/session.go
```diff
@@ -25,14 +25,43 @@ import (
...
 	SessionID       session.ID `storm:"id"`
 	ProviderID      identity.Identity
 	ServiceType     string
 	ProviderCountry string
 	TimeStarted     time.Time
-	// these will be updated while session is ongoing
-	Duration  int // in seconds
-	DataStats stats.SessionStats
+	Status          SessionStatus
+	TimeUpdated     time.Time          // is updated on disconnect event
```
soffokl: If it updated on disconnect event, maybe we should call it `Finished`?

core/connection/session.go
```diff
@@ -25,14 +25,43 @@ import (
...
+	SessionStatusCompleted = SessionStatus(1)
+)
+
 // Session holds structure for saving session history
 type Session struct {
 	SessionID       session.ID `storm:"id"`
 	ProviderID      identity.Identity
 	ServiceType     string
 	ProviderCountry string
 	TimeStarted     time.Time
```
soffokl: Maybe we can call it just `Started` and `Finished`, it has time `time.Time` no need `Time` word at the name of the field.

core/connection/session.go
```diff
@@ -25,14 +25,43 @@ import (
...
-	DataStats stats.SessionStats
+	Status          SessionStatus
+	TimeUpdated     time.Time          // is updated on disconnect event
+	DataStats       stats.SessionStats // is updated on disconnect event
+}
+
+// GetDuration returns delta in seconds (TimeUpdated - TimeStarted)
+func (se *Session) GetDuration() uint64 {
+	if se.TimeUpdated.IsZero() {
+		return 0
```
soffokl: If a session is not finished, `GetDuration` can return `time.Now() - se.TimeStarted` instead of 0.

core/connection/session.go
```diff
@@ -25,14 +25,43 @@ import (
 	"github.com/mysteriumnetwork/node/session"
 )
 
+// SessionStatus represents list of possible session statuses
+type SessionStatus int
```
soffokl: We can just use `iota` here to enumerate statuses.

core/connection/session.go
```diff
@@ -25,14 +25,43 @@ import (
...
+// GetDuration returns delta in seconds (TimeUpdated - TimeStarted)
+func (se *Session) GetDuration() uint64 {
+	if se.TimeUpdated.IsZero() {
+		return 0
+	}
+	return uint64(se.TimeUpdated.Sub(se.TimeStarted).Seconds())
+}
+
+// GetStatus converts status constant to string
+func (se *Session) GetStatus() string {
```
soffokl: I think it should be `String()` method for the SessionStatus type.

core/connection/session_storage.go
```diff
@@ -41,9 +43,10 @@ func (repo *SessionStorage) Save(se Session) error {
 }
 
 // Update updates specified fields of existing session by id
-func (repo *SessionStorage) Update(sessionID session.ID, duration int, dataStats stats.SessionStats) error {
-	// update two fields by sessionID
-	se := Session{SessionID: sessionID, Duration: duration, DataStats: dataStats}
+func (repo *SessionStorage) Update(sessionID session.ID, TimeUpdated time.Time, dataStats stats.SessionStats,
+	status SessionStatus) error {
```
soffokl: How does this formatted, why it splitted to 2 lines?

core/connection/session.go
interro: Later Session object can be updated during session. So I suggest to leave more generic name

core/connection/session.go
interro: removed word `Time`

core/connection/session_storage.go
interro: fixed

core/connection/session.go
interro: is it good idea to use iota with statuses stored in database?

core/connection/session.go
interro: If a session is not finished, that can mean a case where app crashes. So in that case `time.Now() - se.TimeStarted` will provide misleading result.
In current logic, in that case it will provide value 0.

PR #524, opened 2018-11-06
bin/check correct way to run golint

interro:


bin/check
```diff
@@ -6,6 +6,6 @@
 #> bin/check
 
 bin/check_govet
-bin/check_golint
+bin/builder_run bin/check_golint
```
soffokl: why do we need to run `golint` through the builder docker container?

tadovas: It seems that something happened with script - I don't know. Running inside docker was invented to keep script usable on local machines :| and now it seems we need to use docker even locally :(

soffokl: hm, it works for me locally. Could you give more details what is not working for you?

zolia: Can you comment on whats not working without `builder_run` ? We should not increase our build time if its not needed.

interro: for me bin/check_golint is not showing lint warnings

PR #540, opened 2018-11-15
Do not create S3 bucket for outside PR.

soffokl:
Since Travis CI does not provide environment variables for outside PRs, we cannot create S3 bucket there. 
The good thing that we should not do it. 
This PR disables S3 bucket creation for PRs.


tadovas:
Again? :(

PR #541, opened 2018-11-15
Multiple services

vkuznecovas:
`myst service` now runs all the supported services(openvpn, noop). Adjusted the cli accordingly, we're  now able to connect to a provider by specifying the service type as well. It's also possible to start a single service via `myst service openvpn` or `myst service noop`

### DIFF
--- a/cmd/commands/cli/command.go
+++ b/cmd/commands/cli/command.go
@@ -36,12 +36,12 @@ import (
 const cliCommandName = "cli"
 
 // NewCommand constructs CLI based Mysterium UI with possibility to control quiting
-func NewCommand() *cli.Command {
+func NewCommand(defaultServicesToStart []string) *cli.Command {
 	return &cli.Command{
 		Name:  cliCommandName,
 		Usage: "Starts a CLI client with a Tequilapi",
 		Action: func(ctx *cli.Context) error {
-			nodeOptions := cmd.ParseFlagsNode(ctx)
+			nodeOptions := cmd.ParseFlagsNode(ctx, defaultServicesToStart)
 			cmdCLI := &cliApp{
 				historyFile: filepath.Join(nodeOptions.Directories.Data, ".cli_history"),
 				tequilapi:   tequilapi_client.NewClient(nodeOptions.TequilapiAddress, nodeOptions.TequilapiPort),
--- a/cmd/commands/cli/command.go
+++ b/cmd/commands/cli/command.go
@@ -163,17 +163,17 @@ func (c *cliApp) handleActions(line string) {
 func (c *cliApp) connect(argsString string) {
 	options := strings.Fields(argsString)
 
-	if len(options) < 2 {
-		info("Please type in the provider identity. Connect <consumer-identity> <provider-identity> [disable-kill-switch]")
+	if len(options) < 3 {
+		info("Please type in the provider identity. Connect <consumer-identity> <provider-identity> <service-type> [disable-kill-switch]")
 		return
 	}
 
-	consumerID, providerID := options[0], options[1]
+	consumerID, providerID, serviceType := options[0], options[1], options[2]
 
 	var disableKill bool
 	var err error
-	if len(options) > 2 {
-		disableKillStr := options[2]
+	if len(options) > 3 {
+		disableKillStr := options[3]
 		disableKill, err = strconv.ParseBool(disableKillStr)
 		if err != nil {
 			info("Please use true / false for <disable-kill-switch>")
@@ -195,7 +195,7 @@ func (c *cliApp) connect(argsString string) {
 
 	status("CONNECTING", "from:", consumerID, "to:", providerID)
 
-	_, err = c.tequilapi.Connect(consumerID, providerID, connectOptions)
+	_, err = c.tequilapi.Connect(consumerID, providerID, serviceType, connectOptions)
 	if err != nil {
 		warn(err)
 		return
--- a/cmd/commands/cli/command.go
+++ b/cmd/commands/cli/command.go
@@ -36,12 +36,12 @@ import (
 const cliCommandName = "cli"
 
 // NewCommand constructs CLI based Mysterium UI with possibility to control quiting
-func NewCommand(defaultServicesToStart []string) *cli.Command {
+func NewCommand() *cli.Command {
 	return &cli.Command{
 		Name:  cliCommandName,
 		Usage: "Starts a CLI client with a Tequilapi",
 		Action: func(ctx *cli.Context) error {
-			nodeOptions := cmd.ParseFlagsNode(ctx, defaultServicesToStart)
+			nodeOptions := cmd.ParseFlagsNode(ctx)
 			cmdCLI := &cliApp{
 				historyFile: filepath.Join(nodeOptions.Directories.Data, ".cli_history"),
 				tequilapi:   tequilapi_client.NewClient(nodeOptions.TequilapiAddress, nodeOptions.TequilapiPort),
--- a/cmd/commands/service/command.go
+++ b/cmd/commands/service/command.go
@@ -52,15 +52,16 @@ var (
 )
 
 // NewCommand function creates service command
-func NewCommand(licenseCommandName string, defaultServiceTypes []string) *cli.Command {
+func NewCommand(licenseCommandName string) *cli.Command {
+	serviceTypes := []string{"openvpn", "noop"}
 	var di cmd.Dependencies
 	command := &cli.Command{
 		Name:        serviceCommandName,
 		Usage:       "Starts and publishes services on Mysterium Network",
 		ArgsUsage:   " ",
-		Subcommands: getSubcommands(&di, licenseCommandName, defaultServiceTypes),
+		Subcommands: getSubcommands(&di, licenseCommandName, serviceTypes),
 		Action: func(ctx *cli.Context) error {
-			return runServices(ctx, &di, licenseCommandName, defaultServiceTypes)
+			return runServices(ctx, &di, licenseCommandName, serviceTypes)
 		},
 		After: func(ctx *cli.Context) error {
 			return di.Shutdown()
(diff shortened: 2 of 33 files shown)


vkuznecovas:
closes #483

services/noop/connection.go
```diff
@@ -18,33 +18,49 @@
...
 type Connection struct {
-	stateChannel connection.StateChannel
+	isRunning      bool
+	noopConnection sync.WaitGroup
+	stateChannel   connection.StateChannel
 }
 
 // Start implements the connection.Connection interface
 func (c *Connection) Start() error {
+	c.noopConnection.Add(1)
```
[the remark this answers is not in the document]
vkuznecovas: Good catch, will fix

services/noop/bootstrap.go
```diff
@@ -0,0 +1,47 @@
...
+
+package noop
+
+import (
+	"encoding/json"
+
+	dto_discovery "github.com/mysteriumnetwork/node/service_discovery/dto"
+)
+
+// Bootstrap is called on program initialization time and registers various deserializers related to opepnvpn service
```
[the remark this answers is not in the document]
tadovas: Noop service has nothing to do with openvpn

services/noop/service.go
```diff
@@ -30,13 +31,14 @@ import (
...
-	return &Manager{}
+func NewManager(resolver location.Resolver) *Manager {
+	return &Manager{locationResolver: resolver}
 }
 
 // Manager represents entrypoint for Noop service
 type Manager struct {
-	fakeProcess sync.WaitGroup
+	fakeProcess      sync.WaitGroup
+	locationResolver location.Resolver
```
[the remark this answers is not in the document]
tadovas: I guess to lookup country in which service is provided. But country is hardcoded below :(

cmd/commands/cli/command.go
```diff
@@ -163,17 +163,17 @@ func (c *cliApp) handleActions(line string) {
 func (c *cliApp) connect(argsString string) {
 	options := strings.Fields(argsString)
 
-	if len(options) < 2 {
-		info("Please type in the provider identity. Connect <consumer-identity> <provider-identity> [disable-kill-switch]")
+	if len(options) < 3 {
+		info("Please type in the provider identity. Connect <consumer-identity> <provider-identity> <service-type> [disable-kill-switch]")
```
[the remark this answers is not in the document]
vkuznecovas: If we're moving towards a compound ID for proposals it think it only makes sense not to hide this detail on the cli.

core/service/runner.go
```diff
@@ -0,0 +1,70 @@
...
+		return err
+	}
+
+	go func() { errorChannel <- sr.serviceManagers[serviceType].Wait() }()
+
+	return nil
+}
+
+// Kill kills all the service managers
+func (sr *Runner) Kill() []error {
```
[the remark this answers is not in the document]
vkuznecovas: done

services/noop/connection.go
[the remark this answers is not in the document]
vkuznecovas: done

tequilapi/endpoints/connection.go
```diff
@@ -297,7 +303,11 @@ func AddRoutesForConnection(router *httprouter.Router, manager connection.Manage
 }
 
 func toConnectionRequest(req *http.Request) (*connectionRequest, error) {
-	var connectionRequest = connectionRequest{}
+	var connectionRequest = connectionRequest{
+		// This defaults the service type to openvpn, for backwards compatability
```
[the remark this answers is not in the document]
vkuznecovas: fixd

tequilapi/endpoints/connection.go
```diff
@@ -297,7 +303,11 @@ func AddRoutesForConnection(router *httprouter.Router, manager connection.Manage
...
+		// If specified in the request, the value will get overriden
```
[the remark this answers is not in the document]
vkuznecovas: fixd

services/noop/service.go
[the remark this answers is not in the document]
vkuznecovas: correct, somehow forgot it - implemented.

cmd/commands/service/command.go
```diff
@@ -51,51 +52,83 @@ var (
...
 	return command
 }
 
+func runServices(ctx *cli.Context, di *cmd.Dependencies, licenseCommandName string, serviceTypes []string) error {
+	if !ctx.Bool(agreedTermsConditionsFlag.Name) {
+		printTermWarning(licenseCommandName)
+		os.Exit(2)
+	}
+
+	errorChannel := make(chan error, 2+len(serviceTypes))
```
zolia: Why we need 2+len here? Sort comment would be good here.

core/service/manager.go
```diff
@@ -46,10 +46,11 @@ type Service interface {
 	Start(providerID identity.Identity) (dto_discovery.ServiceProposal, session.ConfigProvider, error)
 	Wait() error
 	Stop() error
+	GetType() string
```
zolia: This looks artificial. Maybe its better to get type from serviceProposal returned by manager.service.Start(providerID) if needed? We should know what we are running.

core/service/runner.go
```diff
@@ -0,0 +1,70 @@
...
+	if err := sr.serviceManagers[serviceType].Start(options); err != nil {
+		return err
+	}
+
+	go func() { errorChannel <- sr.serviceManagers[serviceType].Wait() }()
+
+	return nil
+}
+
+// KillAll kills all the service managers
```
zolia:
```suggestion
// KillAll kills all service managers
```

server/mysterium_api_fake.go
```diff
@@ -79,15 +79,19 @@ func (client *ClientFake) PingProposal(proposal dto_discovery.ServiceProposal, s
...
-		var filterMatched = true
+		var providerMatched = true
 		if providerID != "" {
-			filterMatched = filterMatched && (providerID == proposal.ProviderID)
+			providerMatched = providerMatched && (providerID == proposal.ProviderID)
 		}
-		if filterMatched {
+		var serviceMatched = true
+		if serviceType != "" {
+			serviceMatched = serviceMatched && (serviceType == proposal.ServiceType)
```
zolia:
```suggestion
			serviceMatched = (serviceType == proposal.ServiceType)
```

services/noop/service.go
```diff
@@ -30,46 +32,81 @@ import (
...
 }
 
 // Start starts service - does not block
 func (manager *Manager) Start(providerID identity.Identity) (dto_discovery.ServiceProposal, session.ConfigProvider, error) {
+	sessionConfigProvider := func() (session.ServiceConfiguration, error) {
+		return nil, nil
+	}
+
+	if manager.isStarted {
+		return dto_discovery.ServiceProposal{}, sessionConfigProvider, nil
```
zolia: Should it be "ErrAlreadyStarted" or return the same instance?

services/noop/service.go
```diff
@@ -30,46 +32,81 @@ import (
...
+	}
+
 	manager.fakeProcess.Add(1)
```
zolia: We prob should not use 'fake' wording anymore here.

tequilapi/endpoints/connection.go
```diff
@@ -56,6 +56,12 @@ type connectionRequest struct {
 	// example: 0x0000000000000000000000000000000000000002
 	ProviderID string `json:"providerId"`
 
+	// service type
+	// required: false
+	// default: openvpn
+	// example: openvpn
```
zolia: We should see somewhere what available options there are: openvpn, noop

services/noop/service.go
tadovas: Agreed

core/service/manager.go
vkuznecovas: The way I see it, we'll still need the manager(maybe the naming needs a change, but naming is hard). I like that it's responsible for 
1) Registering/pinging the proposal
1) Starting the service it's given
1) Starting to listen for the relevant dialogs

We'll need these for every service, so in my head the component makes perfect sense.

cmd/commands/service/command.go
```diff
@@ -51,51 +52,83 @@ var (
...
+		return err
+	}
+	go func() { errorChannel <- di.Node.Wait() }()
+
+	for _, serviceType := range serviceTypes {
+		options, err := parseFlagsByServiceType(ctx, serviceType)
+		if err != nil {
+			return err
+		}
+		err = di.ServiceRunner.StartServiceByType(serviceType, options, errorChannel)
```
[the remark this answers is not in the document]
vkuznecovas: the error here tells us if the startup of services was successful. The channel is then used to report runtime errors that occur in the underlying service.

vkuznecovas:
Did a rework where a proposal is now passed to the connect method, and the manager no longer fetches it. Fixed most of the comments.

e2e/connection_test.go
```diff
@@ -114,7 +114,7 @@ func consumerConnectFlow(t *testing.T, tequilapi *tequilapi_client.Client, consu
 	})
 	assert.NoError(t, err)
 
-	connectionStatus, err = tequilapi.Connect(consumerID, proposal.ProviderID, endpoints.ConnectOptions{true})
+	connectionStatus, err = tequilapi.Connect(consumerID, proposal.ProviderID, "openvpn", endpoints.ConnectOptions{true})
```
soffokl: Shouldn't we pass here a proposal too?

core/service/manager.go
tadovas: In that case yes - better naming is needed. I didn't expect to see multiple managers :)

cmd/commands/service/command.go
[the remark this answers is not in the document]
tadovas: Hmm. Do we really need to report those errors upstairs? Startup errors are pretty much ok, but runtime errors - well I am not sure, what happens in case of that error - we shutdown whole node? What if other services run perfectly? Maybe some kind of retry mechanism is needed? Don't forget that service will be managed through api in the future - not only by args on startup.

e2e/connection_test.go
vkuznecovas: out of scope, we'll need to fix the api part though, to avoid having to query for the service type in general. I've created #549 for this.

vkuznecovas:
The runners "startServiceByType" now blocks.

e2e/connection_test.go
tadovas: Awesome 👍

server/interface.go
```diff
@@ -28,7 +28,7 @@ import (
 type Client interface {
 	RegisterIdentity(id identity.Identity, signer identity.Signer) (err error)
 
-	FindProposals(providerID string) (proposals []dto_discovery.ServiceProposal, err error)
+	FindProposals(providerID string, serviceType string) (proposals []dto_discovery.ServiceProposal, err error)
```
tadovas: Maybe it's time to add method for single proposal fetch by it's identification

Issue #530, opened 2018-11-09
Final removal of "openvpn" wording


vkuznecovas:
Opening this as a separate issue so we can spread it over a few PR's.

Leftovers:
- [x] TODO Decouple stats.Sender from 'openvpn' states #542 
- [x] TODO Move stats.Sender from 'openvpn' package #542
- [x] TODO Register Openvpn flags as transport "openvpn" arguments #499
- [ ] TODO Bootstrap services in desktop entrypoint only
- [ ] TODO Bootstrap openvpn together with transport
- [ ] TODO Use transport mock in tests instead of real openvpn
- [x] Move `mysteriumnetwork/go-openvpn/openvpn/core` to node repo #529

PR #542, opened 2018-11-15
Decouple stats keeper from openvpn

vkuznecovas:
Updates #530

### DIFF
--- a/client/stats/sender.go
+++ b/client/stats/sender.go
@@ -18,10 +18,10 @@
 package stats
 
 import (
+	"sync"
 	"time"
 
 	log "github.com/cihub/seelog"
-	"github.com/mysteriumnetwork/go-openvpn/openvpn"
 	"github.com/mysteriumnetwork/node/identity"
 	"github.com/mysteriumnetwork/node/server"
 	"github.com/mysteriumnetwork/node/server/dto"
@@ -44,6 +44,9 @@ type RemoteStatsSender struct {
 
 	sendInterval time.Duration
 	done         chan struct{}
+
+	opLock  sync.Mutex
+	started bool
 }
 
 // NewRemoteStatsSender function creates new session stats sender by given options
@@ -63,15 +66,31 @@ func NewRemoteStatsSender(statsKeeper SessionStatsKeeper, mysteriumClient server
 	}
 }
 
-// StateHandler expects connect and disconnect events from the OpenVPN client to start or stop actual sending stats.
-func (rss *RemoteStatsSender) StateHandler(state openvpn.State) {
-	// TODO Decouple stats.Sender from 'openvpn' states
-	switch state {
-	case openvpn.ConnectedState:
-		go rss.intervalSend()
-	case openvpn.ExitingState:
-		close(rss.done)
+// Start starts the sending of stats
+func (rss *RemoteStatsSender) Start() {
+	rss.opLock.Lock()
+	defer rss.opLock.Unlock()
+
+	if rss.started {
+		return
 	}
+
+	rss.done = make(chan struct{})
+	go rss.intervalSend()
+	rss.started = true
+}
+
+// Stop stops the sending of stats
+func (rss *RemoteStatsSender) Stop() {
+	rss.opLock.Lock()
+	defer rss.opLock.Unlock()
+
+	if !rss.started {
+		return
+	}
+
+	close(rss.done)
+	rss.started = false
 }
 
 func (rss *RemoteStatsSender) intervalSend() {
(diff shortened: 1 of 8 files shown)


client/stats/sender.go
```diff
@@ -30,6 +29,16 @@ import (
 
 const statsSenderLogPrefix = "[session-stats-sender] "
 
+// State represents the state changes in connection that the sender is interested in
+type State int
+
+const (
+	// Connected represents the connected state
+	Connected State = iota
```
tadovas: What is wrong with constants defined in connection manager?
Stats sender is aware and handles same connection states

cmd/di.go
```diff
@@ -213,13 +213,18 @@ func (di *Dependencies) bootstrapServiceOpenvpn(nodeOptions node.Options) {
 	}
 	di.ServiceRegistry.Register(service_openvpn.ServiceType, createService)
 
+	statsSenderFactory := func(statsKeeper stats.SessionStatsKeeper, mysteriumClient server.Client, sessionID session.ID, providerID identity.Identity, signer identity.Signer, consumerCountry string, interval time.Duration) service_openvpn.StatsSender {
```
tadovas: Why stats sender has to be factory?

services/openvpn/connection.go
```diff
@@ -51,3 +52,16 @@ func GetStateCallback(stateChannel connection.StateChannel) func(openvpnState op
 		}
 	}
 }
+
+// GetStatsSenderCallback returns a stats sender callback that knows how to work with openvpn state
+func GetStatsSenderCallback(statsHandler func(state stats.State)) func(openvpnState openvpn.State) {
```
tadovas: This seems to be overcomplicated and not necessary

services/openvpn/connection_factory.go
```diff
@@ -75,19 +94,20 @@ func (op *ProcessBasedConnectionFactory) newBytecountMiddleware() management.Mid
 }
 
 func (op *ProcessBasedConnectionFactory) newStateMiddleware(session session.ID, signer identity.Signer, connectionOptions connection.ConnectOptions, stateChannel connection.StateChannel) management.Middleware {
-	// TODO Move stats.Sender from 'openvpn' package
-	statsSender := stats.NewRemoteStatsSender(
+	statsSender := op.statsSenderFactory(
```
tadovas: Why that, statistics sender is leaked into openvpn service, and cannot be simply a part of conn manager? Each connection (a.k.a openvpn or other) should provide clear callbacks - stats, events etc. Conn manager handles them appropriatly

services/openvpn/connection_factory_test.go
```diff
@@ -76,3 +79,21 @@ func TestConnectionFactory_CreatesConnection(t *testing.T) {
 	assert.Nil(t, err)
 	assert.NotNil(t, conn)
 }
+
+type FakeStatsSender struct {
```
tadovas: Mock or stub is more appropriate definition in test world

vkuznecovas:
I've removed the stats sender from openvpn completely. Moved it up to connection manager. The stats sender now only has a start and a stop method, it knows absolutely nothing about states.

core/connection/stub_stat_sender.go
```diff
@@ -0,0 +1,34 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package connection
+
+// StubStatSender represents a stat sender stub
+type StubStatSender struct {
```
tadovas: I guess it's used in tests? Does it really deservers it's own file?

core/connection/manager.go
```diff
@@ -166,6 +185,8 @@ func (manager *connectionManager) startConnection(consumerID identity.Identity,
 		Proposal:      proposal,
 	}
 
+	manager.statsSender = manager.statsSenderFactory(consumerID, sessionID, proposal, time.Minute)
```
tadovas: This stats sender lifecycle is still strage to me. It's constructed on every connection, then on state callback its start/stop methods are called. So its temporal depedency - created on one method, methods called on different. Can that stats sender be made like component injected into connection manager and simply handle stats, and conn start stop .methods internally (by resetting itself when needed or doing separate unrelated goroutines whatever)

vkuznecovas: I agree that it's strange. I think this component shouldn't even be a component, a func would do here. There's no state it needs to keep internally - just send a request with the given data. Maybe it needs to be merged with the keeper, just adding a .Send() to it and renaming it to a stat handler? I don't feel that the schedule part of the component makes it a worthy component at all. I'm also not entirely sure that this is the appropriate PR to rewrite this whole component in general - the idea here was to move it up a level, instead of having it under the openvpn connection.

vkuznecovas:
#567 includes this now, so i'll close it.

PR #555, opened 2018-11-21
Added Wireguard noop support

### LINKED ISSUE #502: Frame WG service as Node PT with Noops

soffokl:
Closes #502

### DIFF
--- a/cmd/commands/service/command.go
+++ b/cmd/commands/service/command.go
@@ -28,6 +28,7 @@ import (
 	service_noop "github.com/mysteriumnetwork/node/services/noop"
 	service_openvpn "github.com/mysteriumnetwork/node/services/openvpn"
 	openvpn_service "github.com/mysteriumnetwork/node/services/openvpn/service"
+	service_wireguard "github.com/mysteriumnetwork/node/services/wireguard"
 	"github.com/urfave/cli"
 )
 
@@ -53,7 +54,7 @@ var (
 
 // NewCommand function creates service command
 func NewCommand(licenseCommandName string) *cli.Command {
-	serviceTypes := []string{"openvpn", "noop"}
+	serviceTypes := []string{"openvpn", "wireguard", "noop"}
 	var di cmd.Dependencies
 	command := &cli.Command{
 		Name:        serviceCommandName,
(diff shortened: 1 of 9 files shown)


cmd/commands/service/command.go
```diff
@@ -28,6 +28,7 @@ import (
 	service_noop "github.com/mysteriumnetwork/node/services/noop"
 	service_openvpn "github.com/mysteriumnetwork/node/services/openvpn"
 	openvpn_service "github.com/mysteriumnetwork/node/services/openvpn/service"
+	service_wireguard "github.com/mysteriumnetwork/node/services/wireguard"
```
tadovas: There is clearly something wrong with package structure inside openvpn :( service_openvpn and openvpn_service? I know it's out of this PR scope, but please don't do it with wireguard too :)

services/wireguard/service.go
```diff
@@ -0,0 +1,151 @@
...
+	// out, err := exec.Command("ip", "address", "add", "dev", "wg0", "192.168.100.1/24").CombinedOutput()
+
+	// TODO configure wireguard interface.
+	// client, err := wireguardctrl.New()
+	// if err != nil {
+	// 	return Config{}, err
+	// }
+
+	// TODO wireguard device configuration like private key, listen port, peer list should survive through restarts.
+	// TODO we need to have some persistent storage for it.
```
zolia: We can create separate ticket for that. Its not high priority IMHO.

services/wireguard/service.go
```diff
@@ -0,0 +1,151 @@
...
+	// client.ConfigureDevice("wg0", wgtypes.Config{
+	// 	PrivateKey: "8C6Pp0cI2tgLeYOMVbnMMLl/zN2blFK+OWGaIxX0bHY=",
+	// 	ListenPort: 52820,
+	// 	Peers:      nil,
+	// })
+
+	// TODO if the wireguard interface already configured we can get required parameters from it.
+	// device, err := client.Device("wg0")
```
zolia: WG device name should be selected only after listing all active devices. Then choosing free number. We should know which device constitutes which session.

tadovas:
After sync with master :)

Issue #556, opened 2018-11-21
Bidirectional data exchange for service provider and consumer


soffokl:
Wireguard requires public keys exchange between server and client to provide service. 
The current implementation provides the only option for a service provider to generate configuration.

We need to have the ability to send extra information(consumer public key for wireguard) from consumer during connection creating.

Waldz:
"session-create" endpoint has possibility to have incoming parameters from consumer side

PR #561, opened 2018-11-22
 Added Wireguard service manager

### LINKED ISSUE #503: Implement service.Manager for WG

soffokl:
Closes #503


services/wireguard/service.go
```diff
@@ -111,41 +132,82 @@ func (manager *Manager) Stop() error {
...
+		ReplacePeers: true,
+	})
+}
+
+func (manager *Manager) deleteInterface(name string) error {
+	return exec.Command("ip", "link", "del", "dev", name).Run()
+}
+
+func (manager *Manager) peerConfig(name, publicIP string) (Config, error) {
+	peerKey, err := wgtypes.GeneratePrivateKey()
```
soffokl: Generating consumer private key on the provider side for POC.
It will be replaced once https://github.com/mysteriumnetwork/node/issues/556 will be resolved.

services/wireguard/service.go
```diff
@@ -49,13 +58,15 @@ type Manager struct {
...
+	wgClient         *wireguardctrl.Client
 }
 
 // Config represent a Wireguard service provider configuration that will be passed to the consumer for establishing a connection
 type Config struct {
 	PublicKey string
-	IP        string
 	Endpoint  string
+	PeerKey   string // TODO peer private key should be generated on consumer side
+	PeerIP    string
```
tadovas: I would like to have more indication which part of config is related to provider and which part to consumer.
I.E. PublicKey -> ProviderPublicKey
Endpoint -> ProviderEndpoint
PeerKey -> ConsumerPrivateKey
PeerIP -> ConsumerIP

services/wireguard/service.go
```diff
@@ -111,41 +132,82 @@ func (manager *Manager) Stop() error {
...
-	// 	ListenPort: 52820,
-	// 	Peers:      nil,
-	// })
-
-	// TODO if the wireguard interface already configured we can get required parameters from it.
-	// device, err := client.Device("wg0")
-	// if err != nil {
-	// 	return Config{}, err
-	// }
+func (manager *Manager) initInterface(name string) error {
```
tadovas: Real tunnel device creation is really OS specific and must be abstracted away from ServiceManager - on linux we can use this approach, on other OSes - userspace impl.

tadovas: It will possibly break e2e

services/wireguard/service.go
soffokl: what if we will make nested JSON?

```JSON
{
	"Provider": {
		"PublicKey": "aaaaaaaaaaaa",
		"Endpoint": "1.2.3.4:12345"
	},
	"Consumer": {
		"PrvivateKey": "bbbbbbbbbbb",
		"IP": "192.168.1.1"
	}
}
```

services/wireguard/service.go
soffokl: Agree, I'll add abstraction level. And for E2E tests we will need to test every transport separately. We have a task for the E2E test for wireguard somewhere.

services/wireguard/service.go
tadovas: Looks better 👍

services/wireguard/network/network.go
```diff
@@ -0,0 +1,44 @@
...
+type Consumer struct {
+	PrivateKey string // TODO peer private key should be generated on consumer side
+	IP         string
+}
+
+var errNotSupported = errors.New("OS is not supported for wireguard transport")
+
+type network struct {
+	name     string
+	publicIP string
```
zolia: This one probably will never be used. Whats its purpose?

zolia: If you mean by it Endpoint address, then it should include port also. Those should not go separately. Lets operate with terms that are used by original project (WG). This endpoint may differ if server has multiple interfaces. We should let service operator decide which to use. If not specified - select automatically, similarly how we detect public IP (outbound IP) atm.

services/wireguard/network/network_linux.go
```diff
@@ -0,0 +1,114 @@
...
+	if err := exec.Command("ip", "address", "replace", "dev", n.name, "192.168.100.1/24").Run(); err != nil {
+		return Provider{}, err
+	}
+
+	if err := exec.Command("ip", "link", "set", "dev", n.name, "up").Run(); err != nil {
+		return Provider{}, err
+	}
+
+	// TODO wireguard provider listen port should be passed as startup argument
+	port := 52820
```
zolia: Whole service endpoint should be passed as startup, not just port.

services/wireguard/network/network_linux.go
```diff
@@ -0,0 +1,114 @@
...
+// Provider starts and configure wireguard network interface for providing service.
+// It returns information required to establish connection to the service.
+func (n *network) Provider() (Provider, error) {
+	if _, err := n.wgClient.Device(n.name); err != nil {
+		if err := exec.Command("ip", "link", "add", "dev", n.name, "type", "wireguard").Run(); err != nil {
+			return Provider{}, err
+		}
+	}
+
+	if err := exec.Command("ip", "address", "replace", "dev", n.name, "192.168.100.1/24").Run(); err != nil {
```
zolia: I would suggest not to use 192.168.x.x, since it has higher probability of clash. Lets use some random from 10.x.x.x as per https://tools.ietf.org/html/rfc1918

```suggestion
	if err := exec.Command("ip", "address", "replace", "dev", n.name, "10.182.47.0/24").Run(); err != nil {
```

services/wireguard/network/network_linux.go
zolia: Having separate wg device for each peer means we have to auto-select port to listen. We just could pass some "base" port not to clash with some mainstream ports. 52820 could be base port.

services/wireguard/network/network_linux.go
```diff
@@ -0,0 +1,114 @@
...
+// Consumer adds service consumer public key to the list of allowed peers.
+// It returns information required to configure a consumer instance to allow connections.
+func (n *network) Consumer() (Consumer, error) {
+	peerKey, err := wgtypes.GeneratePrivateKey()
+	if err != nil {
+		return Consumer{}, err
+	}
+
+	// TODO constant peer IP should be replaced by some generation to allow more than one peer to connect
+	_, peerIP, err := net.ParseCIDR("192.168.100.2/32")
```
zolia: We could have some ResourceHandler service, to handle such OS resources required by peer: wg interface name, PeerAllowedIP, WG port. We could use port number in WG device name.

zolia: This can be a mock in this PR

services/wireguard/service.go
```diff
@@ -43,19 +48,27 @@ func NewManager(locationResolver location.Resolver, ipResolver ip.Resolver) *Man
...
-	PublicKey string
-	IP        string
-	Endpoint  string
+	Provider network.Provider
+	Consumer network.Consumer
+}
+
+// Network represents Wireguard network instance, it provide information
+// required for establishing connection between service provider and consumer.
+type Network interface {
```
zolia: Too generic name here. Would be nice to see a hint that it is an interface. I suggest 'Networker'.

services/wireguard/network/network.go
```diff
@@ -0,0 +1,44 @@
...
+
+// Consumer is a configuration data required to configure service consumer.
+type Consumer struct {
+	PrivateKey string // TODO peer private key should be generated on consumer side
+	IP         string
+}
+
+var errNotSupported = errors.New("OS is not supported for wireguard transport")
+
+type network struct {
```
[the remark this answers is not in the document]
soffokl: Renamed.

services/wireguard/network/network.go
soffokl: Passing now endpoint.

services/wireguard/network/network_linux.go
soffokl: Changed it to the mocked network from 10.0.0.0/8.

services/wireguard/network/network_linux.go
soffokl: Moved port definition to resource handler. It's mocked now but it will be possible to pass it later.

services/wireguard/network/network_linux.go
soffokl: It's a good suggestion, added the mocked resource hander.

services/wireguard/service.go
soffokl: Renamed to `ConfigProvider` to have clear understandig what it does.

soffokl:
@tadovas @zolia it's ready to review.

Gopkg.toml
```diff
@@ -85,3 +85,7 @@
 [[constraint]]
   name = "github.com/asdine/storm"
   version = "2.1.2"
+
+[[constraint]]
+  branch = "master"
```
tadovas: better freeze to concrete commit from master - revision="<commit id>" to avoid surprises later

services/wireguard/endpoint.go
```diff
@@ -0,0 +1,131 @@
...
+
+// ProviderConfig starts and configure wireguard network interface for providing service.
+// It returns information required to establish connection to the service.
+func (ce *connectionEndpoint) ProviderConfig(ip net.IPNet) (Provider, error) {
+	if err := ce.up(ip); err != nil {
+		return Provider{}, err
+	}
+
+	// TODO wireguard provider listen port should be passed as startup argument
+	port := 52820
```
tadovas: It can be set to 0 - os will pick random freely available port

services/wireguard/service.go
```diff
@@ -27,35 +28,46 @@ import (
...
 type Manager struct {
-	process          sync.WaitGroup
-	locationResolver location.Resolver
-	ipResolver       ip.Resolver
-	isStarted        bool
+	isStarted          bool
+	process            sync.WaitGroup
+	locationResolver   location.Resolver
+	ipResolver         ip.Resolver
+	connectionEndpoint ConfigProvider
```
tadovas: connectionEndpoint but its actually ConfigProvider? So it's connection or config provider? Inconsistent naming

services/wireguard/service.go
```diff
@@ -27,35 +28,46 @@ import (
...
-	IP        string
-	Endpoint  string
+	Provider Provider
+	Consumer Consumer
+}
+
+// ConfigProvider represents Wireguard network instance, it provide information
+// required for establishing connection between service provider and consumer.
+type ConfigProvider interface {
+	ProviderConfig(ip net.IPNet) (Provider, error)
```
tadovas: ConfigProvider method ProviderConfig returns Provider which is actually configuration of service provider. As per previous suggestion why is bad to call Provider struct what it actually is ? ProviderConfiguration, the same with Consumer -> ConsumerConfiguration. Another question - why ip parameter is needed to provide config - maybe it either can be set in actual interface implementor or other parameters should be exposed too. Actually I don't understand purpose of this interface, especially Close method (Config provider can be closed, does that imply that some kind of resources need to be released?)

And actual config provider implementation can take resources as depedency and do all that allocation inside. That way details wont be exposed in service manager itself (i.e. IP passing)

services/wireguard/service.go
```diff
@@ -96,7 +123,7 @@ func (manager *Manager) Start(providerID identity.Identity) (dto_discovery.Servi
 	return proposal, sessionConfigProvider, nil
 }
 
-// Wait blocks until service is stopped
+// Wait blocks until service is stopped.
 func (manager *Manager) Wait() error {
 	if !manager.isStarted {
```
tadovas: do we really need these "isStarted" checks? We assume services have a clear lifecycle already defined - create, start (stop in paralle) wait for return

services/wireguard/service.go
```diff
@@ -27,35 +28,46 @@ import (
...
+	process            sync.WaitGroup
+	locationResolver   location.Resolver
+	ipResolver         ip.Resolver
+	connectionEndpoint ConfigProvider
+	resources          resources.Handler
 }
 
-// Config represent a Wireguard service provider configuration that will be passed to the consumer for establishing a connection
+// Config represent a Wireguard service provider configuration that will be passed to the consumer for establishing a connection.
 type Config struct {
```
tadovas: I am still sticking to my previous suggestion -> ServiceConfiguration { 
    ProviderConfiguration,
    ConsumerConfiguration
}
if longer names better express itent and purpose - I go for them

services/wireguard/endpoint.go
tadovas: Ok as it seems it goes to wg config and not directly to listener - maybe a verification is needed if wg accept that :)

soffokl: Port number generation will be done better, it's just a frame of service, `TODO` here is to keep it in mind.

services/wireguard/service.go
soffokl: The problem here is that `ProviderConfig` is not just config generation, it also starts service provider to listen for connections. Same for `ConsumerConfig`, it adds the consumer to the allowed list and returns configuration for the consumer. 

The IP address is required to do this configuration. 

`Close` release interfaces and stops service. 

I'll change it to avoid doing two things in one place.

services/wireguard/service.go
soffokl: `isStarted` here just to support `Wait` method. Otherwise, we will need to periodically check if wireguard interface active and sleep until next check.

services/wireguard/service.go
tadovas: So maybe thats the problem - component does too much unrelated work, which can be split? If we have connection - let's handle connection details there, if we need some details for connection and those details need also to be shared with consumer - let's make configuration item and pass it both to consumer and to that wg connnection

services/wireguard/service.go
tadovas: It can be expressed easily with WaitGroup -> start wg.Add(1), stop -> cleanup and then wg.Done() , wait -> wg.Wait(). No ifs

services/wireguard/endpoint.go
```diff
@@ -0,0 +1,131 @@
...
+
+type connectionEndpoint struct {
+	iface    string
+	endpoint net.UDPAddr
+	wgClient *wireguardctrl.Client
+}
+
+// newConnectionEndpoint creates Wireguard client with predefined interface name and endpoint.
+func newConnectionEndpoint(iface string, endpoint net.UDPAddr) (*connectionEndpoint, error) {
+	wgClient, err := wireguardctrl.New()
```
vkuznecovas: don't we want to pass this as an interface param to the newConnectionEndpoint? We could actually test this then

services/wireguard/endpoint.go
```diff
@@ -0,0 +1,131 @@
...
+// ConsumerConfig adds service consumer public key to the list of allowed peers.
+// It returns information required to configure a consumer instance to allow connections.
+func (ce *connectionEndpoint) ConsumerConfig(ip net.IPNet) (Consumer, error) {
+	peerKey, err := wgtypes.GeneratePrivateKey()
+	if err != nil {
+		return Consumer{}, err
+	}
+
+	err = ce.wgClient.ConfigureDevice(ce.iface, wgtypes.Config{
+		Peers: []wgtypes.PeerConfig{{PublicKey: peerKey.PublicKey(), AllowedIPs: []net.IPNet{ip}}}})
```
zolia: `{ip}}}})` looks like lisp :)

services/wireguard/endpoint.go
```diff
@@ -0,0 +1,131 @@
...
+		return err
+	}
+
+	return exec.Command("ip", "link", "del", "dev", ce.iface).Run()
+}
+
+func (ce *connectionEndpoint) up(net net.IPNet) error {
+	if d, err := ce.wgClient.Device(ce.iface); err != nil || d.Name != ce.iface {
+		if out, err := exec.Command("ip", "link", "add", "dev", ce.iface, "type", "wireguard").CombinedOutput(); err != nil {
+			fmt.Println(string(out), err)
```
zolia: We usually use logger with respective log level.

soffokl: wow, it was a part of debugging, have no idea how it left here. :octocat:

services/wireguard/service_test.go
```diff
@@ -43,78 +40,6 @@ var ipresolverStub = &fakeIPResolver{
...
-	manager.Start(providerID)
-
-	go func() {
-		manager.Wait()
-		assert.Fail(t, "Wait should be blocking")
-	}()
-
-	waitABit()
-}
-
```
zolia: just removed tests? Something can't be added?

soffokl: Reverted tests with stub connection endpoint.

services/wireguard/endpoint.go
soffokl: Removed.

services/wireguard/endpoint.go
soffokl: Changed it a bit, should be more readable now.

services/wireguard/endpoint.go
soffokl: `ConnectionEndpoint` is doing a lot of other low-level things like interface creation. We can move it to another abstraction layer, but in the end, we will have another layer that is hard to test using unit tests. 
I think it should be covered in E2E tests. We have a task for it and it will be done in another PR.

services/wireguard/service.go
soffokl: Changed it to serviceConfig. 

Looks like `Config` is more suitable for go then `Configuration`:
https://golang.org/search?q=Config
https://golang.org/search?q=Configuration

services/wireguard/service.go
soffokl: Left only WaitGroup here.

services/wireguard/service.go
soffokl: Changed it to the things with a single responsibility.

services/wireguard/service.go
soffokl: No more ConfigProvider here.

Gopkg.toml
soffokl: Fixed.

soffokl:
@tadovas @zolia @vkuznecovas 
If you are not totally tired of it, please take a look at it again. 👀

cmd/commands/service/serviceTypes.go
```diff
@@ -0,0 +1,25 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package service
+
+var (
+	serviceTypesAvailable = []string{"openvpn", "noop"}
```
zolia: WG is not available anymore?

soffokl: `// +build !linux`

services/wireguard/consumer.go
```diff
@@ -0,0 +1,50 @@
...
+
+var errNoIPAddress = errors.New("IP address required")
+
+type consumer struct {
+	ip         net.IPNet
+	peer       wgtypes.PeerConfig
+	privateKey wgtypes.Key
+}
+
+func (c consumer) Config() (serviceConfig, error) {
```
zolia: Comment missing

services/wireguard/consumer.go
```diff
@@ -0,0 +1,50 @@
...
+	if len(c.peer.AllowedIPs) == 0 {
+		return serviceConfig{}, errNoIPAddress
+	}
+
+	var config serviceConfig
+
+	config.Provider.IP = c.peer.AllowedIPs[0]
```
zolia: Not sure, but I think its incorrect. Its provider configuration as seen by Consumer? 
Why its called Provider.IP ? It should be AllowedIPs (multi) meaning which IPs will be routed towards WG peer (provider).

zolia: Also `c.peer.AllowedIPs[0]` probably would give here incorrect value. Here it almost always should be 0.0.0.0/0 in our case.

zolia: We could have more extensive explanation here. To remove ambiguity how these fields are / should be used.

soffokl: We are generating `serviceConfig` here, that will be sent to the consumer to establish a connection. 
This config has two blocks: Provider part and Consumer part.

`c.peer` is a consumer's peer - Provider. That is why it goes to `config.Provider` to be sent as service configuration.

soffokl: `c.peer.AllowedIPs[0]` is a Provider local IP address, it's not Consumer IP. It's not empty and needed to allow customer know how to connect to the Provider after tunnel established.

Getting only one addess since do not think that we need to support multiple IP addresses at that moment. One is enough for now.

zolia: For consumer to reach provider, provider's endpoint should be enough. This Provider.IP should not be local provider IP.

tadovas:
@soffokl no we are not tired of reviewing :) actually its vise versa - cannot wait for this PR to land on master

services/wireguard/consumer.go
zolia: Lets keep terms the same as they are used by WG.
We have 3 parameters describing Provider at Consumer side: 
`[Peer]
Endpoint = 94.237.xx.xx:51820
PublicKey = changed here.. 
AllowedIPs = 0.0.0.0/0, ::/0
`

zolia: please see here: 
https://wiki.archlinux.org/index.php/WireGuard 
(Specific use-case: setup a VPN server)

For better described example.

services/wireguard/endpoint.go
```diff
@@ -0,0 +1,143 @@
...
+			AllowedIPs: []net.IPNet{ip},
+		}},
+	}
+
+	if err := ce.wgClient.ConfigureDevice(ce.iface, config); err != nil {
+		return consumer{}, err
+	}
+
+	return consumer{
+		ip:         ip,
```
zolia: we could call this one interfaceAddress or address (as per WG)

zolia: After discussing with @tadovas we may call it either tunnelAddress or interfaceAddress

services/wireguard/resources/handler.go
```diff
@@ -0,0 +1,45 @@
...
+// Handler is mock wireguard resource handler.
+// It will manage lists of network interfaces names, IP addresses and port for endpoints.
+type Handler struct{}
+
+// AllocateInterface provides available name for the wireguard network interface.
+func (h *Handler) AllocateInterface() string { return "myst0" }
+
+// AllocateIP provides available IP address for the wireguard connection.
+func (h *Handler) AllocateIP() net.IPNet {
+	ip, ipnet, _ := net.ParseCIDR("10.182.47.1/24")
```
zolia: Since you will allocate the same address for Provider and for Consumer, network stack will complain about duplicate address and will refuse to work :) We could allocate a subnet, and use .1 for provider. 
Since we will have to allocate separate subnet for each new wg peer (1 device per peer), we can always set .1 for provider and .2 for consumer, just different subnet. This way we will not have to deal with isolation and IP address tracking within subnet, but will need to track used subnets. We could allocate /30 subnets, but this will complicate counting. Thus lets stick with allocation if 10.182.x.1 (2) addresses for now.

services/wireguard/endpoint.go
soffokl: tunnelAddress sounds strange, tunnel is point to point and have 2 addresses commonly. interfaceAddress sounds like interface hardware address. Isn’t consume.ip better choice?

zolia: ip is too generic. Lets use WG terms. WG does not have such notion. It is always good to relate terms with ones used in documentation (manual). Otherwise it takes longer to grasp its usage.

soffokl: > ip is too generic. 
btw, it not just ip, it's always `consumer.ip`, what can it mean except IP address assigned to consumer?

services/wireguard/resources/handler.go
soffokl: Of course, it will not be the same IP addresses for consumer and provider. 
This PR is about a frame for wireguard service. There is no consumer. 

It's enough to have this mocked IP address for this PR. Let's add IP-addresses handling in the next PR, when it will be needed. 

The idea here is to have a pool of IP addresses which can be assigned to the peers(provider or consumer). 

We were planning to start with a single network interface for multiple session, so provider here getting 10.182.47.1, and other consumers will get other available IP-addresses.

services/wireguard/consumer.go
soffokl: This comment is not required since the object is not exported.
But added comment anyway.

services/wireguard/resources/handler.go
zolia: This is where you are wrong. We were not going to have multiple peers per single interface. Whole idea is to have 1 interface per peer. We should declare this as soon as possible so not to have further misconceptions.

soffokl: Ok, then I got you wrong.

services/wireguard/endpoint.go
```diff
@@ -0,0 +1,143 @@
...
+	if err := ce.up(); err != nil {
+		return err
+	}
+
+	key, err := wgtypes.GeneratePrivateKey()
+	if err != nil {
+		return err
+	}
+
+	if err := ce.wgClient.ConfigureDevice(ce.iface, wgtypes.Config{
```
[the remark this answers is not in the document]
soffokl: Fixed.

services/wireguard/resources/handler.go
```diff
@@ -0,0 +1,45 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package resources
+
+import "net"
+
+// Handler is mock wireguard resource handler.
+// It will manage lists of network interfaces names, IP addresses and port for endpoints.
+type Handler struct{}
```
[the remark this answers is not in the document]
soffokl: Changed it to Allocator.

services/wireguard/endpoint.go
```diff
@@ -0,0 +1,143 @@
...
+	"github.com/mysteriumnetwork/node/core/ip"
+	"github.com/mysteriumnetwork/node/services/wireguard/resources"
+)
+
+type connectionEndpoint struct {
+	iface      string
+	ip         net.IPNet
+	endpoint   net.UDPAddr
+	wgClient   *wireguardctrl.Client
+	resources  resources.Handler
```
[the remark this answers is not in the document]
soffokl: Changed here too to resourceAllocator.

services/wireguard/resources/handler.go
zolia: Some motivation for this is described here: https://github.com/mysteriumnetwork/node/wiki/Wireguard:-multiple-vs-single-peer-per-wg-interface

services/wireguard/consumer.go
soffokl: Removed `IP` and now sending only subnets called AllowedIPs.

services/wireguard/resources/handler.go
soffokl: Removed `IP` and now sending only subnets called AllowedIPs.

services/wireguard/resources/allocator.go
```diff
@@ -0,0 +1,45 @@
...
+package resources
+
+import "net"
+
+// Allocator is mock wireguard resource handler.
+// It will manage lists of network interfaces names, IP addresses and port for endpoints.
+type Allocator struct{}
+
+// AllocateInterface provides available name for the wireguard network interface.
+func (h *Allocator) AllocateInterface() string { return "myst0" }
```
vkuznecovas: not to be nitpicky but can we refactor this  (h *Allocator) to an (a *Allocator)? :)

soffokl: we definitely need to do this, thank.

soffokl: Fixed and rebased to latest master.

services/wireguard/endpoint.go
```diff
@@ -0,0 +1,156 @@
...
+
+	return exec.Command("ip", "link", "set", "dev", ce.iface, "up").Run()
+}
+
+func providerIP(subnet net.IPNet) net.IP {
+	subnet.IP[len(subnet.IP)-1] = byte(1)
+	return subnet.IP
+}
+
+func consumerIP(subnet net.IPNet) net.IP {
```
zolia: this one for future use?

soffokl: yep, next PR will use it to get consumer address.

services/wireguard/consumer.go
```diff
@@ -0,0 +1,49 @@
...
+	if len(c.peer.AllowedIPs) == 0 {
+		return serviceConfig{}, errNoIPAddress
+	}
+
+	var config serviceConfig
+
+	config.Provider.Endpoint = *c.peer.Endpoint
+	config.Provider.PublicKey = c.peer.PublicKey
+	config.Consumer.PrivateKey = c.privateKey
+	config.AllowedIPs = c.allowedIPs
```
zolia: From what I see this one will be initialised with '10.182.47.0/24', but is should be '0.0.0.0/0, ::/0', otherwise consumer will reach only local provider address. Or am I wrong?

tadovas: I don't think we need to setup allowed ips at all - we are not planning to do any ip based filtering at the moment - if we leave allowedip untouched nothing will break, and we can test floating ip feature (mobile app switches between different networks - but connection stays intact)

soffokl: Fixed.

PR #564, opened 2018-11-22
redoc image with auto fetch of tequilapi docs

zolia:


bin/docs_docker/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM redocly/redoc AS redoc
+
+# Install packages
+RUN apk add --update --no-cache bash curl \
+    && rm -rf /var/cache/apk/*
```
soffokl: We can ignore `&& rm -rf /var/cache/apk/*` 
https://github.com/gliderlabs/docker-alpine/blob/master/docs/usage.md#disabling-cache

soffokl: btw, do we really need to install `bash` and `curl` here?

tadovas: I guess its just leftover from previous solution - constantly pull docs from inside docker

PR #565, opened 2018-11-22
Mobile node connections configuration is optional now

tadovas:
- Keep mobile node constructor with essential and mandatory params only
- Move openvpn connection registration/configuration to separate method called optionally
- Do the same with wireguard

Introducing new mobile specific service configuration should not affect existing clients (to some extent)

Bonus: @Waldz can now roam freely between wifi and 4G :)


mobile/mysterium/wireguard_connection_setup.go
```diff
@@ -0,0 +1,40 @@
...
+// package independencies we are not reusing the same interface although implementation is rougly the same
+type tunnSetupPlaceholder interface {
+	Establish() int
+}
+
+// WireguardTunnelSetup represents interface for setuping tunnel on caller side (i.e. android api)
+type WireguardTunnelSetup tunnSetupPlaceholder
+
+// OverrideWireguardConnection overrides default wireguard connection implementation to more mobile adapted one
+func (mobNode *MobileNode) OverrideWireguardConnection(wgTunnelSetup WireguardTunnelSetup) {
```
zolia: Why to call it Override? Its simple connection through our bridge.

mobile/mysterium/openvpn_connection_setup.go
```diff
@@ -65,3 +85,45 @@ func (adapter channelToCallbacksAdapter) OnStats(openvpnStats openvpn3.Statistic
 		BytesReceived: uint64(openvpnStats.BytesIn),
 	})
 }
+
+// Openvpn3TunnelSetup is alias for openvpn3 tunnel setup interface exposed to Android/iOS interop
+type Openvpn3TunnelSetup openvpn3.TunnelSetup
+
+// OverrideOpenvpnConnection replaces default openvpn connection factory with mobile related one
+func (mobNode *MobileNode) OverrideOpenvpnConnection(tunnelSetup Openvpn3TunnelSetup) {
```
zolia: Maybe we can call it `MobileOpenvpnConnection`. From user standpoint he is not overriding anything..

mobile/mysterium/wireguard_connection_setup.go
tadovas: Because it exactly what it does - overrides default "openvpn" connection configured in node

mobile/mysterium/openvpn_connection_setup.go
tadovas: See my comment above

PR #566, opened 2018-11-23
Added wireguard connection establishing

### LINKED ISSUE #557: Implement connection.Manager for WG

soffokl:
Closes #557


services/wireguard/connection.go
```diff
@@ -19,27 +19,47 @@ package wireguard
...
 	"github.com/mysteriumnetwork/node/core/connection"
+	"github.com/mysteriumnetwork/node/services/wireguard/network"
 )
 
 // Connection which does no real tunneling
 type Connection struct {
-	isRunning      bool
-	noopConnection sync.WaitGroup
-	stateChannel   connection.StateChannel
+	isRunning    bool
```
tadovas: Do we really need those "isRunning" Our connection is a throw away object, and to actually make it really safe we need more than that.

services/wireguard/network/network_darwin.go
```diff
@@ -28,12 +28,17 @@ func (n *network) Provider() (Provider, error) {
...
 
-// Consumer adds service consumer public key to the list of allowed peers.
+// AddConsumer adds service consumer public key to the list of allowed peers.
 // It returns information required to configure a consumer instance to allow connections.
-func (n *network) Consumer() (Consumer, error) {
+func (n *network) AddConsumer() (Consumer, error) {
 	return Consumer{}, errNotSupported
 }
 
+// Consumer establish network connection between provider and consumer.
```
tadovas: I really don't get those two functions Consumer() and AddConsumer. Either naming is bad, or they are not doing what they need to do.

tadovas: As I understand Consumer and Provider represent configuration passed to service consumer? Than those structs should be named accordingly or maybe gathered to separate package

services/wireguard/service.go
```diff
@@ -63,11 +63,11 @@ type Config struct {
 	Consumer network.Consumer
 }
 
-// Network represents Wireguard network instance, it provide information
-// required for establishing connection between service provider and consumer.
-type Network interface {
+// Provider represents Wireguard network instance that serves consumers.
+// It returns information required for establishing connection between service provider and consumer.
+type Provider interface {
```
tadovas: Provider interface which has method Provider which returns network.Provider :(

tadovas:
Well I think this needs to be more polished before merging - to avoid fixing issues later

services/wireguard/connection.go
soffokl: Removed.

services/wireguard/network/network_darwin.go
soffokl: It was completely refactored.

services/wireguard/service.go
soffokl: It was completely refactored.

services/wireguard/connection.go
```diff
@@ -19,54 +19,58 @@ package wireguard
 
 import (
 	"sync"
-	"time"
 
+	log "github.com/cihub/seelog"
 	"github.com/mysteriumnetwork/node/core/connection"
 )
 
 // Connection which does no real tunneling
```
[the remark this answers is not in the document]
soffokl: Fixed comments.

tadovas:
Ok I am completely lost now. There is another PR #592 which is being merged to this branch, and has some comments (especially wg subpackaging related) which need to be addressed. I also expect this PR to be changed accordingly to our agreement how we decided to move on with those wg depedencies (wireguardctrl). Requesting changes on this one for now.

soffokl:
Reworked whole wireguard package.
@tadovas @zolia @vkuznecovas new review would be nice here.

services/wireguard/endpoint/create_linux.go
```diff
@@ -0,0 +1,65 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package endpoint
+
+import (
+	"encoding/base64"
+
+	"github.com/mdlayher/wireguardctrl/wgtypes"
```
tadovas: I still would like to have this isolated in kernelspace

services/wireguard/endpoint/endpoint.go
```diff
@@ -135,15 +124,13 @@ func (ce *connectionEndpoint) Stop() error {
 }
 
 func (ce *connectionEndpoint) up() error {
-	if d, err := ce.wgClient.Device(ce.iface); err != nil || d.Name != ce.iface {
+	if d, err := ce.wgClient.Device(ce.iface); err != nil || *d.Name() != ce.iface {
```
tadovas: These are kernelspace specific - must be moved

services/wireguard/endpoint/kernelspace/client.go
```diff
@@ -0,0 +1,128 @@
...
+	{IP: net.IPv4zero, Mask: net.CIDRMask(0, 32)},
+	{IP: net.IPv6zero, Mask: net.CIDRMask(0, 128)},
+}
+
+type client struct {
+	wgClient *wireguardctrl.Client
+}
+
+// NewWireguardClient creates new wireguard kernel space client.
+func NewWireguardClient() (*client, error) {
```
tadovas: Exported function returns unexported structure - something is wrong.

services/wireguard/endpoint/kernelspace/client.go
```diff
@@ -0,0 +1,128 @@
...
+	wgClient, err := wireguardctrl.New()
+	if err != nil {
+		return nil, err
+	}
+	return &client{wgClient}, nil
+}
+
+func (c *client) ConfigureDevice(name string, config wg.Device) error {
```
tadovas: config of type wg.Device - naming not clear

services/wireguard/endpoint/kernelspace/client.go
```diff
@@ -0,0 +1,128 @@
...
+		name:       name,
+		privateKey: d.PrivateKey,
+	}, nil
+}
+
+func (c *client) Close() error {
+	return c.wgClient.Close()
+}
+
+type device struct {
```
tadovas: The purpose of this device struct is unclear to me - it looks like simple public/private key wrapper

services/wireguard/endpoint/kernelspace/client.go
```diff
@@ -0,0 +1,128 @@
...
+	key := d.privateKey.String()
+	return &key
+}
+
+func (d device) PublicKey() *string {
+	key := d.privateKey.PublicKey().String()
+	return &key
+}
+
+func (d device) ListenPort() *int           { return nil }
```
tadovas: Implementations of those 3 methods indicates that interface this "device" structure tries to satisfy is too broad and can be split.

services/wireguard/wireguard.go
```diff
@@ -0,0 +1,136 @@
...
+		},
+		consumer{
+			s.Consumer.PrivateKey,
+		},
+		s.Subnet.String(),
+	})
+}
+
+// UnmarshalJSON implements json.Unmarshaler interface to receive human readable configuration.
+func (s *ServiceConfig) UnmarshalJSON(data []byte) error {
```
tadovas: I am not sure we still this complicated unmarshaller - maybe just that config struct bellow could be made exported and simply used?

services/wireguard/endpoint/kernelspace/client.go
soffokl: Removed it.

services/wireguard/wireguard.go
soffokl: The only purpose of this Unmarshaler is to return a human-readable config.
```JSON
{
	"provider": {
		"public_key": "hw253QllKd/qRBdLfc5hfDgyEO6woz6XZf2++N2TNzM=",
		"endpoint": "18.206.227.146:52820"
	},
	"consumer": {
		"private_key": "iMjn+gBtvWWjFaKtTdVizEX0hVBv4YCKKg9IHKKSIV8="
	},
	"subnet": "10.182.47.1/24"
}
```
instead of 
```JSON
{
	"Provider": {
		"PublicKey": "/p6EsyHjMC8kofoXbdIWdtdc/xaVqtsIP5sLPNyklAs=",
		"Endpoint": {
			"IP": "18.206.227.146",
			"Port": 52820,
			"Zone": ""
		}
	},
	"Consumer": {
		"PrivateKey": "IO3+92cVNy0OZqnfyDik3cUtSu3WWaNfa2RVqGgaLXY="
	},
	"Subnet": {
		"IP": "10.182.47.1",
		"Mask": "////AA=="
	}
}
```

soffokl: And I'd prefer to keep `net.UDPAddr` and `net.IPNet` in the struct instead of `string` to keep convetion only in marshaler/unmashaler, but not in every place where I need it.

services/wireguard/endpoint/kernelspace/client.go
soffokl: This unexported struct implement interface and will not be used outside directly.
Replacing it to return the interface instead makes circular dependency.
Not sure if it really worth to change it.

soffokl:
I did a cleanup of the packages. New review required.

services/wireguard/endpoint/kernelspace/client.go
tadovas: I am just saying that exported functions suppose to return exported structures. For example if someone will want to include your structure (wrap it) into another just because of the different interface - it won't be possible. If your function returns structure - that means structure is part of public api and should be exported - yes it can have all the fields private.

services/wireguard/wireguard.go
tadovas: Sounds reasonable. Let's see how it works

PR #576, opened 2018-11-27
simplified the statistics channel adapter for mobile openvpn

### LINKED ISSUE #573: Adjust mobile entrypoint to handle statistics channel
#567 introduced a nasty construct in the [mobile entrypoint](https://github.com/mysteriumnetwork/node/blob/master/mobile/mysterium/openvpn_connection_setup.go#L53). Need to find a proper way to implement it.

vkuznecovas:
CLOSES #573

Got rid of the unneeded struct and pass data to the statistics channel straight away.


tadovas:
We need to add refactoring the same daily task as test writting. Boy scout rule - leave a place cleaner than you found it

tadovas:
Rebase please :)

PR #577, opened 2018-11-27
format discovery factory

vkuznecovas:
formatted the file so that CI would now build.


vkuznecovas:
Ok, already covered in #575, I'll close.

soffokl:
we already have it here: https://github.com/mysteriumnetwork/node/pull/575

PR #584, opened 2018-11-28
Fix session endpoint

### LINKED ISSUE #583: Wrong formatting of session.dateStarted on /sessions endpoint
Session endpoint returns a malformed time format:

```json
 {
            "sessionId": "988072fc-c4cd-4894-b1ea-f9ca0ab8cdb7",
            "providerId": "0x125103195346207995e295a05fe08b4a0f7baa9e",
            "serviceType": "openvpn",
            "providerCountry": "DE",
            "dateStarted": "28118-110-289 116:2828:10",
            "bytesSent": 2320087,
            "bytesReceived": 73911524,
            "duration": 239,
            "status": "Completed"
}
```

vkuznecovas:
Closes #583

Found this while refactoring the session storage.

### DIFF
--- a/tequilapi/endpoints/sessions.go
+++ b/tequilapi/endpoints/sessions.go
@@ -19,6 +19,7 @@ package endpoints
 
 import (
 	"net/http"
+	"time"
 
 	"github.com/julienschmidt/httprouter"
 	"github.com/mysteriumnetwork/node/consumer/session"
@@ -111,7 +112,7 @@ func sessionToDto(se session.Session) SessionDTO {
 		ProviderID:      string(se.ProviderID.Address),
 		ServiceType:     se.ServiceType,
 		ProviderCountry: se.ProviderCountry,
-		DateStarted:     se.Started.Format("2006-01-02 15:04:05"),
+		DateStarted:     se.Started.Format(time.RFC3339),
 		BytesSent:       se.DataStats.BytesSent,
 		BytesReceived:   se.DataStats.BytesReceived,
 		Duration:        se.GetDuration(),
--- a/tequilapi/endpoints/sessions.go
+++ b/tequilapi/endpoints/sessions.go
@@ -111,7 +111,7 @@ func sessionToDto(se session.Session) SessionDTO {
 		ProviderID:      string(se.ProviderID.Address),
 		ServiceType:     se.ServiceType,
 		ProviderCountry: se.ProviderCountry,
-		DateStarted:     se.Started.Format("2018-10-29 16:22:05"),
+		DateStarted:     se.Started.Format("2006-01-02 15:04:05"),
 		BytesSent:       se.DataStats.BytesSent,
 		BytesReceived:   se.DataStats.BytesReceived,
 		Duration:        se.GetDuration(),
--- a/tequilapi/endpoints/sessions_test.go
+++ b/tequilapi/endpoints/sessions_test.go
@@ -18,13 +18,17 @@
 package endpoints
 
 import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"net/http"
+	"net/http/httptest"
 	"testing"
 	"time"
 
 	"github.com/stretchr/testify/assert"
 
 	"github.com/mysteriumnetwork/node/consumer"
-
 	"github.com/mysteriumnetwork/node/consumer/session"
 	"github.com/mysteriumnetwork/node/identity"
 	node_session "github.com/mysteriumnetwork/node/session"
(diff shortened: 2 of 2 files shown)


tequilapi/endpoints/sessions_test.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+
+	resp := httptest.NewRecorder()
+	handlerFunc := NewSessionsEndpoint(ssm).List
+	handlerFunc(resp, req, nil)
+
+	parsedResponse := &SessionsDTO{}
+	err = json.Unmarshal(resp.Body.Bytes(), parsedResponse)
+	assert.Nil(t, err)
+	assert.True(t, reflect.DeepEqual(parsedResponse.Sessions[0], sessionToDto(SessionMock)))
```
soffokl: Maybe we can do it without direct reflect call? 
```suggestion
	assert.ObjectsAreEqualValues(parsedResponse.Sessions[0], sessionToDto(SessionMock))
```
It doing the same but under the hood.

tequilapi/endpoints/sessions_test.go
```diff
@@ -0,0 +1,136 @@
...
+	"net/http"
+	"net/http/httptest"
+	"reflect"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/mysteriumnetwork/node/consumer"
+
```
soffokl: Extra empty line here. 🌚

tequilapi/endpoints/sessions.go
```diff
@@ -111,7 +111,7 @@ func sessionToDto(se session.Session) SessionDTO {
 		ProviderID:      string(se.ProviderID.Address),
 		ServiceType:     se.ServiceType,
 		ProviderCountry: se.ProviderCountry,
-		DateStarted:     se.Started.Format("2018-10-29 16:22:05"),
+		DateStarted:     se.Started.Format("2006-01-02 15:04:05"),
```
tadovas: It looked okey from the first sight. And then I started thinking (maybe this is the real problem 😄 )
What timezone this format uses? What timezone we use when we set actual session start time?
Maybe its better to use one of predefined ISO formats with UTC time zone? In that case session times will be comparable in whole world.
These are not change requests but more like answers we need to answer and to do that maybe we have to look deeper. This looks like good opportunity to do that

vkuznecovas: I've updated the format to RFC3339 which is basicly a stricter ISO 8601.

tequilapi/endpoints/sessions.go
```diff
@@ -111,7 +112,7 @@ func sessionToDto(se session.Session) SessionDTO {
 		ProviderID:      string(se.ProviderID.Address),
 		ServiceType:     se.ServiceType,
 		ProviderCountry: se.ProviderCountry,
-		DateStarted:     se.Started.Format("2018-10-29 16:22:05"),
+		DateStarted:     se.Started.UTC().Format(time.RFC3339),
```
tadovas: I think it would be better to set time in UTC before persisting session itself, not when rendering in presentation layer. I assume RFC3339 formatting includes timezone too?

vkuznecovas: It does - any reason why you don't want the UTC time there?

tadovas: Start time in UTC must be set on session creation - my point is, let's keep times in UTC in database too.

tequilapi/endpoints/sessions.go
```diff
@@ -111,7 +112,7 @@ func sessionToDto(se session.Session) SessionDTO {
 		ProviderID:      string(se.ProviderID.Address),
 		ServiceType:     se.ServiceType,
 		ProviderCountry: se.ProviderCountry,
-		DateStarted:     se.Started.Format("2018-10-29 16:22:05"),
+		DateStarted:     se.Started.Format(time.RFC3339),
```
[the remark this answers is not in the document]
vkuznecovas: You'll be able to parse the RFC3339 in any language easily. It does provide timezone information. One thing that unix timestamp is lacking is the human readability. RFC3339 is both human comprehendible and machine readable.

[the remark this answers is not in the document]
zolia: I would say its a good time to introduce UTC unix timestamp, since we are changing format anyways. Will reduce complexity for users later on.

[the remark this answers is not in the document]
tadovas: @zolia golden rule with time - never ever do direct arithmetics with times. https://en.wikipedia.org/wiki/Leap_second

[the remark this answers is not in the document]
vkuznecovas: I think the RFC3339 is still the way to go. You said it youself - peer will have to parse the timestamp to make sense of it. Why have that step as extra? If you're parsing anyway, you'll be able to parse RFC3339 with the same ease as a timestamp.

IETF seems to recommend RFC3339 as well:
https://tools.ietf.org/html/rfc7493#section-4.3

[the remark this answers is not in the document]
zolia: Ok, theres cons and pros for that. RFC3339 is representative format as common JSON is too. I see a bunch of pitfalls in such usage, but since event IETF recommends that.. well :)

PR #585, opened 2018-11-29
Fixed android build with wireguard

soffokl:


cmd/connections_register_android.go
```diff
@@ -0,0 +1,29 @@
...
+ */
+
+package cmd
+
+import (
+	"github.com/mysteriumnetwork/node/core/node"
+)
+
+func (di *Dependencies) registerConnections(nodeOptions node.Options) {
+	di.registerOpenvpnConnection(nodeOptions)
```
tadovas: This can be removed too. We redefine openvpn connection for android anyway

soffokl: And only noop will be left here. :D
Should I remove both?

tadovas: No. Noop should work on all oses (and it kind of works :D)

soffokl: removed openvpn.

tadovas:
@zolia @vkuznecovas - we need this to fix master build

PR #590, opened 2018-11-30
use openvpn3 reconnect

zolia:


mobile/mysterium/openvpn_connection_setup.go
```diff
@@ -85,8 +87,42 @@ func (adapter channelToCallbacksAdapter) OnStats(openvpnStats openvpn3.Statistic
...
+		return errors.New("session not created yet")
+	}
+
+	return st.session.Reconnect(afterSeconds)
+}
+
+func (st *sessionTracker) handleState(stateEvent connection.StateEvent) {
+	// On disconnected - remove session
+	if stateEvent.State == connection.Disconnecting {
+		st.session.Stop()
```
tadovas: In case of disconnecting session ref should be nil'ed. Not stopped.

mobile/mysterium/openvpn_connection_setup.go
```diff
@@ -85,8 +87,42 @@ func (adapter channelToCallbacksAdapter) OnStats(openvpnStats openvpn3.Statistic
...
 
-// OverrideOpenvpnConnection replaces default openvpn connection factory with mobile related one
-func (mobNode *MobileNode) OverrideOpenvpnConnection(tunnelSetup Openvpn3TunnelSetup) {
+// ReconnectableSession interface exposing reconnect for an active session
+type ReconnectableSession interface {
+	Reconnect(afterSeconds int) error
+}
+
+type sessionTracker struct {
+	session *openvpn3.Session
```
tadovas: This reference is being set/read from possibly different threads. Some kind of sync mechanism is needed, like Mutex

mobile/mysterium/openvpn_connection_setup.go
tadovas:
```suggestion
		st.session=nil
```

mobile/mysterium/openvpn_connection_setup.go
```diff
@@ -85,8 +87,45 @@ func (adapter channelToCallbacksAdapter) OnStats(openvpnStats openvpn3.Statistic
...
+}
+
+func (st *sessionTracker) sessionCreated(s *openvpn3.Session) {
+	st.mux.Lock()
+	st.session = s
+	st.mux.Unlock()
+}
+
+// Reconnect reconnects active session after given time
+func (st *sessionTracker) Reconnect(afterSeconds int) error {
```
soffokl: Maybe we can use here golang native `time.Duration` to avoid future conversion?

mobile/mysterium/openvpn_connection_setup.go
```diff
@@ -85,8 +87,45 @@ func (adapter channelToCallbacksAdapter) OnStats(openvpnStats openvpn3.Statistic
...
+	if st.session == nil {
```
soffokl: Should we protect the reading of `st.session` if we are doing it in other places?

mobile/mysterium/openvpn_connection_setup.go
```diff
@@ -85,8 +87,45 @@ func (adapter channelToCallbacksAdapter) OnStats(openvpnStats openvpn3.Statistic
...
+		return errors.New("session not created yet")
+	}
+
+	return st.session.Reconnect(afterSeconds)
```
soffokl: If this can be run in a multithread way, `st.session` can become `nil` after check above. 
I'd protect the whole body of the function with a mutex.

mobile/mysterium/openvpn_connection_setup.go
```diff
@@ -85,8 +87,45 @@ func (adapter channelToCallbacksAdapter) OnStats(openvpnStats openvpn3.Statistic
...
+}
+
+func (st *sessionTracker) handleState(stateEvent connection.StateEvent) {
+	// On disconnected - remove session
+	if stateEvent.State == connection.Disconnecting {
+		st.session = nil
```
soffokl: Should we protect `st.session` if we are doing it in other places?

mobile/mysterium/openvpn_connection_setup.go
```diff
@@ -85,8 +87,45 @@ func (adapter channelToCallbacksAdapter) OnStats(openvpnStats openvpn3.Statistic
...
+	}
+}
+
+// OverrideOpenvpnConnection replaces default openvpn connection factory with mobile related one returning session that can be reconnected
+func (mobNode *MobileNode) OverrideOpenvpnConnection(tunnelSetup Openvpn3TunnelSetup) ReconnectableSession {
+
```
soffokl: Nitpick: empty line at the beginning of the function.

mobile/mysterium/openvpn_connection_setup.go
zolia: fixed

mobile/mysterium/openvpn_connection_setup.go
zolia: C.reconnect accepts int, so no conversion is needed. From mobile frameworks, more straight type is int too.

mobile/mysterium/openvpn_connection_setup.go
zolia: wrapped `st.session` modification between mutex.

mobile/mysterium/openvpn_connection_setup.go
```diff
@@ -85,8 +87,46 @@ func (adapter channelToCallbacksAdapter) OnStats(openvpnStats openvpn3.Statistic
...
+func (st *sessionTracker) sessionCreated(s *openvpn3.Session) {
+	st.mux.Lock()
+	st.session = s
+	st.mux.Unlock()
+}
+
+// Reconnect reconnects active session after given time
+func (st *sessionTracker) Reconnect(afterSeconds int) error {
+	if st.session == nil {
```
vkuznecovas: does this not need to be locked?

mobile/mysterium/openvpn_connection_setup.go
```diff
@@ -85,8 +87,46 @@ func (adapter channelToCallbacksAdapter) OnStats(openvpnStats openvpn3.Statistic
...
+		return errors.New("session not created yet")
+	}
+
+	return st.session.Reconnect(afterSeconds)
+}
+
+func (st *sessionTracker) handleState(stateEvent connection.StateEvent) {
+	// On disconnected - remove session
+	if stateEvent.State == connection.Disconnecting {
+		st.session = nil
```
vkuznecovas: does this not need to be locked?

mobile/mysterium/openvpn_connection_setup.go
```diff
@@ -105,17 +105,21 @@ func (st *sessionTracker) sessionCreated(s *openvpn3.Session) {
 
 // Reconnect reconnects active session after given time
 func (st *sessionTracker) Reconnect(afterSeconds int) error {
+	st.mux.Lock()
 	if st.session == nil {
 		return errors.New("session not created yet")
```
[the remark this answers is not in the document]
zolia: good point. Fixed.

PR #591, opened 2018-11-30
Feature/wg mobile side

tadovas:
- Add skeleton for wg tunnel setup for mobile side
- Enable userspace wg usage in oses other than linux (darwin, win etc.)

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -93,3 +93,8 @@
 [[constraint]]
   revision = "15a089393cc3b891b2181cd738347dc568a0b4b0"
   name = "github.com/mdlayher/wireguardctrl"
+
+[[constraint]]
+  name = "git.zx2c4.com/wireguard-go"
+  revision = "ffd4865b321335a7b3dca2dc328120d178bae687"
+  source = "github.com/mysteriumnetwork/wireguard-go"
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -96,5 +96,5 @@
 
 [[constraint]]
   name = "git.zx2c4.com/wireguard-go"
-  revision = "29762b79e3c42d3a91f5f119360fd1c84e59b541"
+  revision = "5acb51d173e2938591471974e53d5262f8adfa51"
   source = "github.com/mysteriumnetwork/wireguard-go"
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -96,5 +96,5 @@
 
 [[constraint]]
   name = "git.zx2c4.com/wireguard-go"
-  revision = "f901ba1171a9ed762f252d69ad323ea0c4df1165"
+  revision = "6e19ef17479b8a46e554d7d3e8d66973549facdf"
   source = "github.com/mysteriumnetwork/wireguard-go"
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -96,5 +96,5 @@
 
 [[constraint]]
   name = "git.zx2c4.com/wireguard-go"
-  revision = "ffd4865b321335a7b3dca2dc328120d178bae687"
+  revision = "f901ba1171a9ed762f252d69ad323ea0c4df1165"
   source = "github.com/mysteriumnetwork/wireguard-go"
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -96,5 +96,5 @@
 
 [[constraint]]
   name = "git.zx2c4.com/wireguard-go"
-  revision = "6e19ef17479b8a46e554d7d3e8d66973549facdf"
+  revision = "29762b79e3c42d3a91f5f119360fd1c84e59b541"
   source = "github.com/mysteriumnetwork/wireguard-go"
--- a/mobile/mysterium/wireguard_connection_setup.go
+++ b/mobile/mysterium/wireguard_connection_setup.go
@@ -19,14 +19,30 @@ package mysterium
 
 import (
 	"errors"
+	"os"
+	"sync"
+	"time"
 
+	"git.zx2c4.com/wireguard-go/device"
+	"git.zx2c4.com/wireguard-go/tun"
+	"github.com/cihub/seelog"
+	"github.com/mysteriumnetwork/node/consumer"
 	"github.com/mysteriumnetwork/node/core/connection"
 )
 
 // TODO this probably can be aligned with openvpn3 setup interface as its very close to android api, but for sake of
 // package independencies we are not reusing the same interface although implementation is rougly the same
 type tunnSetupPlaceholder interface {
-	Establish() int
+	NewTunnel()
+	AddTunnelAddress(ip string)
+	AddRoute(route string)
+	AddDNS(ip string)
+	SetBlocking(blocking bool)
+	Establish() (int, error)
+	SetMTU(mtu int)
+	Protect(socket int) error
+	SetSessionName(session string)
+	Close()
 }
 
 // WireguardTunnelSetup represents interface for setuping tunnel on caller side (i.e. android api)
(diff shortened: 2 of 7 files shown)


cmd/commands/service/serviceTypes_linux.go
```diff
@@ -23,7 +23,7 @@ import (
 	"github.com/mysteriumnetwork/node/core/service"
 	service_noop "github.com/mysteriumnetwork/node/services/noop"
 	service_openvpn "github.com/mysteriumnetwork/node/services/openvpn"
-	service_wireguard "github.com/mysteriumnetwork/node/services/wireguard"
+	"github.com/mysteriumnetwork/node/services/wireguard/wgdefinition"
```
soffokl: I see that it was required specifically for mobile, but was so nice to have the same appearance for all wireguard, noop and openvpn. Maybe we can do something similar for all packages in the future? 

And `wgdefinition` looks strange. At least it stutters `wireguard/wgdefinition`.

mobile/mysterium/wireguard_connection_setup.go
```diff
@@ -19,22 +19,158 @@ package mysterium
...
+		//TODO this heavy linfting might go to doInit
+		tun, err := newTunnDevice(wgTunnelSetup)
+		if err != nil {
+			return nil, err
+		}
+
+		devApi := device.UserspaceDeviceApi(tun)
+		socket, err := devApi.GetNetworkSocket()
+		if err != nil {
+			tun.Close()
```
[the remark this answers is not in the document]
tadovas: Actually it has to be devApi instead of tun - it closes tun automatically inside :) good catch.

cmd/commands/service/serviceTypes_linux.go
tadovas: Yea package importing is a bit cumbersome. Mayb wg definition could lay somewhere on the side. Or maybe we can have a separate package with all service definitions registered not - sure. The problem was that wireguard package is so fat with unrelated code that even to register service definition json parse (simple function no wg specifics) requires all wg package to be build with all dependencies (wgctrl!) which does not even compile to armv7 as we already know. I have no clear way to move here - any suggestions would be fine. What I really don't like is that consumer code needs to import whole package with provider code just to register unserializer.

mobile/mysterium/wireguard_connection_setup.go
```diff
@@ -19,22 +19,158 @@ package mysterium
 
 import (
 	"errors"
+	"os"
+	"sync"
+	"time"
 
+	"github.com/mysteriumnetwork/node/services/wireguard/wgdefinition"
```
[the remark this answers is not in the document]
tadovas: I have no idea why goimports behaves like that :(

mobile/mysterium/wireguard_connection_setup.go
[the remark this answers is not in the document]
tadovas: It supposed to be devApi instead of tun. Good catch

mobile/mysterium/wireguard_connection_setup.go
```diff
@@ -19,22 +19,158 @@ package mysterium
...
+			statisticsChannel: statisticsChannel,
+			closed:            &sync.WaitGroup{},
+		}, nil
 	})
 }
+
+func newTunnDevice(wgTunnSetup WireguardTunnelSetup) (tun.TUNDevice, error) {
+	wgTunnSetup.NewTunnel()
+	wgTunnSetup.AddTunnelAddress("10.182.47.5/24")
+	wgTunnSetup.SetMTU(1280)
```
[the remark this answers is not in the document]
tadovas: Yes this must be moved as constant and described as it was taken from wg android project as working param :)

mobile/mysterium/wireguard_connection_setup.go
```diff
@@ -19,22 +19,158 @@ package mysterium
...
+	err := wg.doInit()
+	if err != nil {
+		wg.stateChannel <- connection.NotConnected
+		return err
+	}
+	wg.stateChannel <- connection.Connected
+	return nil
+}
+
+func (wg *wireguardConnection) doInit() error {
```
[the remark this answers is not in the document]
tadovas: No, it will be removed. 👍

mobile/mysterium/wireguard_connection_setup.go
```diff
@@ -19,22 +19,158 @@ package mysterium
...
+		BytesSent:     peerStatistics.Sent,
+		BytesReceived: peerStatistics.Received,
+	}
+}
+
+func (wg *wireguardConnection) doCleanup() {
+	wg.device.Close()
+	wg.device.Wait()
+	wg.stateChannel <- connection.NotConnected
+	close(wg.stateChannel)
```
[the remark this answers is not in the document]
tadovas: Rule of thumb in channel communication -> writer closes channel. That's ok here as state channel is what connection manager inits and passes to lower level transport functionality. Closing channel means signaling reader (connection manager) we are done.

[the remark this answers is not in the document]
tadovas: Well, yes it unrelated change, but want this to be discussed, as I don't see any big problems with current approach. As long as reader doesn't close it its fine, and conn manager doesn't close channel is by convention. Go garbage collects unreferences channels anyway, you don't need to close them excplicitly.

Gopkg.toml
```diff
@@ -93,3 +93,8 @@
 [[constraint]]
   revision = "15a089393cc3b891b2181cd738347dc568a0b4b0"
   name = "github.com/mdlayher/wireguardctrl"
+
+[[constraint]]
+  name = "git.zx2c4.com/wireguard-go"
+  revision = "ffd4865b321335a7b3dca2dc328120d178bae687"
+  source = "github.com/mysteriumnetwork/wireguard-go"
```
vkuznecovas: missing newline

mobile/mysterium/wireguard_connection_setup.go
```diff
@@ -19,22 +19,158 @@ package mysterium
...
+
+		devApi := device.UserspaceDeviceApi(tun)
+		socket, err := devApi.GetNetworkSocket()
+		if err != nil {
+			tun.Close()
+			return nil, err
+		}
+		err = wgTunnelSetup.Protect(int(socket))
+		if err != nil {
+			tun.Close()
```
zolia: should here be devApi.close() too?

mobile/mysterium/wireguard_connection_setup.go
```diff
@@ -19,22 +19,158 @@ package mysterium
...
+	close(wg.stopChannel)
+}
+
+var _ connection.Connection = &wireguardConnection{}
+
+func (wg *wireguardConnection) updateStatistics() {
+	var err error
+	defer func() {
+		if err != nil {
+			seelog.Error("[wg connection] Error updating statistics: ", err)
```
zolia: we should have prefix "[wg connection]" instead.

cmd/commands/service/serviceTypes_linux.go
vkuznecovas: I think we should definitely move the json bootstrap part away from all the services. Maybe put it beside the rest of the package "service_discovery"(which on its own needs a reconsideration). That way we can register the relevant serializers from that package, instead of having each service define their own?

mobile/mysterium/wireguard_connection_setup.go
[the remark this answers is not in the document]
vkuznecovas: It allows for grouping of imports, so if you do click enter anywhere on the imports - it will assume this will be a separate block. It formats the blocks independently :)

mobile/mysterium/wireguard_connection_setup.go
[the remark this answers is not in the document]
vkuznecovas: I like the read only channel idea, but maybe a separate task for that?

cmd/commands/service/serviceTypes_linux.go
tadovas: refactored by @soffokl

mobile/mysterium/wireguard_connection_setup.go
```diff
@@ -19,22 +19,158 @@ package mysterium
...
+	if err != nil {
+		wg.stateChannel <- connection.NotConnected
+		return err
+	}
+	wg.stateChannel <- connection.Connected
+	return nil
+}
+
+func (wg *wireguardConnection) doInit() error {
+	wg.closed.Add(1)
```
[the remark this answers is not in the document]
tadovas: This is how WaitGroup works, this group expresses everything stopped condition. So yes add 1 is at init. Done is where shutdown completes. Renamed to stopCompleted if it makes more sense, but for me personally it's just cosmetics

mobile/mysterium/wireguard_connection_setup.go
```diff
@@ -122,8 +127,11 @@ func setupWireguardDevice(devApi *device.DeviceApi, config *wireguard.ServiceCon
 	}
 
 	err = devApi.AddPeer(device.ExternalPeer{
-		PublicKey:      device.NoisePublicKey(peerPubKeyArr),
-		RemoteEndpoint: parsed,
+		PublicKey:       device.NoisePublicKey(peerPubKeyArr),
+		RemoteEndpoint:  parsed,
+		KeepAlivePeriod: 60,
```
zolia: set this to 20sec.

mobile/mysterium/wireguard_connection_setup.go
```diff
@@ -122,8 +127,11 @@ func setupWireguardDevice(devApi *device.DeviceApi, config *wireguard.ServiceCon
 	}
 
 	err = devApi.AddPeer(device.ExternalPeer{
-		PublicKey:      device.NoisePublicKey(peerPubKeyArr),
-		RemoteEndpoint: parsed,
+		PublicKey:       device.NoisePublicKey(peerPubKeyArr),
+		RemoteEndpoint:  parsed,
```
zolia: parsed?  ParsedEndpoint maybe

mobile/mysterium/wireguard_connection_setup.go
```diff
@@ -122,8 +127,11 @@ func setupWireguardDevice(devApi *device.DeviceApi, config *wireguard.ServiceCon
...
+		KeepAlivePeriod: 60,
+		//all traffic through this peer (unfortunatelly 0.0.0.0/0 didn't work as it was treated as ipv6)
+		AllowedIPs: []string{"0.0.0.0/1", "128.0.0.0/1"},
```
zolia: These should be arbitrary modifiable. We should be getting those from a user somewhere.

mobile/mysterium/wireguard_connection_setup.go
tadovas: renamed to endpoint

mobile/mysterium/wireguard_connection_setup.go
tadovas: Done

mobile/mysterium/wireguard_connection_setup.go
tadovas: In the long perspective - maybe. At the moment - I don't think it's worth the effort. This "allow" to all route can be narrowed to our remote "server" only for better security maybe.

PR #592, opened 2018-12-03
 Human readable wireguard JSON config

soffokl:


services/wireguard/service.go
```diff
@@ -138,3 +140,77 @@ func (manager *Manager) Stop() error {
 	log.Info(logPrefix, "Wireguard service stopped")
 	return nil
 }
+
+// MarshalJSON implements json.Marshaler interface to provide human readable configuration.
+func (s ServiceConfig) MarshalJSON() ([]byte, error) {
```
tadovas: Idea is nice. But there is problem - ServiceConfig which might be shared between consumer and provider has datatypes and depends on wireguardctrl lib, which we decided to hide inside internals of wg (it's specific kernel mode anyway), so I would go with plain and simple strings and leave for consumer/provider to decide how to fill in, or use provided data. I even move it separate subpackage in wireguard - maybe its the best way to move related code to subpackages and reference them in the main package. This way we can reuse parts withouth importing full wireguard package if needed (referencing to bootstrap method in wgdefinition).

soffokl:
changes from this PR merged into https://github.com/mysteriumnetwork/node/pull/566

PR #593, opened 2018-12-03
Reworked the storage layer, added migrations to it.

vkuznecovas:
1) Reworked the storage layer. It now prefers to put things into separate buckets.
1) Refactored the session storage on consumer side.  It now uses strings instead of ints for state.
1) Added a migration layer on top of the storage. Migrations are now run once the database is opened. Could not find an implemented solution for boltDB so wrote a rather simple one myself.
1) God rid of the storage interface on the storage side, added appropriate interfaces on the consumer side instead.


core/storage/boltdb/migrator.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+	return false, nil
+}
+
+func (m *Migrator) saveMigrationRun(migration migrations.Migration) error {
+	return m.db.db.Save(&migration)
+}
+
+func (m *Migrator) migrate(migration migrations.Migration) error {
+	isRun, err := m.isMigrationRun(migration)
```
zolia: i would suggest `isRunning` name

core/storage/boltdb/migrator.go
```diff
@@ -0,0 +1,93 @@
...
+	if err != nil {
+		return err
+	}
+	if isRun {
+		return nil
+	}
+	log.Info(migrationLogPrefix, "running migration", migration.Name)
```
zolia: we probably need some separator here: ""running migration: "

core/storage/boltdb/migrator.go
tadovas: maybe its just english. maybe IsAlreadyApplied would express intent more clearly.

tequilapi/endpoints/sessions.go
```diff
@@ -67,7 +67,7 @@ type sessionsEndpoint struct {
 }
```
soffokl: Could you please also change line 42 
`// example: openpvn` has a typo.

tequilapi/endpoints/sessions.go
```diff
@@ -106,7 +106,7 @@ func AddRoutesForSession(router *httprouter.Router, sessionStorage sessionStorag
 	router.GET("/sessions", sessionsEndpoint.List)
 }
 
-func sessionToDto(se session.Session) SessionDTO {
+func sessionToDto(se session.History) SessionDTO {
 	return SessionDTO{
 		SessionID:       string(se.SessionID),
 		ProviderID:      string(se.ProviderID.Address),
```
soffokl: Unnecessary conversion to string here too.

core/storage/boltdb/migrator.go
```diff
@@ -0,0 +1,93 @@
...
+	return false, nil
+}
+
+func (m *Migrator) saveMigrationRun(migration migrations.Migration) error {
+	return m.db.db.Save(&migration)
+}
+
+func (m *Migrator) migrate(migration migrations.Migration) error {
+	isRun, err := m.isMigrationRun(migration)
+	if err != nil {
```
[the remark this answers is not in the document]
vkuznecovas: If the error is not nil, we want to return it. If the migration is actually run - we want to return nil. I don't think merging makes this any clearer(as i'll have to nest stuff then).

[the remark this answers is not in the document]
soffokl: If `err` not `nil` we will return it
if `err` is `nil` and migration run we will return `nil` since `err` is `nil`. 
If `err` is `nil` and we didn't run we just continue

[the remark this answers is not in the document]
vkuznecovas: Ah, true!

cmd/di.go
```diff
@@ -21,6 +21,8 @@ import (
 	"path/filepath"
 	"time"
 
+	"github.com/mysteriumnetwork/node/core/storage/boltdb/migrations/history"
+
```
soffokl: Extra empty line.

consumer/session/session.go
```diff
@@ -22,6 +22,8 @@ import (
 
 	"github.com/mysteriumnetwork/node/consumer"
 	"github.com/mysteriumnetwork/node/identity"
+	"github.com/mysteriumnetwork/node/service_discovery/dto"
+
```
soffokl: Extra empty line

core/storage/boltdb/migrator_test.go
```diff
@@ -0,0 +1,208 @@
...
+import (
+	"errors"
+	"testing"
+	"time"
+
+	"github.com/asdine/storm"
+	"github.com/stretchr/testify/assert"
+
+	"github.com/mysteriumnetwork/node/core/storage/boltdb/migrations"
+
```
soffokl: Extra empty line.

core/storage/boltdb/boltdbtest/db.go
```diff
@@ -0,0 +1,76 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+// Package boltdbtest contains the utitilies needed for boltdb testing
+package boltdbtest
```
soffokl: Do we really need a dedicated package for test utils?

vkuznecovas: I think it's a good practice. That way you end up reusing code instead of redefining it everywhere.

PR #596, opened 2018-12-06
Enable real forwarding for wireguard service

soffokl:


services/wireguard/endpoint/endpoint.go
```diff
@@ -99,6 +117,10 @@ func (ce *connectionEndpoint) Stop() error {
 		return err
 	}
 
+	if ce.natService != nil {
```
tadovas: can it be nil ? its depedency

soffokl: connectionEndpoint is used both for provider and consumer. 
There is no reason to do anything with NAT for the consumer.
So the answer is yes, it can be nil for the consumer.

services/wireguard/endpoint/endpoint.go
```diff
@@ -68,6 +63,29 @@ func (ce *connectionEndpoint) Start(config *wg.ServiceConfig) error {
 		ce.privateKey = config.Consumer.PrivateKey
 	}
 
+	if ce.ipResolver != nil {
```
tadovas: This shows too much details - could be separate private function with clear name

services/wireguard/endpoint/endpoint.go
tadovas: That means that connection endpoint is too low level component to handle forwarding. Behaviour based on ifs is not a very good practice

soffokl: Moved NAT configuration to the service manager.

soffokl:
Fixed tests and rebased it. Please take a look again.

tadovas:
@zolia @vkuznecovas lets decide

PR #598, opened 2018-12-06
Consumer sends his config to provider

vkuznecovas:
Session creation now includes two steps - create-session and session-acknowledge. Registering the ack handler for each connection allows for skipping of ack where it's not needed.

Any opinions are much appreciated.

The WG implementation is not tested yet, doing it now.


core/connection/manager.go
```diff
@@ -47,12 +48,7 @@ var (
 )
 
 // Creator creates new connection by given options and uses state channel to report state changes
-// Given options:
-//  - session,
-//  - consumer identity
-//  - service provider identity
-//  - service proposal
-type Creator func(ConnectOptions, StateChannel, StatisticsChannel) (Connection, error)
+type Creator func(string, StateChannel, StatisticsChannel) (Connection, error)
```
soffokl: Can we name arguments to explain what should be passed? Just `string` says nothing what should be there.

core/connection/manager.go
```diff
@@ -149,7 +145,25 @@ func (manager *connectionManager) startConnection(consumerID identity.Identity,
 	}
 	cancel = append(cancel, func() { dialog.Close() })
 
-	sessionID, sessionConfig, err := session.RequestSessionCreate(dialog, proposal.ID)
+	stateChannel := make(chan State, 10)
+	statisticsChannel := make(chan consumer.SessionStatistics, 10)
```
soffokl: Why do we have buffer 10 for both channels here?

services/wireguard/connection/connection.go
```diff
@@ -34,11 +35,19 @@ type Connection struct {
 	stateChannel connection.StateChannel
 
 	config             wg.ServiceConfig
+	consumerKey        wg.ConsumerPrivateKey
```
soffokl: Looks like `consumerKey` is unused.

mobile/mysterium/openvpn_connection_setup.go
```diff
@@ -29,22 +29,40 @@ import (
...
 )
 
+type openvpn3SessionFactory func(connection.ConnectOptions) (*openvpn3.Session, error)
+
+var errSessionWrapperNotStarted = errors.New("session wrapper not started")
+
 type sessionWrapper struct {
-	session *openvpn3.Session
+	session        *openvpn3.Session
+	sessionFactory openvpn3SessionFactory
```
tadovas: Type name is okeysh, but sessionFactory -> createSession. Later in code session, err := wrapper.createSession(options) which is more readable

mobile/mysterium/openvpn_connection_setup.go
```diff
@@ -29,22 +29,40 @@ import (
...
 
 func (wrapper *sessionWrapper) Wait() error {
-	return wrapper.session.Wait()
+	if wrapper.session != nil {
+		return wrapper.session.Wait()
+	}
+	return errSessionWrapperNotStarted
+}
+
+func (wrapper *sessionWrapper) GetSessionConfig() (connection.SessionCreationConfig, error) {
```
tadovas: GetConfig() ? Session is repeated too much in this context and its meaning is different. connection.SessionCreationConfig -> simply ConsumerConfig? It doesn't configure session itself but provides option to send underlying transport configuration

services/wireguard/wireguard.go
```diff
@@ -73,14 +73,24 @@ type PeerInfo interface {
 	PublicKey() string
 }
 
+// ConsumerPublicKey is used for sending the public key from consumer to provider
+type ConsumerPublicKey struct {
```
tadovas: Bad naming, wg configuration is not about public key only, but can have different fields. Suggesting WireguardConsumerConfig

session/create_producer.go
```diff
@@ -26,6 +26,7 @@ import (
 
 type createProducer struct {
 	ProposalID int
+	Config     json.RawMessage
```
tadovas: instead of depending heavily on json.RawMessage, a marker interface (with no methods) could be introduced and called ConsumerConfig

core/connection/interface.go
```diff
@@ -27,11 +27,18 @@ import (
 // DialogCreator creates new dialog between consumer and provider, using given contact information
 type DialogCreator func(consumerID, providerID identity.Identity, contact market.Contact) (communication.Dialog, error)
 
+// SessionCreationConfig are the parameters that get sent to the provider
+type SessionCreationConfig interface{}
```
tadovas: It's not about session creation config. It's about passing consumer config parameters to underlying transport - nothing to do with session itself

core/connection/manager.go
```diff
@@ -149,7 +145,25 @@ func (manager *connectionManager) startConnection(consumerID identity.Identity,
...
+	if err != nil {
+		return err
+	}
+
+	sessionCreateConfig, err := connection.GetSessionConfig()
+	if err != nil {
+		return err
+	}
+
+	sessionCreateConfigJSON, err := json.Marshal(sessionCreateConfig)
```
tadovas: I would leave session marshaling to underlying mechanism, and pass returned consumer config structure directly

core/connection/registry.go
```diff
@@ -17,29 +17,32 @@
 
 package connection
 
+// Factory represents a connection constructor
+type Factory func(stateChannel StateChannel, statisticsChannel StatisticsChannel) (Connection, error)
+
 // Registry holds of all plugable connections
 type Registry struct {
-	creators map[string]Creator
+	creators map[string]Factory
```
tadovas: What's wrong with Creator. func type called factory is a bit confusing. Maybe we can introduce interface with single method Create and call it Factory?

mobile/mysterium/openvpn_connection_setup.go
zolia: 👍 for `ConsumerConfig`

core/connection/registry.go
vkuznecovas: done

session/create_producer.go
tadovas: For serialization it's not a problem as long as custom structure implements MarshalJSON interface. For deserialization - looks like the same problem as in communication - concrete message handler should return instance of message type for correct deserialization. Idea can be reused:
struct Message {
    field MarkerInterface
}

message.field = concreteStruct with MarshalJSON - communication serializes it correclty.
on deserialization side:
var message Message
message.field = concreteStruct
json.Unmarshal(&message). Should work, no?

PR #599, opened 2018-12-06
Feature/destroy session

zolia:
Destroys session on failed connect or disconnect.


session/destroy_consumer_test.go
```diff
@@ -0,0 +1,94 @@
...
+import (
+	"testing"
+
+	"github.com/mysteriumnetwork/node/identity"
+	"github.com/stretchr/testify/assert"
+)
+
+// managerFake represents fake manager usually useful in tests
+type managerDestroyFake struct {
+	lastConsumerID identity.Identity
```
soffokl: `lastConsumerID` is unused.

session/destroy_consumer_test.go
```diff
@@ -0,0 +1,94 @@
...
+	lastProposalID int
```
soffokl: `lastProposalID` is unused.

session/destroy_producer_test.go
```diff
@@ -0,0 +1,51 @@
...
+
+var (
+	successfulSessionDestroyResponse = &DestroyResponse{
+		Success: true,
+	}
+)
+
+func TestProducer_RequestSessionDestroy(t *testing.T) {
+	sender := &fakeSender{}
+	sid, _, err := RequestSessionCreate(sender, 123)
```
soffokl: We are not checking this `err` but assigning it.

session/destroy_consumer.go
```diff
@@ -0,0 +1,59 @@
...
+	"github.com/mysteriumnetwork/node/identity"
+)
+
+// destroyConsumer processes session create requests from communication channel.
+type destroyConsumer struct {
+	SessionManager Manager
+	PeerID         identity.Identity
+}
+
+// GetMessageEndpoint returns endpoint there to receive messages
```
soffokl:
```suggestion
// GetMessageEndpoint returns endpoint where to receive messages.
```

session/destroy_consumer_test.go
```diff
@@ -0,0 +1,94 @@
...
+package session
+
+import (
+	"testing"
+
+	"github.com/mysteriumnetwork/node/identity"
+	"github.com/stretchr/testify/assert"
+)
+
+// managerFake represents fake manager usually useful in tests
```
soffokl:
```suggestion
// managerDestroyFake represents fake destroy manager usually useful in tests
```

session/destroy_consumer_test.go
```diff
@@ -0,0 +1,94 @@
...
+	assert.Error(t, err)
+	assert.Exactly(t, destroyResponse(ErrorWrongSessionOwner), sessionResponse)
+}
+
+// Create function creates and returns fake session
+func (manager *managerDestroyFake) Create(consumerID identity.Identity, proposalID int) (Session, error) {
+	return Session{}, nil
+}
+
+func (manager *managerDestroyFake) Destroy(consumerID identity.Identity, serviceID string) error {
```
soffokl: Shouldn't the `serviceID` be `sessionID`?

session/create_consumer_test.go
```diff
@@ -98,3 +98,7 @@ func (manager *managerFake) Create(consumerID identity.Identity, proposalID int)
 	manager.lastProposalID = proposalID
 	return manager.returnSession, manager.returnError
 }
+
+func (manager *managerFake) Destroy(consumerID identity.Identity, serviceID string) error {
```
soffokl: Shouldn't the `serviceID` be `sessionID`?

session/dialog_handler.go
```diff
@@ -41,10 +41,23 @@ func (handler *handler) Handle(dialog communication.Dialog) error {
...
 			SessionManager: handler.sessionManagerFactory(dialog),
 			PeerID:         dialog.PeerID(),
 		},
 	)
+
+	if err != nil {
+		return err
+	}
+
+	err = dialog.Respond(
```
soffokl:
```suggestion
	return dialog.Respond(
```

core/connection/manager.go
```diff
@@ -154,6 +154,8 @@ func (manager *connectionManager) startConnection(consumerID identity.Identity,
 		return err
 	}
 
+	cancel = append(cancel, func() { session.RequestSessionDestroy(dialog, sessionID) })
```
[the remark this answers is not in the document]
zolia: good catch.. fixing it.

session/destroy_producer.go
```diff
@@ -0,0 +1,65 @@
...
+	}
+}
+
+// RequestSessionDestroy requests session destruction and returns response data
+func RequestSessionDestroy(sender communication.Sender, sessionID ID) error {
+	responsePtr, err := sender.Request(&destroyProducer{
+		SessionID: string(sessionID),
+	})
+	if err != nil {
+		return err
```
vkuznecovas: we probably want to log this as well

session/destroy_consumer.go
```diff
@@ -0,0 +1,59 @@
...
+package session
+
+import (
+	"github.com/mysteriumnetwork/node/communication"
+	"github.com/mysteriumnetwork/node/identity"
+)
+
+// destroyConsumer processes session create requests from communication channel.
+type destroyConsumer struct {
+	SessionManager Manager
```
tadovas: You don't need full interface here, only single Destroy method - depend on your own interfaces.

session/destroy_consumer.go
```diff
@@ -0,0 +1,59 @@
...
+func (consumer *destroyConsumer) NewRequest() (requestPtr interface{}) {
+	return &DestroyRequest{}
+}
+
+// Consume handles requests from endpoint and replies with response
+func (consumer *destroyConsumer) Consume(requestPtr interface{}) (response interface{}, err error) {
+	request := requestPtr.(*DestroyRequest)
+
+	err = consumer.SessionManager.Destroy(consumer.PeerID, request.SessionID)
+	return destroyResponse(err), err
```
tadovas: In case of error - either return error or return nil and message which needs to be sent to consumer, not both. According to go idiomatics - if function returns err, other value is considered to be invalid.
My suggestion: only return real fatal errors here - unable to parse etc. In case session Destroy returns error - its enough to log it and return destroyResponse

session/destroy_dto.go
```diff
@@ -0,0 +1,35 @@
...
+package session
+
+import (
+	"github.com/mysteriumnetwork/node/communication"
+)
+
+const endpointSessionDestroy = communication.RequestEndpoint("session-destroy")
+
+// DestroyRequest structure represents message from service consumer to destroy session for given session id
+type DestroyRequest struct {
```
tadovas: These both structures can be in the same file as consumer/producer

session/destroy_producer.go
zolia: fixed

session/manager.go
```diff
@@ -47,19 +51,25 @@ type PromiseProcessor interface {
 	Stop() error
 }
 
+// Manager defines methods for session management
+type Manager interface {
```
tadovas: This interface is needed here - caller should define it, not implementer

zolia: Removed interface. Created Creator interface at caller.

session/manager.go
```diff
@@ -51,12 +51,19 @@ type PromiseProcessor interface {
 	Stop() error
 }
 
+// MemorySessionStorage interface to session storage
+type MemorySessionStorage interface {
```
tadovas: Why it's memory? Simply SessionStorage. Structure satisfying this interface and keeping sessions in memory could be called that

zolia: Since we are in sessions package already. That would give us sessions sessions... 
We could call just "Storage" in that case.

soffokl: I think just `session.Storage` is the best choice.

session/create_consumer.go
```diff
@@ -24,17 +24,17 @@ import (
...
 
-// Manager defines methods for session management
-type Manager interface {
-	Create(consumerID identity.Identity, proposalID int) (Session, error)
-}
-
 // createConsumer processes session create requests from communication channel.
 type createConsumer struct {
-	SessionManager Manager
+	SessionManager Creator
```
tadovas: SessionManager var of type Creator sounds weird. Also it's not needed to be public

zolia: fixed

PR #603, opened 2018-12-10
Added routes configuration for wireguard connection

soffokl:


services/wireguard/endpoint/kernelspace/client.go
```diff
@@ -132,3 +166,10 @@ func stringToKey(key string) (wgtypes.Key, error) {
 	}
 	return wgtypes.NewKey(k)
 }
+
+func ipExec(args ...string) error {
+	if err := exec.Command("ip", args...).Run(); err != nil {
```
zolia: We prob will need to use exec.Command().CombinedOutput here since simple run will swallow error outputs and there will be no way to test what the problem was.

services/wireguard/endpoint/kernelspace/client.go
```diff
@@ -84,20 +92,46 @@ func (c *client) AddPeer(iface string, peer wg.PeerInfo) error {
...
 
 func (c *client) Close() error {
-	if err := exec.Command("ip", "link", "del", "dev", c.iface).Run(); err != nil {
+	d, err := c.wgClient.Device(c.iface)
+	if err != nil || d.Name != c.iface {
+		return err
+	}
+
+	if len(d.Peers) > 0 && d.Peers[0].Endpoint != nil {
+		if err := ipExec("route", "del", d.Peers[0].Endpoint.IP.String()); err != nil {
```
zolia: We should not use `route` command on linux since we are already using `ip`. Route is completely covered by `ip route` subcommand.

soffokl: `ipExec` is calling `ip` tool inside, all function args are arguments for `ip` tool. 
Looks like it is a confusing syntax, and it's not clear.

services/wireguard/endpoint/kernelspace/client.go
zolia: Each external command requiring superuser privileges we run with sudo. `/sbin/ip` is included in linux sudoers.

soffokl: Fixed.

services/wireguard/endpoint/kernelspace/client.go
```diff
@@ -84,20 +92,46 @@ func (c *client) AddPeer(iface string, peer wg.PeerInfo) error {
...
-	if err := exec.Command("ip", "address", "replace", "dev", iface, ipAddr.String()).Run(); err != nil {
+	if err := suExec("ip", "address", "replace", "dev", iface, ipAddr.String()); err != nil {
 		return err
 	}
 
-	return exec.Command("ip", "link", "set", "dev", iface, "up").Run()
+	return suExec("ip", "link", "set", "dev", iface, "up")
+}
+
+func (c *client) addRoute(iface string, endpoint net.UDPAddr) error {
```
tadovas: It's not just addRoute, but it actually adds route to real interface for remote peer, also setups default routing for tunnel - either naming or split this function into two

services/wireguard/endpoint/kernelspace/client.go
```diff
@@ -132,3 +170,10 @@ func stringToKey(key string) (wgtypes.Key, error) {
 	}
 	return wgtypes.NewKey(k)
 }
+
+func suExec(args ...string) error {
```
zolia: IMHO this suExec could be in some utils or some other generic package. This function can and should be reused along the linux (maybe unix) code.

soffokl: Moved it into `utils` package.

services/wireguard/endpoint/kernelspace/client.go
```diff
@@ -84,20 +93,47 @@ func (c *client) AddPeer(iface string, peer wg.PeerInfo) error {
...
-	if err := exec.Command("ip", "address", "replace", "dev", iface, ipAddr.String()).Run(); err != nil {
+	if err := utils.SuExec("ip", "address", "replace", "dev", iface, ipAddr.String()); err != nil {
 		return err
 	}
 
-	return exec.Command("ip", "link", "set", "dev", iface, "up").Run()
+	return utils.SuExec("ip", "link", "set", "dev", iface, "up")
+}
+
+func (c *client) excludeRoute(endpoint net.UDPAddr) error {
```
tadovas: Why UDPAddr is needed to excludeRoute - simple ip should be enough. UDP has nothing to do with routes

services/wireguard/endpoint/kernelspace/client.go
```diff
@@ -73,6 +74,14 @@ func (c *client) AddPeer(iface string, peer wg.PeerInfo) error {
 		if err != nil {
 			return err
 		}
+		if endpoint != nil {
```
vkuznecovas: can we reduce nesting in this method in general? something like 
```golang
if peer == nil {
  // return
}
rest of the code here
```

utils/command.go
```diff
@@ -31,3 +32,12 @@ func SplitCommand(command string, commandArguments string) *exec.Cmd {
 	}
 	return exec.Command(command, trimmedArgs...)
 }
+
+// SuExec executes external command with a sudo privileges.
+// It returns an combined stderr and stdout output and exit code in case of error.
+func SuExec(args ...string) error {
```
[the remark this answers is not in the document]
soffokl: renamed.

services/wireguard/endpoint/kernelspace/client.go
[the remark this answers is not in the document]
soffokl: Changed it to be called on a higher level.

services/wireguard/endpoint/kernelspace/client.go
soffokl: Changed it.

services/wireguard/endpoint/kernelspace/client.go
soffokl: Figured out that we don't need this check at all. Looks like some leftover. 👍

soffokl:
New review required here.

PR #608, opened 2018-12-11
 Initial support for multiple wireguard sessions

soffokl:


services/wireguard/resources/allocator.go
```diff
@@ -17,29 +17,131 @@
...
+import (
+	"errors"
+	"fmt"
+	"net"
+	"strconv"
+	"strings"
+	"sync"
+)
+
+const prefix = "myst"
```
zolia:
```suggestion
const interfacePrefix = "myst"
```

services/wireguard/resources/allocator.go
```diff
@@ -17,29 +17,131 @@
...
+	return Allocator{
+		Ifaces:      make(map[int]struct{}),
+		IPAddresses: make(map[int]struct{}),
+		Port:        make(map[int]struct{}),
+	}
+}
 
 // AllocateInterface provides available name for the wireguard network interface.
-func (a *Allocator) AllocateInterface() string { return "myst0" }
+func (a *Allocator) AllocateInterface() string {
```
zolia: I think it should return an error. Basically "" interface name is an error. Allocation may fail triggering critical error.

zolia:
We need to make sure, that there will be no "mystXX" interfaces leftovers. In case system crashes, or myst process is killed we need to cleanup all "mystXX" interfaces before starting up.

soffokl:
@zolia I have added cleaning for unknown interfaces. Please take a look at it, maybe you will have some thoughts about the implementation.

services/wireguard/resources/allocator.go
soffokl: Added error.

services/wireguard/endpoint/endpoint.go
```diff
@@ -44,8 +45,21 @@ type connectionEndpoint struct {
 // Start starts and configure wireguard network interface for providing service.
 // If config is nil, required options will be generated automatically.
 func (ce *connectionEndpoint) Start(config *wg.ServiceConfig) error {
-	ce.iface = ce.resourceAllocator.AllocateInterface()
-	ce.endpoint.Port = ce.resourceAllocator.AllocatePort()
+	ce.cleanWildInterfaces()
```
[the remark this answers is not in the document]
soffokl: Changed it to `AbandonedInterfaces`.

services/wireguard/resources/allocator.go
```diff
@@ -17,29 +17,155 @@
...
+	ifaces, err := net.Interfaces()
+	if err != nil {
+		return nil, err
+	}
+
+	list := make([]net.Interface, 0)
+	for _, iface := range ifaces {
+		if strings.HasPrefix(iface.Name, interfacePrefix) {
+			ifaceID, err := strconv.Atoi(strings.TrimPrefix(iface.Name, interfacePrefix))
+			if err == nil {
```
vkuznecovas: might want to log in case of error, for instance if we start creating some phantom interfaces we won't even know about it

services/wireguard/endpoint/kernelspace/client.go
```diff
@@ -120,18 +125,7 @@ func addDefaultRoute(iface string) error {
...
-	}
-
-	if len(d.Peers) > 0 && d.Peers[0].Endpoint != nil {
-		if err := utils.SudoExec("ip", "route", "del", d.Peers[0].Endpoint.IP.String()); err != nil {
-			return err
-		}
-	}
-
-	if err := utils.SudoExec("ip", "link", "del", "dev", c.iface); err != nil {
+	if err := c.DestroyDevice(c.iface); err != nil {
```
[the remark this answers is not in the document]
soffokl: If we failed to destroy device we cannot assume that client if fully closed. 
I think we should fail fast, in this case, to allow investigate the problem.

services/wireguard/resources/allocator.go
soffokl: Added logging to the place with actual cleaning, not just listing.

services/wireguard/endpoint/kernelspace/client.go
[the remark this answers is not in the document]
vkuznecovas: I don't think there's anything reasonable the caller can do with the failure therefore, I think it's best to try and cleanup as much as we can and then return the error to the caller, to inform that not everything went well.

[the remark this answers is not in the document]
soffokl: make sense, changed it to return an error only at the end.

services/openvpn/service/factory.go
```diff
@@ -112,21 +112,16 @@ type OpenvpnConfigNegotiator struct {
...
 
-// ConsumeConfig doesn't do anything on the openvpn side, since it's not required here
-func (ocn *OpenvpnConfigNegotiator) ConsumeConfig(json.RawMessage) error {
-	return nil
-}
-
 // ProvideConfig returns the config for user
-func (ocn *OpenvpnConfigNegotiator) ProvideConfig() (session.ServiceConfiguration, error) {
+func (ocn *OpenvpnConfigNegotiator) ProvideConfig(json.RawMessage) (session.ServiceConfiguration, error) {
 	return &ocn.vpnConfig, nil
```
zolia: I don't see anywhere where `ServiceConfiguration` interface implementation would return any kind of error. Not sure if its needed, but we could ALWAYS return a valid 'pseudo' configuration without possibility of error.

tadovas: Maybe error is needed to signal underlying caller that we cannot provide config - therefore drop consumer session or whatever

soffokl: Should we cover it in this PR?

tadovas: Isn't it's working this way already? My comment was more an answer to @zolia

soffokl: yep, correct, it can fail if we cannot provide config.

services/wireguard/endpoint/kernelspace/client.go
```diff
@@ -82,6 +84,10 @@ func (c *client) AddPeer(iface string, peer wg.PeerInfo) error {
 	return c.wgClient.ConfigureDevice(iface, deviceConfig)
 }
 
+func (c *client) DestroyDevice(name string) error {
+	return exec.Command("ip", "link", "del", "dev", name).Run()
```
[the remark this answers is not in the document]
soffokl: Fixed it, missed it for some reasons. Thanks.

services/wireguard/endpoint/kernelspace/client.go
```diff
@@ -120,21 +126,17 @@ func addDefaultRoute(iface string) error {
...
-	}
-
-	if err := utils.SudoExec("ip", "link", "del", "dev", c.iface); err != nil {
-		return err
+	if closeErr != nil && destroyErr == nil {
+		return closeErr
+	} else if closeErr == nil && destroyErr != nil {
+		return destroyErr
+	} else if closeErr != nil && destroyErr != nil {
+		return fmt.Errorf("failed to close wireguard client %v; failed to destroy wireguard device: %v", closeErr, destroyErr)
```
zolia: All these checks looks abundant for me. What if we had another function returning another error here? Would you cross-check all options too? 
I think we could fail fast (return) on first error and defer rest closes as they are.

soffokl: It was working this way initially, changed it after this comment: https://github.com/mysteriumnetwork/node/pull/608#discussion_r241315944

services/wireguard/resources/allocator.go
```diff
@@ -17,29 +17,155 @@
...
+	return *subnet, err
 }
 
 // AllocatePort provides available UDP port for the wireguard endpoint.
-func (a *Allocator) AllocatePort() int { return 52820 }
+func (a *Allocator) AllocatePort() (int, error) {
+	a.mu.Lock()
+	defer a.mu.Unlock()
+
+	for i := 52820; i < 53000; i++ {
```
zolia: We have less range of ports for allocation than IPs. We could relate those with some range constant..

services/wireguard/endpoint/kernelspace/client.go
soffokl: oh, got your point about defering rest closes. Changed it.

services/wireguard/resources/allocator.go
soffokl: Added a const for that.

zolia:
closes #608

PR #612, opened 2018-12-13
Added destroy callback for transport services

soffokl:


services/wireguard/endpoint/endpoint.go
```diff
@@ -110,6 +130,21 @@ func (ce *connectionEndpoint) Stop() error {
...
 }
 
+func (ce *connectionEndpoint) cleanAbandonedInterfaces() error {
+	ifaces, err := ce.resourceAllocator.AbandonedInterfaces()
+	if err != nil {
+		return err
+	}
+
+	for _, iface := range ifaces {
+		if err := ce.wgClient.DestroyDevice(iface.Name); err != nil {
```
vkuznecovas: might want to try and destroy others, before returning the error

services/wireguard/endpoint/kernelspace/client.go
```diff
@@ -97,7 +101,7 @@ func (c *client) up(iface string, ipAddr net.IPNet) error {
 }
 
 func (c *client) Close() error {
-	if err := exec.Command("ip", "link", "del", "dev", c.iface).Run(); err != nil {
+	if err := c.DestroyDevice(c.iface); err != nil {
```
vkuznecovas: if the destroy fails, we wont close wgClient. Is this intended?

soffokl:
@vkuznecovas @zolia @tadovas please take a look on this one. 
Left only `destroyCallback` without renaming `configProvider`

PR #620, opened 2018-12-14
e2e tests for noop and wg

### LINKED ISSUE #614: e2e tests for WG
need to add wg to our e2e

vkuznecovas:
Closes #614

### DIFF
--- a/bin/test_e2e
+++ b/bin/test_e2e
@@ -2,6 +2,20 @@
 
 source bin/localnet/functions.sh
 
+if [ -n "$TRAVIS" ]
+then
+    echo "Travis detected, installing wireguard kernel module"
+    sudo add-apt-repository -y ppa:wireguard/wireguard && \
+    sudo apt-get update && \
+    sudo apt-get install -y linux-headers-$(uname -r) && \
+    sudo apt-get install -y wireguard
+    if [ ! $? -eq 0 ]
+    then
+        print_error "Failed to install WG kernel module"
+        exit 1
+    fi
+fi
+
 projectName="node_e2e_test"
 projectFiles=("e2e/docker-compose.yml")
 
--- a/bin/test_e2e
+++ b/bin/test_e2e
@@ -43,6 +43,7 @@ $dockerComposeCmd run go-runner \
 if [ ! $? -eq 0 ]
 then
     print_error "Tests failed"
+    cat e2e_tests.log
     cleanup $projectName
     exit 1
 fi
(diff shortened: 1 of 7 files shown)


vkuznecovas:
The NATS timeout was a result of the service provider dying. The provider did die due to an out of memory error, which is a result of travis limiting us to 3GB of memory. In order to reduce the memory footprint, the consumer and provider are both now using the lightweight keystore. This allows our e2e suite to fit within 2GB of ram(on my local machine at least).

Thanks @tadovas for pointing me towards #631 which allows for drastic memory footprint reduction.

e2e/connection_test.go
```diff
@@ -150,11 +158,38 @@ func consumerConnectFlow(t *testing.T, tequilapi *tequilapi_client.Client, consu
...
+
+type sessionAsserter func(t *testing.T, session endpoints.SessionDTO)
+
+var serviceTypeAssertionMap = map[string]sessionAsserter{
+	"openvpn":   assertOpenvpn,
+	"noop":      assertNoop,
+	"wireguard": assertWireguard,
+}
+
+func assertOpenvpn(t *testing.T, session endpoints.SessionDTO) {
```
tadovas: What's the point of different service type assertions ? Function is the same :|

vkuznecovas: fixed, now we're asserting `assertStatsNotZero` and `assertStatsZero

bin/test_e2e
```diff
@@ -2,6 +2,20 @@
 
 source bin/localnet/functions.sh
 
+if [ -n "$TRAVIS_BUILD_NUMBER" ]
```
tadovas: Use something from the list of default travis variables https://docs.travis-ci.com/user/environment-variables/#default-environment-variables

bin/test_e2e
```diff
@@ -2,6 +2,20 @@
 
 source bin/localnet/functions.sh
 
+if [ -n "$TRAVIS" ]
+then
+    echo "Travis detected, installing kernel module"
+    sudo add-apt-repository -y ppa:wireguard/wireguard && \
+    sudo apt-get update && \
+    sudo apt install -y linux-headers-$(uname -r) && \
```
soffokl: we can install it in a single line
`sudo apt-get install -y wireguard linux-headers-$(uname -r)`

soffokl: And it's better to use single too, not both `apt` and `apt-get`.

bin/test_e2e
```diff
@@ -2,6 +2,20 @@
 
 source bin/localnet/functions.sh
 
+if [ -n "$TRAVIS" ]
+then
+    echo "Travis detected, installing kernel module"
```
soffokl:
```suggestion
     echo "Travis detected, installing wireguard kernel module"
```

PR #622, opened 2018-12-17
Added session stats sending for wireguard kernelspace implementation

soffokl:


services/wireguard/connection/connection.go
```diff
@@ -105,6 +111,23 @@ func (c *Connection) Stop() {
 	}
 
 	c.stateChannel <- connection.NotConnected
-	c.connection.Done()
+	c.cancel()
```
vkuznecovas: might want to close the statistics channel here, to inform the consumer that we're not gonna send anymore

services/wireguard/endpoint/kernelspace/client.go
```diff
@@ -83,6 +85,22 @@ func (c *client) AddPeer(iface string, peer wg.PeerInfo) error {
 	return c.wgClient.ConfigureDevice(iface, deviceConfig)
 }
 
+func (c *client) PeerStats() (consumer.SessionStatistics, error) {
+	d, err := c.wgClient.Device(c.iface)
+	if err != nil || d.Name != c.iface {
```
vkuznecovas: Aren't we hiding an error here with the `d.Name != c.iface`? If the names don't match, all we'll get is a nil error and zero stats. Should we not at least log this occurrence?

services/wireguard/connection/connection.go
```diff
@@ -54,35 +58,37 @@ func (c *Connection) Start(options connection.ConnectOptions) (err error) {
 		return err
 	}
 
-	c.connection.Add(1)
+	c.ctx, c.cancel = context.WithCancel(context.Background())
```
tadovas: I am not sure context is really needed here as it was introduced for different purposes. Actually there are alot of dicusssions around this topic in golang community. My humble opinion - if it can be avoided - don't use it. I guess simple channel will do the trick anyway.

soffokl: Replaced context by the channel.

soffokl:
Rebased the PR, another approval required.

soffokl:
Fixed build problem after a merge. New review required.

PR #626, opened 2018-12-17
session stats now subscribe to session events

### LINKED ISSUE #621: Session duration is reset after reconnecting in dev-net
Connection for mobile is stable in dev-net, but each time reconnect happens - session duration (received from statistics endpoint) is reset to 0.

vkuznecovas:
Connection manager now emits session events, which the session storage, statistics tracker/reporter now subscribe to. This ensures that internal connection state does not affect the statistics and in return, does not reset the session duration.

Closes #621


consumer/statistics/tracker.go
```diff
@@ -69,17 +70,36 @@ func (sst *SessionStatisticsTracker) markSessionEnd() {
 	sst.sessionStart = nil
 }
 
+// calcStatDiff takes in the old and the new values of statistics, returns the calculated delta
+func (sst *SessionStatisticsTracker) calcStatDiff(old, new uint64) (res uint64) {
```
tadovas: It's just a function - no need of receiver

consumer/statistics/tracker.go
```diff
@@ -69,17 +70,36 @@ func (sst *SessionStatisticsTracker) markSessionEnd() {
...
+	if old > new {
+		return new
+	}
+	return new - old
+}
+
+// GetStatisticsDiff calculates the difference in bytes between the old stats and new
+func (sst *SessionStatisticsTracker) GetStatisticsDiff(old, new consumer.SessionStatistics) consumer.SessionStatistics {
```
tadovas: Same here - no need of receiver

consumer/session/session_storage.go
```diff
@@ -66,17 +66,17 @@ func (repo *Storage) GetAll() ([]History, error) {
...
-func (repo *Storage) ConsumeStateEvent(stateEvent connection.StateEvent) {
-	switch stateEvent.State {
-	case connection.Disconnecting:
-		repo.handleDisconnectingEvent(stateEvent.SessionInfo.SessionID)
-	case connection.Connected:
-		repo.handleConnectedEvent(stateEvent.SessionInfo)
+// ConsumeSessionEvent consumes the session state change events
+func (repo *Storage) ConsumeSessionEvent(sessionEvent connection.SessionEvent) {
+	switch sessionEvent.Status {
+	case connection.SessionStatusEnded:
```
zolia:
```suggestion
	case connection.SessionEndedStatus:
```
is more intuitive.

consumer/statistics/tracker.go
```diff
@@ -69,17 +70,36 @@ func (sst *SessionStatisticsTracker) markSessionEnd() {
...
+		BytesSent:     sst.calcStatDiff(old.BytesSent, new.BytesSent),
+		BytesReceived: sst.calcStatDiff(old.BytesReceived, new.BytesReceived),
+	}
+}
+
 // ConsumeStatisticsEvent handles the connection statistics changes
 func (sst *SessionStatisticsTracker) ConsumeStatisticsEvent(stats consumer.SessionStatistics) {
-	sst.sessionStats = stats
+	diff := sst.GetStatisticsDiff(sst.lastStats, stats)
+	sst.sessionStats.BytesReceived += diff.BytesReceived
```
tadovas: If you are making a function to calc diff, maybe it's worth making a function newStats := AddUp(sessionStats, diff)

tadovas: Those diff and add methods can even be part of SessionStatistics since we are doing calculations with them anyway

PR #628, opened 2018-12-18
Added detection if wireguard kernel space supported

### LINKED ISSUE #623: Wireguard transport should detect kernel space support
Wireguard transport implementation should detect if the wireguard kernel-space module available and use it when it possible. 
If the wireguard kernel-space module is not available we need to switch to user-space implementation.

soffokl:
Closes #623

### DIFF
--- a/services/wireguard/endpoint/create_linux.go
+++ b/services/wireguard/endpoint/create_linux.go
@@ -54,6 +54,6 @@ func NewConnectionEndpoint(ipResolver ip.Resolver, resourceAllocator *resources.
 
 func isKernelSpaceSupported() bool {
 	err := utils.SudoExec("ip", "link", "add", "iswgsupported", "type", "wireguard")
-	utils.SudoExec("ip", "link", "del", "iswgsupported")
+	_ = utils.SudoExec("ip", "link", "del", "iswgsupported")
 	return err == nil
 }
--- a/services/wireguard/endpoint/create_linux.go
+++ b/services/wireguard/endpoint/create_linux.go
@@ -20,15 +20,27 @@
 package endpoint
 
 import (
+	log "github.com/cihub/seelog"
 	"github.com/mysteriumnetwork/node/core/ip"
 	wg "github.com/mysteriumnetwork/node/services/wireguard"
 	"github.com/mysteriumnetwork/node/services/wireguard/endpoint/kernelspace"
+	"github.com/mysteriumnetwork/node/services/wireguard/endpoint/userspace"
 	"github.com/mysteriumnetwork/node/services/wireguard/resources"
+	"github.com/mysteriumnetwork/node/utils"
 )
 
 // NewConnectionEndpoint creates new wireguard connection endpoint.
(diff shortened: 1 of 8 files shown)


services/wireguard/endpoint/create_linux.go
```diff
@@ -40,12 +52,8 @@ func NewConnectionEndpoint(ipResolver ip.Resolver, resourceAllocator *resources.
...
-func GeneratePrivateKey() (string, error) {
-	return kernelspace.GeneratePrivateKey()
-}
-
-// PrivateKeyToPublicKey generates wireguard public key from private key
-func PrivateKeyToPublicKey(key string) (string, error) {
-	return kernelspace.PrivateKeyToPublicKey(key)
+func isKernelSpaceSupported() bool {
+	err := utils.SudoExec("ip", "link", "add", "iswgsupported", "type", "wireguard")
+	utils.SudoExec("ip", "link", "del", "iswgsupported")
```
vkuznecovas: I think we should explicitly state that we understand that an error is returned, but we don't care.
```suggestion
	_ = utils.SudoExec("ip", "link", "del", "iswgsupported")
```

services/wireguard/endpoint/create_linux.go
```diff
@@ -20,15 +20,27 @@
...
+	"github.com/mysteriumnetwork/node/utils"
 )
 
 // NewConnectionEndpoint creates new wireguard connection endpoint.
 func NewConnectionEndpoint(ipResolver ip.Resolver, resourceAllocator *resources.Allocator) (wg.ConnectionEndpoint, error) {
-	wgClient, err := kernelspace.NewWireguardClient()
+	var wgClient wgClient
+	var err error
+
+	if isKernelSpaceSupported() {
```
vkuznecovas: maybe extract this to a separate function, something like GetWGClient?

soffokl: Extracted it.

services/wireguard/endpoint/create_linux.go
```diff
@@ -40,12 +43,17 @@ func NewConnectionEndpoint(ipResolver ip.Resolver, resourceAllocator *resources.
...
+	return userspace.NewWireguardClient()
 }
 
-// PrivateKeyToPublicKey generates wireguard public key from private key
-func PrivateKeyToPublicKey(key string) (string, error) {
-	return kernelspace.PrivateKeyToPublicKey(key)
+func isKernelSpaceSupported() bool {
+	err := utils.SudoExec("ip", "link", "add", "iswgsupported", "type", "wireguard")
+	_ = utils.SudoExec("ip", "link", "del", "iswgsupported")
+	return err == nil
```
tadovas: Logging of error would be nice - in case we need to figure out why exactly it failed - maybe wg is still supported but some little detail needs to be tweaked.

soffokl: Added debug output.

soffokl:
Rebased the PR, another approval required.

PR #644, opened 2018-12-22
Place Info.plist in root framework folder

Arrnas:
Can't find a source for this but AFAIK iOS frameworks need to have the `Info.plist` file in the root folder of the framework. Might need to `rm -rf ${DIR_FRAMEWORK}/Versions` but not sure if that's necessary. Don't have environment set up so can't test this.


tadovas:
Hmm @Arrnas can you give a link or some reference for more info?

PR #650, opened 2019-01-03
Disable wireguard build for windows

soffokl:


zolia:
@soffokl please give us a short description if the task next time after you put up PR. It saves time for everyone before reading what to look for and what this PR's scope.

zolia:
It also gives something to remember for the future and for indexing engines :)

zolia:
It is always good to put up a motivation, why we do this task and what and how it solves. This way we learn to attract possible commits from community and dictate development culture.

PR #651, opened 2019-01-03
Service interface now does not have a wait

### LINKED ISSUE #548: Refactor service interface
At the moment, the service interface looks like the following:

```go
type Service interface {
	Start(providerID identity.Identity) (dto_discovery.ServiceProposal, session.ConfigProvider, error)
	Wait() error
	Stop() error
}
```

The wait seems a bit redundant, the suggestion is to remove the wait, and instead rely on a "serve"
 or run method, that blocks. We can then start it in routines if needed on the consumer side

vkuznecovas:
closes #548 

Removed the wait method from service interface. Start now blocks.

Since we relied on start returning run time variables, the proposal generation has been separated from the starting of the service. The proposal is now returned as a second return value from the service factory instead. Also included the ProvideConfig method to the service, so it's not longer returned from the start but is present on the service itself.

Removed the ip/country resolver dependencies from the services, they are now thinner. These are now resolved at service construction time(see the updated factories in bootstrap).

Also a question - should we keep the start method named start, or do we prefer run/serve?


core/service/manager.go
```diff
@@ -18,9 +18,9 @@
 package service
 
 import (
+	"encoding/json"
 	"errors"
 
-	log "github.com/cihub/seelog"
```
soffokl: `const logPrefix = "[service-manager] "` left and it's unused.

soffokl:
> Also a question - should we keep the start method named start, or do we prefer run/serve?

For me `start` naming is ok here.

cmd/service_bootstrap_desktop.go
```diff
@@ -49,27 +51,79 @@ func (di *Dependencies) BootstrapServices(nodeOptions node.Options) error {
...
+
+	log.Info(logPrefix, "Detected service country: ", loc.Country)
+
+	return
+}
+
 func (di *Dependencies) bootstrapServiceOpenvpn(nodeOptions node.Options) {
-	createService := func(serviceOptions service.Options) (service.Service, error) {
+	createService := func(serviceOptions service.Options) (service.Service, market.ServiceProposal, error) {
+		location, err := di.resolveIPsAndLocation()
```
Waldz: I imagine that things like (location, ip address, etc.) should be asked from `nodeContext`, instead if fetching it twice

cmd/service_bootstrap_desktop.go
```diff
@@ -49,27 +51,79 @@ func (di *Dependencies) BootstrapServices(nodeOptions node.Options) error {
...
+func (di *Dependencies) resolveIPsAndLocation() (loc locationInfo, err error) {
+	pubIP, err := di.IPResolver.GetPublicIP()
+	if err != nil {
+		return
+	}
+	loc.PubIP = pubIP
+
+	outboundIP, err := di.IPResolver.GetOutboundIP()
+	if err != nil {
+		return
```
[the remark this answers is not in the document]
vkuznecovas: It doesn't really matter. If a method returns an error, you should always ignore the values.

cmd/service_bootstrap_desktop.go
vkuznecovas: We don't have a component for this though, there's a location cache, which is rather strange but nothing like that for ip resolution. Created #653 for this. Out of scope for this PR.

Waldz: I am saying that this duplication indicates that its tine to refactor towards better arcitecture

tadovas:
I would go for Serve

vkuznecovas:
Renamed the start to serve

zolia:
need to rebase..

Issue #535, opened 2018-11-12
Better error handling


vkuznecovas:
Under certain conditions, the discovery API fails to return a JSON response. In those cases node will crash with an error of a similar type: `Failed to register proposal, retrying after 1 min. invalid character '<' looking for beginning of value`. This is probably true for most of external calls. Need to improve the handling.

I've added [an issue](https://github.com/mysteriumnetwork/api/issues/86) to the api project to handle errors better there as well.

zolia:
Isn't if fixed?
I remember you were addressing it @vkuznecovas ?

soffokl:
Looks like no, I'm still getting such errors sometimes.

bbrunner11:
I'm seeing this again w/ the latest docker running on ubuntu 16.04 

`2019-06-13T22:54:43.694413208 [Error] [discovery]  Failed to register proposal, retrying after 1 min. Post https://testnet-api.mysterium.network/v1/register_proposal: net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)`

been happening for 7ish hours, was working fine up until then.  Restarting the container does not help.

PR #657, opened 2019-01-07
Better error handling for non JSON responses. #535

soffokl:
We were expecting JSON responses from the remote server, but sometimes error was returned as plain text. 
Now we are logging any details that were received, that makes it more readable. 

Closes: #535


market/mysterium/response.go
```diff
@@ -24,22 +24,20 @@ import (
...
-		var error errorResponse
+		var errorMsg []byte
 
-		err := ParseResponseJSON(response, &error)
-		if err != nil {
-			return err
+		if response.Body != nil {
+			errorMsg, err = ioutil.ReadAll(response.Body)
+			if err != nil {
+				return err
```
tadovas: I would suggest wrapping this error too into our "server response invalid: ....". Because in case of failed body read you will get for example "EOF" only with no context where it failed.
A common message "server respone invalid: ...." should be better for both cases

market/mysterium/response.go
```diff
@@ -22,24 +22,24 @@ import (
...
-		var error errorResponse
+		var errorMsg []byte
 
-		err := ParseResponseJSON(response, &error)
-		if err != nil {
-			return err
+		if response.Body != nil {
+			errorMsg, err = ioutil.ReadAll(response.Body)
+			if err != nil {
+				return errors.Wrap(err, "server response invalid")
```
tadovas: I guess my previous comment was a bit confusing :)
just do fmt.Errorf("server response invalid. Body reading error %s", err) to add more information. That's it

soffokl: But `errors.Wrap` [does](https://godoc.org/github.com/pkg/errors#Wrap) the same. But as an extra option, it allows getting more details if you need id for debugging if you want it. 
By default, it will be the same error as with `fmt.Errorf`.

And some people [recommends](https://dave.cheney.net/2016/04/27/dont-just-check-errors-handle-them-gracefully) to annotate errors with a `errors.Wrap`.

tadovas: In that case we must cease to use fmt.Errorf and use that error package. Sounds like one of team decision and guideline for me

zolia: It would be good to have an actual stack trace, @soffokl , but for this we would probably need to change how we print errors. In general i agree that we need to wrap error causes.

PR #670, opened 2019-01-22
Added initial support for wireguard on windows

soffokl:
This PR depends on https://github.com/mysteriumnetwork/wireguard-go/pull/2
It will contain one extra commit to add https://github.com/mysteriumnetwork/wireguard-go/pull/2 to vendor directory.


services/wireguard/connection/connection.go
```diff
@@ -48,15 +49,15 @@ type Connection struct {
 func (c *Connection) Start(options connection.ConnectOptions) (err error) {
 	var config wg.ServiceConfig
 	if err := json.Unmarshal(options.SessionConfig, &config); err != nil {
-		return err
+		return errors.Wrap(err, "failed to unmarshal connection config")
```
zolia: For future reference. Try not to do refactoring in the same PR which solves some bug or implements some feature. Not to clutter reading. Do it in separate PR.

soffokl: I was using with wrapping for my debugging during developing. It was a pity to remove. :)

services/wireguard/endpoint/userspace/network_windows.go
```diff
@@ -0,0 +1,94 @@
...
+	"net"
+	"os/exec"
+	"strconv"
+
+	"github.com/jackpal/gateway"
+	"github.com/pkg/errors"
+)
+
+func assignIP(iface string, subnet net.IPNet) error {
+	out, err := exec.Command("powershell", "-Command", "netsh interface ip set address name=\""+iface+"\" source=static "+subnet.String()).CombinedOutput()
```
zolia: this powershell will definitely will not be available on pre Win7 versions. We probably need to state that somewhere in docs.. and requirements.

soffokl: Pre Win7 versions of windows officially not supported, I don't think that we should worry about it.

services/wireguard/endpoint/userspace/network_windows.go
```diff
@@ -0,0 +1,94 @@
...
+	return errors.Wrap(err, string(out))
+}
+
+func excludeRoute(ip net.IP) error {
+	gw, err := gateway.DiscoverGateway()
+	if err != nil {
+		return err
+	}
+
+	out, err := exec.Command("powershell", "-Command", "route add "+ip.String()+"/32 "+gw.String()).CombinedOutput()
```
zolia: Does this command generate any output? If it does, then we would have an error returned for normal output.

services/wireguard/endpoint/userspace/network_windows.go
```diff
@@ -0,0 +1,94 @@
...
+	return errors.Wrap(err, string(out))
+}
+
+func addDefaultRoute(name string) error {
+	id, gw, err := interfaceInfo(name)
+	if err != nil {
+		return errors.Wrap(err, "failed to get interfaces info")
```
zolia:
```suggestion
		return errors.Wrap(err, "failed to get info for interface: " + name)
```

services/wireguard/endpoint/userspace/network_windows.go
soffokl: No, we will return an error only if `err != nil`.

services/wireguard/endpoint/userspace/network_windows.go
soffokl: I think it should be `of` instead of `for`, what do you think? 
```suggestion
		return errors.Wrap(err, "failed to get info of interface: " + name)
```

services/wireguard/endpoint/userspace/network_windows.go
```diff
@@ -0,0 +1,94 @@
...
+
+	if out, err := exec.Command("powershell", "-Command", "route add 0.0.0.0/1 "+gw+" if "+id).CombinedOutput(); err != nil {
+		return errors.Wrap(err, string(out))
+	}
+
+	out, err := exec.Command("powershell", "-Command", "route add 128.0.0.0/1 "+gw+" if "+id).CombinedOutput()
+	return errors.Wrap(err, string(out))
+}
+
+func destroyDevice(name string) error {
```
zolia: Should be commented as a stub if not used.

soffokl:
@zolia this is the same driver that we are installing for OpenVPN with our mysterium VPN app, nothing additional required.

services/wireguard/endpoint/userspace/network_windows.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+
+	addrs, err := iface.Addrs()
+	if err != nil {
+		return "", "", errors.Wrap(err, "failed to get interfaces addresses")
+	}
+
+	var ipv4 net.IP
+	for _, addr := range addrs {
+		ip, _, _ := net.ParseCIDR(addr.String())
```
vkuznecovas: Not a fan of hiding the error here. While I'm pretty sure this will work 99.99999% of the time, I can imagine that if we ever hit a case where this does not work as expected, we'll bang our heads here for quite some time. Log?

tadovas: 👍 especially when parsing from such generic type as string

zolia: We convert to string here from well defined address type here.. but well, if we hit go stdlib bug, it might help :)

services/wireguard/endpoint/userspace/network_windows.go
soffokl: Added comment.

services/wireguard/endpoint/userspace/network_windows.go
soffokl: Added logging just in case. :)

PR #675, opened 2019-01-24
[WIP] Wireguard connection now waits for handshake

soffokl:


mobile/mysterium/wireguard_connection_setup.go
```diff
@@ -297,3 +300,25 @@ func (wg *wireguardConnection) runPeriodically(duration time.Duration) {
...
 	}
 }
+
+func (wg *wireguardConnection) waitHandshake() error {
+	for {
+		select {
+		case <-time.After(100 * time.Millisecond):
+			peers, err := wg.device.Peers()
+			if err != nil {
+				return errors.Wrap(err, "failed to wait peer handshake")
```
zolia:
```suggestion
				return errors.Wrap(err, "error while waiting for peer handshake")
```

mobile/mysterium/wireguard_connection_setup.go
```diff
@@ -297,3 +300,25 @@ func (wg *wireguardConnection) runPeriodically(duration time.Duration) {
...
+			}
+			if len(peers) != 1 {
+				return errors.Wrap(errors.New("exactly 1 peer expected"), "failed to wait peer handshake")
```
zolia:
```suggestion
				return errors.Wrap(errors.New("exactly 1 peer expected"), "error while waiting for peer handshake")
```

mobile/mysterium/wireguard_connection_setup.go
```diff
@@ -297,3 +300,25 @@ func (wg *wireguardConnection) runPeriodically(duration time.Duration) {
 		}
 	}
 }
+
+func (wg *wireguardConnection) waitHandshake() error {
+	for {
+		select {
+		case <-time.After(100 * time.Millisecond):
```
zolia:
```suggestion
		case <-time.After(20 * time.Millisecond):
```
These can be checked more frequently imho.

mobile/mysterium/wireguard_connection_setup.go
```diff
@@ -297,3 +300,25 @@ func (wg *wireguardConnection) runPeriodically(duration time.Duration) {
...
+			if len(peers) != 1 {
+				return errors.Wrap(errors.New("exactly 1 peer expected"), "failed to wait peer handshake")
+			}
+			if peers[0].LastHanshake != 0 {
+				return nil
+			}
+
+		case <-wg.stopChannel:
+			wg.doCleanup()
+			return errors.New("stop received")
```
zolia: would be good to add logging prefix, to indicate a module. Too generic message.

services/wireguard/connection/connection.go
```diff
@@ -80,6 +81,12 @@ func (c *Connection) Start(options connection.ConnectOptions) (err error) {
 		return err
 	}
 
+	if err := c.waitHandshake(); err != nil {
+		c.stateChannel <- connection.NotConnected
+		c.connection.Done()
+		return errors.Wrap(err, "failed to wait peer handshake")
```
zolia:
```suggestion
		return errors.Wrap(err, "error while waiting for peer handshake")
```

soffokl: Is it a good idea to start error with a `error` word? Looks like overkill for me.

zolia: in logs usually you do not see that it is an error.(depending on logger), so I dont see why not to state that it is an error.

soffokl: There is a `state` in the logger. And this error not for logging, it for the upper layer, it could be anything.

soffokl: not `state` sorry
`log level`: `error` `info` `warn` `debug` etc

soffokl:
@zolia Suggested changes applied.

services/wireguard/endpoint/kernelspace/client.go
```diff
@@ -85,20 +85,20 @@ func (c *client) AddPeer(iface string, peer wg.PeerInfo) error {
...
 	if len(d.Peers) != 1 {
-		return consumer.SessionStatistics{}, errors.New("exactly 1 peer expected")
+		return consumer.SessionStatistics{}, 0, errors.New("exactly 1 peer expected")
 	}
 
 	return consumer.SessionStatistics{
 		BytesReceived: uint64(d.Peers[0].ReceiveBytes),
 		BytesSent:     uint64(d.Peers[0].TransmitBytes),
-	}, nil
+	}, int(d.Peers[0].LastHandshakeTime.Unix()), nil
```
vkuznecovas: Does it have to be unix time? Why not just return the time and let the caller decide what to do with it?

vkuznecovas: and this pops up in #678

services/wireguard/wireguard.go
```diff
@@ -58,7 +58,7 @@ func (method Payment) GetPrice() money.Money {
 type ConnectionEndpoint interface {
 	Start(config *ServiceConfig) error
 	AddPeer(publicKey string, endpoint *net.UDPAddr) error
-	PeerStats() (consumer.SessionStatistics, error)
+	PeerStats() (stats consumer.SessionStatistics, lastHandshake int, err error)
```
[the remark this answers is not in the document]
soffokl: hm, I thought that last handshake time is a part of the peer stats. Maybe it should go to the `SessionStatistics` or another struct, but it's still stats.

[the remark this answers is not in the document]
tadovas: Yeap. So instead of returning consumer.SessionStatitics which I bet is something defined as common structure, we need to return wg specific stats with last handshake included.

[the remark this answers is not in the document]
soffokl: Replaced by the single struct.

PR #677, opened 2019-01-25
Added cleaning for NAT forwarding rules

soffokl:


nat/interface.go
```diff
@@ -19,9 +19,10 @@ package nat
 
 // NATService describes fake nat service for darwin
 type NATService interface {
-	Add(rule RuleForwarding)
-	Start() error
-	Stop()
+	Enable() error
+	Add(rule RuleForwarding) error
```
zolia: We return an error, but swallow it in the calling code. We should print out a warning, since it might be not a real error.

soffokl: Fixed error handling. Thanks for catching it.

nat/service_iptables.go
```diff
@@ -18,68 +18,79 @@
...
-			log.Warn("Failed to create ip forwarding rule: ", cmd.Args, " Returned exit error: ", err.Error(), " Cmd output: ", string(output))
-			return err
-		}
+	service.mu.Lock()
+	defer service.mu.Unlock()
 
-		log.Info(natLogPrefix, "Forwarding packets from '", rule.SourceAddress, "' to IP: ", rule.TargetIP)
+	for rule := range service.rules {
+		if delErr := iptables("delete", rule); delErr != nil && err == nil {
+			err = delErr
```
vkuznecovas: we're losing error information here, if we have multiple rules and removing all of them fails, all we'll see is the first error encountered. I do realize that in most cases all those will be the same, but we should not ignore them nonetheless.

soffokl: We are doing logging for this error in the `iptables()` function, so we will have all errors logged. I don't think that we should return all errors that occurred. It's enough to return only one error to indicate that some problems happened.

PR #687, opened 2019-01-29
use ethereum port mapping implementation for openvpn port mapping

zolia:
Analysed ethereum project port mapping implementation and since it uses the same huin/goupnp library as we were planning to use and since their methods are exported I decided to reuse their code.

Whats implemented:
- on openvpn service startup launches port mapper, which tries to maintain service port mapping on a gateway 
- on openvpn service shutdown, mapper is terminated and rule on a gateway removed
- since ethereum does not handle the case when a gateway implements only persistent rule handling, rewrote their nat.Map method.
- launch mapper only if node being behind the NAT is detected.
 
Whats lacking:
- WG implementation. Will make it with separate PR.

### DIFF
--- a/nat/mapping/port_mapping.go
+++ b/nat/mapping/port_mapping.go
@@ -33,7 +33,7 @@ const (
 
 // PortMapping maps given port of given protocol from external IP on a gateway to local machine internal IP
 // 'name' denotes rule name added on a gateway.
-func PortMapping(protocol string, port int, name string) chan struct{} {
+func PortMapping(protocol string, port int, name string) func() {
 	mapperQuit := make(chan struct{})
 	go mapPort(portmap.Any(),
 		mapperQuit,
@@ -42,7 +42,7 @@ func PortMapping(protocol string, port int, name string) chan struct{} {
 		port,
 		name)
 
-	return mapperQuit
+	return func() { close(mapperQuit) }
 }
 
 // mapPort adds a port mapping on m and keeps it alive until c is closed.
--- a/nat/mapping/port_mapping.go
+++ b/nat/mapping/port_mapping.go
@@ -1,5 +1,3 @@
-package nat
-
 /*
  * Copyright (C) 2019 The "MysteriumNetwork/node" Authors.
  *
@@ -17,6 +15,8 @@ package nat
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+package mapping
+
 import (
 	"time"
 
@@ -24,7 +24,7 @@ import (
 	portmap "github.com/ethereum/go-ethereum/p2p/nat"
 )
 
-const logPrefix = "[nat] "
+const logPrefix = "[port mapping] "
 
 const (
 	mapTimeout        = 20 * time.Minute
@@ -51,20 +51,13 @@ func mapPort(m portmap.Interface, c chan struct{}, protocol string, extPort, int
 	refresh := time.NewTimer(mapUpdateInterval)
 	defer func() {
 		refresh.Stop()
-		log.Debug(logPrefix, "Deleting port mapping")
-		m.DeleteMapping(protocol, extPort, intPort)
-	}()
-	if err := m.AddMapping(protocol, extPort, intPort, name, mapTimeout); err != nil {
-		log.Debugf("%s, Couldn't add port mapping: %v, retrying with permanent lease", logPrefix, err)
-		if err := m.AddMapping(protocol, extPort, intPort, name, 0); err != nil {
-			// some gateways support only permanent leases
-			log.Debug(logPrefix, "Couldn't add port mapping: ", err)
-		} else {
-			log.Info(logPrefix, "Mapped network port: ", extPort)
+		log.Debug(logPrefix, "Deleting port mapping for port: ", extPort)
+
+		if err := m.DeleteMapping(protocol, extPort, intPort); err != nil {
+			log.Debug(logPrefix, "Couldn't delete port mapping: ", err)
 		}
-	} else {
-		log.Info(logPrefix, "Mapped network port")
-	}
+	}()
+	addMapping(m, protocol, extPort, intPort, name)
 	for {
 		select {
 		case _, ok := <-c:
@@ -72,15 +65,22 @@ func mapPort(m portmap.Interface, c chan struct{}, protocol string, extPort, int
 				return
 			}
 		case <-refresh.C:
-			log.Trace(logPrefix, "Refreshing port mapping")
-			if err := m.AddMapping(protocol, extPort, intPort, name, mapTimeout); err != nil {
-				log.Debugf("%s, Couldn't add port mapping: %v, retrying with permanent lease", logPrefix, err)
-				if err := m.AddMapping(protocol, extPort, intPort, name, 0); err != nil {
-					// some gateways support only permanent leases
-					log.Debug(logPrefix, "Couldn't add port mapping: ", err)
-				}
-			}
+			addMapping(m, protocol, extPort, intPort, name)
 			refresh.Reset(mapUpdateInterval)
 		}
+	}
+}
+
+func addMapping(m portmap.Interface, protocol string, extPort, intPort int, name string) {
+	if err := m.AddMapping(protocol, extPort, intPort, name, mapTimeout); err != nil {
+		log.Debugf("%s, Couldn't add port mapping for port %d: %v, retrying with permanent lease", logPrefix, extPort, err)
+		if err := m.AddMapping(protocol, extPort, intPort, name, 0); err != nil {
+			// some gateways support only permanent leases
+			log.Debugf("%s Couldn't add port mapping for port %d: %v", logPrefix, extPort, err)
+		} else {
+			log.Info(logPrefix, "Mapped network port: ", extPort)
+		}
+	} else {
+		log.Info(logPrefix, "Mapped network port:", extPort)
 	}
 }
--- a/nat/mapping/port_mapping.go
+++ b/nat/mapping/port_mapping.go
@@ -48,25 +48,19 @@ func PortMapping(protocol string, port int, name string) func() {
 // mapPort adds a port mapping on m and keeps it alive until c is closed.
 // This function is typically invoked in its own goroutine.
 func mapPort(m portmap.Interface, c chan struct{}, protocol string, extPort, intPort int, name string) {
-	refresh := time.NewTimer(mapUpdateInterval)
 	defer func() {
-		refresh.Stop()
 		log.Debug(logPrefix, "Deleting port mapping for port: ", extPort)
 
 		if err := m.DeleteMapping(protocol, extPort, intPort); err != nil {
 			log.Debug(logPrefix, "Couldn't delete port mapping: ", err)
 		}
 	}()
-	addMapping(m, protocol, extPort, intPort, name)
 	for {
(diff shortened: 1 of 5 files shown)


nat/port_mapping.go
```diff
@@ -0,0 +1,86 @@
...
+ */
+
+import (
+	"time"
+
+	log "github.com/cihub/seelog"
+	portmap "github.com/ethereum/go-ethereum/p2p/nat"
+)
+
+const logPrefix = "[nat] "
```
soffokl: We already have `natLogPrefix` defined in the package, with the same value.

nat/port_mapping.go
```diff
@@ -0,0 +1,86 @@
...
+
+const (
+	mapTimeout        = 20 * time.Minute
+	mapUpdateInterval = 15 * time.Minute
+)
+
+// PortMapping maps given port of given protocol from external IP on a gateway to local machine internal IP
+//  'name' denotes rule name added on a gateway
```
soffokl:
```suggestion
// 'name' denotes rule name added on a gateway.
```

nat/port_mapping.go
```diff
@@ -0,0 +1,86 @@
...
+	return mapperQuit
+}
+
+// mapPort adds a port mapping on m and keeps it alive until c is closed.
+// This function is typically invoked in its own goroutine.
+func mapPort(m portmap.Interface, c chan struct{}, protocol string, extPort, intPort int, name string) {
+	refresh := time.NewTimer(mapUpdateInterval)
+	defer func() {
+		refresh.Stop()
+		log.Debug(logPrefix, "Deleting port mapping")
```
soffokl: It could be very useful to specify `name` and `extPort:intPort` in the log to see what was deleted.

nat/port_mapping.go
```diff
@@ -0,0 +1,86 @@
...
+		m.DeleteMapping(protocol, extPort, intPort)
```
soffokl: `DeleteMapping` returns an error, it could be helpful to lot it in case of problems.

nat/port_mapping.go
```diff
@@ -0,0 +1,86 @@
...
+	}()
+	if err := m.AddMapping(protocol, extPort, intPort, name, mapTimeout); err != nil {
+		log.Debugf("%s, Couldn't add port mapping: %v, %s", logPrefix, err, "retrying with permanent lease")
```
soffokl:
```suggestion
		log.Debugf("%s, Couldn't add port mapping: %v, retrying with permanent lease", logPrefix, err)
```

nat/port_mapping.go
```diff
@@ -0,0 +1,86 @@
...
+		if err := m.AddMapping(protocol, extPort, intPort, name, 0); err != nil {
+			// some gateways support only permanent leases
+			log.Debug(logPrefix, "Couldn't add port mapping: ", err)
```
soffokl: It could be very useful to specify `name` and `extPort:intPort` in the log to see what port was requested.

nat/port_mapping.go
```diff
@@ -0,0 +1,86 @@
...
+		} else {
+			log.Info(logPrefix, "Mapped network port: ", extPort)
+		}
+	} else {
+		log.Info(logPrefix, "Mapped network port")
```
soffokl: It could be very useful to specify `name` and `extPort:intPort` in the log to see what port was requested.

nat/port_mapping.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+	for {
+		select {
+		case _, ok := <-c:
+			if !ok {
+				return
+			}
+		case <-refresh.C:
```
soffokl: It looks like [time.After()](https://golang.org/pkg/time/#After) could be a simpler solution here. 
Is there any specific case for using exactly `time.NewTimer()`?

nat/port_mapping.go
```diff
@@ -0,0 +1,86 @@
...
+			log.Trace(logPrefix, "Refreshing port mapping")
```
soffokl: It could be very useful to specify `name` and `extPort:intPort` in the log to see what port was requested.

nat/port_mapping.go
```diff
@@ -0,0 +1,86 @@
...
+			if err := m.AddMapping(protocol, extPort, intPort, name, mapTimeout); err != nil {
```
soffokl: This block exactly repeat the same block on mapping creation, maybe extract it into own function?

nat/port_mapping.go
```diff
@@ -0,0 +1,86 @@
...
+				log.Debugf("%s, Couldn't add port mapping: %v, %s", logPrefix, err, "retrying with permanent lease")
```
soffokl:
```suggestion
				log.Debugf("%s, Couldn't add port mapping: %v, retrying with permanent lease", logPrefix, err)
```

nat/port_mapping.go
```diff
@@ -0,0 +1,86 @@
...
+				if err := m.AddMapping(protocol, extPort, intPort, name, 0); err != nil {
+					// some gateways support only permanent leases
+					log.Debug(logPrefix, "Couldn't add port mapping: ", err)
```
soffokl: It could be very useful to specify name and extPort:intPort in the log to see what port was requested.

services/openvpn/service/manager.go
```diff
@@ -28,6 +28,7 @@ import (
 	"github.com/mysteriumnetwork/node/nat"
 	openvpn_service "github.com/mysteriumnetwork/node/services/openvpn"
 	"github.com/mysteriumnetwork/node/session"
+
```
soffokl: Looks like an unnecessary extra line.

services/openvpn/service/manager.go
```diff
@@ -48,6 +49,7 @@ type SessionConfigNegotiatorFactory func(secPrimitives *tls.Primitives, outbound
 // Manager represents entrypoint for Openvpn service with top level components
 type Manager struct {
 	natService nat.NATService
+	mapperQuit chan struct{}
```
soffokl: I'm not sure if the transport service should be aware of the port mapper. Maybe it should be handled on some upper layer?

tadovas: Actualy instead of passing something like port mapping, what about exposing interface where service can request port pair allocation - one for passing to consumer (in session config) another for binding transport itself. It can then close or release that port pair, notifying port mapper that's it's no longer needed. Bonus points - in case of binding on public ip you can pass noop implemention - which simply always returns preconfigured ports.

nat/port_mapping.go
zolia: added handling.

nat/port_mapping.go
zolia: true. extracted.

nat/port_mapping.go
zolia: true, but moved to own package.

nat/port_mapping.go
zolia: added

services/openvpn/service/manager.go
zolia: I don't see a point to abuse DI and lift it somewhere else since it is not stand alone service as such, but maintains mapping just for concrete service and such maintenance goes down together with a service. I really don't think that there is a point making it self sustained service, lets leave it as simple as possible.

tadovas: The only point here - hide implementation details. If services needs publically accesible port - just give it to him, nothing more nothing less

zolia: As per @tadovas suggestion. I do agree that we need dynamic port handling. That is service should request some randomly generated port from local resource pool. But this relates to many parts of the code and imho should not be addressed in this PR. Created #689 ticket for this.

nat/mapping/port_mapping.go
```diff
@@ -0,0 +1,86 @@
...
+const logPrefix = "[port mapping] "
+
+const (
+	mapTimeout        = 20 * time.Minute
+	mapUpdateInterval = 15 * time.Minute
+)
+
+// PortMapping maps given port of given protocol from external IP on a gateway to local machine internal IP
+// 'name' denotes rule name added on a gateway.
+func PortMapping(protocol string, port int, name string) chan struct{} {
```
tadovas: I would return a function instead of channel here - it's used to release resources anyway. port allocation in style
releasePorts := PortMapping(...)
... service runs...
releasePorts();
looks more readable

services/openvpn/service/manager.go
```diff
@@ -89,11 +99,14 @@ func (manager *Manager) Serve(providerID identity.Identity) (err error) {
 	if err = manager.vpnServer.Start(); err != nil {
 		return
 	}
+
 	return manager.vpnServer.Wait()
 }
 
 // Stop stops service
 func (manager *Manager) Stop() (err error) {
+	close(manager.mapperQuit)
```
tadovas: See my comment on "return function instead of channel"

nat/mapping/port_mapping.go
zolia: true, this looks a bit better. Changed.

nat/mapping/port_mapping.go
```diff
@@ -0,0 +1,86 @@
...
+	defer func() {
+		refresh.Stop()
+		log.Debug(logPrefix, "Deleting port mapping for port: ", extPort)
+
+		if err := m.DeleteMapping(protocol, extPort, intPort); err != nil {
+			log.Debug(logPrefix, "Couldn't delete port mapping: ", err)
+		}
+	}()
+	addMapping(m, protocol, extPort, intPort, name)
+	for {
```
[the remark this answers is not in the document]
soffokl: so, my suggestion to make it like this: 

```golang
	for {
		addMapping(m, protocol, extPort, intPort, name)
		select {
		case <-c:
			return
		case <-time.After(mapUpdateInterval):
		}
	}
```

nat/mapping/port_mapping.go
```diff
@@ -0,0 +1,86 @@
...
+		select {
+		case _, ok := <-c:
+			if !ok {
+				return
+			}
+		case <-refresh.C:
```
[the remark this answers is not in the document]
zolia: That was partial code that Ethereum used. I thought their concern was that this timer is being not garbage collected, but from what I understand it should cleanup after each timer hit, isn't it? 
Rewrote according your suggestion.

nat/mapping/port_mapping.go
```diff
@@ -0,0 +1,86 @@
...
+			addMapping(m, protocol, extPort, intPort, name)
+			refresh.Reset(mapUpdateInterval)
+		}
+	}
+}
+
+func addMapping(m portmap.Interface, protocol string, extPort, intPort int, name string) {
```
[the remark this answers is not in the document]
zolia: true, reduced.

nat/mapping/port_mapping.go
```diff
@@ -0,0 +1,78 @@
...
+			log.Debug(logPrefix, "Couldn't delete port mapping: ", err)
+		}
+	}()
+	for {
+		addMapping(m, protocol, extPort, intPort, name)
+		select {
+		case <-c:
+			return
+		case <-time.After(mapUpdateInterval):
+			addMapping(m, protocol, extPort, intPort, name)
```
soffokl: This line can be removed since we already have the same line at the beginning of the loop.
No need to do it twice.

zolia: true, missed it somehow. thanx.

PR #692, opened 2019-01-31
Feature/basic value transfer to provider skeleton

vkuznecovas:
I know it's quite a large one. Here's the gist of it:

Consumer initiates session creation. The session creation request message is expanded with consumer information, providing the consumer myst version as well as the issuerID. 

The issuerID represents the identity of the party that will sign the promises, this can be different from the consumer - we can hide the hard parts of payments this way).

The version information was included just as a bonus and nothing is being checked yet, but in the future we might want to check for compatibility issues.

The provider then validates that this issuerID is in fact valid(this step is not implemented in this PR). It responds to the session creation request by sending an extended CreateResponse object. This object is extended with PaymentInfo object. The PaymentInfo contains the LastPromise and FreeCredit fields.

FreeCredit represents the amount of currency the provider gives for free. This will be set to a small amount, to allow for a few moments of free usage for successful session establishment.

The LastPromise object contains two fields: SequenceID and Amount. The Sequence is the ID of the last promise the provider is aware of for this consumer. The amount represents the amount promised by that promise.

After this, the session is established.

The provider then starts sending a balance message to the consumer via `session-balance` topic. The message is sent periodically. These are fire and forget requests, so no response is expected for the message. 

The session balance consists of two fields: balance and sequenceID. The balance field represents the amount of currently available balance that the consumer is allowed to consume. The sequenceID represents the sequenceID of the current promise. 

After sending a session balance message, the provider expects the consumer to reply with a promise message within a given time frame. If the consumer does not send the message, the session is terminated.

The consumer checks the balance that it received. If it's happy with it - it sends a promise message to the consumer via `session-promise` topic.

If he's not happy with what the provider sends - he terminates the session. Take note that the verification steps are not implemented in this PR nor for provider, nor for consumer. 

The message that's sent via the `session-promise` topic has Amount, SequenceID and Signature fields for the issued promise.

When the balance is running low, the consumer will send a promise with an increased amount of balance for the provider. 

The provider validates the signature, saves the promise(storage mechanism is not implemented yet) and increases the balance. The provider will also decide when to request for a new promise. This is indicated by an increase of the sequenceID in the balance message. 

This PR is not a full solution by any means, but it is a beginning.


core/connection/stubs_test.go
```diff
@@ -231,6 +209,8 @@ func (foc *connectionMock) StateCallback(callback func(state fakeState)) {
 	foc.stateCallback = callback
 }
 
+const fakeDialogLog = "[fake dialog] "
```
soffokl: Do we really need to have a log in tests?

e2e/wallet_helper.go
```diff
@@ -180,7 +180,7 @@ func newCliWallet(owner common.Address, passphrase string, ks *keystore.KeyStore
 
 	tokensContract, err := mysttoken.NewMystTokenTransactor(tokenAddress, ehtClient)
```
soffokl: This `err` looks unchecked, we probably need to check it if it's important.

session/balance/balance_messaging.go
```diff
@@ -0,0 +1,119 @@
...
+
+const endpointBalance = "session-balance"
+const messageEndpointBalance = communication.MessageEndpoint(endpointBalance)
+
+// Sender is responsible for sending the balance messages
+type Sender struct {
+	sender communication.Sender
+}
+
+// NewBalanceSender returns a new instance of the balance sennder
```
soffokl:
```suggestion
// NewBalanceSender returns a new instance of the balance sender
```

session/balance/balance_messaging.go
```diff
@@ -0,0 +1,119 @@
...
+func NewBalanceSender(sender communication.Sender) *Sender {
+	return &Sender{
+		sender: sender,
+	}
+}
+
+// Send sends the given balance message
+func (bs *Sender) Send(bm Message) error {
+	err := bs.sender.Send(&balanceMessageProducer{BalanceMessage: bm})
```
soffokl:
```suggestion
	return bs.sender.Send(&balanceMessageProducer{BalanceMessage: bm})
```

session/balance/balance_tracker.go
```diff
@@ -0,0 +1,81 @@
...
+
+package balance
+
+import (
+	"time"
+
+	"github.com/mysteriumnetwork/node/money"
+)
+
+const balanceTrackerPrefix = "[balance-tracker] "
```
soffokl: `balanceTrackerPrefix` is unused.

session/balance/balance_tracker.go
```diff
@@ -0,0 +1,81 @@
...
+}
+
+// AmountCalculator is able to deduce the amount required for payment from a given duraiton
+type AmountCalculator interface {
+	TotalAmount(duration time.Duration) money.Money
+}
+
+// ProviderBalanceTracker is responsible for tracking the balance on the provider side
+type ProviderBalanceTracker struct {
+	sender           PeerSender
```
soffokl: `sender` is unused.

session/balance/balance_tracker.go
```diff
@@ -0,0 +1,81 @@
...
+	Send(Message) error
+}
+
+// TimeKeeper keeps track of time for payments
+type TimeKeeper interface {
+	StartTracking()
+	Elapsed() time.Duration
+}
+
+// AmountCalculator is able to deduce the amount required for payment from a given duraiton
```
soffokl:
```suggestion
// AmountCalculator is able to deduce the amount required for payment from a given duration
```

session/create_dto.go
```diff
@@ -32,19 +33,41 @@ var (
 
 // CreateRequest structure represents message from service consumer to initiate session for given proposal id
 type CreateRequest struct {
-	ProposalId int             `json:"proposal_id"`
-	Config     json.RawMessage `json:"config"`
+	ProposalId   int             `json:"proposal_id"`
```
soffokl: > struct field ProposalId should be ProposalID (golint)

session/dto.go
```diff
@@ -22,12 +22,18 @@ import "github.com/mysteriumnetwork/node/identity"
...
 // Session structure holds all required information about current session between service consumer and provider
 type Session struct {
-	ID         ID
-	Config     ServiceConfiguration
-	ConsumerID identity.Identity
-	Done       chan struct{}
+	ID              ID
+	ConsumerID      identity.Identity
+	Done            chan struct{}
+	DestroyCallback DestroyCallback
```
soffokl: Are we having a `DestroyCallback` back in Session struct?

session/manager_test.go
```diff
@@ -30,11 +30,12 @@ var (
...
+	consumerID = identity.FromAddress("deadbeef")
+
 	expectedID      = ID("mocked-id")
 	expectedSession = Session{
 		ID:         expectedID,
-		Config:     expectedSessionConfig,
-		ConsumerID: identity.FromAddress("deadbeef"),
+		ConsumerID: consumerID,
 	}
 	lastSession Session
```
soffokl: This `lastSession` is unused can we remove it in this PR? 
`const expectedSessionConfig = "config_string"` is unused too.

session/payment/noop/noop_session_balance.go
```diff
@@ -0,0 +1,46 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+// Package noop contains the non operational payment and balance trackers
+package noop
+
+// SessionBalance doesnt really track the balance
```
soffokl:
```suggestion
// SessionBalance doesn't really track the balance 
```

session/payment/noop/noop_session_balance.go
```diff
@@ -0,0 +1,46 @@
...
+
+// NewSessionBalance returns a noop session balance tracker
+func NewSessionBalance() *SessionBalance {
+	return &SessionBalance{
+		stopChan: make(chan struct{}),
+	}
+}
+
+// Start starts the noop session balance tracker
+func (nsb *SessionBalance) Start() error {
```
soffokl: A body of the func could be simple: 
```golang
<-nsb.stopChan
return nil
```

session/payment/session_balance.go
```diff
@@ -0,0 +1,124 @@
...
+	Validate(promise.Message) bool
+}
+
+// PeerBalanceSender knows how to send a balance message to the peer
+type PeerBalanceSender interface {
+	Send(balance.Message) error
+}
+
+// ErrPromiseWaitTimeout indicates that we waited for a promise long enough, but with no result
+var ErrPromiseWaitTimeout = errors.New("Did not get a new promise")
```
soffokl: [Error strings should not be capitalized...](https://github.com/golang/go/wiki/CodeReviewComments#error-strings)
```suggestion
var ErrPromiseWaitTimeout = errors.New("did not get a new promise")
```

session/payment/session_balance.go
```diff
@@ -0,0 +1,124 @@
...
+
+// ErrPromiseValidationFailed indicates that an invalid promise was sent
+var ErrPromiseValidationFailed = errors.New("Promise validation failed")
```
soffokl: [Error strings should not be capitalized...](https://github.com/golang/go/wiki/CodeReviewComments#error-strings)
```suggestion
var ErrPromiseValidationFailed = errors.New("promise validation failed")
```

session/payment/session_balance.go
```diff
@@ -0,0 +1,124 @@
...
+// NewSessionBalance creates a new instance of provider payment orchestrator
+func NewSessionBalance(
+	peerBalanceSender PeerBalanceSender,
+	balanceTracker BalanceTracker,
+	promiseChan chan promise.Message,
+	period time.Duration,
+	promiseWaitTimeout time.Duration,
+	promiseValidator PromiseValidator) *SessionBalance {
+	return &SessionBalance{
+		stop:               make(chan struct{}, 1),
```
soffokl: Why do we need a buffered `stop` channel?

session/payment/session_balance.go
```diff
@@ -0,0 +1,124 @@
...
+				return err
+			}
+		}
+	}
+}
+
+func (ppo *SessionBalance) sendBalance() error {
+	balance := ppo.balanceTracker.GetBalance()
+	// TODO: Maybe retry a couple of times?
+	err := ppo.peerBalanceSender.Send(balance)
```
soffokl:
```suggestion
	return ppo.peerBalanceSender.Send(balance)
```

session/payment/session_balance.go
```diff
@@ -0,0 +1,124 @@
...
+		// TODO: Change balance
+	case <-time.After(ppo.promiseWaitTimeout):
+		return ErrPromiseWaitTimeout
+	}
+	return nil
+}
+
+// Stop stops the payment orchestrator
+func (ppo *SessionBalance) Stop() {
+	ppo.stop <- struct{}{}
```
soffokl: Maybe just close the channel?

session/payment/session_balance_test.go
```diff
@@ -0,0 +1,125 @@
...
+		orch.Stop()
+	}()
+	err := orch.Start()
+	assert.Nil(t, err)
+}
+
+func Test_ProviderPaymentOchestratorSendsBalance(t *testing.T) {
+	orch := NewMockSessionBalance()
+	defer orch.Stop()
+	go func() { _ = orch.Start() }()
```
soffokl:
```suggestion
	go orch.Start()
```

session/payment/session_payments.go
```diff
@@ -0,0 +1,93 @@
...
+
+// SessionPayments orchestrates the ping pong of balance received from provider -> promise sent to provider flow
+type SessionPayments struct {
+	stop              chan struct{}
+	balanceChan       chan balance.Message
+	peerPromiseSender PeerPromiseSender
+	promiseTracker    PromiseTracker
+}
+
+// NewSessionPayments returns a new instnace of consumer payment orchestrator
```
soffokl:
```suggestion
// NewSessionPayments returns a new instance of consumer payment orchestrator 
```

session/payment/session_payments.go
```diff
@@ -0,0 +1,93 @@
...
+func NewSessionPayments(balanceChan chan balance.Message, peerPromiseSender PeerPromiseSender, promiseTracker PromiseTracker) *SessionPayments {
+	return &SessionPayments{
+		stop:              make(chan struct{}, 1),
```
soffokl: Another buffered `stop` channel.

session/payment/session_payments.go
```diff
@@ -0,0 +1,93 @@
...
+			if err != nil {
+				return err
+			}
+		}
+	}
+}
+
+// Stop stops the payment orchestrator
+func (cpo *SessionPayments) Stop() {
+	cpo.stop <- struct{}{}
```
soffokl: We can close the channel here too.

session/payment/session_payments_test.go
```diff
@@ -0,0 +1,125 @@
...
+		time.Sleep(time.Nanosecond * 10)
+		cpo.Stop()
+	}()
+	err := cpo.Start()
+	assert.Nil(t, err)
+}
+
+func Test_SessionPayments_SendsPromiseOnBalance(t *testing.T) {
+	cpo := NewTestSessionPayments(balanceChannel, promiseSender, promiseTracker)
+	go func() { cpo.Start() }()
```
soffokl:
```suggestion
	go cpo.Start()
```

session/promise/issued_promise_tracker_test.go
```diff
@@ -0,0 +1,102 @@
...
+
+package promise
+
+import (
+	"testing"
+
+	"github.com/ethereum/go-ethereum/common"
+
+	"github.com/stretchr/testify/assert"
+
```
soffokl: Some strange formatting with empty lines.

session/promise/issuers/keystoreissuer.go
```diff
@@ -0,0 +1,72 @@
...
+// NewLocalIssuer creates local issuer based on provided identity signer
+func NewLocalIssuer(signer identity.Signer) *LocalIssuer {
+	return &LocalIssuer{
+		paymentsSigner: paymentsSignerAdapter{
+			identitySigner: signer,
+		},
+	}
+}
+
+// Issue issues thet given promise
```
soffokl:
```suggestion
// Issue issues that given promise
```

session/promise/issuers/keystoreissuer.go
```diff
@@ -0,0 +1,72 @@
...
+var _ promise.Issuer = LocalIssuer{}
+
+// this is ugly adapter to make identity.Signer from node usable in payments package
+// it's a bit confusing as both interfaces has the same name and method, but only params and return values differ
+type paymentsSignerAdapter struct {
+	identitySigner identity.Signer
+}
+
+func (psa paymentsSignerAdapter) Sign(data ...[]byte) ([]byte, error) {
+	var message []byte
```
soffokl: We can preallocate it with a `len(data)` to make it a bit faster. Not sure if we should worry about performance here.

session/promise/promise_messaging.go
```diff
@@ -0,0 +1,120 @@
...
+// NewSender returns a new instance of promise sender
+func NewSender(sender communication.Sender) *Sender {
+	return &Sender{
+		sender: sender,
+	}
+}
+
+// Send send the given promise message
+func (ps *Sender) Send(pm Message) error {
+	err := ps.sender.Send(&MessageProducer{Message: pm})
```
soffokl:
```suggestion
	return ps.sender.Send(&MessageProducer{Message: pm})
```

session/promise/validators/promise_signature_validator_test.go
```diff
@@ -0,0 +1,71 @@
...
+ */
+
+package validators
+
+import (
+	"testing"
+
+	"github.com/ethereum/go-ethereum/common/hexutil"
+	"github.com/stretchr/testify/assert"
+
```
soffokl: Extra empty line.

session/promise/validators/promise_signature_validator_test.go
```diff
@@ -0,0 +1,71 @@
...
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/mysteriumnetwork/node/session/promise"
+	"github.com/mysteriumnetwork/payments/promises"
+	"github.com/mysteriumnetwork/payments/test_utils"
+
```
soffokl: Extra empty line.

session/time_tracker.go
```diff
@@ -0,0 +1,49 @@
...
+
+// StartTracking starts tracking the time
+func (tt *TimeTracker) StartTracking() {
+	tt.started = true
+	tt.startTime = tt.getTime()
+}
+
+// Elapsed gets the total duration of time that has passed since we've started
+func (tt TimeTracker) Elapsed() time.Duration {
+	if !tt.started {
```
soffokl: We avoid using `started` field. 
We can check [time.IsZero()](https://golang.org/pkg/time/#Time.IsZero) here: 
```suggestion
	if tt.startTime.IsZero() {
```

cmd/di.go
```diff
@@ -288,14 +309,39 @@ func (di *Dependencies) bootstrapNodeComponents(nodeOptions node.Options) {
...
 	return func(dialog communication.Dialog) *session.Manager {
+		providerBalanceTrackerFactory := func(consumer, provider, issuer identity.Identity) session.BalanceTracker {
+			// if the flag aint set, just return a noop balance tracker
+			if !nodeOptions.ExperimentPayments {
+				return payments_noop.NewSessionBalance()
+			}
+
+			timeTracker := session.NewTracker(time.Now)
+			// TODO: set the time and proper payment info
+			payment := dto.PaymentPerTime{
```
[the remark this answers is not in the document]
vkuznecovas: This will probably have to be per service. Will figure this out at a later date.

cmd/di.go
```diff
@@ -288,14 +309,39 @@ func (di *Dependencies) bootstrapNodeComponents(nodeOptions node.Options) {
...
+				},
+				Duration: time.Minute,
+			}
+			amountCalc := session.AmountCalc{PaymentDef: payment}
+			sender := balance.NewBalanceSender(dialog)
+			promiseChan := make(chan promise.Message, 1)
+			listener := promise.NewListener(promiseChan)
+			dialog.Receive(listener.GetConsumer())
+
+			tracker := balance.NewProviderBalanceTracker(&timeTracker, amountCalc, time.Second*5, 100)
```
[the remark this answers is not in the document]
vkuznecovas: It's not a real value yet, we're just setting it here for testing. Will have to fix this as well in the future. Will add a TODO.

core/connection/stubs_test.go
vkuznecovas: Makes it easier to catch issues in e2e tests.

session/dto.go
vkuznecovas: rebased a few times, so it must have snook back in somehow

session/promise/issuers/keystoreissuer.go
vkuznecovas: Don't think you're gonna be issuing promises at lightning speed.

session/time_tracker.go
tadovas: Actually there are slight differences what time IsZero actually reports and what we rely on. More about that - golang monotonic vs wall clock on google. And here I prefer to have explicit flag for two main reasons:
1. It shows intent or reason more clearly - return zero duration if not started
2. IsZero function doesn't say anything about monotonic clock values - and we heavily rely on that

tadovas: Update: It's probably save to use IsZero here, as it also checks nsec part of monotonic clock - however there are no guarantees it will change in the future. Anyway I would stick for first argument. @vkuznecovas its your call

cmd/di.go
```diff
@@ -244,11 +251,24 @@ func (di *Dependencies) bootstrapNodeComponents(nodeOptions node.Options) {
 		return dialogEstablisher.EstablishDialog(providerID, contact)
 	}
 
-	promiseIssuerFactory := func(issuerID identity.Identity, dialog communication.Dialog) connection.PromiseIssuer {
-		if nodeOptions.ExperimentPromiseCheck {
-			return promise_noop.NewPromiseIssuer(issuerID, dialog, di.SignerFactory(issuerID))
+	paymentIssuerFactory := func(
```
zolia: We should have a separate NewPaymentIssuerFactory() method, not to clutter tidy bootstrapNodeComponents() function.

cmd/di.go
```diff
@@ -244,11 +251,24 @@ func (di *Dependencies) bootstrapNodeComponents(nodeOptions node.Options) {
...
 		}
-		return &promise_noop.FakePromiseEngine{}
+
+		bl := balance.NewListener(messageChan)
+		ps := promise.NewSender(dialog)
+		issuer := issuers.NewLocalIssuer(di.SignerFactory(consumer))
+		tracker := promise.NewConsumerTracker(initialState, consumer, provider, issuer)
+		payments := session_payment.NewSessionPayments(messageChan, ps, tracker)
+		err := dialog.Receive(bl.GetConsumer())
+		return payments, err
```
zolia: we could use errors.Wrap(err, "fail to receive from consumer") or something like that.

cmd/di.go
```diff
@@ -266,7 +286,7 @@ func (di *Dependencies) bootstrapNodeComponents(nodeOptions node.Options) {
 	di.ConnectionRegistry = connection.NewRegistry()
 	di.ConnectionManager = connection.NewManager(
 		dialogFactory,
-		promiseIssuerFactory,
+		paymentIssuerFactory,
```
zolia: why 'promiseIssuerFactory' is bad?

cmd/di.go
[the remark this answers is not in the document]
zolia: We either should open a separate ticket with TODO things or make a stub here with pending TODO descriptions.

session/amount_calculator.go
```diff
@@ -0,0 +1,44 @@
...
+type AmountCalc struct {
+	PaymentDef dto.PaymentPerTime
+}
+
+// TotalAmount gets the total amount of money to pay given the duration
+func (ac AmountCalc) TotalAmount(duration time.Duration) money.Money {
+	// time.Duration holds info in nanoseconds internally anyway (with max duration of 290 years) so we are probably safe here
+	// however - careful testing of corner cases is needed
+	// another question - in case of amount of 15 seconds, and price 10 myst per minute, total amount will be rounded to zero
+	// add 1 in case it's bad
```
zolia: Yes, there should always be some minimal sum to charge. Even if you used 0.000001 of a cent, 1 cent should be charged.

zolia: There should be some notion of minimal charge unit.

core/connection/manager.go
```diff
@@ -157,7 +169,25 @@ func (manager *connectionManager) startConnection(consumerID identity.Identity,
 		return err
 	}
 
-	sessionID, sessionConfig, err := session.RequestSessionCreate(dialog, proposal.ID, sessionCreateConfig)
+	messageChan := make(chan balance.Message, 1)
+
+	// TODO: load initial promise state
```
zolia: We could have a separate ticket for that, we could even link that in such TODO comment.

cmd/di.go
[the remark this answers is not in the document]
vkuznecovas: I'll start opening tickets under the promise engine epic

core/connection/manager.go
```diff
@@ -61,12 +64,21 @@ type Publisher interface {
 	Publish(topic string, args ...interface{})
 }
 
+// PaymentManager handles the payments for service
+type PaymentManager interface {
```
zolia: Imho too abstract naming. It would be good, that interface naming we set could be easily relatable with original structure they represent. Without this relation is very hard to read code. Imho original naming with 'PaymentIssuer' or 'PromiseIssuer' would match better. Would be best if I search for 'NamedInterfaceAtUserSide' and I am able to find original struct where this interface is implemented. Not to traverse 100 dependency injections and look for it.

session/amount_calculator_test.go
```diff
@@ -0,0 +1,45 @@
...
+		PaymentDef: dto.PaymentPerTime{
+			Duration: time.Minute,
+			Price: money.Money{
+				Amount:   100,
+				Currency: money.CURRENCY_MYST,
+			},
+		},
+	}
+
+	elapsed := time.Minute*3 + time.Second*15
```
zolia:
```suggestion
	elapsed := 3*time.Minute + 15*time.Second
```
Write as you read..

session/amount_calculator_test.go
```diff
@@ -0,0 +1,45 @@
...
+
+	totalAmount := aCalc.TotalAmount(elapsed)
+
+	assert.Equal(t, uint64(300), totalAmount.Amount)
```
zolia: Do I understand correctly, that we have 100 MYST / min price here, and duration of 3 min and 15 sec? Why we expect 300 price and not 325 ?

tadovas: Because we are calcalting price per minute, not per second and these seconds a dropped in favour or consumer.

soffokl: Can this be abused by the user by stopping session every 59 seconds, and starting a new one?

vkuznecovas: In theory - yes. In reality, you'll be asked for a promise rather quickly by the provider and he'll be killing the connection if you don't provide one. You could get a few seconds for free, probably.

tadovas: Not actually - as consumer is required to pay in advance to keep balance positive

cmd/di.go
vkuznecovas: moved

cmd/di.go
vkuznecovas: mainly wanted to illustrate the fact that it's not only issuing promises as it will also have to keep track of balance, etc.

session/balance/balance_tracker.go
```diff
@@ -0,0 +1,78 @@
...
+	Elapsed() time.Duration
+}
+
+// AmountCalculator is able to deduce the amount required for payment from a given duration
+type AmountCalculator interface {
+	TotalAmount(duration time.Duration) money.Money
+}
+
+// ProviderBalanceTracker is responsible for tracking the balance on the provider side
+type ProviderBalanceTracker struct {
```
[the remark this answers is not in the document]
vkuznecovas: Moved to a provider package

core/connection/manager.go
vkuznecovas: renamed to payment issuer

session/amount_calculator_test.go
zolia: So the actual elapsed time cannot cross billing unit boundary? In this case 3 min.

session/create_dto.go
```diff
@@ -32,19 +33,41 @@ var (
...
 
 // SessionDto structure represents session information data within session creation response (session id and configuration options for underlying service type)
 type SessionDto struct {
 	ID     ID              `json:"id"`
 	Config json.RawMessage `json:"config"`
 }
+
+// ConsumerInfo represents the consumer related information
+type ConsumerInfo struct {
+	MystClientVersion string            `json:"mystClientVersion"`
```
zolia: Cannot find where this 'MystClientVersion' is used. Need clarification and probably a separate task if its gonna be used. Payments itself should not depend on consumer version. If we would support more then one payment method (or protocol) they I would say we would need this version.

vkuznecovas: You're correct - it's not used anymore. Was introduced here as we'll probably want to know what consumer version is on, and maybe adjust provider behaviour accordingly. Since we were already touching this session creation it only seemed appropriate to add this field.

session/manager.go
```diff
@@ -107,13 +113,29 @@ func (manager *Manager) Create(consumerID identity.Identity, proposalID int, con
...
+	go func() {
+		<-sessionInstance.Done
+		balanceTracker.Stop()
+	}()
+
+	go func() {
+		err := balanceTracker.Start()
+		if err != nil {
+			log.Error(managerLogPrefix, "balance tracker error: ", err)
+			destroyErr := manager.Destroy(consumerID, string(sessionInstance.ID))
```
zolia: We convert sessionID to string here, then in Destroy method we convert it back :), maybe refactor Destroy method..

session/payment/factory/payments.go
```diff
@@ -0,0 +1,72 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package factory
```
zolia: Loop dependencies? Why we need this single file package?

vkuznecovas: Moving it to a separate package so that we don't flood the DI.

session/payment/session_balance.go
```diff
@@ -0,0 +1,123 @@
...
+			if err != nil {
+				return err
+			}
+		}
+	}
+}
+
+func (ppo *SessionBalance) sendBalance() error {
+	balance := ppo.balanceTracker.GetBalance()
+	// TODO: Maybe retry a couple of times?
```
zolia: Its not a essential information, we can accept some loses here.

session/payment/session_balance.go
```diff
@@ -0,0 +1,123 @@
...
+}
+
+func (ppo *SessionBalance) receivePromiseOrTimeout() error {
+	select {
+	case pm := <-ppo.promiseChan:
+		if !ppo.promiseValidator.Validate(pm) {
+			return ErrPromiseValidationFailed
+		}
+		// TODO: Save the promise
+		// TODO: Change balance
```
zolia: We should create a ticket and link its number here.

session/payment/session_balance.go
```diff
@@ -0,0 +1,123 @@
...
+	promiseChan chan promise.Message,
+	period time.Duration,
+	promiseWaitTimeout time.Duration,
+	promiseValidator PromiseValidator) *SessionBalance {
+	return &SessionBalance{
+		stop:               make(chan struct{}),
+		peerBalanceSender:  peerBalanceSender,
+		balanceTracker:     balanceTracker,
+		promiseChan:        promiseChan,
+		period:             period,
```
zolia: i would suggest 'chargePeriod'. Also, this should be exposed so that node operator could change it. We prob can create a separate ticket for this too.

session/payment/session_balance_test.go
```diff
@@ -0,0 +1,125 @@
...
+		time.Millisecond*1,
+		time.Millisecond*1,
+		MPV,
+	)
+}
+
+func Test_ProviderPaymentOchestratorStartStop(t *testing.T) {
+	orch := NewMockSessionBalance()
+	go func() {
+		time.Sleep(time.Nanosecond * 10)
```
zolia: This probably will not guarantee that Stop will not be executed before Start.

zolia: Simple defer should suffice since we don't have result for Stop.

vkuznecovas: I am yet to see a failure on local machine or CI

session/payment/session_balance.go
vkuznecovas: What do you mean exposed? It's currently settable through the NewSessionBalance

session/manager.go
vkuznecovas: I'll create a separate task for the destroy then, since it's used in a couple of places.

session/payment/session_balance_test.go
```diff
@@ -0,0 +1,125 @@
...
+		orch.Stop()
+	}()
+	err := orch.Start()
+	assert.Nil(t, err)
+}
+
+func Test_ProviderPaymentOchestratorSendsBalance(t *testing.T) {
+	orch := NewMockSessionBalance()
+	defer orch.Stop()
+	go orch.Start()
```
zolia: This one may not be executed at all imho since test might be finished before start is executed.

vkuznecovas: The sleep below ensures that it is

session/payment/session_balance_test.go
```diff
@@ -0,0 +1,125 @@
...
+
+func Test_ProviderPaymentOchestratorInvalidPromise(t *testing.T) {
+	MPV.isValid = false
+	orch := NewMockSessionBalance()
+	defer orch.Stop()
+
+	go func() {
+		err := orch.Start()
+		assert.Equal(t, ErrPromiseValidationFailed, err)
+		MPV.isValid = true
```
zolia: I would put this in defer function if we need to be sure to cleanup the state, since it might be not the only place where we might crash.

vkuznecovas: I've remade the NewMockSessionBalance to accept a mock promise validator as a param then. There's no global state and nothing to defer/set that way

session/payment/session_payments.go
```diff
@@ -0,0 +1,93 @@
...
+)
+
+// PeerPromiseSender knows how to send a promise message to the peer
+type PeerPromiseSender interface {
+	Send(promise.Message) error
+}
+
+// PromiseTracker keeps track of promises
+type PromiseTracker interface {
+	AlignStateWithProvider(providerState promise.State) error
```
zolia: I personally feel word 'Aligned' as a curse :), but if you don't afraid it, you can use it ;)

session/payment/session_payments.go
```diff
@@ -0,0 +1,93 @@
...
+	IssuePromiseWithAddedAmount(amountToAdd int64) (promises.IssuedPromise, error)
```
zolia: We have a term for this: 'ExtendPromise'.

tadovas: It's not always promise exteding, it can generate a new promise (new id) as per request by service provider.

session/payment/session_payments.go
```diff
@@ -0,0 +1,93 @@
...
+			err := cpo.promiseTracker.AlignStateWithProvider(promise.State{
+				// TODO: figure out the int64/uint64 mess
+				Seq:    int64(balance.SequenceID),
+				Amount: int64(balance.Balance),
+			})
+			if err != nil {
+				return err
+			}
+			// TODO: figure out the int64/uint64 mess
+			issuedPromise, err := cpo.promiseTracker.ExtendPromise(int64(balance.Balance))
```
zolia: We should have a logic here. Current dump balance addition should be pointed out and separate task should exists to implement this logic. If we get balance not enough for next charge period we should extend promise with enough  credits just to cover future expected communication. Consumers confidence (reliance on provider) should reflected on this logic. Probably we should have some external parameter to control this from application settings.

vkuznecovas: Added a note for this in https://github.com/mysteriumnetwork/node/issues/717

session/manager.go
vkuznecovas: https://github.com/mysteriumnetwork/node/issues/705

core/connection/manager.go
vkuznecovas: https://github.com/mysteriumnetwork/node/issues/714

session/amount_calculator.go
vkuznecovas: https://github.com/mysteriumnetwork/node/issues/723 added this as a separate task as we might want to think about this a bit more.

session/promise/promise_extra_data.go
```diff
@@ -0,0 +1,36 @@
...
+
+package promise
+
+import (
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/crypto"
+	"github.com/mysteriumnetwork/payments/promises"
+)
+
+// ExtraData represents the extra data in the promise
```
zolia: We should add more info on what this extra data might be. For docs sake at least.

tadovas: This is exactly what it is - any extradata which will be hashed and validatable on service provider side, currently it can be service consumer identity, in case some 3d party signs promises on behalf of service consumer. Think of it more like a place holder for anything arbitrary. Though I agree that better docs needed

zolia: Currently we send a consumer address, but we need more motivation for this, otherwise we just could have a separate field in a promise as ConsumerAddress

tadovas: It's not for sending info with promise. It's for validation. And no we don't send consumer address .- we put it in extra data and validate promise signature.

zolia: True

PR #699, opened 2019-02-06
Events statistics using ELK: startup event

donce:
This collects provider startup event, including node version. By default this is enabled, but you can opt-out using `statistics.disable` flag.

Sample event json:
```json
{
  "_index": "myst-local-2019.02.06",
  "_type": "json",
  "_id": "AZ7UwWgBOzaCW2l7PZLk",
  "_version": 1,
  "_score": null,
  "_source": {
    "@version": "1",
    "@timestamp": "2019-02-06T08:02:18.623Z",
    "application": {
      "version": "source.dev-build",
      "name": "myst"
    },
    "context": {
      "role": "provider"
    },
    "createdAt": 1549440138,
    "eventName": "startup",
    "headers": {
      "request_uri": "//",
      "http_host": "metrics.mysterium.network:8091",
      "request_path": "//",
      "content_type": "application/json",
      "request_method": "POST",
      "http_accept": "application/json",
      "http_user_agent": "goclient-v0.1",
      "http_connection": "close",
      "http_accept_encoding": "gzip",
      "http_version": "HTTP/1.1",
      "content_length": "135"
    }
  },
  "fields": {
    "@timestamp": [
      "2019-02-06T08:02:18.623Z"
    ]
  },
  "highlight": {
    "application.name": [
      "@kibana-highlighted-field@myst@/kibana-highlighted-field@"
    ]
  },
  "sort": [
    1549440138623
  ]
}
```

This is very simple and straightforward implementation - adding retry mechanism, more events and etc should be done in separate PRs.

### DIFF
--- a/cmd/di.go
+++ b/cmd/di.go
@@ -74,7 +74,7 @@ type Storage interface {
 	Close() error
 }
 
-// Dependencies is DI container for top level components which is reusedin several places
+// Dependencies is DI container for top level components which is reused in several places
 type Dependencies struct {
 	Node *node.Node
 
--- a/cmd/di.go
+++ b/cmd/di.go
@@ -39,6 +39,7 @@ import (
 	"github.com/mysteriumnetwork/node/core/service"
 	"github.com/mysteriumnetwork/node/core/storage/boltdb"
 	"github.com/mysteriumnetwork/node/core/storage/boltdb/migrations/history"
+	"github.com/mysteriumnetwork/node/events"
 	"github.com/mysteriumnetwork/node/identity"
 	identity_registry "github.com/mysteriumnetwork/node/identity/registry"
 	"github.com/mysteriumnetwork/node/logconfig"
@@ -99,6 +100,7 @@ type Dependencies struct {
 	StatisticsTracker  *statistics.SessionStatisticsTracker
 	StatisticsReporter *statistics.SessionStatisticsReporter
 	SessionStorage     *consumer_session.Storage
+	EventsSender       *events.Sender
 
 	EventBus EventBus.Bus
 
@@ -363,6 +365,8 @@ func (di *Dependencies) bootstrapNetworkComponents(options node.OptionsNetwork) 
 	di.NetworkDefinition = network
 	di.MysteriumAPI = mysterium.NewClient(network.DiscoveryAPIAddress)
 	di.MysteriumMorqaClient = oracle.NewMorqaClient(network.QualityOracle)
+	eventsTransport := events.NewELKTransport(network.ELKAddress, 10*time.Second)
+	di.EventsSender = &events.Sender{Transport: eventsTransport}
 
 	log.Info("Using Eth endpoint: ", network.EtherClientRPC)
 	if di.EtherClient, err = blockchain.NewClient(network.EtherClientRPC); err != nil {
--- a/cmd/di.go
+++ b/cmd/di.go
@@ -372,7 +372,7 @@ func (di *Dependencies) bootstrapNetworkComponents(options node.OptionsNetwork) 
 
 	var eventsTransport events.Transport
 	if options.DisableStatistics {
-		eventsTransport = events.NewNilTransport()
+		eventsTransport = events.NewNoopTransport()
 	} else {
 		eventsTransport = events.NewELKTransport(network.ELKAddress, 10*time.Second)
 	}
--- a/cmd/di.go
+++ b/cmd/di.go
@@ -39,15 +39,15 @@ import (
 	"github.com/mysteriumnetwork/node/core/service"
 	"github.com/mysteriumnetwork/node/core/storage/boltdb"
 	"github.com/mysteriumnetwork/node/core/storage/boltdb/migrations/history"
-	"github.com/mysteriumnetwork/node/events"
 	"github.com/mysteriumnetwork/node/identity"
 	identity_registry "github.com/mysteriumnetwork/node/identity/registry"
 	"github.com/mysteriumnetwork/node/logconfig"
 	"github.com/mysteriumnetwork/node/market"
-	"github.com/mysteriumnetwork/node/market/metrics"
+	market_metrics "github.com/mysteriumnetwork/node/market/metrics"
 	"github.com/mysteriumnetwork/node/market/metrics/oracle"
 	"github.com/mysteriumnetwork/node/market/mysterium"
 	"github.com/mysteriumnetwork/node/metadata"
+	"github.com/mysteriumnetwork/node/metrics"
 	"github.com/mysteriumnetwork/node/money"
 	"github.com/mysteriumnetwork/node/nat"
 	service_noop "github.com/mysteriumnetwork/node/services/noop"
@@ -81,7 +81,7 @@ type Dependencies struct {
 
 	NetworkDefinition    metadata.NetworkDefinition
 	MysteriumAPI         *mysterium.MysteriumAPI
-	MysteriumMorqaClient metrics.QualityOracle
+	MysteriumMorqaClient market_metrics.QualityOracle
 	EtherClient          *ethclient.Client
 
 	NATService           nat.NATService
@@ -100,7 +100,7 @@ type Dependencies struct {
 	StatisticsTracker  *statistics.SessionStatisticsTracker
 	StatisticsReporter *statistics.SessionStatisticsReporter
 	SessionStorage     *consumer_session.Storage
-	EventsSender       *events.Sender
+	MetricsSender      *metrics.Sender
 
 	EventBus EventBus.Bus
 
@@ -362,15 +362,15 @@ func (di *Dependencies) bootstrapNetworkComponents(options node.OptionsNetwork) 
 		network.EtherClientRPC = options.EtherClientRPC
 	}
 
-	if options.StatisticsAddress != metadata.DefaultNetwork.ELKAddress {
-		network.ELKAddress = options.StatisticsAddress
+	if options.MetricsAddress != metadata.DefaultNetwork.MetricsAddress {
+		network.MetricsAddress = options.MetricsAddress
 	}
 
 	di.NetworkDefinition = network
 	di.MysteriumAPI = mysterium.NewClient(network.DiscoveryAPIAddress)
 	di.MysteriumMorqaClient = oracle.NewMorqaClient(network.QualityOracle)
 
-	di.EventsSender = events.CreateSender(options.DisableStatistics, network.ELKAddress)
+	di.MetricsSender = metrics.CreateSender(options.DisableMetrics, network.MetricsAddress)
 
 	log.Info("Using Eth endpoint: ", network.EtherClientRPC)
 	if di.EtherClient, err = blockchain.NewClient(network.EtherClientRPC); err != nil {
--- a/cmd/di.go
+++ b/cmd/di.go
@@ -362,10 +362,20 @@ func (di *Dependencies) bootstrapNetworkComponents(options node.OptionsNetwork) 
 		network.EtherClientRPC = options.EtherClientRPC
 	}
 
+	if options.StatisticsAddress != metadata.DefaultNetwork.ELKAddress {
+		network.ELKAddress = options.StatisticsAddress
+	}
+
 	di.NetworkDefinition = network
 	di.MysteriumAPI = mysterium.NewClient(network.DiscoveryAPIAddress)
 	di.MysteriumMorqaClient = oracle.NewMorqaClient(network.QualityOracle)
-	eventsTransport := events.NewELKTransport(network.ELKAddress, 10*time.Second)
+
+	var eventsTransport events.Transport
+	if options.DisableStatistics {
+		eventsTransport = events.NewNilTransport()
+	} else {
+		eventsTransport = events.NewELKTransport(network.ELKAddress, 10*time.Second)
+	}
 	di.EventsSender = &events.Sender{Transport: eventsTransport}
 
 	log.Info("Using Eth endpoint: ", network.EtherClientRPC)
--- a/cmd/di.go
+++ b/cmd/di.go
@@ -100,7 +100,6 @@ type Dependencies struct {
 	StatisticsTracker  *statistics.SessionStatisticsTracker
 	StatisticsReporter *statistics.SessionStatisticsReporter
 	SessionStorage     *consumer_session.Storage
-	MetricsSender      *metrics.Sender
 
 	EventBus EventBus.Bus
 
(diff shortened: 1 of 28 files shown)


cmd/di.go
```diff
@@ -329,10 +331,22 @@ func (di *Dependencies) bootstrapNetworkComponents(options node.OptionsNetwork)
 		network.EtherClientRPC = options.EtherClientRPC
 	}
 
+	if options.StatisticsAddress != metadata.DefaultNetwork.ELKAddress {
```
soffokl: I don't like `ELK` name here since it only ElasticSearch and there is no Logstash and Kibana.

events/nil_transport.go
```diff
@@ -0,0 +1,30 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package events
+
+// NewNilTransport creates transport which ignores requests to send events
+func NewNilTransport() Transport {
```
vkuznecovas: The convention in this repo seems to be that we use the word noop to indicate a non functional component. Can you please rename?

events/elk_transport.go
```diff
@@ -0,0 +1,77 @@
...
+)
+
+// NewELKTransport creates transport allowing to send events to ELK through HTTP
+func NewELKTransport(elkUrl string, timeout time.Duration) Transport {
+	return &elkTransport{http: newMysteriumHTTPTransport(timeout), elkUrl: elkUrl}
+}
+
+type elkTransport struct {
+	http   mysterium.HTTPTransport
+	elkUrl string
```
soffokl: > struct field elkUrl should be elkURL (golint)

events/elk_transport.go
```diff
@@ -0,0 +1,77 @@
...
+	"net/http"
+	"strings"
+	"time"
+
+	"github.com/mysteriumnetwork/node/market/mysterium"
+	"github.com/mysteriumnetwork/node/requests"
+)
+
+// NewELKTransport creates transport allowing to send events to ELK through HTTP
+func NewELKTransport(elkUrl string, timeout time.Duration) Transport {
```
soffokl: > func parameter elkUrl should be elkURL (golint)

cmd/di.go
```diff
@@ -91,6 +92,7 @@ type Dependencies struct {
 	StatisticsTracker  *statistics.SessionStatisticsTracker
 	StatisticsReporter *statistics.SessionStatisticsReporter
 	SessionStorage     *consumer_session.Storage
+	EventsSender       *events.Sender
```
vkuznecovas: Is it really an EventsSender? To me it seems more appropriate to name it Metrics.Sender or something like that. Maybe rename the whole events package to metrics?

events/elk_transport.go
```diff
@@ -0,0 +1,77 @@
...
+	if err != nil {
+		return err
+	}
+
+	body := string(bodyBytes)
+	if strings.ToUpper(body) != "OK" {
+		return errors.New("Unexpected response body: " + body)
+	}
+
+	return err
```
soffokl: It will always be `nil` at that moment. 
```suggestion
	return nil
```

events/elk_transport.go
```diff
@@ -0,0 +1,77 @@
...
+		return err
+	}
+
+	response, err := transport.http.Do(req)
+	if err != nil {
+		return err
+	}
+
+	if response.StatusCode != 200 {
+		return errors.New("Unexpected elk response status: " + response.Status)
```
soffokl: [Error strings should not be capitalized...](https://github.com/golang/go/wiki/CodeReviewComments#error-strings)
```suggestion
		return errors.New("unexpected elk response status: " + response.Status)
```

events/elk_transport.go
```diff
@@ -0,0 +1,77 @@
...
+	}
+
+	bodyBytes, err := ioutil.ReadAll(response.Body)
+	if err != nil {
+		return err
+	}
+
+	body := string(bodyBytes)
+	if strings.ToUpper(body) != "OK" {
+		return errors.New("Unexpected response body: " + body)
```
soffokl: [Error strings should not be capitalized...](https://github.com/golang/go/wiki/CodeReviewComments#error-strings)
```suggestion
		return errors.New("unexpected response body: " + body)
```

cmd/di.go
```diff
@@ -329,10 +331,22 @@ func (di *Dependencies) bootstrapNetworkComponents(options node.OptionsNetwork)
...
+	if options.StatisticsAddress != metadata.DefaultNetwork.ELKAddress {
+		network.ELKAddress = options.StatisticsAddress
+	}
+
 	di.NetworkDefinition = network
 	di.MysteriumAPI = mysterium.NewClient(network.DiscoveryAPIAddress)
 	di.MysteriumMorqaClient = oracle.NewMorqaClient(network.QualityOracle)
 
+	var eventsTransport events.Transport
+	if options.DisableStatistics {
```
vkuznecovas: I'd rather this be an opt-in instead of an opt-out since we're working on privacy. But that's just me.

events/sender.go
```diff
@@ -0,0 +1,74 @@
...
+const RoleProvider = "provider"
+
+// Sender builds events and sends them using given transport
+type Sender struct {
+	Transport Transport
+}
+
+// Transport allows sending events
+type Transport interface {
+	sendEvent(event) error
```
vkuznecovas: By making this private you're forcing the transport to live in the same package. What if I actually want it to live in a sub package?

events/sender.go
```diff
@@ -0,0 +1,74 @@
...
+type applicationInfo struct {
+	Name    string `json:"name"`
+	Version string `json:"version"`
+}
+
+type startupContext struct {
+	Role string `json:"role"`
+}
+
+const appName = "myst"
```
soffokl: I'd move consts and types to the top of the file.

cmd/di.go
```diff
@@ -329,10 +331,22 @@ func (di *Dependencies) bootstrapNetworkComponents(options node.OptionsNetwork)
...
+		eventsTransport = events.NewNilTransport()
+	} else {
+		eventsTransport = events.NewELKTransport(network.ELKAddress, 10*time.Second)
+	}
+	di.EventsSender = &events.Sender{Transport: eventsTransport}
```
vkuznecovas: could also create a sub package in the events(or preferably the metrics) package, call it a factory and just implement the whole block there. Here we'd only have a single call to the factory with params and we would not make the DI any more heavy than it is.

events/elk_transport.go
vkuznecovas: You'd probably want to include more info here not just the status. Or at the very least log additional info.

events/sender.go
vkuznecovas: Talking of which, might be a good idea to move the transports to a separate sub package, called transports

events/sender.go
vkuznecovas: I'd move these to the top of the document. Feels a bit weird that they are stuck at the bottom.

cmd/di.go
vkuznecovas: You'll also want to make sure that we disable these on the e2e suite if we stay with the opt-out.

metadata/network.go
```diff
@@ -36,15 +37,17 @@ var TestnetDefinition = NetworkDefinition{
 	"nats://testnet-broker.mysterium.network",
 	"https://ropsten.infura.io",
 	"https://testnet-morqa.mysterium.network/api/v1",
+	"http://metrics.mysterium.network:8091",
```
soffokl: I'm not a fan of sending events directly to ElasticSearch. There are several reasons for that:
1) If we have open ElasticSearch to everyone, it can be easily overloaded, no matter how many nodes we have;
1) If we will need to replace ElasticSearch with some other solution one day, we will need to update every client for that;
1) Any one can fake/compromise these stats. 

I'd do some simple own service that will receive events and send it to the ElasticSearch. In this case we can control access and replace backend db(ElasticSearch) at any moment. 
It could be a part of discovery service or another one.

vkuznecovas: What we could also consider is adding an endpoint to current morqa.

events/elk_transport.go
donce: Strange - why `bin/check_golint` doesn't warn me about this?

soffokl: It's probably because it has a flag `--min_confidence=1` in our scripts:
https://github.com/mysteriumnetwork/node/blob/master/bin/check_golint#L42

cmd/di.go
donce: I was thinking about that, but we already have `metrics` package (`market.metrics`), so naming 2 things with same package names seemed confusing and importing both packages will require named imports.

cmd/di.go
donce: We discussed this with @zolia that we probably won't receive any stats if it is an opt-in. Do you think we all need to discuss more about this topic?

events/elk_transport.go
donce: Added response body as well

cmd/di.go
donce: Renamed to `metrics`

events/sender.go
donce: But that's the idea - `Transport` is public, so you can manage `Transport` instances outside of package, deciding on the transport you want to use. But `Transport.sendEvent` shouldn't be used directly, so not exposing this method enforces users to use this as it is intended - only through `Sender`

cmd/di.go
donce: Extracted this as `metrics.CreateSender`

cmd/di.go
donce: Discussed this once more time all together, decided to keep it opt-out.

cmd/di.go
donce: Renamed name to `ElasticSearch`, and remove the name altogether where appropriate. But it makes sense to name transport with the name (`ElasticSearchTransport`), since the implementation is specific to elastic search.

metadata/network.go
donce: We'll agreed on creating this as a separate service. Created separate task for that: https://github.com/mysteriumnetwork/node/issues/701.
We discussed this with @zolia, that it can be done separately, because we're already sending statistics directly to ElasticSearch in all desktop and mobile apps, so we'll need to migrate all those as well - so let's do that separately in other PRs, step by step :)

donce:
@vkuznecovas @soffokl all issues were address - most of them fixed, few probably needs some discussion

metrics/sender.go
```diff
@@ -0,0 +1,74 @@
...
+// SendStartupEvent sends startup event
+func (sender *Sender) SendStartupEvent(role, version string) {
+	appInfo := applicationInfo{Name: appName, Version: version}
+	event := event{Application: appInfo, EventName: startupEventName, Context: startupContext{Role: role}, CreatedAt: time.Now().Unix()}
+
+	sender.sendEvent(event)
+}
+
+func (sender *Sender) sendEvent(event event) {
+	go func() {
```
vkuznecovas: you'd better off getting rid of the goroutine here. Just return the error and let the consumer decide if he wants to send it in a separate routine or not.

donce: I don't want to block application initiation just because statistics are being sent - sending statistics should be done in the background by default.

Now it probably doesn't make a lot of sense here, but later on, we'll have to implement retrying mechanism - especially for connection events, which happen when network is changing and often those requests fail.

That's why I am testing that sender doesn't block execution - with retry mechanism, we'll just have to extend those tests, responsibility of managing execution will be kept in sender.

donce: Maybe actually having retry mechanism in this PR would solve this? Because I understand that responsibilities distribution in current code is not that clear :D

vkuznecovas: If you get rid of the goroutine the component becomes flexible, it can be used in synchronous execution and as a background task. Here you're hardcoding it as a background task with no say on the consumer side. 

You can just happily use go ...sendEvent(...) and have the behaviour you want. 


As far as retrying goes - it would probably be best to leave it for the consumer to decide. Maybe some metrics are more relevant than others and they should be retried, while others can be allowed to fail once in a while? My point here is that the consumer should be the one making these assumptions, not the underlying sender.

donce: Yeah, flexibility is important and I would like to keep requests being async by default.

Maybe we can have both - what if we had `AsyncTransport` just for making requests in the background?

Both `AsyncTransport(transport Transport)` and `ElasticSearchTransport(url)` would implement same `Transport` interface, so they could be composed together.

By default, we could construct `Sender` with `AsyncTransport(ElasticSearchTransport)` to get async functionality by default, but we can change that later if we need to do sync events somewhere.

Later we can extend `AsyncTransport` to `RetryTransport` (or whatever, I am bad at naming :D) - nothing else would change, just this transport.

Would you be fine with this approach?

vkuznecovas: This misses the point though, you'll have two separate dependencies depending on the behaviour you want. I'd rather have a single component that handles the metrics sending in the way I as the consumer would like. Maybe I want a single request to be sent asynchronously and another one sent synchronously. 

I'm not gonna block the PR for this though.

cmd/flags_directory.go
```diff
@@ -21,7 +21,7 @@ import (
 	"os"
 	"path/filepath"
 
-	"github.com/mitchellh/go-homedir"
+	homedir "github.com/mitchellh/go-homedir"
```
tadovas: Why this is needed?

core/node/options_network.go
```diff
@@ -32,4 +32,7 @@ type OptionsNetwork struct {
 	EtherPaymentsAddress string
 
 	QualityOracle string
+
+	DisableMetrics bool
```
tadovas: It has nothing to do with network options.
Network options are ment to be related to our network itself - blockchain, market etc.

metrics/sender.go
```diff
@@ -0,0 +1,74 @@
...
+const appName = "myst"
+const startupEventName = "startup"
+
+// Sender builds events and sends them using given transport
+type Sender struct {
+	Transport Transport
+}
+
+// Transport allows sending events
+type Transport interface {
```
tadovas: Exported interface with all private methods - looks strange

cmd/flags_directory.go
donce: `goimports` requires it

tadovas: Oh its just alias. 👍

metrics/sender.go
donce: This allows passing around instancef of Transport, but hides internals - transports are meant only to be passed to `Sender` without exposing internals.
Later on, we can have `RetryTransport` which wraps another transport (i.e. HTTP!).

core/node/options_network.go
donce: Will fix that.

events/sender.go
tadovas: I would highly recommend designing metrics sending async right away. To make it as less intrusive as possible and not accidently hurt performance of any system components. A good problem example was - when consumer for any reason losses internet connection (DNS functionality to be exact) and tries to disconnect from node - disconnect hangs until final stats send timeouts - and thats not good.

metrics/elastic_search_transport.go
```diff
@@ -0,0 +1,77 @@
...
+	url  string
+}
+
+func (transport *elasticSearchTransport) sendEvent(event event) error {
+	req, err := requests.NewPostRequest(transport.url, "/", event)
+	if err != nil {
+		return err
+	}
+
+	response, err := transport.http.Do(req)
```
zolia: I don't see response being closed.

metrics/sender.go
zolia: Rule of thumb in go is not to hide go routine in a function, not to make assumptions on how function works. Is a matter of caller how he wants to run it.
As of additional transport interfaces - I don't think there is a point in increasing complexity if not strictly required atm.

metrics/elastic_search_transport.go
zolia: defer response.Body.Close() is needed here.

donce:
@tadovas @zolia fixed comments

PR #704, opened 2019-02-07
disconnect is now solely responsible for calling the cleanup

vkuznecovas:
We had an issue where certain cleanup operations could be called multiple times due to the fact that they were called both on the cancel side and the connection waiter. Now, only the disconnect method calls the cleanup operation.

### DIFF
--- a/core/connection/manager.go
+++ b/core/connection/manager.go
@@ -140,7 +140,7 @@ func (manager *connectionManager) startConnection(consumerID identity.Identity, 
 		}
 		if err != nil {
 			log.Info(managerLogPrefix, "Cancelling connection initiation", err)
-			_ = manager.Disconnect()
+			logDisconnectError(manager.Disconnect())
 		}
 	}()
 
@@ -290,7 +290,7 @@ func (manager *connectionManager) connectionWaiter(connection Connection) {
 		log.Info(managerLogPrefix, "Connection exited")
 	}
 
-	_ = manager.Disconnect()
+	logDisconnectError(manager.Disconnect())
 }
 
 func (manager *connectionManager) waitForConnectedState(stateChannel <-chan State, sessionID session.ID) error {
@@ -320,7 +320,7 @@ func (manager *connectionManager) consumeConnectionStates(stateChannel <-chan St
 	}
 
 	log.Debug(managerLogPrefix, "State updater stopCalled")
-	_ = manager.Disconnect()
+	logDisconnectError(manager.Disconnect())
 }
 
 func (manager *connectionManager) consumeStats(statisticsChannel <-chan consumer.SessionStatistics) {
@@ -340,5 +340,11 @@ func (manager *connectionManager) onStateChanged(state State) {
 		manager.setStatus(statusConnected(manager.sessionInfo.SessionID, manager.sessionInfo.Proposal))
 	case Reconnecting:
 		manager.setStatus(statusReconnecting())
+	}
+}
+
+func logDisconnectError(err error) {
+	if err != nil && err != ErrNoConnection {
+		log.Error(managerLogPrefix, "Disconnect error", err)
 	}
 }
--- a/core/connection/manager.go
+++ b/core/connection/manager.go
@@ -82,10 +82,12 @@ type connectionManager struct {
 
 	//these are populated by Connect at runtime
 	ctx             context.Context
-	mutex           sync.RWMutex
 	status          Status
+	statusLock      sync.RWMutex
 	sessionInfo     SessionInfo
 	cleanConnection func()
+
+	discoLock sync.Mutex
 }
 
 // NewManager creates connection manager with given dependencies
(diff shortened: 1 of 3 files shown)


core/connection/manager_test.go
```diff
@@ -140,14 +140,15 @@ func (tc *testContext) TestStatusReportsConnectingWhenConnectionIsInProgress() {
...
-func (tc *testContext) TestStatusReportsDisconnectingThenNotConnected() {
+func (tc *testContext) TestStatusReportsNotConnected() {
 	tc.fakeConnectionFactory.mockConnection.onStopReportStates = []fakeState{}
 	err := tc.connManager.Connect(consumerID, activeProposal, ConnectParams{})
 	assert.NoError(tc.T(), err)
 	assert.Equal(tc.T(), statusConnected(establishedSessionID), tc.connManager.Status())
 
 	assert.NoError(tc.T(), tc.connManager.Disconnect())
-	assert.Equal(tc.T(), statusDisconnecting(), tc.connManager.Status())
+	// TODO: this switch now happens way too quickly to catch
```
Waldz: :(

tadovas: What do you mean too quickly? Disconnecting state stays as long as underlying connection hasn't completed Wait() call. As we use mock (or fake) underlying connection, it can easily be emulated with simple channel blocking:
assert.( ... DisconnectingState)
tc.mockConnection.completeWait()
assert.(.... DisconnectedState)
That was the plan. Now I have a strange feeling that implementation broke the test somehow, not that test itself is broken

vkuznecovas:
I've put the managers status behind a lock so it's not accessible simultaneously by many routines. Put disconnect behind its own lock, and added a mechanism that can make stop linger around in the connection manager mock.

core/connection/manager.go
```diff
@@ -130,22 +128,19 @@ func (manager *connectionManager) Connect(consumerID identity.Identity, proposal
...
 	defer func() {
 		manager.cleanConnection = func() {
-			manager.status = statusDisconnecting()
 			cancelCtx()
 			for i := range cancel { // Cancelling in a reverse order to keep correct workflow.
 				cancel[len(cancel)-i-1]()
 			}
 		}
 		if err != nil {
 			log.Info(managerLogPrefix, "Cancelling connection initiation")
```
zolia:
```suggestion
			log.Info(managerLogPrefix, "Cancelling connection initiation: ", err)
```
eating error.

core/connection/manager.go
```diff
@@ -130,22 +128,19 @@ func (manager *connectionManager) Connect(consumerID identity.Identity, proposal
...
-			defer manager.cleanConnection()
+			_ = manager.Disconnect()
```
zolia: Why not to log an error?

core/connection/manager.go
```diff
@@ -82,10 +82,12 @@ type connectionManager struct {
 
 	//these are populated by Connect at runtime
 	ctx             context.Context
-	mutex           sync.RWMutex
 	status          Status
 	sessionInfo     SessionInfo
 	cleanConnection func()
+
+	statusLock sync.RWMutex
```
zolia: should go right after field it is guarding. That why sometimes field name can be not used in its mutex.

core/connection/manager.go
```diff
@@ -277,15 +282,15 @@ func warnOnClean() {
...
+func (manager *connectionManager) connectionWaiter(connection Connection) {
 	err := connection.Wait()
 	if err != nil {
 		log.Warn(managerLogPrefix, "Connection exited with error: ", err)
 	} else {
 		log.Info(managerLogPrefix, "Connection exited")
 	}
 
-	dialog.Close()
+	_ = manager.Disconnect()
```
zolia: same here.. afraid of underscores ;)

core/connection/manager.go
vkuznecovas: The only error we're possibly returning from disconnect is the `ErrNoConnection`, which is going to flood the logs needlessly, since these disconnects will get triggered by multiple goroutines(possibly).

core/connection/manager.go
vkuznecovas: The only error we're possibly returning from disconnect is the `ErrNoConnection`, which is going to flood the logs needlessly, since these disconnects will get triggered by multiple goroutines(possibly).

vkuznecovas: I've added a method to log the error and ignore the ErrNoConnection

core/connection/manager.go
```diff
@@ -240,25 +235,35 @@ func (manager *connectionManager) startConnection(consumerID identity.Identity,
...
+func (manager *connectionManager) setStatus(cs Status) {
+	manager.statusLock.Lock()
+	manager.status = cs
+	manager.statusLock.Unlock()
+}
+
 func (manager *connectionManager) Disconnect() error {
-	manager.mutex.RLock()
-	defer manager.mutex.RUnlock()
+	manager.discoLock.Lock()
```
soffokl: Do we really need a second mutex for `Disconnect`? 
We already have a mutex for status, we can check status here to make sure we are in a right state.

vkuznecovas: I think it's required here due to the fact that the status is a r/w lock. We could get into a situation where 2 goroutines are calling disconnect here. What could happen then is:
1st routine checks the status state, let's say it's connected. So it get's that. It's execution get's delayed at this point in time.
2nd routine checks the status - we're still connected, alrighty.

Both  1st and 2nd routines start executing the cleanConnection, which I want to avoid.

The status though should definitely be behind a r/wlock since it needs to be more responsive for external calls to check the status.

PR #737, opened 2019-02-13
port forwarding autoconfiguration for WG

zolia:
Uses IGD (upnp) and PMP protocols to enable port forwarding for WG
 - Reuses the same Ethereum port mapping code
 - Maps port for each WG session separately since WG instance sits on separate port
 - Instruct client to postpone connect for some time for mappings to get activated

If client connects while mappings are not active - FW throttles (blacklists) incoming requests for some random time and WG session fails to establish.  In case WG provider is behind NAT we instruct client to delay connect for mappings to get active.

### DIFF
--- a/cmd/service_bootstrap_desktop.go
+++ b/cmd/service_bootstrap_desktop.go
@@ -24,12 +24,14 @@ import (
 	"github.com/mysteriumnetwork/node/communication"
 	nats_dialog "github.com/mysteriumnetwork/node/communication/nats/dialog"
 	nats_discovery "github.com/mysteriumnetwork/node/communication/nats/discovery"
+	"github.com/mysteriumnetwork/node/core/location"
 	"github.com/mysteriumnetwork/node/core/node"
 	"github.com/mysteriumnetwork/node/core/service"
 	"github.com/mysteriumnetwork/node/identity"
 	"github.com/mysteriumnetwork/node/market"
 	"github.com/mysteriumnetwork/node/market/proposals/registry"
 	"github.com/mysteriumnetwork/node/nat"
+	"github.com/mysteriumnetwork/node/nat/mapping"
 	service_noop "github.com/mysteriumnetwork/node/services/noop"
 	service_openvpn "github.com/mysteriumnetwork/node/services/openvpn"
 	openvpn_discovery "github.com/mysteriumnetwork/node/services/openvpn/discovery"
@@ -39,13 +41,7 @@ import (
 
 const logPrefix = "[service bootstrap] "
 
-type locationInfo struct {
-	OutIP   string
-	PubIP   string
-	Country string
-}
-
-func (di *Dependencies) resolveIPsAndLocation() (loc locationInfo, err error) {
+func (di *Dependencies) resolveIPsAndLocation() (loc location.ServiceLocationInfo, err error) {
 	pubIP, err := di.IPResolver.GetPublicIP()
 	if err != nil {
 		return
@@ -80,8 +76,18 @@ func (di *Dependencies) bootstrapServiceOpenvpn(nodeOptions node.Options) {
 		currentLocation := market.Location{Country: location.Country}
 		transportOptions := serviceOptions.(openvpn_service.Options)
 
+		mapPort := func() func() {
+			if location.OutIP != location.PubIP {
+				return mapping.PortMapping(
+					transportOptions.OpenvpnProtocol,
+					transportOptions.OpenvpnPort,
+					"Myst node openvpn port mapping")
+			}
+			return func() {}
+		}
+
 		proposal := openvpn_discovery.NewServiceProposalWithLocation(currentLocation, transportOptions.OpenvpnProtocol)
-		return openvpn_service.NewManager(nodeOptions, transportOptions, location.PubIP, location.OutIP, location.Country, di.ServiceSessionStorage, di.NATService), proposal, nil
+		return openvpn_service.NewManager(nodeOptions, transportOptions, location, di.ServiceSessionStorage, di.NATService, mapPort), proposal, nil
 	}
 	di.ServiceRegistry.Register(service_openvpn.ServiceType, createService)
 }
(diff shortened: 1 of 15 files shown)


services/openvpn/service/manager.go
```diff
@@ -65,58 +65,63 @@ type Manager struct {
...
 }
 
+func (m *Manager) mapPort() {
+	if m.outboundIP != m.publicIP {
+		m.releasePorts = mapping.PortMapping(
+			m.serviceOptions.OpenvpnProtocol,
+			m.serviceOptions.OpenvpnPort,
+			"Myst node openvpn port mapping")
+	}
+	m.releasePorts = func() {}
```
soffokl: We are replacing `m.releasePorts` every time. Probably some return required in the if block.

services/wireguard/endpoint/endpoint.go
```diff
@@ -22,12 +22,14 @@ import (
 	"net"
 
 	log "github.com/cihub/seelog"
+	"github.com/mysteriumnetwork/node/nat/mapping"
```
soffokl: I still have a feeling that mapping should not go into service, and all port manipulations should be done outside this package. `¯\_(ツ)_/¯`
In this case we don't need to pass both `publicIP` and `outbandIP` into connection endpoint.

services/wireguard/endpoint/endpoint.go
```diff
@@ -92,6 +100,18 @@ func (ce *connectionEndpoint) Start(config *wg.ServiceConfig) error {
 	return ce.wgClient.ConfigureDevice(ce.iface, deviceConfig, ce.ipAddr)
 }
 
+func (ce *connectionEndpoint) mapPort(port int) {
+	if ce.outboundIP != ce.publicIP {
+		ce.connectDelay = connectDelayForPortMap
```
soffokl: This looks like a hack, are we sure that 2000ms will be good for every router? Why 2000ms?

services/wireguard/endpoint/endpoint.go
```diff
@@ -39,13 +41,16 @@ type wgClient interface {
...
+	iface              string
+	privateKey         string
+	publicIP           string
+	outboundIP         string
+	ipAddr             net.IPNet
+	endpoint           net.UDPAddr
+	resourceAllocator  *resources.Allocator
+	wgClient           wgClient
+	releasePortMapping func()
+	connectDelay       int // connect delay in milliseconds
```
soffokl: I know what problem are you trying to solve, but it will not be clear for another person who will try to read this code. We probably need to add some description of why we need this delay.

services/wireguard/endpoint/endpoint.go
zolia: 2000 just by experimentation only. I cannot think of nice enough way to determine if router really activated this forwarding. Rule itself gets accepted, but looks like it is still not active for sometime. IDG protocol is too dum in this case and we don't have any status. 
I believe it might be heavy dependent on a gateway manufacturer how long does it take to activate such rule and we should perform tests with various models.
If you have some router differing from ours ( Mikrotik ) you can test this too.

services/wireguard/endpoint/endpoint.go
```diff
@@ -22,12 +22,14 @@ import (
...
 
 	log "github.com/cihub/seelog"
+	"github.com/mysteriumnetwork/node/nat/mapping"
 	wg "github.com/mysteriumnetwork/node/services/wireguard"
 	"github.com/mysteriumnetwork/node/services/wireguard/key"
 	"github.com/mysteriumnetwork/node/services/wireguard/resources"
 )
 
 const logPrefix = "[wireguard-connection-endpoint] "
+const connectDelayForPortMap = 2000
```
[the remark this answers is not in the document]
vkuznecovas: might also need to add it to the docs after implementing.

mobile/mysterium/wireguard_connection_setup.go
```diff
@@ -220,6 +225,11 @@ func (wg *wireguardConnection) Start(options connection.ConnectOptions) error {
 	wg.device = device
 	wg.stateChannel <- connection.Connecting
 
+	if config.Consumer.ConnectDelay > 0 {
+		log.Infof("%s delaying connect for %v milliseconds", logPrefix, config.Consumer.ConnectDelay)
```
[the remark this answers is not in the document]
zolia: comment added.

services/wireguard/endpoint/endpoint.go
zolia: Since you raised this 2nd time, i refactored it to be injectable for both services.

cmd/service_bootstrap_desktop.go
```diff
@@ -80,8 +76,18 @@ func (di *Dependencies) bootstrapServiceOpenvpn(nodeOptions node.Options) {
 		currentLocation := market.Location{Country: location.Country}
 		transportOptions := serviceOptions.(openvpn_service.Options)
 
+		mapPort := func() func() {
```
vkuznecovas: This makes me uncomfortable. Since we're already using this in two places(and probably more to come)  I think it makes sense to go a step further, and maybe have a func like this in the mapping package:

```golang
func GetPortMappingFunc(pubIP,outIP, protocol,port, description) func() {
    if pubIP != outIP {
		return mapping.PortMapping(protocol, port, description)
    }
    return func() {}
}
```

In that case, it will be more concise to inject these to the corresponding services, as the map port would become something like this:

```golang
mapPort := func() {
    return mapping.GetPortMappingFunc(location.OutIP, locationPubIp, ....)
} 
```

services/wireguard/endpoint/endpoint.go
```diff
@@ -21,13 +21,16 @@ import (
...
+	"github.com/mysteriumnetwork/node/core/location"
+
 	log "github.com/cihub/seelog"
 	wg "github.com/mysteriumnetwork/node/services/wireguard"
 	"github.com/mysteriumnetwork/node/services/wireguard/key"
 	"github.com/mysteriumnetwork/node/services/wireguard/resources"
 )
 
 const logPrefix = "[wireguard-connection-endpoint] "
+const connectDelayForPortMap = 200
```
vkuznecovas: still hardcoded :)

cmd/service_bootstrap_desktop.go
zolia: will do.

PR #738, opened 2019-02-14
Mocked Tequilapi endpoint for services

soffokl:


tequilapi/endpoints/connection_test.go
```diff
@@ -34,7 +34,7 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-type fakeManager struct {
+type fakeConnectionManager struct {
```
vkuznecovas: fakeConnectionManager -> mockConnectionManager. We probably should not use the `fake` anymore.

soffokl: renamed

tequilapi/endpoints/service.go
```diff
@@ -0,0 +1,230 @@
...
+
+	resp.WriteHeader(http.StatusCreated)
+	se.ServiceGet(resp, req, params)
+}
+
+// ServiceStop stops service on the node.
+// swagger:operation DELETE /service/:id Service serviceStop
+// ---
+// summary: Stops service
+// description: Stops current service
```
donce: So if I understand correctly, if service is `Starting`, than this will try to cancel service before it starts?
If so, can we add a notice for this case here?

tequilapi/endpoints/service.go
```diff
@@ -0,0 +1,230 @@
...
+// responses:
+//   202:
+//     description: Service Stopped
```
donce: So if I understand correctly, it is guaranteed that `ServiceStop` will finish only after `State` will be `NotRunning`? I'm talking about real implementation, not mock one.

tequilapi/endpoints/service.go
```diff
@@ -0,0 +1,230 @@
...
+//     description: List of running services
+//     schema:
+//       "$ref": "#/definitions/ServiceListDTO"
+func (se *ServiceEndpoint) ServiceList(resp http.ResponseWriter, _ *http.Request, _ httprouter.Params) {
+	statusResponse := toServiceListResponse(se.serviceState)
+	utils.WriteAsJSON(statusResponse, resp)
+}
+
+// ServiceGet provides info for requested service on the node.
+// swagger:operation GET /services/:id Service serviceGet
```
donce: It's a bit unclear what is service id - how to get it?

tequilapi/endpoints/service.go
soffokl: I think it should stop both `Running` and `Starting` states of the service.

Let's say you were using the system as a consumer only, then decided to provide `openvpn` service. You need to start `openvpn` service by `POST "openvpn" to /services`.

After some time you decided to stop providing `openvpn` and switch to `wireguard`. You need to do `DELETE /services/openvpn` and `POST "wireguard" to /services`. 

All these steps should be done without node restart.

tequilapi/endpoints/service.go
soffokl: hm, probably description is not clear here. 202 response says that request accepted for processing, but it's not finished. 
Let me fix description for something better.

tequilapi/endpoints/service.go
soffokl: ID is a unique identifier for the service. It can be get from the list of services `GET /services`

There are two possible options that I see now: 

1. Something like UUID just to have a unique identifier;
1. Service type like `openvpn`, `wireguard`, `noop`. Since there is no point to provide more than one services of the particular type.

@mysteriumnetwork/core what do you think about this ID, what should we use? 

I personally like the second option more if we are really not planning to provide several services of the single type from one node.

donce: > ID is a unique identifier for the service. It can be get from the list of services `GET /services`

But I don't see any id in `serviceInfo` struct, which is returned from `GET /services`:
```go
	type serviceInfo struct {
	Proposal proposalRes `json:"proposal"`
	// example: Running
	Status  string         `json:"status"`
	Options serviceOptions `json:"options"`
}
```

> Service type like `openvpn`, `wireguard`, `noop`
This makes sense, but in that case `ServiceRequestDTO` should not require `providerID` as well - ID should be consistent across different endpoints.

soffokl: `serviceInfo` includes a proposal, which has both `providerID` and `ServiceType`. 

For the starting service, `ServiceRequestDTO` requires `providerID` to define under which identity service should be started. We cannot skip it.

Waldz: I dont think we should mix `proposalId` with `serviceInstanceId`

soffokl: The question is do we need serviceInstanceID at all, or it’s enough to have only serviceType as identifier?

Waldz: We talked about that  with @zolia. No point to add any limitations from technical side:
- if several Openvpn services can be started, why do deny it
- if we decide to limit that, we should do it in `serviceCreate` validators

soffokl: Added mocked ID for services

PR #741, opened 2019-02-14
Use basic type for proposalRes.Metrics to fix tequilapi documentation

donce:
Using `json.RawMessage` means that swagger will use description from `RawMessage` object, which is not very useful:

screenshot 2019-02-14 at 17 07 43
screenshot 2019-02-14 at 17 07 26

If we want to have our own description, we can achieve that by using primitive data type (`[]byte`):

screenshot 2019-02-14 at 17 06 56
screenshot 2019-02-14 at 17 07 03

Description works. As for example, it is not perfect, because `metrics` field is shown as a `string`, but it's much more useful and less confusing than `[0]` :D


vkuznecovas:
failed tests indicate that something aint right

Issue #731, opened 2019-02-13
Move 'myst service' subcommand to Tequilapi endpoint `POST /service`


soffokl:
`service` subcommand for `myst` binary should use Tequilapi endpoint for manipulating services of the node.

PR #744, opened 2019-02-19
Pluggable services manager

soffokl:
This PR will also include https://github.com/mysteriumnetwork/node/issues/731 
The service manager was changed in this PR, and it's better to change `service` subcommand here as well.


tequilapi/endpoints/service.go
```diff
@@ -164,11 +175,40 @@ func (se *ServiceEndpoint) ServiceStart(resp http.ResponseWriter, req *http.Requ
...
+
+	for _, instance := range se.serviceManager.List() {
+		proposal := instance.Proposal()
+		if proposal.ProviderID == sr.ProviderID && proposal.ServiceType == sr.ServiceType {
+			utils.SendErrorMessage(resp, "Service already running", http.StatusConflict)
+			return
+		}
+	}
+
+	if err := se.identityManager.Unlock(sr.ProviderID, sr.Passphrase); err != nil {
```
soffokl: Not sure if we need to do unlock automatically, or force a user to request unlock before starting services.

tequilapi/endpoints/service.go
```diff
@@ -106,13 +115,15 @@ func (se *ServiceEndpoint) ServiceList(resp http.ResponseWriter, _ *http.Request
 //     schema:
 //       "$ref": "#/definitions/ErrorMessageDTO"
 func (se *ServiceEndpoint) ServiceGet(resp http.ResponseWriter, _ *http.Request, params httprouter.Params) {
-	statusResponse := toServiceInfoResponse(se.serviceState)
-	statusResponse.ID = params.ByName("id")
-	if statusResponse.ID != "6ba7b810-9dad-11d1-80b4-00c04fd430c8" && statusResponse.ID != "7037d6f6-1b92-4513-8cb4-8fd1a944b996" {
-		utils.SendErrorMessage(resp, "Service not found", http.StatusNotFound)
-		return
+	for _, instance := range se.serviceManager.List() {
```
vkuznecovas: this lookup could also be made easier, if we'd have a map instead of a list there :)

tequilapi/endpoints/service.go
vkuznecovas: I think the auto unlock is a good option, but we need appropriate handling to let the user know if the unlock failed because the passphrase provided was invalid or there was no passphrase provided

tequilapi/endpoints/service.go
```diff
@@ -189,18 +229,22 @@ func (se *ServiceEndpoint) ServiceStart(resp http.ResponseWriter, req *http.Requ
 //     schema:
 //       "$ref": "#/definitions/ErrorMessageDTO"
 func (se *ServiceEndpoint) ServiceStop(resp http.ResponseWriter, _ *http.Request, params httprouter.Params) {
-	if params.ByName("id") != "6ba7b810-9dad-11d1-80b4-00c04fd430c8" && params.ByName("id") != "7037d6f6-1b92-4513-8cb4-8fd1a944b996" {
-		utils.SendErrorMessage(resp, "Service not found", http.StatusNotFound)
-		return
+	for _, instance := range se.serviceManager.List() {
```
vkuznecovas: this lookup could also be made easier, if we'd have a map instead of a list there :)

tequilapi/endpoints/service.go
```diff
@@ -164,11 +175,40 @@ func (se *ServiceEndpoint) ServiceStart(resp http.ResponseWriter, req *http.Requ
...
+		return
+	}
+
+	options, err := optionsParser(sr.Options)
+	if err != nil {
+		utils.SendErrorMessage(resp, "Invalid options", http.StatusBadRequest)
+		return
+	}
+
+	for _, instance := range se.serviceManager.List() {
```
Waldz: You need some helper function in `servicePool` here

tequilapi/endpoints/service.go
```diff
@@ -164,11 +175,40 @@ func (se *ServiceEndpoint) ServiceStart(resp http.ResponseWriter, req *http.Requ
...
+		proposal := instance.Proposal()
+		if proposal.ProviderID == sr.ProviderID && proposal.ServiceType == sr.ServiceType {
```
Waldz: This looks like validator, acts like validator, should be a validator :)

tequilapi/endpoints/service.go
Waldz: I would like to follow atomic action in `POST /service` endpoint, same as we have in `POST /connection`.
That means caller of endpoint should care about identify unlocking.

e.g. CLI command `myst service` does such actions:
- lists current identities
- Autoselect provider identity and unlock it
- Call Service start endpoint

tequilapi/endpoints/service.go
Waldz: And I suggest to have have helper functions on `Service.Pool` side. This would leave `ServiceEndpoint` thin -  thin controller and fat Service from MVC principle

tequilapi/endpoints/service.go
```diff
@@ -215,17 +259,23 @@ func toServiceRequest(req *http.Request) (*serviceRequest, error) {
...
 }
 
-func toServiceListResponse(state service.State) serviceList {
-	return serviceList{
-		{ID: "6ba7b810-9dad-11d1-80b4-00c04fd430c8", Status: string(state)},
-		{ID: "7037d6f6-1b92-4513-8cb4-8fd1a944b996", Status: string(state)},
+func toServiceListResponse(instances []*service.Instance) (res serviceList) {
+	for _, instance := range instances {
+		res = append(res, serviceInfo{
+			Status:   string(service.Running),
```
Waldz: Starting should happen just before `se.serviceManager.Start`. And I miss `service.Instance` states to be isolated and handled on `service.Manager` side

tequilapi/endpoints/service.go
soffokl: Removed auto unlock for Tequilapi endpoint. We are getting an explicit error if identity was not unlocked.

cmd/di.go
```diff
@@ -140,6 +140,9 @@ func (di *Dependencies) Bootstrap(nodeOptions node.Options) error {
 
 	di.bootstrapIdentityComponents(nodeOptions)
 	di.bootstrapLocationComponents(nodeOptions.Location, nodeOptions.Directories.Config)
+	if err := di.BootstrapServices(nodeOptions); err != nil {
```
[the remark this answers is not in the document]
soffokl: Changed it to work on mobile too.

core/service/manager.go
```diff
@@ -78,49 +79,69 @@ type Manager struct {
...
+		}
 
-	discovery.Wait()
-	return nil
+		discovery.Wait()
+	}()
+	return instance, nil
 }
 
-// Kill stops service
```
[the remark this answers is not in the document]
soffokl: Added it.

tequilapi/endpoints/service.go
soffokl: Changed it.

cmd/di.go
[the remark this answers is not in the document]
soffokl: Changed it.

core/service/manager.go
```diff
@@ -78,49 +79,69 @@ type Manager struct {
...
+func (manager *Manager) List() []*Instance {
```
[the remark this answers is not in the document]
soffokl: The point was to keep the single service manager that will manage services, start, stop or list it.
Added several extra dependencies instead of one will make service endpoint more complicated.

core/service/pool.go
```diff
@@ -48,12 +52,22 @@ func NewPool() *Pool {
...
-func (sr *Pool) Add(instance *Instance) {
-	sr.instances = append(sr.instances, instance)
+func (p *Pool) Add(instance *Instance) {
+	p.instances = append(p.instances, instance)
+}
+
+// Del removes a service from running instances pool
+func (p *Pool) Del(instance *Instance) {
+	for i, item := range p.instances {
+		if instance == item {
```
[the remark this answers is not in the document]
soffokl: Using an ID and map now.

core/service/pool.go
```diff
@@ -65,15 +79,36 @@ func (sr *Pool) Stop(instance *Instance) error {
...
 	}
 
 	return errStop.Errorf("Some instances did not stop: %v", ". ")
 }
+
+func (p *Pool) List() []*Instance {
+	return p.instances
+}
+
+func (i *Instance) Proposal() market.ServiceProposal {
```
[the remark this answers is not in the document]
soffokl: Done.

services/noop/flags.go
```diff
@@ -18,14 +18,23 @@
...
 
-// ParseFlags function fills in Noop options from CLI context
-func ParseFlags(_ *cli.Context) service.Options {
+// ParseCLIFlags function fills in Noop options from CLI context
+func ParseCLIFlags(_ *cli.Context) service.Options {
 	return Options{}
 }
+
+// ParseJSONOptions function fills in Noop options from JSON request
+func ParseJSONOptions(request json.RawMessage) (service.Options, error) {
```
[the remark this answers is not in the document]
soffokl: Renamed it to `options.go` since we have options filling there, both for arguments and JSON requests. It's about service options only, not the whole service request.

tequilapi/endpoints/service.go
soffokl: Changed it.

tequilapi/endpoints/service.go
soffokl: Extracted it.

tequilapi/endpoints/service.go
soffokl: The state is now in the `service.Instance`.

tequilapi/endpoints/service.go
```diff
@@ -189,58 +222,82 @@ func (se *ServiceEndpoint) ServiceStart(resp http.ResponseWriter, req *http.Requ
...
 	}
 	return errors
 }
 
 // ServiceManager represents service manager that will be used for manipulation node services.
 type ServiceManager interface {
-	Start(providerID identity.Identity, serviceType string, options service.Options) (err error)
+	Start(providerID identity.Identity, serviceType string, options service.Options) (service.ID, error)
+	Stop(id service.ID) error
+	Service(id service.ID) *service.Instance
```
zolia: IMHO it should be a verb. Interface provides actions.

tequilapi/endpoints/service.go
```diff
@@ -35,11 +35,17 @@ type serviceRequest struct {
...
+	// required: true
+	Passphrase string `json:"passphrase"`
+
 	// service type. Possible values are "openvpn", "wireguard" and "noop"
 	// required: false
 	// default: openvpn
 	// example: openvpn
 	ServiceType string `json:"serviceType"`
+
+	Options json.RawMessage `json:"options"`
```
Waldz: Documentation :(

cmd/commands/service/command.go
```diff
@@ -66,24 +65,25 @@ func NewCommand(licenseCommandName string) *cli.Command {
...
 
+			errorChannel := make(chan error, 3)
 			nodeOptions := cmd.ParseFlagsNode(ctx)
 			if err := di.Bootstrap(nodeOptions); err != nil {
 				return err
 			}
-			if err := di.BootstrapServices(nodeOptions); err != nil {
-				return err
-			}
+			go func() { errorChannel <- di.Node.Wait() }()
```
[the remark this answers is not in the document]
Waldz: - Looks like `errorChannel` was duplicated and not aligned  with `runErrors` (let's leave just this one)
- Looks like `RegisterSignalCallback` was duplicated also

tequilapi/endpoints/service.go
```diff
@@ -164,11 +174,34 @@ func (se *ServiceEndpoint) ServiceStart(resp http.ResponseWriter, req *http.Requ
...
+		return
+	}
+
+	instance := se.serviceManager.Service(id)
 
 	resp.WriteHeader(http.StatusCreated)
-	statusResponse := toServiceInfoResponse(se.serviceState)
-	statusResponse.ID = "6ba7b810-9dad-11d1-80b4-00c04fd430c8"
+	statusResponse := toServiceInfoResponse(id, instance)
+	statusResponse.Status = string(service.Starting)
```
Waldz: `toServiceInfoResponse()` already has this fields mapping, dont map 2nd time

tequilapi/endpoints/service.go
```diff
@@ -189,58 +222,82 @@ func (se *ServiceEndpoint) ServiceStart(resp http.ResponseWriter, req *http.Requ
...
+	err = json.NewDecoder(req.Body).Decode(&sr)
+	return sr, err
 }
 
-func toServiceInfoResponse(state service.State) serviceInfo {
+func toServiceInfoResponse(id service.ID, instance *service.Instance) serviceInfo {
 	return serviceInfo{
-		Status: string(state),
+		ID:       string(id),
+		Status:   string(service.Running),
```
Waldz: State shouldn't be hardcoded here

tequilapi/endpoints/service.go
```diff
@@ -189,58 +222,82 @@ func (se *ServiceEndpoint) ServiceStart(resp http.ResponseWriter, req *http.Requ
...
 }
 
-func toServiceListResponse(state service.State) serviceList {
-	return serviceList{
-		{ID: "6ba7b810-9dad-11d1-80b4-00c04fd430c8", Status: string(state)},
-		{ID: "7037d6f6-1b92-4513-8cb4-8fd1a944b996", Status: string(state)},
+func toServiceListResponse(instances map[service.ID]*service.Instance) serviceList {
+	res := make([]serviceInfo, 0)
+	for id, instance := range instances {
+		res = append(res, serviceInfo{
```
Waldz: You could reuse `toServiceInfoResponse()` here

core/service/manager.go
```diff
@@ -78,49 +79,75 @@ type Manager struct {
...
+	if err != nil {
+		return id, err
+	}
 
-	discovery.Wait()
-	return nil
+	go func() {
+		instance.state = Running
+		err = service.Serve(providerID)
+		instance.state = NotRunning
```
Waldz: State  is handled very strange here, why it's becoming to `NotRunning` immediately

core/service/manager.go
```diff
@@ -78,49 +79,75 @@ type Manager struct {
...
+		if err != nil {
+			log.Error("Service serve failed: ", err)
+		}
+
```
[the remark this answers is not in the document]
Waldz: That's exactly why `Serve()` was called before `servicePool.Add()`. Why reordering was done?

core/service/manager.go
```diff
@@ -78,49 +79,69 @@ type Manager struct {
...
 		discovery:    discovery,
-	})
+	}
+
+	manager.servicePool.Add(&instance)
+
+	go func() {
+		err = service.Serve(providerID)
+		if err != nil {
+			log.Error("Service serve failed: ", err)
```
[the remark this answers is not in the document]
soffokl: We are returning errors only for the preparation step if we failed to do initial starting. Serving service does not return an error, it only prints a log.

cmd/commands/service/command.go
```diff
@@ -120,45 +119,37 @@ func (c *serviceCommand) Run(ctx *cli.Context) (err error) {
...
+		go sc.runService(providerID, serviceType, options)
 	}
 
 	return nil
 }
 
-func (c *serviceCommand) runService(providerID identity.Identity, serviceType string, options service.Options) {
-	err := c.di.ServicesManager.Start(providerID, serviceType, options)
-	if err == service.ErrorLocation {
-		printLocationWarning("myst")
```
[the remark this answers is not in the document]
soffokl: We are not starting `di.ServiceManager.Start()` in the `service` subcommand. We are doing call to `tequilapi.ServiceStart` which is not waits service start, and cannot return `service.ErrorLocation` error.

core/service/pool.go
```diff
@@ -18,42 +18,52 @@
...
+	if err != nil {
+		return id, err
+	}
+
+	p.instances[id] = instance
+	return id, nil
+}
+
+// Del removes a service from running instances pool
+func (p *Pool) Del(id ID) {
```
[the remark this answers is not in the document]
soffokl: `p.Del()` is called at the end of the `p.Stop()` to remove it from the pool. 
If an error happened we will have service instance in the pool in the "NotRunning" state, and `p.Stop()` will delete it from the pool.

core/service/manager.go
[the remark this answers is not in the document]
soffokl: `Serve()` is a blocking operation. before `servicePool.Add()` was after `Serve()` and never worked. It was never reached in a normal state. 

On failed serve, we are marking service as "NotRunning" and keeping it in the pool to let a user ability to get this information from the Tequilapi endpoint. And the user can delete it from the pool by calling service `stop`.

tequilapi/endpoints/service.go
soffokl: It can be called `GetService`, but it sounds like [it's neither idiomatic nor necessary to put Get into the getter's name](https://golang.org/doc/effective_go.html#Getters).

core/service/manager.go
soffokl: `Serve` is blocking operation, the state became `NotRunning` only when service finished to `Serve`.

cmd/commands/service/command.go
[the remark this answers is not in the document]
soffokl: `errorChannel` will reach return if an error happened in one of the components.
`cmdService.Run(ctx)` and `di.Node.Wait()` can return error during the work. 
If none of them returned an error we are waiting for `cmd.RegisterSignalCallback(func() { errorChannel <- nil })`

cmd/commands/service/command.go
```diff
@@ -66,24 +65,25 @@ func NewCommand(licenseCommandName string) *cli.Command {
...
+				tequilapi: client.NewClient(nodeOptions.TequilapiAddress, nodeOptions.TequilapiPort),
+				runErrors: errorChannel,
 			}
-			return cmdService.Run(ctx)
+
+			go func() {
+				errorChannel <- cmdService.Run(ctx)
+			}()
+
+			return <-errorChannel
```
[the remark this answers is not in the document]
soffokl: We have 3 components that can return an error, but we are reading only the first one, others became blocked to send error waiting for reader forever. 
With a buffered channel we are receiving all errors not blocking writers.

cmd/commands/service/serviceTypes.go
```diff
@@ -24,15 +22,17 @@ import (
 	service_noop "github.com/mysteriumnetwork/node/services/noop"
 	service_openvpn "github.com/mysteriumnetwork/node/services/openvpn"
 	openvpn_service "github.com/mysteriumnetwork/node/services/openvpn/service"
+	service_wireguard "github.com/mysteriumnetwork/node/services/wireguard"
+	wireguard_service "github.com/mysteriumnetwork/node/services/wireguard/service"
```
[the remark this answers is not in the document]
soffokl: Some old naming style left here. Renamed it now.

tequilapi/endpoints/service.go
soffokl: Added description that every service will have own list of options.

tequilapi/endpoints/service.go
soffokl: Fixed.

tequilapi/endpoints/service.go
soffokl: Fixed.

tequilapi/endpoints/service.go
soffokl: Done.

cmd/commands/service/command.go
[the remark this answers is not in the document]
zolia: But as you noted above, if we get an error in any of sub components, we instantly return with `return <-errorChannel`. Other sub components are in routines and probably is does not matter if they will block. Execution will be finished anyways.

core/service/manager.go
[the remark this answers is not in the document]
zolia: OK, makes sense.

cmd/commands/service/command.go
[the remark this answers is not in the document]
soffokl: hm, right. Initially, it was added a long time ago because it was not able to stop properly. I'll test different cases with non-buffered channels to make sure how it works.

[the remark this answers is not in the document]
soffokl: Changed it to unbuffered channel.

cmd/commands/service/command.go
```diff
@@ -120,45 +125,36 @@ func (c *serviceCommand) Run(ctx *cli.Context) (err error) {
 	// 1 for the signal callback
 	// 1 for the node.Wait()
 	// 1 for each of the services
-	c.runErrors = make(chan error, 2+len(serviceTypes))
-	go c.runNode(ctx)
-	c.runServices(ctx, providerID, serviceTypes)
-
-	cmd.RegisterSignalCallback(func() { c.runErrors <- nil })
+	sc.runErrors = make(chan error, 2+len(serviceTypes))
```
Waldz: `errorChannel` and `runErrors` still not aligned and looks duplicated

soffokl: merged it together.

cmd/commands/service/command.go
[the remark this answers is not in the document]
Waldz: IMHO, not detected Location should 400 be error in Tequila `POST /services` endpoint

zolia:
branch outdated..

cmd/commands/service/command.go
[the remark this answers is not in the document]
soffokl: Changed it to be a 400 error.

PR #745, opened 2019-02-19
We now send the initial promise info to the consumer 

### LINKED ISSUE #714: Initial promise state
Currently, the paymentIssuerFactory in the connection manager starts up with an empty promise.State{}. This needs to be preloaded with the amount the provider gives in the PaymentInfo struct.

vkuznecovas:
Closes #714

### DIFF
--- a/core/connection/manager.go
+++ b/core/connection/manager.go
@@ -249,7 +249,7 @@ func (manager *connectionManager) startConnection(
 
 	defer func() {
 		if err != nil {
-			log.Info(managerLogPrefix, "Cancelling connection initiation", err)
+			log.Info(managerLogPrefix, "Cancelling connection initiation: ", err)
 			logDisconnectError(manager.Disconnect())
 		}
 	}()
--- a/core/connection/manager.go
+++ b/core/connection/manager.go
@@ -81,11 +81,12 @@ type connectionManager struct {
 	eventPublisher       Publisher
 
 	//these are populated by Connect at runtime
-	ctx             context.Context
-	status          Status
-	statusLock      sync.RWMutex
-	sessionInfo     SessionInfo
-	cleanConnection func()
+	ctx         context.Context
+	status      Status
+	statusLock  sync.RWMutex
+	sessionInfo SessionInfo
+	cleanup     []func()
+	cancel      func()
 
 	discoLock sync.Mutex
 }
@@ -102,8 +103,8 @@ func NewManager(
 		paymentIssuerFactory: paymentIssuerFactory,
 		newConnection:        connectionCreator,
 		status:               statusNotConnected(),
-		cleanConnection:      warnOnClean,
 		eventPublisher:       eventPublisher,
+		cleanup:              make([]func(), 0),
 	}
 }
 
@@ -112,7 +113,8 @@ func (manager *connectionManager) Connect(consumerID identity.Identity, proposal
 		return ErrAlreadyExists
 	}
 
-	manager.ctx, manager.cleanConnection = context.WithCancel(context.Background())
+	manager.ctx, manager.cancel = context.WithCancel(context.Background())
+
 	manager.setStatus(statusConnecting())
 	defer func() {
 		if err != nil {
@@ -120,36 +122,12 @@ func (manager *connectionManager) Connect(consumerID identity.Identity, proposal
 		}
 	}()
 
-	err = manager.startConnection(consumerID, proposal, params)
-	if err == context.Canceled {
-		return ErrConnectionCancelled
-	}
-	return err
-}
-
-func (manager *connectionManager) startConnection(consumerID identity.Identity, proposal market.ServiceProposal, params ConnectParams) (err error) {
-	cancelCtx := manager.cleanConnection
-
-	var cancel []func()
-	defer func() {
-		manager.cleanConnection = func() {
-			cancelCtx()
-			for i := range cancel { // Cancelling in a reverse order to keep correct workflow.
-				cancel[len(cancel)-i-1]()
-			}
-		}
-		if err != nil {
-			log.Info(managerLogPrefix, "Cancelling connection initiation", err)
-			logDisconnectError(manager.Disconnect())
-		}
-	}()
-
 	providerID := identity.FromAddress(proposal.ProviderID)
-	dialog, err := manager.newDialog(consumerID, providerID, proposal.ProviderContacts[0])
+
+	dialog, err := manager.createDialog(consumerID, providerID, proposal.ProviderContacts[0])
 	if err != nil {
 		return err
 	}
-	cancel = append(cancel, func() { dialog.Close() })
 
 	stateChannel := make(chan State, 10)
 	statisticsChannel := make(chan consumer.SessionStatistics, 10)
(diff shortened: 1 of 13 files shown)


core/connection/manager.go
```diff
@@ -166,32 +166,37 @@ func (manager *connectionManager) startConnection(consumerID identity.Identity,
...
-		IssuerID:          consumerID,
-		MystClientVersion: metadata.VersionAsString(),
+	var promiseState promise.State
+	if paymentInfo != nil {
+		promiseState.Amount = paymentInfo.LastPromise.Amount
+		promiseState.Seq = paymentInfo.LastPromise.SequenceID
 	}
 
-	sessionID, sessionConfig, err := session.RequestSessionCreate(dialog, proposal.ID, sessionCreateConfig, consumerInfo)
+	payments, err := manager.paymentIssuerFactory(promiseState, messageChan, dialog, consumerID, providerID)
```
Waldz: `startConnection()` should start connection, I strongly disagree to grow it with unrelated logic.

Waldz:
```
func (manager *connectionManager) Connect(
  ..
  manager.startConnection(consumerID, proposal, params)
  go manager.payForService(dialog, session, connectionSubscription)`
)

File: session/create_consumer.go [hunk 2]
```diff
@@ -70,14 +80,33 @@ func (consumer *createConsumer) Consume(requestPtr interface{}) (response interf
 				destroyCallback()
 			}()
 		}
-		return responseWithSession(sessionInstance, config, nil), nil
+		return responseWithSession(sessionInstance, config, consumer.loadPaymentInfo(issuerID)), nil
```
Waldz: Communication consumer/provider is meant to be thin in term of business logic, it's just a transport layer with mapping to communication DTOs. So then we can refactor them easily later.

Back in a day I followed MVC principle *thin Controller, fat Service* here:

- `createConsumer` - is meant to be a Controller here. So thin and just calls `consumer.SessionManager.Create()`
- `SessionManager. sessionInstance` - is meant to be a Service

core/connection/manager.go
vkuznecovas: refactored

session/create_consumer.go
vkuznecovas: We have 2 services, the lookup of last promise and session creation. They are called, and the response is remapped to a dto.

core/connection/manager.go
```diff
@@ -206,29 +223,48 @@ func (manager *connectionManager) startConnection(consumerID identity.Identity,
...
 
+	return s, paymentInfo, nil
+}
+
+func (manager *connectionManager) startConnection(
+	connection Connection,
+	consumerID identity.Identity,
+	proposal market.ServiceProposal,
+	params ConnectParams,
+	sessionDTO session.SessionDto,
```
soffokl: I'd name it just `session` without `DTO`.

vkuznecovas: conflicts with the package name

core/connection/manager.go
```diff
@@ -159,39 +137,83 @@ func (manager *connectionManager) startConnection(consumerID identity.Identity,
...
 
-	// TODO: load initial promise state
-	payments, err := manager.paymentIssuerFactory(promise.State{}, messageChan, dialog, consumerID, providerID)
+	payments, err := manager.paymentIssuerFactory(promiseState, messageChan, dialog, consumerID, providerID)
 	if err != nil {
 		return err
 	}
 
-	cancel = append(cancel, func() { payments.Stop() })
+	manager.cleanup = append(manager.cleanup, func() { payments.Stop() })
```
zolia: should be enough:
```suggestion
	manager.cleanup = append(manager.cleanup, payments.Stop)
```
We should write function wrapper only in cases we have not a real function but some channel. There are more such places with append.

vkuznecovas: fixed this one, all the other places do need a wrapper since we're using `func()` as the array type, and all the others tend to return the error. We could switch this to using the `func() error` for the array type, but I don't think we'd save much of anything

core/connection/manager.go
zolia: can be `sessionData`

session/create_consumer.go
```diff
@@ -70,14 +80,33 @@ func (consumer *createConsumer) Consume(requestPtr interface{}) (response interf
...
 	default:
 		return responseInternalError, nil
 	}
 }
 
+func (consumer *createConsumer) loadPaymentInfo(issuerID identity.Identity) *PaymentInfo {
+	sp, err := consumer.promiseLoader.GetLastPromise(issuerID)
+	if err != nil {
+		log.Trace(consumerLogPrefix, "could not load promise info, defaulting to nil payment info", err)
+		return nil
```
zolia: Here we probably should return empty, but initialised zero `PaymentInfo`. Further logic would not have to deal with nil pointer.

session/create_consumer.go
```diff
@@ -70,14 +80,33 @@ func (consumer *createConsumer) Consume(requestPtr interface{}) (response interf
...
-		return responseWithSession(sessionInstance, config, nil), nil
+		return responseWithSession(sessionInstance, config, consumer.loadPaymentInfo(issuerID)), nil
 	case ErrorInvalidProposal:
 		return responseInvalidProposal, nil
 	default:
 		return responseInternalError, nil
 	}
 }
 
+func (consumer *createConsumer) loadPaymentInfo(issuerID identity.Identity) *PaymentInfo {
```
zolia: You are testing private method in this case. Imho it should be tested, but maybe it can go to some payment related package and made public? Promise is cross session persistent. It is not something unique to a single session.

core/connection/manager.go
zolia: but thats a bit smelly then :) we might be swallowing errors. Maybe we can collect '[]func() error' then and print out errors when interating through cleanup? Would be bonus not having those nasty wrappers and all error accounted.

session/promise/storage.go
```diff
@@ -23,11 +23,13 @@ import (
 	"sync"
 	"time"
 
+	"github.com/ethereum/go-ethereum/log"
```
soffokl:
```suggestion
	log "github.com/cihub/seelog"
```

vkuznecovas: oh god damn

session/create_consumer.go
Waldz: Still not fixed, we need to make `SessionManager.Create()` the only class to make connections in network

session/create_producer_test.go
```diff
@@ -47,8 +47,9 @@ func TestProducer_RequestSessionCreate(t *testing.T) {
 	sender := &fakeSender{}
 	sid, config, err := RequestSessionCreate(sender, 123, []byte{}, ConsumerInfo{})
```
soffokl: I guess `sid` was for `sessionID`, but now it looks strange: `sid.ID`.

session/destroy_producer_test.go
```diff
@@ -36,7 +36,7 @@ func TestProducer_RequestSessionDestroy(t *testing.T) {
 	assert.NoError(t, err)
 
 	destroySender := &fakeDestroySender{}
-	err = RequestSessionDestroy(destroySender, sid)
+	err = RequestSessionDestroy(destroySender, sid.ID)
```
soffokl: I guess `sid` was for `sessionID`, but now it looks strange: `sid.ID`.

session/create_producer_test.go
soffokl: And looks like it's not a `config` anymore?

vkuznecovas: changed both

PR #746, opened 2019-02-19
Payments/missing storage fields

vkuznecovas:
Added the consumer/receiver fields to the promise storage, as those will be needed.

### DIFF
--- a/cmd/di.go
+++ b/cmd/di.go
@@ -302,7 +302,7 @@ func newSessionManagerFactory(
 	nodeOptions node.Options,
 ) session.ManagerFactory {
 	return func(dialog communication.Dialog) *session.Manager {
-		providerBalanceTrackerFactory := func(consumer, receiver, issuer identity.Identity) (session.BalanceTracker, error) {
+		providerBalanceTrackerFactory := func(consumerID, receiverID, issuerID identity.Identity) (session.BalanceTracker, error) {
 			// if the flag ain't set, just return a noop balance tracker
 			if !nodeOptions.ExperimentPayments {
 				return payments_noop.NewSessionBalance(), nil
@@ -328,8 +328,8 @@ func newSessionManagerFactory(
 
 			// TODO: the ints and times here need to be passed in as well, or defined as constants
 			tracker := balance_provider.NewBalanceTracker(&timeTracker, amountCalc, 0)
-			validator := validators.NewIssuedPromiseValidator(consumer, receiver, issuer)
-			return session_payment.NewSessionBalance(sender, tracker, promiseChan, time.Second*5, time.Second*1, validator, promiseStorage, consumer, receiver, issuer), nil
+			validator := validators.NewIssuedPromiseValidator(consumerID, receiverID, issuerID)
+			return session_payment.NewSessionBalance(sender, tracker, promiseChan, time.Second*5, time.Second*1, validator, promiseStorage, consumerID, receiverID, issuerID), nil
 		}
 		return session.NewManager(
 			proposal,
--- a/cmd/di.go
+++ b/cmd/di.go
@@ -302,7 +302,7 @@ func newSessionManagerFactory(
 	nodeOptions node.Options,
 ) session.ManagerFactory {
 	return func(dialog communication.Dialog) *session.Manager {
-		providerBalanceTrackerFactory := func(consumer, provider, issuer identity.Identity) (session.BalanceTracker, error) {
+		providerBalanceTrackerFactory := func(consumer, receiver, issuer identity.Identity) (session.BalanceTracker, error) {
 			// if the flag ain't set, just return a noop balance tracker
 			if !nodeOptions.ExperimentPayments {
 				return payments_noop.NewSessionBalance(), nil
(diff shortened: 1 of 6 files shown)


session/promise/storage_test.go
```diff
@@ -47,7 +47,7 @@ var (
 func Test_Storage_IssuesOneForUnknownID(t *testing.T) {
 	ms := newMockStorage(nil)
 	s := NewStorage(ms)
-	res, err := s.GetNewSeqIDForIssuer(id)
+	res, err := s.GetNewSeqIDForIssuer(id, id, id)
```
soffokl: `(id, id, id)` is scary stuff. :)

vkuznecovas: I'll refactor the tests a bit to look a bit less intimidating :D

session/payment/session_balance.go
```diff
@@ -31,9 +32,10 @@ import (
 
 // PromiseStorage stores the promises and issues new sequenceID's
 type PromiseStorage interface {
-	GetNewSeqIDForIssuer(issuerID identity.Identity) (uint64, error)
+	GetNewSeqIDForIssuer(consumer, receiver, issuer identity.Identity) (uint64, error)
```
Waldz: Same here

session/payment/session_balance.go
```diff
@@ -126,14 +135,73 @@ func (sb *SessionBalance) Start() error {
...
+		return consumerPromise, nil
+	}
+
 	sb.sequenceID = lastPromise.SequenceID
 	return lastPromise, nil
 }
 
+var errNoPromiseForConsumer = errors.New("no promise for consumer")
+
+func (sb *SessionBalance) findPromiseForConsumer() (promise.StoredPromise, error) {
```
Waldz: Feels like sorting and filtering helpers should be on `PromiseStorage` side

cmd/di.go
```diff
@@ -328,8 +328,8 @@ func newSessionManagerFactory(
 
 			// TODO: the ints and times here need to be passed in as well, or defined as constants
 			tracker := balance_provider.NewBalanceTracker(&timeTracker, amountCalc, 0)
-			validator := validators.NewIssuedPromiseValidator(consumer, provider, issuer)
-			return session_payment.NewSessionBalance(sender, tracker, promiseChan, time.Second*5, time.Second*1, validator, promiseStorage, issuer), nil
+			validator := validators.NewIssuedPromiseValidator(consumer, receiver, issuer)
```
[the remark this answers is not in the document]
vkuznecovas: this is the ID of the party who'll receive the promised amount.

session/payment/session_balance.go
vkuznecovas: moved

session/promise/storage.go
```diff
@@ -146,6 +147,45 @@ func (s *Storage) GetAllKnownIssuers() []identity.Identity {
...
+func (s *Storage) FindPromiseForConsumer(issuerID, consumerID identity.Identity) (StoredPromise, error) {
+	s.Lock()
+	defer s.Unlock()
+	promises, err := s.getAllPromisesForIssuer(issuerID)
+	if err != nil {
+		return StoredPromise{}, err
+	}
+
+	// sort by sequenceID, descending
+	sort.Slice(promises, func(i, j int) bool {
```
soffokl: We probably can just call `sortPromisesDesc` here.

session/promise/storage.go
```diff
@@ -142,6 +147,45 @@ func (s *Storage) GetAllKnownIssuers() []identity.Identity {
...
+		// otherwise, we're free to extend
+		if promises[i].ConsumerID == consumerID {
+			return promises[i], nil
+		}
+	}
+
+	return StoredPromise{}, errNoPromiseForConsumer
+}
+
+func (s *Storage) sortPromisesDesc(promises []StoredPromise) {
```
soffokl: Not sure if it should be a method of the `*Storage`. It feels like we are sorting it inside the storage.

session/promise/storage.go
vkuznecovas: fixd

session/promise/storage.go
vkuznecovas: fixd

session/payment/session_balance.go
```diff
@@ -124,12 +136,43 @@ func (sb *SessionBalance) Start() error {
...
+
+	If it does not - we'll need to do a thorough lookup since there's a case here,
+	where we might have issued a tiny amount that the provider did not clear,
+	but did clear the following promises.
+
+	In this case we'll need to issue a new id.
+	If none of the promises were cleared, we can reuse the old one.
+
+	TODO: This could turn out to be expensive, and we might just want to skip this check and issue a new ID instead.
+	*/
```
zolia: I suggest not to invest time too much here for now since our new payment model does this a bit differently with a hash instead of seq id. In such case we could have a separte hash for each consumer even if promise issuer is the same.

vkuznecovas: Since it's already implemented might as well just leave it here till we get more details on how it will actually work.

PR #747, opened 2019-02-22
Added service to CLI subcommand

### LINKED ISSUE #733: CLI demo subcommand 'service start' starts service through Tequilapi
CLI `service` subcommand should use Tequilapi endpoints for manipulating services of the node.

soffokl:
Closes #733

### DIFF
--- a/cmd/commands/cli/command.go
+++ b/cmd/commands/cli/command.go
@@ -18,6 +18,7 @@
 package cli
 
 import (
+	"encoding/json"
 	"fmt"
 	"log"
 	"path/filepath"
@@ -137,6 +138,7 @@ func (c *cliApp) handleActions(line string) {
 		{command: "license", handler: c.license},
 		{command: "registration", handler: c.registration},
 		{command: "proposals", handler: c.proposals},
+		{command: "service", handler: c.service},
 	}
 
 	for _, cmd := range staticCmds {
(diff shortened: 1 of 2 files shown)


cmd/commands/cli/command.go
```diff
@@ -159,20 +161,107 @@ func (c *cliApp) handleActions(line string) {
 	}
 }
 
+func (c *cliApp) service(argsString string) {
+	var action, providerID, serviceType, id string
+	var options json.RawMessage
+
+	help := func() {
```
vkuznecovas: i'd rather this was called printHelp

cmd/commands/cli/command.go
```diff
@@ -159,20 +161,107 @@ func (c *cliApp) handleActions(line string) {
...
+
+	args := strings.Fields(argsString)
+	if len(args) == 0 {
+		help()
+		return
+	}
+
+	action = args[0]
+	switch action {
+	case "start":
```
vkuznecovas: could we extract all the cases to separate functions? we'd have smaller blocks.

soffokl: Extracted.

cmd/commands/cli/command.go
```diff
@@ -159,20 +161,111 @@ func (c *cliApp) handleActions(line string) {
...
 
+func (c *cliApp) service(argsString string) {
+	printHelp := func() {
+		fmt.Println("service <action> [args]")
+		fmt.Println("	start	<ProviderID> <ServiceType> [options]")
+		fmt.Println("	stop	<ServiceID>")
+		fmt.Println("	list")
+		fmt.Println("	get	<ServiceID>")
+		fmt.Println("")
+		fmt.Println(`example: service start 0x7d5ee3557775aed0b85d691b036769c17349db23 openvpn {"port":1190, "protocol":"UDP"}`)
```
zolia: Are we sure we want to pass options this way? Such approach might not be good for scripting and command line usage. Why not use a common way?

```suggestion
		fmt.Println(`example: service start 0x7d5ee3557775aed0b85d691b036769c17349db23 openvpn 
 --port=1190 --protocol=UDP`)
```

zolia: I like how iproute2 utility organizes its commands / arguments. In that case we wouldn't even need '--'s just: 
` service start 0x7d5ee3557775aed0b85d691b036769c17349db23 openvpn port 1190 protocol UDP`

cmd/commands/cli/command.go
```diff
@@ -159,20 +161,111 @@ func (c *cliApp) handleActions(line string) {
 	}
 }
 
+func (c *cliApp) service(argsString string) {
+	printHelp := func() {
+		fmt.Println("service <action> [args]")
+		fmt.Println("	start	<ProviderID> <ServiceType> [options]")
+		fmt.Println("	stop	<ServiceID>")
+		fmt.Println("	list")
+		fmt.Println("	get	<ServiceID>")
```
Waldz: `status` maybe?

cmd/commands/cli/command.go
Waldz: and all options should be optional

soffokl: All options are optional and have the same representation as for service subcommand now.

cmd/commands/cli/command.go
soffokl: Changed it to `status`.

cmd/commands/cli/command.go
```diff
@@ -159,20 +168,112 @@ func (c *cliApp) handleActions(line string) {
 	}
 }
 
+func (c *cliApp) service(argsString string) {
+	printHelp := func() {
+		fmt.Println("service <action> [args]")
```
vkuznecovas: we could just use a multiline string as a const, something like

```golang
const serviceHelp = `service <action> [args]
	start	<ProviderID> <ServiceType> [options]
...
`
```

The print then becomes way shorter, and does not clutter the method.

cmd/commands/cli/command.go
```diff
@@ -159,20 +168,112 @@ func (c *cliApp) handleActions(line string) {
...
+		return
+	}
+
+	service, err := c.tequilapi.ServiceStart(providerID, serviceType, opts)
+	if err != nil {
+		info("Failed to start service: ", err)
+		return
+	}
+
+	status(service.Status, "ID: "+service.ID, "ProviderID: "+service.Proposal.ProviderID, "ServiceType: "+service.Proposal.ServiceType)
```
vkuznecovas: can we reformat this a bit? IMHO it would look a bit cleaner this way

```golang
status(service.Status, 
fmt.Sprintf("ID: %v",service.ID),
fmt.Sprintf("ProviderID: %v",service.Proposal.ProviderID).
..
```

cmd/commands/cli/command.go
```diff
@@ -159,20 +168,112 @@ func (c *cliApp) handleActions(line string) {
...
+
+func (c *cliApp) serviceList() {
+	services, err := c.tequilapi.Services()
+	if err != nil {
+		info("Failed to get a list of services: ", err)
+		return
+	}
+
+	for _, service := range services {
+		status(service.Status, "ID: "+service.ID, "ProviderID: "+service.Proposal.ProviderID, "ServiceType: "+service.Proposal.ServiceType)
```
vkuznecovas: same as above

cmd/commands/cli/command.go
```diff
@@ -159,20 +168,112 @@ func (c *cliApp) handleActions(line string) {
...
+}
+
+func (c *cliApp) serviceGet(id string) {
+	service, err := c.tequilapi.Service(id)
+	if err != nil {
+		info("Failed to get service info: ", err)
+		return
+	}
+
+	status(service.Status, "ID: "+service.ID, "ProviderID: "+service.Proposal.ProviderID, "ServiceType: "+service.Proposal.ServiceType)
```
vkuznecovas: same as above

cmd/commands/cli/command.go
soffokl: Changed it.

cmd/commands/cli/command.go
soffokl: Changed all of them.

cmd/commands/cli/command.go
```diff
@@ -26,14 +28,28 @@ import (
...
 
 const cliCommandName = "cli"
 
+const serviceHelp = `service <action> [args]
+	start	<ProviderID> <ServiceType> [options]
+	stop	<ServiceID>
+	list
+	status	<ServiceID>
+
+	example: service start 0x7d5ee3557775aed0b85d691b036769c17349db23 openvpn --openvpn.port=1194 --openvpn.proto=UDP`
```
Waldz: These instructions looks false

cmd/commands/cli/command.go
```diff
@@ -26,14 +28,28 @@ import (
...
+	wireguard_service "github.com/mysteriumnetwork/node/services/wireguard/service"
 	tequilapi_client "github.com/mysteriumnetwork/node/tequilapi/client"
 	"github.com/mysteriumnetwork/node/utils"
 	"github.com/urfave/cli"
 )
 
 const cliCommandName = "cli"
 
+const serviceHelp = `service <action> [args]
+	start	<ProviderID> <ServiceType> [options]
```
Waldz: `start	<ProviderID> <ServiceType> [<OpenvpnPort>] [<OpenvpnProto>]`

soffokl: But it can be any options depending on the `ServiceType`. Do you want to hardcode all allowed now flags?

cmd/commands/cli/command.go
soffokl: I didn't get it. Why does it look false?

cmd/commands/cli/command.go
Waldz: ok, clear now

cmd/commands/cli/command.go
Waldz: Probably I got mistaken here

PR #748, opened 2019-02-25
Payments/consumer side balance tracking

### LINKED ISSUE #716: Adjust provider/consumer balance after receiving/sending a promise
Currently, once the promise is sent, not much is done with it. We need to make sure that the balance accounts for the increase in payment for the provider. The consumer will probably want to keep track of this as well.

### LINKED ISSUE #717: Consumer-side balance tracking
At the moment, the consumer does not keep balance on his side. Basically needs to do a similar thing to the provider side balance tracking.

We are also passing a bad param to extendPromise method. This needs to be passed in correctly, once we have a proper balance keeper.

See:
https://github.com/mysteriumnetwork/node/pull/692/files/1fc5cf0315198d2385255252410d8ebf98631cbe#r254980944

Adding the same comment below as well:

We should have a logic there. Current dump balance addition should be pointed out and separate task should exists to implement this logic. If we get balance not enough for next charge period we should extend promise with enough credits just to cover future expected communication. Consumers confidence (reliance on provider) should reflected on this logic. Probably we should have some external parameter to control this from application settings.

vkuznecovas:
marking this as a draft since I need the previous 2 to get merged so i can rebase this properly and merge. Closes #717 and Closes #716

### DIFF
--- a/core/connection/manager.go
+++ b/core/connection/manager.go
@@ -21,6 +21,7 @@ import (
 	"context"
 	"errors"
 	"sync"
+	"time"
 
 	log "github.com/cihub/seelog"
 	"github.com/mysteriumnetwork/node/communication"
@@ -28,6 +29,8 @@ import (
 	"github.com/mysteriumnetwork/node/firewall"
 	"github.com/mysteriumnetwork/node/identity"
 	"github.com/mysteriumnetwork/node/market"
+	"github.com/mysteriumnetwork/node/money"
+	"github.com/mysteriumnetwork/node/services/openvpn/discovery/dto"
 	"github.com/mysteriumnetwork/node/session"
 	"github.com/mysteriumnetwork/node/session/balance"
 	"github.com/mysteriumnetwork/node/session/promise"
@@ -70,7 +73,12 @@ type PaymentIssuer interface {
 }
 
 // PaymentIssuerFactory creates a new payment issuer from the given params
-type PaymentIssuerFactory func(initialState promise.State, messageChan chan balance.Message, dialog communication.Dialog, consumer, provider identity.Identity) (PaymentIssuer, error)
+type PaymentIssuerFactory func(
+	initialState promise.PaymentInfo,
+	paymentDefinition dto.PaymentPerTime,
+	messageChan chan balance.Message,
+	dialog communication.Dialog,
+	consumer, provider identity.Identity) (PaymentIssuer, error)
 
 type connectionManager struct {
 	//these are passed on creation
@@ -154,15 +162,24 @@ func (manager *connectionManager) Connect(consumerID identity.Identity, proposal
 }
 
 func (manager *connectionManager) launchPayments(paymentInfo *promise.PaymentInfo, dialog communication.Dialog, consumerID, providerID identity.Identity) error {
-	var promiseState promise.State
+	var promiseState promise.PaymentInfo
 	if paymentInfo != nil {
-		promiseState.Amount = paymentInfo.LastPromise.Amount
-		promiseState.Seq = paymentInfo.LastPromise.SequenceID
+		promiseState.FreeCredit = paymentInfo.FreeCredit
+		promiseState.LastPromise = paymentInfo.LastPromise
 	}
 
 	messageChan := make(chan balance.Message, 1)
 
-	payments, err := manager.paymentIssuerFactory(promiseState, messageChan, dialog, consumerID, providerID)
+	// TODO: set the time and proper payment info
+	payment := dto.PaymentPerTime{
+		Price: money.Money{
+			Currency: money.CurrencyMyst,
+			Amount:   uint64(10),
+		},
+		Duration: time.Minute,
+	}
+
+	payments, err := manager.paymentIssuerFactory(promiseState, payment, messageChan, dialog, consumerID, providerID)
 	if err != nil {
 		return err
 	}
--- a/session/payment/factory/payments.go
+++ b/session/payment/factory/payments.go
@@ -18,10 +18,14 @@
 package factory
 
 import (
+	"time"
+
 	"github.com/mysteriumnetwork/node/communication"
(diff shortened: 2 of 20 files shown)


core/connection/manager.go
```diff
@@ -21,13 +21,16 @@ import (
...
+	"time"
 
 	log "github.com/cihub/seelog"
 	"github.com/mysteriumnetwork/node/communication"
 	"github.com/mysteriumnetwork/node/consumer"
 	"github.com/mysteriumnetwork/node/firewall"
 	"github.com/mysteriumnetwork/node/identity"
 	"github.com/mysteriumnetwork/node/market"
+	"github.com/mysteriumnetwork/node/money"
+	"github.com/mysteriumnetwork/node/services/openvpn/discovery/dto"
```
soffokl: Again I'm worried that we are using openvpn specific `dto.PaymentPerTime`.

vkuznecovas: I'll create a separate ticket to move it to somewhere more appropriate.

vkuznecovas: https://github.com/mysteriumnetwork/node/issues/761

session/payment/session_balance.go
```diff
@@ -136,43 +134,20 @@ func (sb *SessionBalance) Start() error {
 }
 
 func (sb *SessionBalance) loadInitialPromiseState() (promise.StoredPromise, error) {
-	lastPromise, err := sb.promiseStorage.GetLastPromise(sb.issuerID)
+	lastPromise, err := sb.promiseStorage.FindPromiseForConsumer(sb.issuerID, sb.consumerID)
 	if err != nil {
-		if err.Error() == errBoltNotFound.Error() {
+		if err.Error() == errBoltNotFound.Error() || err.Error() == errNoPromiseForConsumer.Error() {
```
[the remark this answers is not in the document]
vkuznecovas: Two errors with the same message are not equal. See: https://play.golang.org/p/rxwh2j6okm2

I could get rid of this issue by having a direct import of the bolt package, but I don't want the balance to know anything about the underlying DB. Therefore I have to hide it by declaring my own variable. I could just store it in a string though, not sure if that would be more readable though. I feel like mirroring the same error here is an OK solution to this problem.

session/promise/storage.go
```diff
@@ -193,14 +199,36 @@ func (s *Storage) GetAllPromisesFromIssuer(issuerID identity.Identity) ([]Stored
...
-		return &PaymentInfo{}
+		if err == errNoPromiseForConsumer {
+			lastPromise, lastPromiseErr := s.getLastPromise(issuerID)
+			if lastPromiseErr != nil {
+				log.Trace(promiseLogPrefix, "could not load promise info, defaulting to zero payment info ", lastPromiseErr)
+				return defaultPaymentInfo
+			}
+
+			newID := lastPromise.SequenceID + 1
+			log.Trace(promiseLogPrefix, "found payments for other consumers, will issue promiseID ", newID)
```
Waldz: Quite unreadable here, can we move sequence increment logic to storage level and lets have proper locks for that

session/promise/storage_test.go
```diff
@@ -195,17 +201,53 @@ func Test_LoadPaymentInfo_LoadsProperInfo(t *testing.T) {
...
 	assert.Equal(t, uint64(300), pi.LastPromise.Amount)
 	assert.Equal(t, uint64(200), pi.FreeCredit)
 }
 
+func Test_LoadPaymentInfo_WithPreviousConsumersIssuesNewID(t *testing.T) {
+	newMock := map[string][]StoredPromise{
+		getBucketNameFromIssuer(issuerID): {
+			{
+				SequenceID: 10,
+				AddedAt:    timeMock,
```
Waldz: Where is `consumerID` in mocked DB state?

session/payment/factory/payments.go
```diff
@@ -51,21 +57,30 @@ func noopPaymentIssuerFactory(initialState promise.State,
...
 		messageChan chan balance.Message,
 		dialog communication.Dialog,
 		consumer, provider identity.Identity) (connection.PaymentIssuer, error) {
 
 		bl := balance.NewListener(messageChan)
 		ps := promise.NewSender(dialog)
 		issuer := issuers.NewLocalIssuer(signerFactory(consumer))
-		tracker := promise.NewConsumerTracker(initialState, consumer, provider, issuer)
-		payments := payment.NewSessionPayments(messageChan, ps, tracker)
+		tracker := promise.NewConsumerTracker(promise.State{
```
Waldz: Looks like too much glue logic in factory, such fat factories demands to be tested.
Or can we move them out somehow?

session/promise/storage_test.go
vkuznecovas: the defaults act as a separate consumer, but i've made it more explicit anwyay.

session/payment/factory/payments.go
vkuznecovas: I've added a test for the only piece of logic we had there.

session/payment/factory/payments.go
```diff
@@ -51,22 +57,37 @@ func noopPaymentIssuerFactory(initialState promise.State,
...
 		dialog communication.Dialog,
 		consumer, provider identity.Identity) (connection.PaymentIssuer, error) {
 
 		bl := balance.NewListener(messageChan)
 		ps := promise.NewSender(dialog)
 		issuer := issuers.NewLocalIssuer(signerFactory(consumer))
-		tracker := promise.NewConsumerTracker(initialState, consumer, provider, issuer)
-		payments := payment.NewSessionPayments(messageChan, ps, tracker)
+
+		promiseState := mapInitialStateToPromiseState(initialState)
```
Waldz: Do this initialisation inside `promise.NewConsumerTracker()`

Waldz:
What's the status here?

vkuznecovas:
We should merge it. That's the status :D

PR #755, opened 2019-02-27
Hide tequilapi from unknown parties by customising CORS

donce:
Currently, any website can access locally running tequilapi, so if you go to `http://any-evil-side.com`, javascript code provided by them will have access to tequilapi - they will be able to control your node.

This are just defaults - make a config option, which later on could be extended and made configurable using CLI args (will create a separate issue for that).

### DIFF
--- a/cmd/di.go
+++ b/cmd/di.go
@@ -291,7 +291,7 @@ func (di *Dependencies) bootstrapNodeComponents(nodeOptions node.Options) {
 
 	identity_registry.AddIdentityRegistrationEndpoint(router, di.IdentityRegistration, di.IdentityRegistry)
 
-	httpAPIServer := tequilapi.NewServer(nodeOptions.TequilapiAddress, nodeOptions.TequilapiPort, router, corsConfig)
+	httpAPIServer := tequilapi.NewServer(nodeOptions.TequilapiAddress, nodeOptions.TequilapiPort, router, corsPolicy)
 	metricsSender := metrics.CreateSender(nodeOptions.DisableMetrics, nodeOptions.MetricsAddress)
 
 	di.Node = node.NewNode(di.ConnectionManager, httpAPIServer, di.LocationOriginal, metricsSender)
@@ -417,7 +417,7 @@ func (di *Dependencies) bootstrapLocationComponents(options node.OptionsLocation
 	di.LocationOriginal = location.NewLocationCache(di.LocationDetector)
 }
 
-var corsConfig = tequilapi.WhitelistedCorsConfig{
+var corsPolicy = tequilapi.WhitelistingCorsPolicy{
 	DefaultTrustedOrigin:  "https://mysterium.network",
 	AllowedOriginSuffixes: []string{"wallet.mysterium.network", "wallet-dev.mysterium.network", "localhost"},
 }
--- a/cmd/di.go
+++ b/cmd/di.go
@@ -417,7 +417,7 @@ func (di *Dependencies) bootstrapLocationComponents(options node.OptionsLocation
 	di.LocationOriginal = location.NewLocationCache(di.LocationDetector)
 }
 
-var corsConfig = tequilapi.CorsConfig{
+var corsConfig = tequilapi.WhitelistedCorsConfig{
 	DefaultTrustedOrigin:  "https://mysterium.network",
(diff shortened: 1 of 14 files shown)


cmd/di.go
```diff
@@ -416,3 +416,8 @@ func (di *Dependencies) bootstrapLocationComponents(options node.OptionsLocation
 	di.LocationDetector = location.NewDetector(di.IPResolver, di.LocationResolver)
 	di.LocationOriginal = location.NewLocationCache(di.LocationDetector)
 }
+
+var corsConfig = tequilapi.CorsConfig{
+	DefaultTrustedOrigin:  "https://mysterium.network",
+	AllowedOriginSuffixes: []string{"mysterium.network", "localhost"},
```
Waldz: Too many domains, lets trust just what we need

tadovas: Another thing - hardcoding trusted values without possibility to override them on startup - is a breaking change for existing api clients.

tequilapi/http_middlewares.go
```diff
@@ -24,37 +24,60 @@ import (
...
+	return corsHandler{originalHandler: original, corsConfig: corsConfig}
 }
 
-func allowAllCorsActions(resp http.ResponseWriter) {
-	resp.Header().Set("Access-Control-Allow-Origin", "*")
+func allowSpecifiedCorsActions(resp http.ResponseWriter, req *http.Request, corsConfig CorsConfig) {
+	requestOrigin := req.Header.Get("Origin")
+	allowedOrigin := requestOrigin
+	if !isOriginAllowed(requestOrigin, corsConfig.AllowedOriginSuffixes) {
+		trustedOrigin := corsConfig.DefaultTrustedOrigin
```
tadovas: What's the point of intermediate variable here?

cmd/di.go
donce: The idea behind having to trust the whole "mysterium.network" is that we don't need to update it every time we want to add or change a new sub-domain, i.e. adding another wallet environment - but anyways, changed it to specific sub-domains because I don't see big difference :D

As for posibility to override those values - created a separate task for that: https://github.com/mysteriumnetwork/node/issues/758
It is a breaking change, but I believe that it's worth breaking supporting unknown websites :D

Waldz: It's a security issue, we need to estimate risk of ANY `localhost` talking to API

tadovas: Some site can decide to register your identity for you

tequilapi/http_middlewares.go
```diff
@@ -24,37 +24,59 @@ import (
 
 type corsHandler struct {
 	originalHandler http.Handler
+	corsConfig      CorsConfig
+}
+
+// CorsConfig allows customizing CORS (Cross-Origin Resource Sharing) behaviour - whitelisting only specific domains
+type CorsConfig struct {
+	DefaultTrustedOrigin  string
+	AllowedOriginSuffixes []string
```
tadovas: A proposal for design improvement.
Instead of passing around "naked" CorsConfig struct without methods, create an interface with single method: AllowedOrigin(..requestedOrigin..), that way we can have different implementations for different cases like AllowAll, AllowSpecific, etc., corsHandler can depend on this interface only and put returned value inside headers when needed.

cmd/di.go
```diff
@@ -416,3 +416,8 @@ func (di *Dependencies) bootstrapLocationComponents(options node.OptionsLocation
 	di.LocationDetector = location.NewDetector(di.IPResolver, di.LocationResolver)
 	di.LocationOriginal = location.NewLocationCache(di.LocationDetector)
 }
+
+var corsConfig = tequilapi.CorsConfig{
+	DefaultTrustedOrigin:  "https://mysterium.network",
+	AllowedOriginSuffixes: []string{"wallet.mysterium.network", "wallet-dev.mysterium.network", "localhost"},
```
Waldz: "localhost" is needed just for `--localnet` case, lets add new CLI argument `--wallet.address`

cmd/di.go
Waldz: "localhost" is needed just for `--localnet` case, lets add new CLI argument `--wallet.address`

donce: "localhost" is needed always - Electron makes requests using browser, so it respects CORS.

Waldz: PR seems good as CORS engine feature, needs final dots on i to finalise final security

PR #759, opened 2019-02-28
Improve service endpoint validations

Waldz:
Changes:
- Service start validates all fields in request with 422 error code
- Improve Swagger descriptions

### DIFF
--- a/tequilapi/endpoints/service.go
+++ b/tequilapi/endpoints/service.go
@@ -26,6 +26,7 @@ import (
 	"github.com/mysteriumnetwork/node/identity"
 	"github.com/mysteriumnetwork/node/tequilapi/utils"
 	"github.com/mysteriumnetwork/node/tequilapi/validation"
+	"github.com/pkg/errors"
 )
 
 // swagger:model ServiceRequestDTO
@@ -47,7 +48,7 @@ type serviceRequest struct {
 
 	// service options. Every service has a unique list of allowed options.
 	// required: false
-	Options json.RawMessage `json:"options"`
+	Options interface{} `json:"options"`
 }
 
 // swagger:model ServiceListDTO
@@ -73,11 +74,14 @@ type serviceOptions struct {
 // ServiceEndpoint struct represents /service resource and it's sub-resources
 type ServiceEndpoint struct {
 	serviceManager ServiceManager
-	optionsParser  map[string]func(json.RawMessage) (service.Options, error)
+	optionsParser  map[string]ServiceOptionsParser
 }
 
+// ServiceOptionsParser parses request to service specific options
+type ServiceOptionsParser func(json.RawMessage) (service.Options, error)
+
 // NewServiceEndpoint creates and returns service endpoint
-func NewServiceEndpoint(serviceManager ServiceManager, optionsParser map[string]func(json.RawMessage) (service.Options, error)) *ServiceEndpoint {
+func NewServiceEndpoint(serviceManager ServiceManager, optionsParser map[string]ServiceOptionsParser) *ServiceEndpoint {
 	return &ServiceEndpoint{
 		serviceManager: serviceManager,
 		optionsParser:  optionsParser,
@@ -164,7 +168,7 @@ func (se *ServiceEndpoint) ServiceGet(resp http.ResponseWriter, _ *http.Request,
 //     schema:
 //       "$ref": "#/definitions/ErrorMessageDTO"
 func (se *ServiceEndpoint) ServiceStart(resp http.ResponseWriter, req *http.Request, params httprouter.Params) {
-	sr, err := toServiceRequest(req)
+	sr, err := se.toServiceRequest(req)
 	if err != nil {
 		utils.SendError(resp, err, http.StatusBadRequest)
 		return
@@ -176,24 +180,12 @@ func (se *ServiceEndpoint) ServiceStart(resp http.ResponseWriter, req *http.Requ
 		return
 	}
 
-	optionsParser, ok := se.optionsParser[sr.ServiceType]
-	if !ok {
-		utils.SendErrorMessage(resp, "Invalid service type", http.StatusBadRequest)
-		return
-	}
-
-	options, err := optionsParser(sr.Options)
-	if err != nil {
-		utils.SendErrorMessage(resp, "Invalid options", http.StatusBadRequest)
-		return
-	}
-
 	if se.isAlreadyRunning(sr) {
 		utils.SendErrorMessage(resp, "Service already running", http.StatusConflict)
 		return
 	}
 
-	id, err := se.serviceManager.Start(identity.FromAddress(sr.ProviderID), sr.ServiceType, options)
+	id, err := se.serviceManager.Start(identity.FromAddress(sr.ProviderID), sr.ServiceType, sr.Options)
 	if err == service.ErrorLocation {
 		utils.SendError(resp, err, http.StatusBadRequest)
 		return
@@ -252,7 +244,7 @@ func (se *ServiceEndpoint) isAlreadyRunning(sr serviceRequest) bool {
 }
 
 // AddRoutesForService adds service routes to given router
-func AddRoutesForService(router *httprouter.Router, serviceManager ServiceManager, optionsParser map[string]func(json.RawMessage) (service.Options, error)) {
+func AddRoutesForService(router *httprouter.Router, serviceManager ServiceManager, optionsParser map[string]ServiceOptionsParser) {
 	serviceEndpoint := NewServiceEndpoint(serviceManager, optionsParser)
 
 	router.GET("/services", serviceEndpoint.ServiceList)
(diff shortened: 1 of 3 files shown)


tequilapi/endpoints/service.go
```diff
@@ -56,28 +52,32 @@ type serviceList []serviceInfo
 // swagger:model ServiceInfoDTO
 type serviceInfo struct {
 	// example: 6ba7b810-9dad-11d1-80b4-00c04fd430c8
-	ID       string      `json:"id"`
+	ID string `json:"id"`
+
 	Proposal proposalRes `json:"proposal"`
-	// example: Running
-	Status  string         `json:"status"`
-	Options serviceOptions `json:"options"`
```
soffokl: If we are removing "Options" from the service info, what should be the way to figure out what options were used for starting a service?

tequilapi/endpoints/service.go
```diff
@@ -56,28 +52,32 @@ type serviceList []serviceInfo
...
-	optionsParser  map[string]func(json.RawMessage) (service.Options, error)
+	optionsParser  map[string]ServiceOptionsParser
 }
 
+// ServiceOptionsParser parses request to service specific options
+type ServiceOptionsParser func(json.RawMessage) (service.Options, error)
+
+var (
+	// serviceTypeInvalid represents service type which is unknown to node
+	serviceTypeInvalid = "<invalid>"
```
soffokl: If it represents `unknown` services time, maybe it should be called "unknown"?

tequilapi/endpoints/service.go
```diff
@@ -261,9 +246,53 @@ func AddRoutesForService(router *httprouter.Router, serviceManager ServiceManage
...
+	}
+
+	optionsParser, ok := se.optionsParser[serviceType]
+	if !ok {
+		return nil
+	}
+
+	options, err := optionsParser(*value)
+	if err != nil {
+		return serviceOptionsInvalid
```
soffokl: I'd prefer to return `err` here, to give more details of what was invalid in the request.

tequilapi/endpoints/service.go
```diff
@@ -116,14 +116,14 @@ func (se *ServiceEndpoint) ServiceList(resp http.ResponseWriter, _ *http.Request
 //       "$ref": "#/definitions/ErrorMessageDTO"
 func (se *ServiceEndpoint) ServiceGet(resp http.ResponseWriter, _ *http.Request, params httprouter.Params) {
 	id := service.ID(params.ByName("id"))
-	service := se.serviceManager.Service(id)
 
-	if service == nil {
-		utils.SendErrorMessage(resp, "Requested service not found", http.StatusNotFound)
+	serviceInstance := se.serviceManager.Service(id)
+	if serviceInstance == nil {
+		utils.SendErrorMessage(resp, "Requested serviceInstance not found", http.StatusNotFound)
```
soffokl: Isn't `Requested service not found` was more human readable?

tequilapi/endpoints/service.go
```diff
@@ -116,14 +116,14 @@ func (se *ServiceEndpoint) ServiceList(resp http.ResponseWriter, _ *http.Request
 //       "$ref": "#/definitions/ErrorMessageDTO"
 func (se *ServiceEndpoint) ServiceGet(resp http.ResponseWriter, _ *http.Request, params httprouter.Params) {
 	id := service.ID(params.ByName("id"))
-	service := se.serviceManager.Service(id)
 
-	if service == nil {
-		utils.SendErrorMessage(resp, "Requested service not found", http.StatusNotFound)
+	serviceInstance := se.serviceManager.Service(id)
```
soffokl: From my point of view `service` was more suitable than `serviceInstance`. But it's up to you to keep `serviceInstance`.

tequilapi/endpoints/service.go
```diff
@@ -261,9 +246,53 @@ func AddRoutesForService(router *httprouter.Router, serviceManager ServiceManage
...
 }
 
-func toServiceRequest(req *http.Request) (sr serviceRequest, err error) {
-	err = json.NewDecoder(req.Body).Decode(&sr)
-	return sr, err
+func (se *ServiceEndpoint) toServiceRequest(req *http.Request) (serviceRequest, error) {
+	var jsonData struct {
+		ProviderID  string           `json:"providerId"`
+		ServiceType string           `json:"serviceType"`
+		Options     *json.RawMessage `json:"options"`
```
soffokl: Why `Options` should be a pointer here?

tequilapi/endpoints/service.go
Waldz: Anyways it was not implemented, I am just removing dead code here

tequilapi/endpoints/service.go
Waldz: It was bulk renaming typo

tequilapi/endpoints/service.go
Waldz: It was bulk renaming typo

tequilapi/endpoints/service.go
Waldz: It clashes with package name `service`

tequilapi/endpoints/service.go
Waldz: Becauce it's optional argument - if You don't pass it, defaults is being used

tequilapi/endpoints/service.go
donce: We are going to need to know at least `serviceType` (wireguard/openvpn/etc) - since you're removing dead-code this seems fine, but we do need this implemented.

tequilapi/endpoints/service.go
```diff
@@ -261,9 +246,53 @@ func AddRoutesForService(router *httprouter.Router, serviceManager ServiceManage
...
+		ProviderID:  jsonData.ProviderID,
+		ServiceType: se.toServiceType(jsonData.ServiceType),
+		Options:     se.toServiceOptions(jsonData.ServiceType, jsonData.Options),
+	}
+	return sr, nil
+}
+
+func (se *ServiceEndpoint) toServiceType(value string) string {
+	if value == "" {
+		return ""
```
donce: Is empty string a valid service type? Shouldn't we return `serviceTypeInvalid`?

tequilapi/endpoints/service.go
donce: Looks awkward to introduce new special service type just for separating parsing and validation steps - why not just invoke `.toServiceType` inside `.validateServiceRequest` to avoid adding temporary fictional service types?

tequilapi/endpoints/service.go
donce: Resource is called `service`(`.ServiceGet()` method, `GET /services/:id` path), please use terms consistently at the API level. 
```suggestion
		utils.SendErrorMessage(resp, "Requested service not found", http.StatusNotFound)
```

tequilapi/endpoints/service_test.go
```diff
@@ -74,15 +74,15 @@ func Test_AddRoutesForServiceAddsRoutes(t *testing.T) {
 	}{
 		{
 			http.MethodGet, "/services", "",
-			http.StatusOK, `[{"id":"11111111-9dad-11d1-80b4-00c04fd430c0","proposal":{"id":1,"providerId":"0xProviderId","serviceType":"testprotocol","serviceDefinition":{"locationOriginate":{"asn":"LT","country":"Lithuania","city":"Vilnius"}}},"status":"NotRunning","options":{"protocol":"","port":0}}]`,
+			http.StatusOK, `[{"id":"11111111-9dad-11d1-80b4-00c04fd430c0","proposal":{"id":1,"providerId":"0xProviderId","serviceType":"testprotocol","serviceDefinition":{"locationOriginate":{"asn":"LT","country":"Lithuania","city":"Vilnius"}}},"status":"NotRunning"}]`,
```
donce: Is `NotRunning` a status? If service is not running, it should return `404` in `GET /services/:id` and should not be included in `GET /services`

tequilapi/endpoints/service.go
Waldz: I did not degrade - there was no error also. Created separate ticket to improve that and more #763

tequilapi/endpoints/service.go
Waldz: Registered to ticket #764

Waldz: Registered to ticket #764

soffokl: `serviceType` exists in the `Proposal`.

tequilapi/endpoints/service.go
Waldz: This is pattern that was already fallowed `node`'s codebase and yes it's a bit Goish, but we Gophers at the end

tequilapi/endpoints/service.go
Waldz: No point to align internal structures to exposed ones. API should not leak internal structures, what's why we map `service.Instance` -> `tequilapi.ServiceInfo`

tequilapi/endpoints/service_test.go
Waldz: `/services/:id` is meant to search for service instances, 404 seems logical to be related with authorisation of resource, should we link it with state of resource.
Service instance should be removed from service pool, then it stops.
Anyways it's a different scope

tequilapi/endpoints/service.go
Waldz: No, empty value is non given value, and we dont event want to parse it with further logic in function

PR #760, opened 2019-03-01
IP forwarding for Windows with ICS

soffokl:


nat/service_ics.go
```diff
@@ -0,0 +1,147 @@
...
+			return iface.Name, nil
+		}
+	}
+
+	return "", errors.New("interface not found")
+}
+
+func powerShell(cmd string) ([]byte, error) {
+	out, err := exec.Command("powershell", "-Command", cmd).CombinedOutput()
+	return out, errors.Wrap(err, string(out))
```
zolia: You wrap perfectly legit Command output as an error all the time. Maybe simple printout would do for normal output, on error, you can warp it like you did.

nat/service_ics.go
```diff
@@ -0,0 +1,147 @@
...
+	"github.com/pkg/errors"
+)
+
+type serviceICS struct {
+	ifaces             map[string]struct{} // list in internal interfaces with enabled internet connection sharing
+	remoteAccessStatus string
+}
+
+func (nat *serviceICS) Enable() error {
+	status, err := powerShell("Get-Service RemoteAccess | foreach { $_.StartType }")
```
zolia: What this 'StartType' means?

nat/service_ics.go
soffokl: If the `err` is `nil` output will be ignored, it adds an output to the real errors only, to add some context and avoid errors like `exit status 1` with no context.

nat/service_ics.go
```diff
@@ -0,0 +1,147 @@
...
+		$netShare = New-Object -ComObject HNetCfg.HNetShare;
+		$c = $netShare.EnumEveryConnection |? { $netShare.NetConnectionProps.Invoke($_).Name -eq "` + ifaceName + `" };
+		$config = $netShare.INetSharingConfigurationForINetConnection.Invoke($c);
+		$config.DisableSharing()`)
+
+	return err
+}
+
+func getPublicInterfaceName() (string, error) {
+	out, err := powerShell(`Get-WmiObject -Class Win32_IP4RouteTable | where { $_.destination -eq '0.0.0.0' -and $_.mask -eq '0.0.0.0'} | foreach { $_.InterfaceIndex }`)
```
zolia: Have you tested this with win10 home only? I suspect we might have trouble executing those on other versions. We probably need  a way to identify win version and execute appropriately or test on all win7+ versions.

nat/service_ics.go
soffokl: It's a service start type:
image

nat/service_ics.go
```diff
@@ -0,0 +1,147 @@
...
+	if err != nil {
+		return "", err
+	}
+
+	ifaceID, err := strconv.Atoi(strings.TrimSpace(string(out)))
+	if err != nil {
+		return "", err
+
```
zolia: extra space

nat/service_ics.go
soffokl: Yes, tested on Windows 10 only.
Isn't Windows 7 too old to support it?

nat/service_ics.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+
+	ifaces, err := net.Interfaces()
+	if err != nil {
+		return "", err
```
zolia: We could use errors.warp here pinpoint more specific place with "failed to fetch system interfaces" or so. Same above.

nat/service_ics.go
```diff
@@ -0,0 +1,147 @@
...
+	if err != nil {
+		return err
+	}
+
+	// Enable internet connection sharing for the public interface.
+	_, err = powerShell(`regsvr32 /s hnetcfg.dll;
+		$netShare = New-Object -ComObject HNetCfg.HNetShare;
+		$c = $netShare.EnumEveryConnection |? { $netShare.NetConnectionProps.Invoke($_).Name -eq "` + ifaceName + `" };
+		$config = $netShare.INetSharingConfigurationForINetConnection.Invoke($c);
+		$config.EnableSharing(0)`)
```
zolia: `$config.EnableSharing(0)` 0 - enables sharing? :)

soffokl: Yep. 
`0` here is a type of sharing. 0 - public, 1 - private.

nat/service_ics.go
zolia: google "windows versions market share", almost 34% atm ;)

nat/service_ics.go
zolia: nice table

nat/service_ics.go
soffokl: Tested it on Windows 7, it works with no issue.

nat/service_ics.go
```diff
@@ -0,0 +1,224 @@
...
+	for _, iface := range ifaces {
+		if iface.Index == ifaceID {
+			return iface.Name, nil
+		}
+	}
+
+	return "", errors.New("interface not found")
+}
+
+func powerShell(cmd string) ([]byte, error) {
```
Waldz: Is it possible to mock `exec.Command` and have some test with common situations simulated?

nat/service_ics.go
Waldz: Constants would improve readability here

nat/service_ics.go
soffokl: Yeah, we can do some mocked tests.

nat/service_ics.go
```diff
@@ -0,0 +1,224 @@
...
+
+// Add enables internet connection sharing for the local interface.
+func (nat *serviceICS) Add(rule RuleForwarding) error {
+	// TODO firewall rule configuration should be added here for new connections.
+	ifaceName, err := getInterfaceBySubnet(rule.SourceAddress)
+	if err != nil {
+		return errors.Wrap(err, "failed to get public interface name")
+	}
+
+	_, err = powerShell(`regsvr32 /s hnetcfg.dll;
```
vkuznecovas: this whole super string seems like a perfect candidate for a template since we're using it twice.

nat/service_ics.go
```diff
@@ -0,0 +1,224 @@
...
+
+// Del disables internet connection sharing for the local interface.
+func (nat *serviceICS) Del(rule RuleForwarding) error {
+	// TODO firewall rule configuration should be added here for cleaning up unused connections.
+	ifaceName, err := getInterfaceBySubnet(rule.SourceAddress)
+	if err != nil {
+		return errors.Wrap(err, "failed to find suitable interface")
+	}
+
+	_, err = powerShell(`regsvr32 /s hnetcfg.dll;
```
vkuznecovas: or maybe even thrice? would need to think about it though as we're getting more complex now lol.

nat/service_ics.go
```diff
@@ -0,0 +1,224 @@
...
+		return "", errors.Wrap(err, "failed to get a list of network interfaces")
+	}
+
+	for _, iface := range ifaces {
+		addrs, err := iface.Addrs()
+		if err != nil {
+			return "", errors.Wrap(err, "failed to get list of interface addresses")
+		}
+
+		for _, addr := range addrs {
```
vkuznecovas: this could get extracted to a separate function that's easily testable.

nat/service_ics.go
soffokl: Yeah, I was thinking to do it too. Extracted it now.

nat/service_ics.go
soffokl: Added some testing for it.

nat/service_ics.go
```diff
@@ -0,0 +1,234 @@
...
+	if err != nil {
+		return errors.Wrap(err, "failed to find suitable interface")
+	}
+
+	err = nat.applySharingConfig(disableSharing, ifaceName)
+	if err != nil {
+		return errors.Wrap(err, "failed to disable internet connection sharing for internal interface")
+	}
+
+	nat.mu.Lock()
```
Waldz: Should not locking be before actions with interfaces?

nat/service_ics.go
```diff
@@ -0,0 +1,234 @@
...
+
+type serviceICS struct {
+	mu                 sync.Mutex
+	ifaces             map[string]RuleForwarding // list in internal interfaces with enabled internet connection sharing
+	remoteAccessStatus string
+	powerShell         func(cmd string) ([]byte, error)
+}
+
+// Enable enables internet connection sharing for the public interface.
+func (nat *serviceICS) Enable() error {
```
Waldz: Would be so nice to have single-purpose functions like `enableRemoteAccess()` and `Enable()` - which just glues everything together

nat/service_ics.go
soffokl: This lock is just protecting map from concurrent access. 
If we want to have locks for the whole configuration process, it will require more changes.

nat/service_ics.go
```diff
@@ -0,0 +1,240 @@
...
+	if err := nat.enableRemoteAccessService(); err != nil {
```
Waldz: 💯

Waldz:
LGTM

PR #762, opened 2019-03-01
[Feature] Collect NAT mapping (upnp) stats

donce:
#843

### DIFF
--- a/metrics/factory.go
+++ b/metrics/factory.go
@@ -19,12 +19,10 @@ package metrics
 
 import (
 	"time"
-
-	"github.com/mysteriumnetwork/node/metadata"
 )
 
-// CreateSender creates metrics sender with appropriate transport
-func CreateSender(disableMetrics bool, metricsAddress string) *Sender {
+// NewSender creates metrics sender with appropriate transport
+func NewSender(disableMetrics bool, metricsAddress, applicationVersion string) *Sender {
 	var transport Transport
 	if disableMetrics {
 		transport = NewNoopTransport()
@@ -32,5 +30,5 @@ func CreateSender(disableMetrics bool, metricsAddress string) *Sender {
 		transport = NewElasticSearchTransport(metricsAddress, 10*time.Second)
 	}
 
-	return &Sender{Transport: transport, ApplicationVersion: metadata.VersionAsString()}
+	return &Sender{Transport: transport, ApplicationVersion: applicationVersion}
 }
--- a/metrics/factory.go
+++ b/metrics/factory.go
@@ -17,7 +17,11 @@
 
 package metrics
 
-import "time"
+import (
+	"time"
+
+	"github.com/mysteriumnetwork/node/metadata"
+)
 
 // CreateSender creates metrics sender with appropriate transport
 func CreateSender(disableMetrics bool, metricsAddress string) *Sender {
@@ -28,5 +32,5 @@ func CreateSender(disableMetrics bool, metricsAddress string) *Sender {
 		transport = NewElasticSearchTransport(metricsAddress, 10*time.Second)
 	}
 
-	return &Sender{Transport: transport}
+	return &Sender{Transport: transport, ApplicationVersion: metadata.VersionAsString()}
 }
--- a/metrics/factory.go
+++ b/metrics/factory.go
@@ -22,7 +22,7 @@ import (
 )
 
 // NewSender creates metrics sender with appropriate transport
-func NewSender(disableMetrics bool, metricsAddress, applicationVersion string) *Sender {
+func NewSender(disableMetrics bool, metricsAddress, appVersion string) *Sender {
 	var transport Transport
 	if disableMetrics {
 		transport = NewNoopTransport()
@@ -30,5 +30,5 @@ func NewSender(disableMetrics bool, metricsAddress, applicationVersion string) *
 		transport = NewElasticSearchTransport(metricsAddress, 10*time.Second)
 	}
 
-	return &Sender{Transport: transport, ApplicationVersion: applicationVersion}
+	return &Sender{Transport: transport, AppVersion: appVersion}
 }
--- a/metrics/sender.go
+++ b/metrics/sender.go
@@ -23,7 +23,8 @@ import (
 
 const appName = "myst"
 const startupEventName = "startup"
-const natMappingResultEventName = "nat_mapping_result"
+const natMappingSuccessEventName = "nat_mapping_success"
+const natMappingFailEventName = "nat_mapping_fail"
 
 // Sender builds events and sends them using given transport
 type Sender struct {
@@ -48,8 +49,8 @@ type applicationInfo struct {
 	Version string `json:"version"`
 }
 
-type natMappingResultContext struct {
-	success bool
+type natMappingFailContext struct {
+	error string
 }
 
 // SendStartupEvent sends startup event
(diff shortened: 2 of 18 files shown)


nat/mapping/port_mapping.go
```diff
@@ -73,14 +83,19 @@ func mapPort(m portmap.Interface, c chan struct{}, protocol string, extPort, int
...
-func addMapping(m portmap.Interface, protocol string, extPort, intPort int, name string) {
-	if err := m.AddMapping(protocol, extPort, intPort, name, mapTimeout); err != nil {
-		log.Debugf("%s, Couldn't add port mapping for port %d: %v, retrying with permanent lease", logPrefix, extPort, err)
-		if err := m.AddMapping(protocol, extPort, intPort, name, 0); err != nil {
-			// some gateways support only permanent leases
-			log.Debugf("%s Couldn't add port mapping for port %d: %v", logPrefix, extPort, err)
-			return
-		}
+func addMapping(m portmap.Interface, protocol string, extPort, intPort int, name string) error {
+	err := m.AddMapping(protocol, extPort, intPort, name, mapTimeout)
```
vkuznecovas: I personally feel the old version of the method was way more readable, the `err == nil` check feels super odd in what is usually an `error != nil` world.

zolia:
branch not  up to date

metrics/sender.go
```diff
@@ -46,10 +49,28 @@ type applicationInfo struct {
 	Version string `json:"version"`
 }
 
+type natMappingFailContext struct {
+	error string
```
vkuznecovas: please don't have a prop called error with string for type. Either call it errorInfo, or use error as type.

metrics/factory.go
```diff
@@ -17,7 +17,11 @@
...
 
-import "time"
+import (
+	"time"
+
+	"github.com/mysteriumnetwork/node/metadata"
+)
 
 // CreateSender creates metrics sender with appropriate transport
 func CreateSender(disableMetrics bool, metricsAddress string) *Sender {
```
Waldz: `NewSender()`

metrics/factory.go
```diff
@@ -17,7 +17,11 @@
 
 package metrics
 
-import "time"
+import (
+	"time"
+
+	"github.com/mysteriumnetwork/node/metadata"
```
Waldz: My advice is not to couple `metadata` and `metrics` components, because they can be perfectly uncoupled

nat/mapping/port_mapping.go
```diff
@@ -18,10 +18,12 @@
 package mapping
 
 import (
+	"fmt"
 	"time"
 
 	log "github.com/cihub/seelog"
 	portmap "github.com/ethereum/go-ethereum/p2p/nat"
+	"github.com/mysteriumnetwork/node/metrics"
```
Waldz: Lets not couple low level packages with application logic it self. How could we achieve that?

nat/mapping/port_mapping.go
```diff
@@ -64,7 +67,14 @@ func mapPort(m portmap.Interface, c chan struct{}, protocol string, extPort, int
...
 	}()
 	for {
-		addMapping(m, protocol, extPort, intPort, name)
+		err := addMapping(m, protocol, extPort, intPort, name)
+		if err != nil {
+			log.Infof("%s, Mapping for port %d failed: %s", logPrefix, extPort, err)
+			metricsSender.SendNATMappingFailEvent(err)
+		} else {
+			log.Info("%s, Mapped network port: %d", logPrefix, extPort)
+			metricsSender.SendNATMappingSuccessEvent()
```
zolia: Theres one bad issue with this implementation. Since mapPort is executed in routine every `mapUpdateInterval` this means we will be getting as much statistics updates. This might be overhelming for out metrics collector. For NAT statistics analysis it should be sufficient for us to see just a single report after service gets started. 

My suggestion as to decouple from metrics sender here and to solve this problem is to use an EventBus.

I am working currently introducing event bus to port_mapper for communication with other NAT submodules, but these could be reused them. I would suggent to stall this PR and to make it work for both NAT methods reusing EventBus.

nat/mapping/port_mapping.go
donce: I personally found it very hard to understand all the cases in the previous example - but we're all different and it might be just me.
My motivation was to remove nesting by returning early - i.e. if we had to more more edge-cases, we would need more nested `if`s in first approach, and in my approach - only single method-level `return` would be added.

http://www.itamarweiss.com/personal/2018/02/28/return-early-pattern.html
https://softwareengineering.stackexchange.com/questions/18454/should-i-return-from-a-function-early-or-use-an-if-statement

Do others find this more difficult to understand as well?

vkuznecovas: I realise that multiple nested levels are a bit fishy but in this case we're not nesting that much.

nat/mapping/port_mapping.go
donce: By using `EventBus`, which @zolia is working on :)

donce:
Closing this PR until @zolia finishes `EventBus`

nat/mapping/port_mapping.go
Waldz: I had similar ideas in mind here also https://github.com/mysteriumnetwork/node/pull/762#discussion_r262380471

nat/mapping/port_mapping.go
```diff
@@ -72,7 +72,12 @@ func mapPort(m portmap.Interface, c chan struct{}, protocol string, extPort, int
 		}
 	}()
 	for {
-		addMapping(m, protocol, extPort, intPort, name, publisher)
+		err := addMapping(m, protocol, extPort, intPort, name, publisher)
```
zolia: Why we need another error printouts? We already have it inside addMapping..

nat/mapping/port_mapping.go
```diff
@@ -81,16 +86,17 @@ func mapPort(m portmap.Interface, c chan struct{}, protocol string, extPort, int
...
 			// some gateways support only permanent leases
-			publisher.Publish(traversal.EventTopic, traversal.EventFailure)
+			publisher.Publish(traversal.EventTopic, traversal.BuildFailureEvent(err))
 			log.Debugf("%s Couldn't add port mapping for port %d: %v", logPrefix, extPort, err)
-			return
+			return err
 		}
 	}
-	publisher.Publish(traversal.EventTopic, traversal.EventSuccess)
+	publisher.Publish(traversal.EventTopic, traversal.BuildSuccessEvent())
```
zolia: We could discriminate between what kind of success: with mapTimeout or with persistant (0). This would give us additional information in the future.

nat/mapping/port_mapping.go
donce: Yeah, removed this.

nat/mapping/port_mapping.go
donce: I think we've agreed to add more stuff later, when/if needed.

vkuznecovas:
I dunno, I feel like the current NatEventTracker should not be responsible for reporting those as metrics. Maybe we could have a MetricsReporter, that instead subscribes to multiple topics(for now it would only be NAT, but we could hook up more senders in the future once needed). Just a thought.

nat/traversal/events_tracker.go
```diff
@@ -26,29 +27,53 @@ import (
...
 }
 
 // ConsumeNATEvent consumes a NAT event
 func (et *EventsTracker) ConsumeNATEvent(event Event) {
-	log.Info("got NAT event: ", event)
+	log.Info(eventsTrackerLogPrefix, "got NAT event: ", event)
+
+	err := et.sendNATEvent(event)
+	if err != nil {
+		log.Warn(eventsTrackerLogPrefix, "sending event failed", err)
```
soffokl: We need to have an explicit delimiter between arguments. 
```suggestion
		log.Warn(eventsTrackerLogPrefix, "sending event failed: ", err)
```

PR #767, opened 2019-03-04
Allow to skip service options for service starting

Waldz:


services/openvpn/service/options.go
```diff
@@ -57,8 +57,12 @@ func ParseFlags(ctx *cli.Context) service.Options {
 }
 
 // ParseJSONOptions function fills in Openvpn options from JSON request
-func ParseJSONOptions(request json.RawMessage) (service.Options, error) {
+func ParseJSONOptions(request *json.RawMessage) (service.Options, error) {
+	if request == nil {
+		return Options{}, nil
```
vkuznecovas: could use named return params here to avoid declaring the opts, and using naked returns instead.

services/wireguard/service/options.go
```diff
@@ -50,8 +50,12 @@ func ParseFlags(ctx *cli.Context) service.Options {
 }
 
 // ParseJSONOptions function fills in Openvpn options from JSON request
-func ParseJSONOptions(request json.RawMessage) (service.Options, error) {
+func ParseJSONOptions(request *json.RawMessage) (service.Options, error) {
+	if request == nil {
+		return Options{}, nil
```
vkuznecovas: could use named return params here to avoid declaring the opts, and using naked returns instead.

services/openvpn/service/options.go
Waldz: Imposible, because return type is `Options interface{}`

PR #772, opened 2019-03-05
Expose `providerId`, `type`, `options` in service endpoints

Waldz:


tequilapi/client/client.go
```diff
@@ -253,8 +253,8 @@ func (client *Client) GetSessions() (SessionsDTO, error) {
 	return sessions, err
 }
 
-// Services returns all running services.
-func (client *Client) Services() (services Services, err error) {
+// ServiceListDTO returns all running services.
```
soffokl: It looks like another bulk rename affected unnecessary comments too.

tequilapi/client/client.go
```diff
@@ -265,8 +265,8 @@ func (client *Client) Services() (services Services, err error) {
 	return services, err
 }
 
-// Service returns a service information by the requested id.
-func (client *Client) Service(id string) (service Service, err error) {
+// ServiceInfoDTO returns a service information by the requested id.
```
soffokl: It looks like another bulk rename affected unnecessary comments too.

tequilapi/client/dto.go
```diff
@@ -131,13 +131,15 @@ type SessionDTO struct {
...
-type Services []Service
-
-// Service represents a service information.
-type Service struct {
-	ID       string          `json:"id"`
-	Status   string          `json:"status"`
-	Proposal ProposalDTO     `json:"proposal"`
-	Options  json.RawMessage `json:"options"`
+// ServiceListDTO represents a list of running services on the node
+type ServiceListDTO []ServiceInfoDTO
```
soffokl: I'm not sure if we need to name DTOs as `DTO` and arrays as `List`.
Is it a right golang-way?

tequilapi/endpoints/service.go
```diff
@@ -293,10 +305,14 @@ func (se *ServiceEndpoint) toServiceOptions(serviceType string, value *json.RawM
 }
 
 func toServiceInfoResponse(id service.ID, instance *service.Instance) serviceInfo {
+	proposal := instance.Proposal()
 	return serviceInfo{
-		ID:       string(id),
-		Status:   string(instance.State()),
-		Proposal: proposalToRes(instance.Proposal()),
+		ID:         string(id),
+		ProviderID: proposal.ProviderID,
```
soffokl: Now we have `ProviderID` and `ServiceType` twice in the same DTO. 
One day someone will change it in one place and will forget to change in another place. 
That is why I'd prefer to keep it only in one place.

tequilapi/endpoints/service_test.go
```diff
@@ -73,16 +87,67 @@ func Test_AddRoutesForServiceAddsRoutes(t *testing.T) {
...
+			`[{
+				"id": "11111111-9dad-11d1-80b4-00c04fd430c0",
+				"providerId": "0xProviderId",
+				"type": "testprotocol",
+				"options": {"foo": "bar"},
+				"status": "NotRunning",
+				"proposal": {
+					"id": 1,
+					"providerId": "0xProviderId",
+				    "serviceType": "testprotocol",
```
soffokl: These lines looks not aligned right.

tequilapi/client/dto.go
vkuznecovas: More than likely it's not: https://dave.cheney.net/practical-go/presentations/qcon-china.html#_dont_name_your_variables_for_their_types

Waldz: I am just keeping them aligned with Swagger definitions and being consistent with other endpoints.
Most likely we want to generate Tequilapi client from Swagger definitions later, instead of writing it handy

tequilapi/endpoints/service.go
Waldz: `proposal` is data structure of marketplace, and I expose fields, which caller used to start the service:
- I don't like that by exposing thru `proposal.ServiceType` we depends on inside effects what it matches now
- Is non-user-friendly for caller to find service by service type

Waldz:
all comments approached

core/service/pool.go
```diff
@@ -100,23 +100,37 @@ func (p *Pool) Instance(id ID) *Instance {
...
+	state          State
+	serviceOptions Options
+	service        RunnableService
+	proposal       market.ServiceProposal
+	dialogWaiter   communication.DialogWaiter
+	discovery      *discovery_registry.Discovery
+}
+
+// ServiceOptions returns options used to start service
+func (i *Instance) ServiceOptions() Options {
```
vkuznecovas: Could we perhaps name this GetOptions instead? I feel it describes the action better.

tequilapi/endpoints/service.go
```diff
@@ -54,10 +54,22 @@ type serviceInfo struct {
...
-	Proposal proposalRes `json:"proposal"`
+	// provider identity
+	// example: 0x0000000000000000000000000000000000000002
+	ProviderID string `json:"providerId"`
+
+	// service type. Possible values are "openvpn", "wireguard" and "noop"
+	// example: openvpn
+	Type string `json:"type"`
+
+	// options with which service was started. Every service has a unique list of allowed options.
```
vkuznecovas: we could expand on this comment, adding a link to where you could find the unique list - if we had one.

tequilapi/endpoints/service_test.go
```diff
@@ -28,10 +28,24 @@ import (
...
 	"github.com/mysteriumnetwork/node/core/service"
 	"github.com/mysteriumnetwork/node/identity"
+	"github.com/mysteriumnetwork/node/market"
 	"github.com/stretchr/testify/assert"
 )
 
-var mockServiceID = service.ID("6ba7b810-9dad-11d1-80b4-00c04fd430c8")
+var (
+	mockServiceID      = service.ID("6ba7b810-9dad-11d1-80b4-00c04fd430c8")
+	mockServiceOptions = struct {
```
vkuznecovas: these inline struct declarations always feel a bit weird.

tequilapi/endpoints/service_test.go
```diff
@@ -28,10 +28,24 @@ import (
...
+		Foo string `json:"foo"`
+	}{"bar"}
```
vkuznecovas: please provide a named param, not just "bar"

tequilapi/endpoints/service.go
Waldz: Not my scope here. We're thinking to implement emdpoint `/service/types` which will expose service types and details how to start them

Waldz:
comments fixed

tequilapi/client/client.go
```diff
@@ -253,8 +253,8 @@ func (client *Client) GetSessions() (SessionsDTO, error) {
 	return sessions, err
 }
 
-// Services returns all running services.
```
soffokl: Just a notice, I'm using these dots at the end of sentence/comment by the recommendation of [Effective GO](https://golang.org/doc/effective_go.html#commentary).
For example, standard packages are using it everywhere: https://golang.org/src/errors/errors.go

Waldz: ok, when we need to make it consistent everywhere

PR #775, opened 2019-03-06
Service start endpoint defaults options when starting a service

Waldz:


services/openvpn/service/options.go
```diff
@@ -51,18 +56,22 @@ func RegisterFlags(flags *[]cli.Flag) {
...
-		OpenvpnProtocol: ctx.String(protocolFlag.Name),
-		OpenvpnPort:     ctx.Int(portFlag.Name),
+		Protocol: ctx.String(protocolFlag.Name),
+		Port:     ctx.Int(portFlag.Name),
 	}
 }
 
 // ParseJSONOptions function fills in Openvpn options from JSON request
 func ParseJSONOptions(request *json.RawMessage) (service.Options, error) {
+	opts := Options{
```
vkuznecovas: All good and well. One thing I'm thinking is maybe we could make a defaultOptions variable, that holds a struct with the default values. We can then reuse those values where needed from the defaultOptions variable, and return the defaultOptions struct directly. Might save a few lines of code. Not gonna block this PR for it though, implement only if you like it(but if you do so, please in all the places it's required).

Waldz: introduced

PR #776, opened 2019-03-06
Allow all localhost port origins, extend cors policy to regexp

donce:
Electron uses `localhost:9080` origin in development - and it seems port is dynamic, so we can solve this by supporting all ports on localhost.

### DIFF
--- a/tequilapi/mysterium_cors_policy_factory.go
+++ b/tequilapi/mysterium_cors_policy_factory.go
@@ -18,7 +18,7 @@
 package tequilapi
 
 func NewMysteriumCorsPolicy() CorsPolicy {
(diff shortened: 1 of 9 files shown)


Waldz:
Could You to description example of headers send by Electron to API

tequilapi/mysterium_cors_policy_factory.go
```diff
@@ -15,21 +15,11 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-// Package tequilapi Tequila API
```
vkuznecovas: This is probably required for the swagger generation

PR #782, opened 2019-03-07
service manager now removes the service from the pool if it stops

vkuznecovas:
Before this change, the service could crash and still stay in the service list. Now it gets removed. To do this, the pool needed to be added under a lock, since a few racing goroutines would access it.

### DIFF
--- a/core/service/manager.go
+++ b/core/service/manager.go
@@ -49,10 +49,10 @@ type DialogWaiterFactory func(providerID identity.Identity, serviceType string) 
 type DialogHandlerFactory func(market.ServiceProposal, session.ConfigNegotiator) communication.DialogHandler
 
 // DiscoveryFactory initiates instance which is able announce service discoverability
-type DiscoveryFactory func() DiscoveryService
+type DiscoveryFactory func() Discovery
 
-// DiscoveryService registers the service to the discovery api periodically
-type DiscoveryService interface {
+// Discovery registers the service to the discovery api periodically
+type Discovery interface {
 	Start(ownIdentity identity.Identity, proposal market.ServiceProposal)
 	Stop()
 	Wait()
--- a/core/service/manager.go
+++ b/core/service/manager.go
@@ -25,7 +25,6 @@ import (
 	"github.com/mysteriumnetwork/node/communication"
 	"github.com/mysteriumnetwork/node/identity"
 	"github.com/mysteriumnetwork/node/market"
-	discovery_registry "github.com/mysteriumnetwork/node/market/proposals/registry"
 	"github.com/mysteriumnetwork/node/session"
 )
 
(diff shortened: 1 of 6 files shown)


core/service/manager.go
```diff
@@ -122,14 +128,21 @@ func (manager *Manager) Start(providerID identity.Identity, serviceType string,
...
-		err = service.Serve(providerID)
-		instance.state = NotRunning
-		if err != nil {
-			log.Error("Service serve failed: ", err)
+		serveErr := service.Serve(providerID)
+		if serveErr != nil {
+			log.Error("Service serve failed: ", serveErr)
+		}
+
+		// instance.state = NotRunning
```
donce: Do we still need this?

core/service/manager.go
```diff
@@ -122,14 +128,21 @@ func (manager *Manager) Start(providerID identity.Identity, serviceType string,
...
+		serveErr := service.Serve(providerID)
+		if serveErr != nil {
+			log.Error("Service serve failed: ", serveErr)
+		}
+
 		instance.state = NotRunning
-		if err != nil {
-			log.Error("Service serve failed: ", err)
+
+		stopErr := manager.servicePool.Stop(id)
```
soffokl: Keeping failed service in pool was left for purpose: 
> On failed serve, we are marking service as "NotRunning" and keeping it in the pool to let a user ability to get this information from the Tequilapi endpoint. And the user can delete it from the pool by calling service stop.

https://github.com/mysteriumnetwork/node/pull/744/files#r259858610

soffokl: We have a state to show the service status. If we are going to keep only running services, there is no need to keep the `state` field in the struct.

donce: From API perspective, it's very confusing to have 2 different states for not-running service:
- CRUD resource exists, status is `NOT_RUNNING`
- CRUD resource does not exist (404 when getting resource)

So we've agreed to remove `NotRunning` state value - `state` could be kept there for states such as `STARTING`, `RUNNING`. In the future, we might have more states.

soffokl: I was thinking from the API user perspective: 
1. You are starting service, checking it in the list, see nothing there.
1. Starting it again, and see nothing in the list again. 
1. Starting thinking that API does not work. 

But if you see it in the list, but with a failed state, you understand that service crashed.

donce: If we start the service, `CREATE /services` endpoint returns the service, which already contains `STARTING` `state`. Then we start polling `state` to know when it changes (we need to do polling in either case).
- if it changes to `RUNNING`, that means it's connected successful
- if it changes to `NOT_RUNNING`, that means it didn't connect successful - we show that to the user.

Both options seem viable to me, but I would stick the the one we agreed, since I've already started implementing that - and it seems easier finish implementing it now :D

vkuznecovas: There's an issue with having it in the list from the non api perspective though. You'll end up with a ghost entry in a map that really doesn't say much - except that it failed. It will hang there forever until you either restart/explicitly delete the entry. If we're gonna have quite a few services available here, and start starting/stopping them once in a while this could get quite costly in terms of memory.

core/service/manager.go
```diff
@@ -50,7 +49,14 @@ type DialogWaiterFactory func(providerID identity.Identity, serviceType string)
 type DialogHandlerFactory func(market.ServiceProposal, session.ConfigNegotiator) communication.DialogHandler
 
 // DiscoveryFactory initiates instance which is able announce service discoverability
-type DiscoveryFactory func() *discovery_registry.Discovery
+type DiscoveryFactory func() DiscoveryService
+
+// DiscoveryService registers the service to the discovery api periodically
+type DiscoveryService interface {
```
zolia: Imho its ambiguous naming.. I would call it ServiceDiscovery, but since we are in a service package we could just call it 'Discovery'.

core/service/manager.go
zolia: True for @donce and @vkuznecovas comments, I would vote for zomby entries removal too.

Issue #672, opened 2019-01-23
Cleanup inactive sessions on timeout


soffokl:
At that moment we are never cleaning inactive sessions. This is causing a problem when we have a number of sessions that are waiting for a connection on a provider side forever. 

We need to add timeout for inactive sessions to close it if there was no any activity from the consumer side. 

One option could be looking on transferred data, and if it never changes mark session as inactive. 

Another option could be adding heartbeats between consumer and provider to make sure that session still alive.

soffokl:
This is resolved in the scope of payment:
https://github.com/mysteriumnetwork/node/pull/811

PR #787, opened 2019-03-12
Temporary solution for wireguard resources rotation.

soffokl:
Once we are waiting for proper session cleanup in https://github.com/mysteriumnetwork/node/issues/672

This PR will allow rotating old wireguard resources (interfaces, subnets, listen ports) to allow new connections.

### DIFF
--- a/services/wireguard/service/service.go
+++ b/services/wireguard/service/service.go
@@ -21,9 +21,8 @@ import (
 	"encoding/json"
 	"sync"
 
-	"github.com/mysteriumnetwork/node/core/location"
-
 	log "github.com/cihub/seelog"
+	"github.com/mysteriumnetwork/node/core/location"
 	"github.com/mysteriumnetwork/node/identity"
 	"github.com/mysteriumnetwork/node/market"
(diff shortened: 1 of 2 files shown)


services/wireguard/service/service.go
```diff
@@ -110,7 +114,49 @@ func (manager *Manager) ProvideConfig(publicKey json.RawMessage) (session.Servic
...
+// TODO this is a temporary solution to cleanup oldest used wireguard resources.
+// TODO it should be removed once payment bases session cleanup implemented.
+func (manager *Manager) cleanOldEndpoints() {
+	manager.mu.Lock()
+	defer manager.mu.Unlock()
+
+	if len(manager.list) >= resources.MaxResources-1 {
+		log.Warn(logPrefix, "We have reached a maximum number of interfaces. Cleaning up oldest one.")
+
+		f := *manager.list[0]
```
zolia: It will be oldest endpoint, but not oldest active endpoint? I suppose it might be enough for a temp solution.

zolia:
We will need a backport to stable branch.

PR #804, opened 2019-03-15
Wireguard provider for Windows uses a single interface

soffokl:


services/wireguard/resources/allocator_windows.go
```diff
@@ -0,0 +1,108 @@
...
+
+// Allocator is mock wireguard resource handler.
+// It will manage lists of network interfaces names, IP addresses and port for endpoints.
+type Allocator struct {
+	IPAddresses map[int]struct{}
+	mu          sync.Mutex
+}
+
+// NewAllocator creates new resource pool for wireguard connection.
+func NewAllocator() Allocator {
```
vkuznecovas:
```suggestion
func NewAllocator() *Allocator {
  return &Allocator{
   IPAddresses: make(map[int]struct{}),
  }
}
```

this way I won't have to do this on the consumer side if this is abstracted as an interface

Waldz:
I am bad reviewer for this, dont understand a thing :)

services/wireguard/resources/allocator_windows.go
```diff
@@ -0,0 +1,108 @@
...
+
+import (
+	"errors"
+	"fmt"
+	"net"
+	"sync"
+)
+
+// MaxResources sets the limit to the maximum number of wireguard connections.
+const MaxResources = 255
```
zolia: most likely it should be 253 IPs. Normally 1 is reserved for gw and 255 for broadcast and those should not be assigned to client.

services/wireguard/resources/allocator_windows.go
```diff
@@ -0,0 +1,108 @@
...
+// NewAllocator creates new resource pool for wireguard connection.
+func NewAllocator() Allocator {
+	return Allocator{
+		IPAddresses: make(map[int]struct{}),
+	}
+}
+
+// AbandonedInterfaces returns a list of abandoned interfaces that exist in the system,
+// but was not allocated by the Allocator.
+func (a *Allocator) AbandonedInterfaces() ([]net.Interface, error) {
```
zolia: What is the purpose of this function? When those abandoned interfaces come up?

zolia: We need TODO here at least.

services/wireguard/resources/allocator_windows.go
zolia: ok, you use it as exclusive in code..

services/wireguard/resources/allocator_windows.go
```diff
@@ -0,0 +1,108 @@
...
+	return nil
+}
+
+// ReleaseIPNet releases IP address.
+func (a *Allocator) ReleaseIPNet(ipnet net.IPNet) error {
+	a.mu.Lock()
+	defer a.mu.Unlock()
+
+	ip4 := ipnet.IP.To4()
+	if len(ip4) != net.IPv4len {
```
zolia: `ipnet.IP.To4()` should simply retur nil here if it is not a valid IP. Should we still compare it with `net.IPv4len`?

services/wireguard/resources/allocator_windows.go
```diff
@@ -0,0 +1,108 @@
...
+	if _, ok := a.IPAddresses[i]; !ok {
+		return errors.New("allocated subnet not found")
+	}
+
+	delete(a.IPAddresses, i)
+	return nil
+}
+
+// ReleasePort releases UDP port.
+func (a *Allocator) ReleasePort(port int) error {
```
zolia: We should mark such functions that hey are just a stub for WIN and does not need to be implemented.

services/wireguard/service/service_unix.go
```diff
@@ -0,0 +1,133 @@
...
+	portMap func(port int) (releasePortMapping func()),
+	options Options) *Manager {
+
+	resourceAllocator := resources.NewAllocator()
+	return &Manager{
+		natService: natService,
+
+		publicIP:        location.PubIP,
+		outboundIP:      location.OutIP,
+		currentLocation: location.OutIP,
```
zolia:
```suggestion
		currentLocation: location.Country,
```

services/wireguard/resources/allocator_windows.go
soffokl: Changed the description of the function.

services/wireguard/resources/allocator_windows.go
soffokl: Changed it to compare with `nil` instead of `net.IPv4len`

services/wireguard/resources/allocator_windows.go
soffokl: Changed the description of the function.

tadovas:
wireguard-go ready

services/wireguard/endpoint/endpoint.go
```diff
@@ -35,7 +34,8 @@ type wgClient interface {
 	ConfigureDevice(name string, config wg.DeviceConfig, subnet net.IPNet) error
 	ConfigureRoutes(iface string, ip net.IP) error
 	DestroyDevice(name string) error
-	AddPeer(name string, peer wg.PeerInfo) error
+	AddPeer(name string, peer wg.PeerInfo, allowedIP ...string) error
+	DelPeer(name string, publicKey string) error
```
zolia:
```suggestion
	RemovePeer(name string, publicKey string) error
```

Issue #768, opened 2019-03-04
Internet Connection Sharing for Windows automatically set IP-address configuration for internal interface


soffokl:
Windows automatically set IP-address for the TUN interface to 192.168.137.1. 
Sometimes this happens after we finished interface configuration, sometimes before it.

We need to figure out how to keep settings of the TUN interface unchanged.

PR #806, opened 2019-03-18
Reassign interface address for Windows ICS

soffokl:
Closes: #768

### DIFF
--- a/nat/service_ics.go
+++ b/nat/service_ics.go
@@ -39,7 +39,8 @@ type serviceICS struct {
 	ifaces             map[string]RuleForwarding // list in internal interfaces with enabled internet connection sharing
 	remoteAccessStatus string
 	powerShell         func(cmd string) ([]byte, error)
-	setICSAddresses    func(ip string) error
+	setICSAddresses    func(config map[string]string) (map[string]string, error)
+	oldICSConfig       map[string]string
 }
 
 // Enable enables internet connection sharing for the public interface.
@@ -81,7 +82,11 @@ func (ics *serviceICS) Add(rule RuleForwarding) error {
 	}
 
 	ip := incrementIP(ipnet.IP)
-	if err := ics.setICSAddresses(ip.String()); err != nil {
+	ics.oldICSConfig, err = ics.setICSAddresses(map[string]string{
+		"ScopeAddress":          ip.String(),
+		"ScopeAddressBackup":    ip.String(),
+		"StandaloneDhcpAddress": ip.String()})
+	if err != nil {
 		return errors.Wrap(err, "failed to set ICS IP-address range")
 	}
 
@@ -124,6 +129,9 @@ func (ics *serviceICS) Del(rule RuleForwarding) error {
 
 // Disable disables internet connection sharing for the public interface.
 func (ics *serviceICS) Disable() (resErr error) {
+	if _, err := ics.setICSAddresses(ics.oldICSConfig); err != nil {
+		return errors.Wrap(err, "failed to revert ICS IP-address range")
+	}
 	for iface, rule := range ics.ifaces {
 		if err := ics.Del(rule); err != nil {
 			log.Errorf("%s Failed to cleanup internet connection sharing for '%s' interface: %v", natLogPrefix, iface, err)
--- a/nat/service_ics.go
+++ b/nat/service_ics.go
@@ -1,5 +1,3 @@
-//+build windows
-
 /*
  * Copyright (C) 2019 The "MysteriumNetwork/node" Authors.
  *
@@ -28,7 +26,6 @@ import (
 
 	log "github.com/cihub/seelog"
 	"github.com/pkg/errors"
-	"golang.org/x/sys/windows/registry"
 )
 
 const (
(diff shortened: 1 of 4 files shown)


nat/service_ics.go
```diff
@@ -113,6 +121,11 @@ func (nat *serviceICS) Del(rule RuleForwarding) error {
 
 // Disable disables internet connection sharing for the public interface.
 func (nat *serviceICS) Disable() (resErr error) {
+	if nat.ipAddrWatchdog != nil {
```
tadovas: These kind of checks and nil assigments are bug sources especially in multiple go routines. Is it really necessary to do it? Just init channel on constructor and close it when neeed

zolia: We probably should not be able to call 'Disable' if service is not enabled. `ipAddrWatchdog` then never would be nil.

tadovas: The concept must be improved anyway. Maybe watchdog go routine can do all cleanup when signal to stop is received, and exported Disable just closes the channel - that way it would be simpler and without a need for complex synchronization.

nat/service_ics.go
```diff
@@ -206,6 +219,57 @@ func (nat *serviceICS) getInternalInterfaceName() (string, error) {
 	return ifaceName, nil
 }
 
+func (nat *serviceICS) interfaceIPWatchdog() {
+	for {
+		select {
+		case <-time.After(5 * time.Second):
```
zolia: Yeach, this looks really a hack :( We should search for help probably here. Maybe we can write to some windows forum or stackoverflow with this issue. We will not be able to be confident with this.

nat/service_ics.go
```diff
@@ -206,6 +219,57 @@ func (nat *serviceICS) getInternalInterfaceName() (string, error) {
...
+					log.Warnf("%s Failed to assign IP-address '%s' to interface '%d'", natLogPrefix, ipnet.String(), ifaceName)
+				}
+			}
+		case <-nat.ipAddrWatchdog:
+			return
+		}
+	}
+}
+
+func (nat *serviceICS) setInterfaceAddr(name string, subnet *net.IPNet) error {
```
Waldz: `nat` is very wrong receiver name for `service..`:
- it's coupled with interface it's implementing, but interfaces should be on caller side only
- it tends to conflict with some package

nat/service_ics.go
zolia: @vkuznecovas suggest maybe to make this a bit more deterministic, we can check interface's IP right after it is configured? Did you test time span when this change from windows occurs?

nat/service_ics.go
```diff
@@ -206,6 +219,57 @@ func (nat *serviceICS) getInternalInterfaceName() (string, error) {
...
+		if addr.String() == subnet.String() {
+			return nil // Interface address already assigned, nothing to do.
+		}
+	}
+
+	out, err := nat.powerShell("netsh interface ip set address name=\"" + name + "\" source=static " + subnet.String())
+	return errors.Wrap(err, string(out))
+}
+
+func incrementIP(ip net.IP) {
```
Waldz: - should return newly calculated value `incrementIP(ip IP) IP`
- or should accept IP reference `incrementIP(ipPtr *IP)` if you're changing value to new value

nat/service_ics.go
soffokl: That problem is that IP is not assigned by ICS immediately after enabling sharing. It happens after some time. 
During my tests with manual IP assignment, it was not changed back after several hours. 
So, the current implementation sets it once and checking it every 5 seconds.

soffokl: The another possible way is to change ICS configuration in Windows registry to assigne our addresses through ICS. 
But we are limited to `/24` subnet using this way.

soffokl: Changed to the implementation for setting IP as ICS config.

nat/service_ics.go
soffokl: renamed

nat/service_ics.go
soffokl: It is pointer anyway. Changed it returns the new value to make it more clear.

nat/service_ics_windows.go
```diff
@@ -0,0 +1,35 @@
...
+	k, err := registry.OpenKey(registry.LOCAL_MACHINE, `System\CurrentControlSet\Services\SharedAccess\Parameters`, registry.QUERY_VALUE|registry.SET_VALUE)
+	if err != nil {
+		return err
+	}
+	defer k.Close()
+
+	if err := k.SetStringValue("ScopeAddress", ip); err != nil {
+		return err
+	}
+	return k.SetStringValue("ScopeAddressBackup", ip)
```
zolia: I would also set `StandaloneDhcpAddress` to the same address, not sure how defaults would behave if someone would try try get IP through DHCP requets without it.

soffokl: Hm, I don't have this parameter in my test Windows. I'll take a look.

soffokl: Looks like it is missing on Windows 10.

soffokl: In our case DHCP with TUN devices, Wireguard allowed IPs is not the best choice.

soffokl: Changed it to set and restore keys if it originally exists.

PR #811, opened 2019-03-20
payment pingpong now enabled by default

vkuznecovas:
enables the payment pingpong by default.

added a retry to send, increased the payment increments to 240secs and a 120sec timeout. Made the waiting for timeout  cancellable.

### DIFF
--- a/cmd/di.go
+++ b/cmd/di.go
@@ -275,7 +275,7 @@ func (di *Dependencies) bootstrapNodeComponents(nodeOptions node.Options) {
 	di.ConnectionRegistry = connection.NewRegistry()
 	di.ConnectionManager = connection.NewManager(
 		dialogFactory,
-		payment_factory.PaymentIssuerFactoryFunc(nodeOptions, di.SignerFactory, time.Second*20),
+		payment_factory.PaymentIssuerFactoryFunc(nodeOptions, di.SignerFactory, payment_factory.SendRetryDuration),
 		di.ConnectionRegistry.CreateConnection,
 		di.EventBus,
 	)
@@ -335,7 +335,7 @@ func newSessionManagerFactory(
 			// TODO: the ints and times here need to be passed in as well, or defined as constants
 			tracker := balance.NewBalanceTracker(&timeTracker, amountCalc, 0)
 			validator := validators.NewIssuedPromiseValidator(consumerID, receiverID, issuerID)
-			return session_payment.NewSessionBalance(sender, tracker, promiseChan, time.Second*240, time.Second*120, validator, promiseStorage, consumerID, receiverID, issuerID), nil
+			return session_payment.NewSessionBalance(sender, tracker, promiseChan, payment_factory.BalanceSendPeriod, payment_factory.PromiseWaitTimeout, validator, promiseStorage, consumerID, receiverID, issuerID), nil
 		}
 		return session.NewManager(
 			proposal,
--- a/cmd/di.go
+++ b/cmd/di.go
@@ -275,7 +275,7 @@ func (di *Dependencies) bootstrapNodeComponents(nodeOptions node.Options) {
 	di.ConnectionRegistry = connection.NewRegistry()
 	di.ConnectionManager = connection.NewManager(
 		dialogFactory,
-		payment_factory.PaymentIssuerFactoryFunc(nodeOptions, di.SignerFactory, payment_factory.SendRetryDuration),
+		payment_factory.PaymentIssuerFactoryFunc(nodeOptions, di.SignerFactory),
 		di.ConnectionRegistry.CreateConnection,
 		di.EventBus,
 	)
--- a/cmd/di.go
+++ b/cmd/di.go
@@ -51,6 +51,7 @@ import (
 	"github.com/mysteriumnetwork/node/money"
 	"github.com/mysteriumnetwork/node/nat"
 	service_noop "github.com/mysteriumnetwork/node/services/noop"
+	"github.com/mysteriumnetwork/node/services/openvpn"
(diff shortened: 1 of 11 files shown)


cmd/di.go
```diff
@@ -274,7 +275,7 @@ func (di *Dependencies) bootstrapNodeComponents(nodeOptions node.Options) {
 	di.ConnectionRegistry = connection.NewRegistry()
 	di.ConnectionManager = connection.NewManager(
 		dialogFactory,
-		payment_factory.PaymentIssuerFactoryFunc(nodeOptions, di.SignerFactory),
+		payment_factory.PaymentIssuerFactoryFunc(nodeOptions, di.SignerFactory, time.Second*20),
```
zolia: I would do it as a constant in payments package, would also hint naming for some number.

cmd/di.go
```diff
@@ -332,7 +335,7 @@ func newSessionManagerFactory(
 			// TODO: the ints and times here need to be passed in as well, or defined as constants
 			tracker := balance.NewBalanceTracker(&timeTracker, amountCalc, 0)
 			validator := validators.NewIssuedPromiseValidator(consumerID, receiverID, issuerID)
-			return session_payment.NewSessionBalance(sender, tracker, promiseChan, time.Second*5, time.Second*1, validator, promiseStorage, consumerID, receiverID, issuerID), nil
+			return session_payment.NewSessionBalance(sender, tracker, promiseChan, time.Second*240, time.Second*120, validator, promiseStorage, consumerID, receiverID, issuerID), nil
```
zolia: Here too config constants would be good.

session/payment/session_payments.go
```diff
@@ -45,22 +47,26 @@ type SessionPayments struct {
...
 		balanceChan:       balanceChan,
 		peerPromiseSender: peerPromiseSender,
 		promiseTracker:    promiseTracker,
 		balanceTracker:    balanceTracker,
+		sendRetryDuration: sendRetryDuration,
 	}
 }
 
 // balanceDifferenceThreshold determines the threshold where we'll cancel the session if there's a missmatch larger than the threshold provided between the provider and the consumer balances
 const balanceDifferenceThreshold uint64 = 20
```
zolia: I don't see it being used anywhere. If balance miss-match would occur, we should print out bald warning. 
In our case when we would pay for time, this should not occur.

If we would make payment by data transferred, that would mean we would have increasing balance miss-match. In such a case static threshold would not make sense, since difference would continuously increase. I would propose % difference,  not absolute.

session/payment/session_payments.go
```diff
@@ -103,14 +109,18 @@ func (cpo *SessionPayments) issuePromise(balance balance.Message) error {
...
-		Signature:  fmt.Sprintf("0x%v", hex.EncodeToString(issuedPromise.IssuerSignature)),
-	})
+
+	err = cpo.sendWithRetry(func() error {
+		return cpo.peerPromiseSender.Send(promise.Message{
+			Amount:     issuedPromise.Promise.Amount,
+			SequenceID: issuedPromise.Promise.SeqNo,
+			Signature:  fmt.Sprintf("0x%v", hex.EncodeToString(issuedPromise.IssuerSignature)),
+		})
+	}, 5, cpo.sendRetryDuration)
```
zolia: constant..

session/payment/session_payments.go
vkuznecovas: #812

session/payment/session_balance.go
```diff
@@ -116,23 +116,39 @@ func (sb *SessionBalance) Start() error {
...
 		}
 	}
 }
 
+func (sb *SessionBalance) sendBalanceExpectPromise() error {
+	err := sb.sendBalance()
+	if err != nil {
+		return err
+	}
+	err = sb.receivePromiseOrTimeout()
```
soffokl:
```suggestion
	return sb.receivePromiseOrTimeout()
```

cmd/di.go
```diff
@@ -306,8 +307,10 @@ func newSessionManagerFactory(
 ) session.ManagerFactory {
 	return func(dialog communication.Dialog) *session.Manager {
 		providerBalanceTrackerFactory := func(consumerID, receiverID, issuerID identity.Identity) (session.BalanceTracker, error) {
-			// if the flag ain't set, just return a noop balance tracker
-			if !nodeOptions.ExperimentPayments {
+			// We want backwards compatibility for openvpn on desktop providers, so no payments for them.
+			// Splitting this as a separate case just for that reason.
+			// TODO: remove this one day.
+			if proposal.ServiceType == openvpn.ServiceType {
```
Waldz: Proposal has `paymentType` - field specially for feature toggle like this:
- `openvpn` services can announce `paymentType=none`
- `noop` & `wireguard` services can announce `paymentType=per_time`

In such way, both Provider & Consumer can explicitly know if ping-pong game should be played

session/payment/session_payments.go
```diff
@@ -124,6 +136,26 @@ func (cpo *SessionPayments) validateBalanceDifference(balance uint64) error {
 	return nil
 }
 
+func (cpo *SessionPayments) sendWithRetry(send func() error, attempts int, delay time.Duration) error {
```
Waldz: I dont see this retry as necessary because promised are already repeated each Xs
I see this retry as retry of retries :)

cmd/di.go
vkuznecovas: Tried implementing this - there's an issue. Old consumers won't be able to decode the new paymentType, therefore they won't see the nodes. So this is a no-go for us. Leaving the same check here for backwards compatibility.

PR #815, opened 2019-03-22
Smaller timeouts for broker reconnect

soffokl:


services/wireguard/service/service_unix.go
```diff
@@ -112,7 +107,7 @@ func (manager *Manager) ProvideConfig(publicKey json.RawMessage) (session.Servic
 		}
 	}
 
-	return config, manager.once(destroy), nil
+	return config, destroy, nil
```
tadovas: This seems to be unrelated to PR

communication/nats/sender.go
```diff
@@ -94,3 +95,12 @@ func (sender *senderNATS) Request(producer communication.RequestProducer) (respo
 
 	return responsePtr, nil
 }
+
+func checkConnection(conn Connection) {
```
tadovas: I would like to see this function part of connection. I.e. conn.Check

communication/nats/discovery/config.go
```diff
@@ -23,6 +23,6 @@ import "time"
 const (
 	BrokerPort          = 4222
 	BrokerMaxReconnect  = -1
-	BrokerReconnectWait = 4 * time.Second
+	BrokerReconnectWait = 1 * time.Second
```
tadovas: It would be nice all these tweaks to be configurable from cmd line, with sensible defaults

soffokl: The issue for that added: https://github.com/mysteriumnetwork/node/issues/818

services/wireguard/service/service_unix.go
soffokl: Well, this is revert of the hack that was added because session destroys never worked. This PR fix session destroys, so these changes can be reverted. 
I can do it in another PR, but why not to include it here? :)

tadovas: Didn't notice the relation. All good 👍

communication/nats/discovery/config.go
tadovas: 👍

communication/nats/sender.go
soffokl: Changed it to be part of the connection.

communication/nats/discovery/address.go
```diff
@@ -136,3 +144,14 @@ func (address *AddressNATS) GetContact() market.Contact {
 		},
 	}
 }
+
+type connection struct {
+	*nats_lib.Conn
```
tadovas: 💯 points for delegate pattern

PR #817, opened 2019-03-25
Added Windows firewall configuration

### LINKED ISSUE #757: Support for Windows firewall auto configuration
Since we are going to support providing services on Windows, we need to be able to auto-configure the Windows internal firewall in case if it's enabled.

soffokl:
Closes #757

### DIFF
--- a/firewall/firewall_windows.go
+++ b/firewall/firewall_windows.go
@@ -21,7 +21,12 @@ import (
 	"errors"
 	"fmt"
 
+	log "github.com/cihub/seelog"
 	"github.com/mysteriumnetwork/node/utils"
+)
+
+const (
+	firewallLogPrefix = "[firewall] "
 )
 
 // AddInboundRule adds new inbound rule to the platform specific firewall.
@@ -29,12 +34,17 @@ func AddInboundRule(proto string, port int) error {
 	name := fmt.Sprintf("myst-%d:%s", port, proto)
 	cmd := fmt.Sprintf(`New-NetFirewallRule -DisplayName "%s" -Direction Inbound -LocalPort %d -Protocol %s -Action Allow`, name, port, proto)
 
-	if isInboundRuleExist(name) {
+	if inboundRuleExists(name) {
 		return nil
 	}
 
 	_, err := utils.PowerShell(cmd)
-	return err
+	if err != nil {
+		log.Trace(firewallLogPrefix, "Failed to add firewall rule: ", err)
+		return err
+	}
+
+	return nil
 }
 
 // RemoveInboundRule removes inbound rule from the platform specific firewall.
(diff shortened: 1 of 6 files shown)


firewall/firewall_windows.go
```diff
@@ -0,0 +1,61 @@
...
+
+	if !isInboundRuleExist(name) {
+		return errors.New("firewall rule not found")
+	}
+
+	_, err := utils.PowerShell(cmd)
+	return err
+}
+
+func isInboundRuleExist(name string) bool {
```
vkuznecovas:
```suggestion
func inboundRuleExists(name string) bool {
```

firewall/firewall_windows.go
```diff
@@ -0,0 +1,61 @@
...
+	cmd := fmt.Sprintf(`Get-NetFirewallRule -DisplayName "%s"`, name)
+
+	if _, err := utils.PowerShell(cmd); err != nil {
```
vkuznecovas: does this log errors inside? cause otherwise we're losing info, might want to warn/trace it.

services/wireguard/service/service_windows.go
```diff
@@ -131,6 +143,8 @@ func (manager *Manager) Serve(providerID identity.Identity) error {
 func (manager *Manager) Stop() error {
 	manager.wg.Done()
 
+	manager.cleanup()
```
vkuznecovas: this introduces a race condition. The cleanup might still be nil if I call serve stop in rapid succession. Need to put this under a lock with a proper check.

firewall/firewall_windows.go
soffokl: Added some more logging for Window firewall configuration.

services/wireguard/service/service_windows.go
soffokl: Changed it to be deferred right after creation.

soffokl:
I have just found that it does not properly for Windows 7. :(
Marking this as WIP.

soffokl:
Now firewall configuration supported for both Windows 10 and Windows 7.

utils/command_windows.go
```diff
@@ -0,0 +1,31 @@
...
+
+import (
+	"fmt"
+	"os/exec"
+)
+
+func PowerShell(cmd string) ([]byte, error) {
+	out, err := exec.Command("powershell", "-Command", cmd).CombinedOutput()
+	if err != nil {
+		return nil, fmt.Errorf("'powershell -Command %v': %v output: %s", cmd, err, out)
```
vkuznecovas: you might want to explicitly cast out to string as it might just display an array of bytes.

firewall/firewall_windows.go
```diff
@@ -0,0 +1,77 @@
...
+// AddInboundRule adds new inbound rule to the platform specific firewall.
+func AddInboundRule(proto string, port int) error {
+	name := fmt.Sprintf("myst-%d:%s", port, proto)
+	cmd := fmt.Sprintf(`netsh advfirewall firewall add rule name="%s" dir=in action=allow protocol=%s localport=%d`, name, proto, port)
+
+	if inboundRuleExists(name) {
+		return nil
+	}
+
+	_, err := utils.PowerShell(cmd)
```
zolia: You always swallow output from PowerShell. Not sure of its a good strategy. Maybe this output could explain us some miss execution. Maybe exec command will not return an error, but shell output would basically mean some issue. Maybe we can add this output to Trace level ?

soffokl: If command exited with 0 exit code we print nothing. If the exit code is not 0 we will return it as an error and it will be printed as `Trace`. 
I don't think that we need to print the output of successful commands.

utils/command_windows.go
soffokl: Fixed it.

firewall/firewall_windows.go
donce: We have lots of logs already, it's quite hard to understand them without deep knowledge of `core`. I think we should have proper filtering of log levels (i.e. skip `Trace` by default) before adding even more noise.

PR #820, opened 2019-03-25
payout set command in cli

interro:
`payout set <identity> <ethAddress>`

### DIFF
--- a/cmd/commands/cli/command.go
+++ b/cmd/commands/cli/command.go
@@ -358,13 +358,9 @@ func (c *cliApp) unlock(argsString string) {
 }
 
 func (c *cliApp) payout(argsString string) {
-	const usage = "payout command:\n    set"
-	if len(argsString) == 0 {
-		info(usage)
-		return
-	}
-
 	args := strings.Fields(argsString)
+
+	const usage = "payout command:\n    set"
 	if len(args) < 1 {
 		info(usage)
 		return
@@ -372,14 +368,7 @@ func (c *cliApp) payout(argsString string) {
 
 	action := args[0]
 	switch action {
-	case "set": // Known sub-commands.
-	default:
-		warnf("Unknown sub-command '%s'\n", action)
-		fmt.Println(usage)
-		return
-	}
-
-	if action == "set" {
+	case "set":
 		payoutSignature := "payout set <identity> <ethAddress>"
 		if len(args) < 2 {
 			info("Press tab to select identity.\n", payoutSignature)
@@ -401,6 +390,11 @@ func (c *cliApp) payout(argsString string) {
 		}
 
 		success(fmt.Sprintf("Payout address %s registered.", ethAddress))
+
+	default:
+		warnf("Unknown sub-command '%s'\n", action)
+		fmt.Println(usage)
+		return
 	}
 }
 
--- a/cmd/commands/cli/command.go
+++ b/cmd/commands/cli/command.go
@@ -150,6 +150,7 @@ func (c *cliApp) handleActions(line string) {
 		{command: "connect", handler: c.connect},
 		{command: "unlock", handler: c.unlock},
 		{command: "identities", handler: c.identities},
+		{command: "payout", handler: c.payout},
 		{command: "version", handler: c.version},
 		{command: "license", handler: c.license},
 		{command: "registration", handler: c.registration},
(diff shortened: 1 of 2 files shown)


tequilapi/client/client.go
```diff
@@ -228,6 +228,24 @@ func (client *Client) Unlock(identity, passphrase string) error {
...
+		ethAddress,
+	}
+
+	response, err := client.http.Put(path, payload)
+	if err != nil {
+		return err
+	}
+	defer response.Body.Close()
+
+	return nil
```
soffokl: Shouldn't we validate at least HTTP status here?

cmd/commands/cli/command.go
```diff
@@ -356,6 +357,53 @@ func (c *cliApp) unlock(argsString string) {
...
+	action := args[0]
+	switch action {
+	case "set": // Known sub-commands.
+	default:
+		warnf("Unknown sub-command '%s'\n", action)
+		fmt.Println(usage)
+		return
+	}
+
+	if action == "set" {
```
soffokl: This block can be a part of the `switch case "set"`, we don't need to check it twice.

cmd/commands/cli/command.go
```diff
@@ -356,6 +357,53 @@ func (c *cliApp) unlock(argsString string) {
...
+		payoutSignature := "payout set <identity> <ethAddress>"
+		if len(args) < 2 {
+			info("Press tab to select identity.\n", payoutSignature)
+			return
+		}
+
+		var identity, ethAddress string
+		if len(args) == 3 {
```
soffokl: There are 4 checks:
`if len(argsString) == 0 {`
`if len(args) < 1 {`
`if len(args) < 2 {`
`if len(args) == 3 {`

But looks like only one requried `if len(args) != 3 { return error }`

interro: You are right - only one is requried `if len(args) == 3 { ... }` to check if command, sub-command and parameter structure is valid.
All other checks are for console to help user to navigate through commands and parameters.

interro: Reduced checks

cmd/commands/cli/command.go
interro: Moved to switch

tequilapi/client/client.go
interro: Why it should be validated here?

cmd/commands/cli/command.go
```diff
@@ -356,6 +357,47 @@ func (c *cliApp) unlock(argsString string) {
 	success(fmt.Sprintf("Identity %s unlocked.", identity))
 }
 
+func (c *cliApp) payout(argsString string) {
+	args := strings.Fields(argsString)
+
+	const usage = "payout command:\n    set"
+	if len(args) < 1 {
```
vkuznecovas: I think this is more explicit and clear:

```suggestion
	if len(args) == 0 {
```

cmd/commands/cli/command.go
```diff
@@ -356,6 +357,47 @@ func (c *cliApp) unlock(argsString string) {
...
+		info(usage)
+		return
+	}
+
+	action := args[0]
+	switch action {
+	case "set":
+		payoutSignature := "payout set <identity> <ethAddress>"
+		if len(args) < 2 {
+			info("Press tab to select identity.\n", payoutSignature)
```
vkuznecovas: The information should be expanded upon here, you don't really need to press tab, you can enter it manually as well... So maybe something like: "Please provide an identity. You can select one by pressing tab.". It's more generic and useful in both cases.

cmd/commands/cli/command.go
```diff
@@ -356,6 +357,47 @@ func (c *cliApp) unlock(argsString string) {
...
+		}
+
+		err := c.tequilapi.Payout(identity, ethAddress)
+		if err != nil {
+			warn(err)
+			return
+		}
+
+		success(fmt.Sprintf("Payout address %s registered.", ethAddress))
+
```
vkuznecovas: redundant newline

tequilapi/client/client.go
vkuznecovas: because you can get stuff like 404, 503 and what not from the put response. Not having an error is not an indicator of success here.

interro: [`parseResponseError`](https://github.com/mysteriumnetwork/node/blob/master/tequilapi/client/http_client.go#L128) will handle statuses like 404, 503 by creating error

cmd/commands/cli/command.go
interro: `+`

cmd/commands/cli/command.go
interro: expanded message

cmd/commands/cli/command.go
interro: `+`

PR #821, opened 2019-03-26
Add endpoint to get payout info

donce:
For https://github.com/mysteriumnetwork/mysterium-vpn/issues/458

### DIFF
--- a/requests/request.go
+++ b/requests/request.go
@@ -41,30 +41,30 @@ func NewGetRequest(apiURI, path string, params url.Values) (*http.Request, error
 
 // NewPostRequest generates http Post request
 func NewPostRequest(apiURI, path string, requestBody interface{}) (*http.Request, error) {
-	bodyBytes, err := encodeToJSON(requestBody)
+	encodedBody, err := encodeToJSON(requestBody)
 	if err != nil {
 		return nil, err
 	}
-	return newRequest(http.MethodPost, apiURI, path, bodyBytes)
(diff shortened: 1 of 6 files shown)


requests/request.go
```diff
@@ -50,12 +50,16 @@ func NewPostRequest(apiURI, path string, requestBody interface{}) (*http.Request
 
 // NewSignedRequest signs payload and generates http request
 func NewSignedRequest(httpMethod, apiURI, path string, requestBody interface{}, signer identity.Signer) (*http.Request, error) {
-	bodyBytes, err := encodeToJSON(requestBody)
-	if err != nil {
-		return nil, err
+	var bodyBytes []byte = nil
```
soffokl: I'd name it just `body`, we don't need to include type into the variable name.

requests/request.go
```diff
@@ -96,3 +105,12 @@ func newRequest(method, apiURI, path string, body []byte) (*http.Request, error)
 
 	return req, nil
 }
+
+func getBodySignature(bodyBytes []byte, signer identity.Signer) (identity.Signature, error) {
```
soffokl: The same naming problem here with `bodyBytes`.

requests/request.go
donce: Renamed to `encodedBody`, because we have `requestBody` in the same scope, so having `requestBody` and `body` in the same scope seemed confusing.

PR #837, opened 2019-04-01
Bug/powershell bom outputs

vkuznecovas:
* Powershell outputs are now stripped of BOM(both out and error) as this was causing fmt.Print and seelog to crash.
* Replaced `Get-Service RemoteAccess | foreach { $_.StartType }` with `(Get-WmiObject Win32_Service -filter "Name='RemoteAccess'").StartMode` as the latter is compatible with older powershell versions(looking at you 2.0) as well as the current ones.

### DIFF
--- a/nat/service_ics.go
+++ b/nat/service_ics.go
@@ -24,7 +24,6 @@ import (
 	"sync"
 
 	log "github.com/cihub/seelog"
-	"github.com/mysteriumnetwork/node/utils"
 	"github.com/pkg/errors"
 )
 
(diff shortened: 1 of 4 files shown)


nat/service_ics.go
```diff
@@ -59,11 +58,17 @@ func (ics *serviceICS) Enable() error {
 }
 
 func (ics *serviceICS) enableRemoteAccessService() error {
-	status, err := ics.powerShell("Get-Service RemoteAccess | foreach { $_.StartType }")
+	status, err := ics.powerShell(`(Get-WmiObject Win32_Service -filter "Name='RemoteAccess'").StartMode`)
```
soffokl: Are we replacing because of `foreach` or because of `Get-Service`? Asking because we have `Set-Service` and `foreach` in other places too.

vkuznecovas: Because of `Get-Service` here concretely. In 2.0 the Get-Service is not aware of the StartType property and does not have anything that's equivalent.

Issue #532, opened 2018-11-12
session-create broker random topic 


zolia:
Currently we have following format for registering new session with a broker: 
`0x207bab3168645844c8f53a5bc220557e622701fb.0x8e586ca51c0f0e83564747bfd4d73bcd90119fc7.session-create`

`ProviderID.ConsumerID.session-create`

This poses security threat since attacker might subscribe via broker to a-priori know topic.

Replace predictable topic name with some random.

PR #846, opened 2019-04-03
Using unique communication topic for every session

soffokl:
Closes https://github.com/mysteriumnetwork/node/issues/532


communication/nats/dialog/dialog_establisher.go
```diff
@@ -67,31 +66,32 @@ func (establisher *dialogEstablisher) EstablishDialog(
 	peerCodec := establisher.newCodecForPeer(peerID)
 
 	peerSender := establisher.newSenderToPeer(peerAddress, peerCodec)
-	err = establisher.negotiateDialog(peerSender)
+	topic, err := establisher.negotiateDialog(peerSender)
```
tadovas: maybe negiotiatTopic? It strange to see dialog negotiation before new dialog creation (the line below)

soffokl: Right, renamed it.

PR #850, opened 2019-04-03
NAT proxy for OpenVPN

zolia:
Implements NATProxy for OpenVPN.

* if '--experiment-natpunching' is enabled and port auto-configuration fails node starts sending NAT pings
* backwards compatible with clients without NATPinger functionality (or disabled)
* NATProxy implementation
* NATProxy for TCP protocol is bypassed
* Removed restartable openvpn service code since not required now

What lacks: 
* Think how to test this
* More testing with actual clients
* Speed tests


nat/traversal/pinger.go
```diff
@@ -116,9 +123,10 @@ func (p *Pinger) Start() {
 				continue
 			}
 
-			p.pingReceived <- struct{}{}
+			// p.pingReceived <- struct{}{}
```
Waldz: rudiment?

services/openvpn/service/factory.go
```diff
@@ -135,19 +135,22 @@ func newSessionConfigNegotiatorFactory(networkOptions node.OptionsNetwork, servi
...
-	vpnConfig      openvpn_service.VPNConfig
+	vpnConfig      *openvpn_service.VPNConfig
 }
 
 // ProvideConfig returns the config for user
 func (ocn *OpenvpnConfigNegotiator) ProvideConfig(json.RawMessage) (session.ServiceConfiguration, session.DestroyCallback, error) {
 	localPort := ocn.determineClientPort()
 	ocn.vpnConfig.LocalPort = localPort
-	return &ocn.vpnConfig, nil, nil
+	log.Info("remote port: ", ocn.vpnConfig.RemotePort)
```
Waldz: Would be good to have prefix for logs. Or is it a leftover of debugging?

services/openvpn/client.go
```diff
@@ -89,6 +90,13 @@ func (c *Client) GetConfig() (connection.ConsumerConfig, error) {
 		return nil, errors.Wrap(err, "failed to get consumer config")
 	}
 	c.publicIP = ip
+
+	switch c.natPinger.(type) {
+	case *traversal.NoopPinger:
+		log.Infof("noop pinger detected, returning nil client config.")
+		return nil, nil
+	}
```
donce: It looks like configuration depends on implementation of `traversal.Pinger` - maybe it should provide the config to avoid type-conditionals?

nat/traversal/pinger.go
```diff
@@ -104,43 +130,34 @@ func (p *Pinger) Start() {
...
+			go p.natProxy.handOff(serviceType, conn)
 		}
 	}
 }
 
-// Stop stops the nat pinger
+// Stop noop method
 func (p *Pinger) Stop() {
-	p.once.Do(func() { close(p.stop) })
+	// noop method - NATPinger should not stop
```
vkuznecovas: This is bad, you're leaving an infinite for loop hanging in some goroutine for no reason. We should stop gracefully.

services/openvpn/service/manager_test.go
```diff
@@ -21,6 +21,8 @@ import (
 	"encoding/json"
 	"testing"
 
+	"github.com/mysteriumnetwork/node/core/port"
+
```
vkuznecovas: redundant whitespace

services/type.go
```diff
@@ -0,0 +1,21 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package services
+
+// ServiceType identifies service type
+type ServiceType string
```
vkuznecovas: does this really deserve its own file?

session/create_consumer.go
```diff
@@ -20,6 +20,8 @@ package session
 import (
 	"encoding/json"
 
+	"github.com/mysteriumnetwork/node/nat/traversal"
+
```
vkuznecovas: redundant whitespace

session/create_consumer.go
```diff
@@ -61,7 +63,13 @@ func (consumer *createConsumer) NewRequest() (requestPtr interface{}) {
 func (consumer *createConsumer) Consume(requestPtr interface{}) (response interface{}, err error) {
 	request := requestPtr.(*CreateRequest)
 
-	config, destroyCallback, err := consumer.configProvider(request.Config)
+	var pingerPort int
+	config, destroyCallback, err := consumer.configProvider(
+		request.Config,
+		func(port int) int {
```
vkuznecovas: this confuses me, you're taking in a port, and returning the same port?

session/create_consumer_test.go
```diff
@@ -22,14 +22,16 @@ import (
 	"errors"
 	"testing"
 
+	"github.com/mysteriumnetwork/node/nat/traversal"
+
```
vkuznecovas: redundant whitespace

session/manager.go
```diff
@@ -23,6 +23,8 @@ import (
 	"sync"
 	"time"
 
+	"github.com/davecgh/go-spew/spew"
+
```
vkuznecovas: redundant whitespace

session/manager.go
```diff
@@ -159,15 +158,8 @@ func (manager *Manager) Create(consumerID identity.Identity, issuerID identity.I
...
-	// start NAT pinger here, do not block - configuration should be returned to consumer
-	// start NAT pinger, get hole punched, launch service.
-	//  on session-destroy - shutdown service and wait for session-create
-	// TODO: We might want to start a separate openvpn daemon if node is behind the NAT
-
-	// We need to know that session creation is already in-progress here
-
-	// postpone vpnServer start until NAT hole is punched
-	manager.natPingerChan(requestConfig)
+	spew.Dump(pingerParams)
```
vkuznecovas: don't need to dump the whole struct here

nat/traversal/nat_proxy.go
```diff
@@ -0,0 +1,96 @@
...
+	return &NATProxy{
+		servicePorts: make(map[services.ServiceType]int),
+	}
+}
+
+// handOff traffic incoming through NATPinger punched hole should be handed off to NATPoxy
+func (np *NATProxy) handOff(serviceType services.ServiceType, incomingConn *net.UDPConn) {
+	proxyConn, err := np.getConnection(serviceType)
+	if err != nil {
+		log.Error(logPrefix, "failed to connect to NATProxy: ", err)
```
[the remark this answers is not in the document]
zolia: Yes, basically theres no point to proceed. As for @donce, we don't return error here, since we run in throw away go routine in case of failure, just print an error for each connect attempt.

nat/traversal/nat_proxy.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+
+	return conn, nil
+}
+
+func (np *NATProxy) isAvailable(serviceType services.ServiceType) bool {
+	if np.servicePorts[serviceType] > 0 {
+		return true
+	}
+	return false
```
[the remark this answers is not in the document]
zolia: `NATProxy.isAvailable(serviceType)`

you should read this method as: "is there available NATProxy for specified service type?".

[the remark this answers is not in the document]
zolia: @donce we are asserting here that we do not have '0' as a valid value in a map, since '0' would not work here.

services/openvpn/client.go
zolia: Not sure if it is a bad solution, since introducing another config dependency will additionally span throughout the code. Does if worth it?

nat/traversal/nat_proxy.go
```diff
@@ -0,0 +1,96 @@
...
+		defer proxyConn.Close()
+		totalBytes, err := io.Copy(incomingConn, proxyConn)
+		if err != nil {
+			log.Error(logPrefix, "failed to copy stream to NATProxy: ", err)
+		}
+		log.Info(logPrefix, "total bytes outgoing to client: ", totalBytes)
+	}()
+}
+
+func (np *NATProxy) registerServicePort(serviceType services.ServiceType, protocol string, port int) {
```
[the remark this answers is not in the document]
zolia: true, rebase leftovers.

session/create_consumer.go
zolia: here can be any number, but basically this `configProvider` spans as interface to service type configProviders and there we already return a value.

zolia: made a comment in code.

nat/traversal/nat_proxy.go
```diff
@@ -0,0 +1,87 @@
...
+	udpAddr, err := net.ResolveUDPAddr("udp4", fmt.Sprintf("127.0.0.1:%d", np.servicePorts[serviceType]))
+	if err != nil {
+		return nil, err
+	}
+
+	return net.DialUDP("udp", nil, udpAddr)
+}
+
+func (np *NATProxy) isAvailable(serviceType services.ServiceType) bool {
+	if np.servicePorts[serviceType] > 0 {
```
soffokl:
```suggestion
	return np.servicePorts[serviceType] > 0
```

nat/traversal/pinger.go
```diff
@@ -33,69 +36,92 @@ const prefix = "[NATPinger] "
...
 type NatEventWaiter interface {
 	WaitForEvent() Event
 }
 
+// ConfigParser is able to parse a config from given raw json
+type ConfigParser interface {
+	Parse(config json.RawMessage) (ip string, port int, serviceType services.ServiceType, err error)
+}
+
+type portSupplier interface {
```
soffokl: We gave this interface in multiple places. Can't we define it once somewhere? 
There are many examples of such definitions in a standard library: https://golang.org/pkg/io/#Reader

nat/traversal/pinger.go
```diff
@@ -33,69 +36,92 @@ const prefix = "[NATPinger] "
...
-// Start starts NAT pinger and waits for pingTarget to ping
+type natProxy interface {
+	handOff(serviceType services.ServiceType, conn *net.UDPConn)
+	registerServicePort(serviceType services.ServiceType, port int)
+	isAvailable(serviceType services.ServiceType) bool
+}
+
+// Params contains session parameters needed to NAT ping remote peer
+type Params struct {
+	RequestConfig json.RawMessage
```
soffokl: Why is it called `RequesConfig`? 
Is it a config of the request?

zolia: This config comes with session setup request.

nat/traversal/pinger.go
```diff
@@ -33,69 +35,98 @@ const prefix = "[NATPinger] "
...
+				log.Warn(prefix, errors.Wrap(err, fmt.Sprintf("unable to parse ping message: %v", pingParams)))
+			}
+
+			if !p.natProxy.isAvailable(serviceType) {
+				log.Warn(prefix, serviceType, " NATProxy is not available for this transport protocol")
+				continue
 			}
 
-			log.Infof("%s ping target received: IP: %v, port: %v", prefix, IP, port)
+			log.Infof("%sping target received: IP: %v, port: %v", prefix, IP, port)
```
Waldz: prefix missing

zolia: its not, its format string, prefix at the end.

services/openvpn/service/manager_test.go
vkuznecovas: this whitespace is still redundant

cmd/service_bootstrap_desktop.go
```diff
@@ -132,6 +132,12 @@ func (di *Dependencies) bootstrapServiceOpenvpn(nodeOptions node.Options) {
 		}
 
 		proposal := openvpn_discovery.NewServiceProposalWithLocation(currentLocation, transportOptions.Protocol)
+
+		var portSupplier port.ServicePortSupplier = di.PortPool
+		if transportOptions.Port != 0 && locationInfo.PubIP == locationInfo.OutIP {
```
tadaskay: I'm a bit surprised nobody noticed. This actually changes behavior in most cases (because most clients will have a router), and `openvpn.port` flag/option is ignored and thus it's always randomized (which causes problems for desktop clients, so for now we decided to use static port `1194` - passing it through options when starting service via tequilapi).

Why this check was needed: `locationInfo.PubIP == locationInfo.OutIP`? I can't find any useful info in commit messages. Any ideas, @zolia , @vkuznecovas ?

soffokl: `locationInfo.PubIP == locationInfo.OutIP` supposed to check if we are behind the NAT. 
If IP on the host and public IP is the same, we, probably, don't need to do any NAT traversal since IP is reachable without it.

tadaskay: True, but I think we should respect user flags despite this.

zolia:
> You're adding hard coupling between the packages where there need not be any
For future reference. I see upsides and downsides of such solution. In this case, if we have package which is very very feature isolated and self sustained i think such package could expose more generic interfaces. In this case `port` package IMHO is exactly such a package.

Issue #829, opened 2019-03-28
Clean sessions from storage when service is stopped


Waldz:
*Problem:*
Now when Provider stops service, `GET /service-sessions` endpoint still returns previous sessions

Need:
- either implement `session.Instance.Status` to distinguish sessions from closed ones 
- either clear "current sessions" from `SessionStorage`
- also consider destroying session on Consumer side

zolia:
On app we can see that after first connect it instantly sets counter to 4 and after disconnect it leaves at 3.

PR #851, opened 2019-04-03
Destroy sessions of stopped service

donce:
Fixes https://github.com/mysteriumnetwork/node/issues/829

### DIFF
--- a/cmd/service_bootstrap_desktop.go
+++ b/cmd/service_bootstrap_desktop.go
@@ -168,7 +168,8 @@ func (di *Dependencies) bootstrapServiceComponents(nodeOptions node.Options) {
 	}
 	newDialogHandler := func(proposal market.ServiceProposal, configProvider session.ConfigNegotiator) communication.DialogHandler {
 		sessionManagerFactory := newSessionManagerFactory(
-			proposal, di.ServiceSessionStorage,
+			proposal,
+			di.ServiceSessionStorage,
 			di.PromiseStorage,
 			nodeOptions,
 			di.NATPinger.PingTarget,
@@ -185,5 +186,6 @@ func (di *Dependencies) bootstrapServiceComponents(nodeOptions node.Options) {
 		newDialogHandler,
 		newDiscovery,
 		di.NATPinger,
+		di.ServiceSessionStorage,
 	)
 }
--- a/cmd/service_bootstrap_desktop.go
+++ b/cmd/service_bootstrap_desktop.go
@@ -186,6 +186,11 @@ func (di *Dependencies) bootstrapServiceComponents(nodeOptions node.Options) {
 		newDialogHandler,
 		newDiscovery,
 		di.NATPinger,
-		di.ServiceSessionStorage,
+		di.EventBus,
 	)
+
+	serviceCleaner := service.Cleaner{SessionStorage: di.ServiceSessionStorage}
+	if err := di.EventBus.Subscribe(service.StopTopic, serviceCleaner.Cleanup); err != nil {
+		log.Error(logPrefix, "failed to subscribe service cleaner")
+	}
 }
--- a/cmd/service_bootstrap_desktop.go
+++ b/cmd/service_bootstrap_desktop.go
@@ -166,7 +166,7 @@ func (di *Dependencies) bootstrapServiceComponents(nodeOptions node.Options) {
 			di.IdentityRegistry,
 		), nil
 	}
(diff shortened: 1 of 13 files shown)


core/service/manager_test.go
```diff
@@ -21,6 +21,7 @@ import (
 	"errors"
 	"testing"
 
+	"github.com/asaskevich/EventBus"
```
Waldz: What kind of dependency is here, I think we have own implementation of event wo external dependencies

session/storage_memory.go
```diff
@@ -81,3 +81,13 @@ func (storage *StorageMemory) Remove(id ID) {
 	defer storage.lock.Unlock()
 	delete(storage.sessions, id)
 }
+
+// RemoveForProposal removes all sessions with given proposal id
+func (storage *StorageMemory) RemoveForProposal(proposalID int) {
```
Waldz: Why filtering by proposalId, it should search by unique index (like `session.id`) and there is a function for that already `storage.Remove(id)`

session/manager.go
```diff
@@ -131,6 +131,7 @@ func (manager *Manager) Create(consumerID identity.Identity, issuerID identity.I
 	sessionInstance.ConsumerID = consumerID
 	sessionInstance.done = make(chan struct{})
 	sessionInstance.Config = config
+	sessionInstance.ProposalID = proposalID
```
Waldz: I don't think we this field in session, it's a serial number for proposal but not unique index itself

core/service/manager.go
```diff
@@ -171,7 +182,18 @@ func (manager *Manager) Kill() error {
...
+	if instance == nil {
+		return errors.New("service not found")
+	}
+
+	err := manager.servicePool.Stop(id)
+	if err != nil {
+		return err
+	}
+
+	manager.eventPublisher.Publish(StopTopic, instance)
```
Waldz: I expect all session stop logic to live in `servicePool.Stop()`. No point to do searching twice, it's already done there

core/service/manager_test.go
donce: It doesn't seem that we have our own implementation - we're even using the interface from this external repo in `di.go`:
https://github.com/mysteriumnetwork/node/blob/master/cmd/di.go#L25

session/storage_memory.go
donce: When stopping service, we need to find sessions which belong to this service.
I've done this outside of `StorageMemory` at first, but then that method was feature-envy. So I've moved that method here.

session/manager.go
donce: It's not supposed to be unique index, we have session id for that. This field allows finding which sessions belong to service being stopped.

core/service/manager.go
donce: At first I've done all cleanup directly, but then I've added this event as request from @vkuznecovas .
The idea is that all the management is getting too complex, so we want to decouple different components/features by using event bus, signalling all events (service start, stop and possibly others).

At this point, this event doesn't seem very useful, but it's a start towards decoupling :)

cmd/service_bootstrap_desktop.go
```diff
@@ -185,5 +186,11 @@ func (di *Dependencies) bootstrapServiceComponents(nodeOptions node.Options) {
 		newDialogHandler,
 		newDiscovery,
 		di.NATPinger,
+		di.EventBus,
 	)
+
+	serviceCleaner := service.Cleaner{SessionStorage: di.ServiceSessionStorage}
+	if err := di.EventBus.Subscribe(service.StopTopic, serviceCleaner); err != nil {
```
zolia: We probably we should call a Cleanup function, not pass an instance of Cleaner

core/service/manager.go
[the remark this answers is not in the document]
donce: Well, we call it `Topic` everywhere, even though we're using `EventBus` term for the message queue. Actually, this naming doesn't come from us - it comes from asaskevich/event-bus repo, which we are coupled to somehow.

I've created a separate task to decouple from this implementation, and then we will be able to have our own naming: https://github.com/mysteriumnetwork/node/issues/854

core/service/manager_test.go
donce: Created an issue for this not to polute this PR with more changes (this PR is already up for 2 days!): https://github.com/mysteriumnetwork/node/issues/854

session/manager.go
donce: Oh, now I got what you meant - will fix this.

Issue #689, opened 2019-01-30
generate service ports dynamically


zolia:
In order to reduce possible DOS attacks and to reduce port clashes on hosts and gateways we need to generate random ports for our services.

 - ATM WG uses multi-interface setup (except windows), which means it administers port pool from a range. 
 - OpenVPN uses single interface, thus single port. 

Create a port pool service, which administers port pool. Services would request a port from such pool. Port pool service would return a random port from its pool and mark it as used. After service is shut down, port should be freed. 

Tasklist:
- [x] Openvpn https://github.com/mysteriumnetwork/node/pull/866
- [x] Wireguard https://github.com/mysteriumnetwork/node/pull/880

Waldz:
Leftover #770

PR #866, opened 2019-04-08
Dynamically generate service port for OpenVPN

tadaskay:
Ref: https://github.com/mysteriumnetwork/node/issues/689

`core/port/Pool` tests random TCP/UDP ports for availability and returns for service use (range 40000-50000 by default). This PR covers OpenVPN part.

Updated docs: https://github.com/mysteriumnetwork/docs/pull/29


cmd/di.go
```diff
@@ -19,6 +19,7 @@ package cmd
 
 import (
 	"encoding/json"
+	"github.com/mysteriumnetwork/node/port"
```
zolia: i would go with 
```suggestion
	"github.com/mysteriumnetwork/node/core/pools/port"
```

port/pool.go
```diff
@@ -0,0 +1,46 @@
...
+type Port struct {
+	Number int
+}
+
+func (p *Port) Release() {
+	log.Debug("releasing port back to pool: ", p.Number)
+}
+
+type Pool struct {
+	start, cap int
```
zolia: Do not abbreviate in structs, unless its acronym or very know abbreviation. Cap may mean max, top not only amount. `capacity` not that long word after all  :)

port/pool.go
```diff
@@ -0,0 +1,46 @@
...
+	log "github.com/cihub/seelog"
+	"math/rand"
+	"time"
+)
+
+type Port struct {
+	Number int
+}
+
+func (p *Port) Release() {
```
zolia: All exported methods should have documentation.

cmd/service_bootstrap_desktop.go
```diff
@@ -114,19 +115,31 @@ func (di *Dependencies) bootstrapServiceOpenvpn(nodeOptions node.Options) {
 		currentLocation := market.Location{Country: locationInfo.Country}
 		transportOptions := serviceOptions.(openvpn_service.Options)
 
-		mapPort := func() func() {
+		mapPort := func(port int) func() {
 			return mapping.GetPortMappingFunc(
 				locationInfo.PubIP,
 				locationInfo.OutIP,
 				transportOptions.Protocol,
-				transportOptions.Port,
```
zolia: We cannot remove transportOptions altogether. User should be left with an option to configure their own, static port. Default value should be removed, or set to '0' and only if it is '0' we should activate pool logic.

port/pool.go
```diff
@@ -0,0 +1,46 @@
...
+func NewPool() *Pool {
+	return &Pool{
+		start: 40000,
+		cap:   10000,
+		seed:  rand.NewSource(time.Now().UnixNano()),
+	}
+}
+
+func (pool *Pool) Start() *Pool {
+	//log.Debug("starting port pool")
```
zolia: Commented code usually will not go into master

port/pool.go
```diff
@@ -0,0 +1,46 @@
+package port
+
+import (
+	log "github.com/cihub/seelog"
+	"math/rand"
+	"time"
+)
+
+type Port struct {
```
zolia: simple `type Port int` should be enough in current situation.

port/pool.go
zolia: If you acquire through pool, you should return through pool too. Imho Port struct is not very needed here at all.

core/port/pool.go
```diff
@@ -0,0 +1,50 @@
...
+	start, capacity int
+	seed            rand.Source
+}
+
+// NewPool creates a port pool that will provide ports from range 40000-50000
+func NewPool() *Pool {
+	return &Pool{
+		start:    40000,
+		capacity: 10000,
+		seed:     rand.NewSource(time.Now().UnixNano()),
```
vkuznecovas: why store the seed and not the rand itself?

core/port/port.go
```diff
@@ -0,0 +1,38 @@
...
+}
+
+// Num returns port's numeric value
+func (p *Port) Num() int {
+	return p.number
+}
+
+// Release releases the port back to the pool.
+// After it is called, port is considered disposed and should not be used.
+func (p *Port) Release() {
```
vkuznecovas: should this not be on the pool itself?

services/openvpn/service/manager.go
```diff
@@ -59,10 +59,16 @@ type NATEventGetter interface {
...
+	Acquire() port.Port
+}
+
 // Manager represents entrypoint for Openvpn service with top level components
 type Manager struct {
 	natService     nat.NATService
-	mapPort        func() (releasePortMapping func())
+	mapPort        func(int) (releasePortMapping func())
+	ports          PortSupplier
+	port           port.Port
```
vkuznecovas: i don't think this needs to be stored on the struct, explained below

services/openvpn/service/manager.go
```diff
@@ -91,23 +97,24 @@ func (m *Manager) Serve(providerID identity.Identity) (err error) {
 		return errors.Wrap(err, "failed to add NAT forwarding rule")
 	}
 
-	m.releasePorts = m.mapPort()
+	m.port = m.ports.Acquire()
```
vkuznecovas: afaik the serve is blocking. in this case, we can just do a 
```golang
port := m.ports.Acquire()
defer port.Release()
```
same can be done with the firewall and we then don't have to extend the struct with the redundant field.

tadaskay: Although `Release` functionality in the latest version has been removed, it is still needed on the manager instance, because of `Stop`:

`firewall.RemoveInboundRule(m.serviceOptions.Protocol, m.port.Num())`

core/port/pool.go
tadaskay: I'm not entirely sure, what's the difference?

core/port/port.go
tadaskay: In the latest version, this is removed. Anyway, this was placed on the port rather than the pool to make it more cohesive (you may release a port only after acquiring it)

cmd/service_bootstrap_desktop.go
tadaskay: Fixed

cmd/di.go
tadaskay: Moved to `node/core/port`. Pool is an implementation detail, so I wanted to avoid that.

core/port/pool.go
tadaskay: Changed to `rand`, just in case.

services/openvpn/service/manager.go
```diff
@@ -91,23 +98,27 @@ func (m *Manager) Serve(providerID identity.Identity) (err error) {
 		return errors.Wrap(err, "failed to add NAT forwarding rule")
 	}
 
-	m.releasePorts = m.mapPort()
+	m.port, err = m.ports.Acquire(m.serviceOptions.Protocol)
+	if err != nil {
+		return errors.Wrap(err, "failed to acquire an unused port")
+	}
+	m.releasePorts = m.mapPort(m.port.Num())
```
vkuznecovas: can still get rid of the m.port and m.releasePorts. We'll have less port variables, makes it easier to understand as I'm a bit confused now by the naming, especially this part: 
```golang
m.releasePorts = m.mapPort(m.port.Num())
```

So i suggest the following changes:

```golang
port, err := m.ports.Acquire(m.serviceOptions.Protocol)
if err != nil {
	return errors.Wrap(err, "failed to acquire an unused port")
}
defer m.mapPort(m.port.Num())()
```

then for the firewall part:

```golang
log.Info(logPrefix, "starting openvpn server on port: ", m.port.Num())
if err := firewall.AddInboundRule(m.serviceOptions.Protocol, port.Num()); err != nil {
		return errors.Wrap(err, "failed to add firewall rule")
}
defer func() {
        if err := firewall.RemoveInboundRule(m.serviceOptions.Protocol, port.Num()); err != nil {
		log.Error(logPrefix, "Failed to delete firewall rule for OpenVPN", err)
	}
} ()
```

core/port/pool.go
```diff
@@ -0,0 +1,69 @@
...
+	"time"
+
+	log "github.com/cihub/seelog"
+	"github.com/pkg/errors"
+)
+
+// Pool hands out ports for service use
+type Pool struct {
+	start, capacity int
+	rng             *rand.Rand
```
zolia: i would use `rnd`

core/port/pool.go
```diff
@@ -0,0 +1,69 @@
...
+func (pool *Pool) Acquire(protocol string) (Port, error) {
+	p := pool.randomPort()
+	if !available(protocol, p) {
+		var err error
+		p, err = pool.seekAvailablePort(protocol)
+		if err != nil {
+			return 0, err
+		}
+	}
+	log.Debugf(logPrefix+"supplying %s port %d", protocol, p)
```
zolia:
```suggestion
	log.Debugf("%s supplying %s port %d", logPrefix, protocol, p)
```

core/port/pool_test.go
```diff
@@ -0,0 +1,65 @@
...
+	err = listenUdp(udpPort.Num())
+	assert.NoError(t, err)
+}
+
+func listenTcp(port int) error {
+	listener, err := net.Listen("tcp", ":"+strconv.Itoa(port))
+	if err != nil {
+		return err
+	}
+	_ = listener.Close()
```
zolia:
```suggestion
	listener.Close()
```

core/port/pool_test.go
```diff
@@ -0,0 +1,65 @@
...
+func listenUdp(port int) error {
+	udpAddr, err := net.ResolveUDPAddr("udp", ":"+strconv.Itoa(port))
+	if err != nil {
+		return err
+	}
+	udpConn, err := net.ListenUDP("udp", udpAddr)
+	if err != nil {
+		return err
+	}
+	_ = udpConn.Close()
```
zolia:
```suggestion
	udpConn.Close()
```

core/port/pool.go
```diff
@@ -0,0 +1,69 @@
...
+		start:    40000,
+		capacity: 10000,
+		rng:      rand.New(rand.NewSource(time.Now().UnixNano())),
+	}
+}
+
+// Acquire returns an unused port in pool's range
+func (pool *Pool) Acquire(protocol string) (Port, error) {
+	p := pool.randomPort()
+	if !available(protocol, p) {
```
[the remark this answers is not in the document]
tadaskay: This wouldn't work: return types are incompatible, and `seekAvailablePort` error case isn't handled. Here is a working equivalent with less nesting:

```go
	p := pool.randomPort()
	if available(protocol, p) {
		log.Debugf(logPrefix+"supplying %s port %d", protocol, p)
		return Port(p), nil
	}
	p, err := pool.seekAvailablePort(protocol)
	if err == nil {
		log.Debugf(logPrefix+"supplying %s port %d", protocol, p)
		return Port(p), nil
	}
	return 0, err
```

Surely, it's less nesting, but harder to comprehend and repetitive, don't you think?

core/port/pool.go
tadaskay: I would like to think of `logPrefix` like a log organization related part, thus wouldn't it make sense to keep it separate from actual message arguments? Ideally, I would create a local wrapper, so that actual logging calls don't need to care about it, but I got only a couple of log calls in the package thus I don't think that's necessary.

core/port/pool_test.go
tadaskay: Isn't it a good practice in golang to ignore errs explicitly? Goland's default ruleset gives me a warning about this one.

services/openvpn/service/manager.go
tadaskay: Moving firewall teardown part to the `Serve` as a deferred function looks good to me. Now, regarding the first part, I don't think that:
```go
m.mapPort(m.port.Num())()
```
conveys meaning better than 
```
releasePorts()
```
does. In fact, it looks like an obfuscaton to me. Perhaps there's a better way?

core/port/pool.go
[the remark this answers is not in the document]
vkuznecovas:
```golang
        p = pool.randomPort()
	if available(protocol, p) {
                log.Debugf(logPrefix+"supplying %s port %d", protocol, p)
		return Port(p), nil
	}
	p, err = pool.seekAvailablePort(protocol)
        log.Debugf(logPrefix+"supplying %s port %d", protocol, p)
        return Port(p), errors.Wrap(err, "supplying port failed")
```

core/port/availability.go
```diff
@@ -0,0 +1,52 @@
...
+			return false
+		}
+		_ = conn.Close()
+	default:
+		listener, err := net.Listen(protocol, addr)
+		if err != nil {
+			log.Infof(logPrefix+"port %v for protocol %v is busy", port, protocol)
+			return false
+		}
+		_ = listener.Close()
```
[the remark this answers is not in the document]
tadaskay: Discussed offline: decided to go with `defer conn.Close()`

core/port/availability.go
```diff
@@ -0,0 +1,52 @@
...
+	case "udp", "udp4", "udp6":
+		udpAddr, err := net.ResolveUDPAddr(protocol, addr)
+		if err != nil {
+			panic(errors.Wrap(err, "unable to resolve UDP address - this is a bug"))
+		}
+		conn, err := net.ListenUDP(protocol, udpAddr)
+		if err != nil {
+			return false
+		}
+		_ = conn.Close()
```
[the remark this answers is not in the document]
tadaskay: Discussed offline: decided to go with `defer conn.Close()`

core/port/availability.go
```diff
@@ -0,0 +1,52 @@
...
+
+	switch protocol {
+	case "udp", "udp4", "udp6":
+		udpAddr, err := net.ResolveUDPAddr(protocol, addr)
+		if err != nil {
+			panic(errors.Wrap(err, "unable to resolve UDP address - this is a bug"))
+		}
+		conn, err := net.ListenUDP(protocol, udpAddr)
+		if err != nil {
+			return false
```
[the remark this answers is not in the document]
tadaskay: That's the purpose of this method: if there's an error listening on the port, we consider it unavailable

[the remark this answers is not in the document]
tadaskay: Will add a logging statement to log the `err`

core/port/availability.go
```diff
@@ -0,0 +1,52 @@
...
+)
+
+func available(protocol string, port int) bool {
+	addr := ":" + strconv.Itoa(port)
+
+	switch protocol {
+	case "udp", "udp4", "udp6":
+		udpAddr, err := net.ResolveUDPAddr(protocol, addr)
+		if err != nil {
+			panic(errors.Wrap(err, "unable to resolve UDP address - this is a bug"))
```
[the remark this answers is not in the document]
tadaskay: Sure

core/port/availability.go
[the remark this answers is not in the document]
tadaskay: Discussed offline: decided to go with defer conn.Close()

core/port/availability.go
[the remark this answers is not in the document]
tadaskay: Discussed offline: decided to go with defer conn.Close()

core/port/pool_test.go
tadaskay: Discussed offline: decided to go with defer conn.Close()

core/port/pool_test.go
tadaskay: Discussed offline: decided to go with defer conn.Close()

services/openvpn/service/manager.go
tadaskay: Got rid of them. Thanks!

core/port/pool.go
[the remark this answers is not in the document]
tadaskay: Reworked as discussed offline. Thanks

services/openvpn/service/manager.go
tadaskay: Removed functionality from `Stop()`, replaced with defers.

tadaskay:
@zolia Looks like travis PR check is stuck here, too. I don't have the rights to access settings.

core/port/pool_test.go
```diff
@@ -0,0 +1,65 @@
...
+
+	// when
+	udpPort, _ := pool.Acquire("udp")
+	// then
+	err = listenUdp(udpPort.Num())
+	assert.NoError(t, err)
+}
+
+func listenTcp(port int) error {
+	listener, err := net.Listen("tcp", ":"+strconv.Itoa(port))
```
vkuznecovas:
```golang
listener, err := net.Listen("tcp", ":"+strconv.Itoa(port))
defer listener.Close()
return err
```

core/port/pool_test.go
```diff
@@ -0,0 +1,65 @@
...
+	defer listener.Close()
+	return nil
+}
+
+func listenUdp(port int) error {
+	udpAddr, err := net.ResolveUDPAddr("udp", ":"+strconv.Itoa(port))
+	if err != nil {
+		return err
+	}
+	udpConn, err := net.ListenUDP("udp", udpAddr)
```
vkuznecovas:
```golang
udpConn, err := net.ListenUDP("udp", udpAddr)
defer udpConn.Close()
return err
```

core/port/pool_test.go
tadaskay: Looks slick, but listener may be `nil`

core/port/pool_test.go
tadaskay: Same as above

PR #874, opened 2019-04-11
Remove redundant protocol requirement for acquiring a port

tadaskay:
Port can be tested using any protocol (in this case, I'm using TCP)

### DIFF
--- a/core/port/availability.go
+++ b/core/port/availability.go
@@ -25,29 +25,19 @@ import (
 	"github.com/pkg/errors"
 )
 
-func available(protocol string, port int) (bool, error) {
-	addr := ":" + strconv.Itoa(port)
-
-	switch protocol {
-	case "udp", "udp4", "udp6":
-		udpAddr, err := net.ResolveUDPAddr(protocol, addr)
-		if err != nil {
-			return false, errors.Wrap(err, "unable to resolve UDP address")
-		}
-		conn, err := net.ListenUDP(protocol, udpAddr)
-		if err != nil {
-			log.Infof("%s cannot listen on UDP port %v: %v", logPrefix, port, err)
-			return false, nil
-		}
-		defer conn.Close()
-	default:
-		listener, err := net.Listen(protocol, addr)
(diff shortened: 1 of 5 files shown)


core/port/availability.go
```diff
@@ -22,32 +22,14 @@ import (
...
-		listener, err := net.Listen(protocol, addr)
-		if err != nil {
-			log.Infof("%s cannot listen on TCP port %v: %v", logPrefix, port, err)
-			return false, nil
-		}
-		defer listener.Close()
+func available(port int) bool {
+	listener, err := net.Listen("tcp", ":"+strconv.Itoa(port))
+	if err != nil {
+		log.Infof("%s cannot listen on TCP port %v: %v", logPrefix, port, err)
```
zolia: I would write here TCP, since for us it does not matter. We just need to know if port is free.
```suggestion
		log.Infof("%s port %v: is taken %v", logPrefix, port, err)
```

core/port/pool_test.go
```diff
@@ -30,13 +30,13 @@ func TestAcquiredPortsAreUsable(t *testing.T) {
...
 	// when
-	tcpPort, _ := pool.Acquire("tcp")
+	tcpPort, _ := pool.Acquire()
 	// then
 	err := listenTcp(tcpPort.Num())
 	assert.NoError(t, err)
 
 	// when
-	udpPort, _ := pool.Acquire("udp")
+	udpPort, _ := pool.Acquire()
```
zolia: Not sure we need both cases with 'tcp' and with 'udp'. One is enough to test pool.

core/port/availability.go
tadaskay: Missed this, thanks.

core/port/pool_test.go
tadaskay: Missed this, thanks.

tadaskay:
@zolia @vkuznecovas made the changes as discussed, take a look

PR #879, opened 2019-04-15
NAT status endpoint

ignasbernotas:


cmd/di.go
```diff
@@ -337,6 +337,7 @@ func (di *Dependencies) bootstrapNodeComponents(nodeOptions node.Options) {
 	tequilapi_endpoints.AddRoutesForServiceSessions(router, di.ServiceSessionStorage)
 	tequilapi_endpoints.AddRoutesForPayout(router, di.IdentityManager, di.SignerFactory, di.MysteriumAPI)
 	tequilapi_endpoints.AddRoutesForAccessPolicies(router)
+	tequilapi_endpoints.AddRoutesForNat(router, di.NATTracker)
```
vkuznecovas:
```suggestion
	tequilapi_endpoints.AddRoutesForNAT(router, di.NATTracker)
```

donce: Aren't abbreviations supposed to be upper case?
```suggestion
	tequilapi_endpoints.AddRoutesForNAT(router, di.NATTracker)
```

tequilapi/endpoints/nat.go
```diff
@@ -0,0 +1,86 @@
...
+
+import (
+	"github.com/julienschmidt/httprouter"
+	"github.com/mysteriumnetwork/node/nat/traversal"
+	"github.com/mysteriumnetwork/node/tequilapi/utils"
+	"net/http"
+)
+
+// swagger:model NatStatusDTO
+type NatStatusDTO struct {
```
vkuznecovas:
```suggestion
type NATStatusDTO struct {
```

tequilapi/endpoints/nat.go
```diff
@@ -0,0 +1,86 @@
...
+	Status string `json:"status"`
+	Error  string `json:"error"`
+}
+
+type NatEvents interface {
```
vkuznecovas:
```suggestion
type NATEventReporter interface {
```

tequilapi/endpoints/nat.go
```diff
@@ -0,0 +1,86 @@
...
+	LastEvent() traversal.Event
+}
+
+// ServiceEndpoint struct represents management of service resource and it's sub-resources
+type NatEndpoint struct {
```
vkuznecovas:
```suggestion
type NATEndpoint struct {
```

tequilapi/endpoints/nat.go
```diff
@@ -0,0 +1,86 @@
...
+	natEvents NatEvents
+}
+
+// NewNatEndpoint creates and returns nat endpoint
+func NewNatEndpoint(natEvents NatEvents) *NatEndpoint {
```
vkuznecovas:
```suggestion
func NewNATEndpoint(natEvents NatEvents) *NatEndpoint {
```

tequilapi/endpoints/nat.go
```diff
@@ -0,0 +1,86 @@
...
+// summary: Shows NAT status
+// description: Nat status returns the last known NAT event
+// responses:
+//   200:
+//     description: NAT status and/or error
+//     schema:
+//       "$ref": "#/definitions/NatStatusDTO"
+//   204:
+//     description: No status available
+func (ne *NatEndpoint) NatStatus(resp http.ResponseWriter, _ *http.Request, _ httprouter.Params) {
```
vkuznecovas:
```suggestion
func (ne *NatEndpoint) NATStatus(resp http.ResponseWriter, _ *http.Request, _ httprouter.Params) {
```

tequilapi/endpoints/nat.go
```diff
@@ -0,0 +1,86 @@
...
+		return
+	}
+
+	statusResponse := toNatStatusResponse(event)
+
+	utils.WriteAsJSON(statusResponse, resp)
+}
+
+// AddRoutesForService adds service routes to given router
+func AddRoutesForNat(router *httprouter.Router, natEvents NatEvents) {
```
vkuznecovas:
```suggestion
func AddRoutesForNAT(router *httprouter.Router, natEvents NatEvents) {
```

tequilapi/endpoints/nat.go
```diff
@@ -0,0 +1,86 @@
...
+	natEndpoint := NewNatEndpoint(natEvents)
+
+	router.GET("/nat/status", natEndpoint.NatStatus)
+}
+
+func toNatStatusResponse(event traversal.Event) NatStatusDTO {
```
vkuznecovas:
```suggestion
func toNATStatusResponse(event traversal.Event) NatStatusDTO {
```

tequilapi/endpoints/nat.go
donce: `NatEvents` is not what you think it is (probably because of bad naming) - AFAIK, it's actually port mapping events (upnp) which are used to trigger NAT. I.e. NatEvents gets triggered when UPNP fails, and then NAT is actually started.

You probably need to create new events when Nat traversal using hole-punching fails, and catch only those events.

tequilapi/endpoints/nat_test.go
```diff
@@ -0,0 +1,119 @@
...
+	assert.EqualValues(t, expected, parsedResponse)
+}
+
+func makeStatusRequestAndReturnResponse(t *testing.T, mockedTracker natTrackerMock) (*httptest.ResponseRecorder, error) {
+	req, err := http.NewRequest(
+		http.MethodGet,
+		"/irrelevant",
+		nil,
+	)
+	if err != nil {
```
vkuznecovas: since you're passing the t here anyway, might as well just `assert.nil(t, err)` here without returning it to the caller.

ignasbernotas: Yeah I thought about that, but it would be like a hidden assertion :D I feared someone would comment on that:D

vkuznecovas: I don't think it's hidden. If you're sending the test context, it only makes sense to use it.

PR #880, opened 2019-04-15
Generate service ports dynamically for wireguard

tadaskay:
Ref: #689

Uses shared port pool if `wireguard.listen.ports` is not specified. Can use specific range, e.g. to use old defaults use `--wireguard.listen.ports 52820:53075`

Tested with OSX provider, multiple android consumers


services/wireguard/resources/allocator_windows.go
```diff
@@ -61,7 +60,7 @@ func (a *Allocator) AllocateIPNet() (net.IPNet, error) {
 	a.mu.Lock()
 	defer a.mu.Unlock()
 
-	for i := 1; i < MaxResources; i++ {
+	for i := 2; i < a.maxConnections+2; i++ {
```
vkuznecovas: could you perhaps add a comment explaining this loop? why is i 2, why a.maxConnections+2? perhaps adding new variables could also help.

services/wireguard/service/service_windows.go
```diff
@@ -38,9 +39,16 @@ func NewManager(
...
 	natService nat.NATService,
 	portMap func(port int) (releasePortMapping func()),
-	options Options) *Manager {
+	options Options,
+	portPool portSupplier,
+) *Manager {
 
-	resourceAllocator := resources.NewAllocator(options.PortMin, options.Subnet)
+	portSupplier := portPool
+	if options.PortMin != 0 {
```
vkuznecovas: we should probably just accept a portSupplier interface as a param. We then don't have to perform this check here and can just consume the port supplier.

services/wireguard/service/service_unix.go
```diff
@@ -39,9 +40,16 @@ func NewManager(
...
 	natService nat.NATService,
 	portMap func(port int) (releasePortMapping func()),
-	options Options) *Manager {
+	options Options,
+	portPool resources.PortSupplier,
+) *Manager {
 
-	resourceAllocator := resources.NewAllocator(options.PortMin, options.PortMax, options.Subnet)
+	portSupplier := portPool
+	if options.PortMin != 0 {
```
vkuznecovas: we should probably just accept a portSupplier interface as a param. We then don't have to perform this check here and can just consume the port supplier.

services/wireguard/resources/allocator_unix.go
```diff
@@ -126,7 +124,12 @@ func (a *Allocator) AllocatePort() (int, error) {
 	a.mu.Lock()
 	defer a.mu.Unlock()
 
-	for i := a.portMin; i < a.portMax; i++ {
+	if err := a.initializePortMin(); err != nil {
```
soffokl: Are we checking only min port availability, but continue increasing it without any check?

services/wireguard/resources/allocator_unix.go
```diff
@@ -126,7 +124,12 @@ func (a *Allocator) AllocatePort() (int, error) {
...
+		return 0, err
+	}
+
+	portMax := a.portMin + a.maxConnections
```
soffokl: Can it go out of the available pool?
let say we have 
```
		start:    40000,
		capacity: 10000,
```
We acquired 49999 port, and adding 255 to it will be out of the range.

services/wireguard/service/options_defaults.go
```diff
@@ -0,0 +1,32 @@
+//+build !windows
+
+/*
+ * Copyright (C) 2018 The "MysteriumNetwork/node" Authors.
```
soffokl:
```suggestion
 * Copyright (C) 2019 The "MysteriumNetwork/node" Authors.
```

services/wireguard/service/options_defaults.go
```diff
@@ -0,0 +1,32 @@
...
+ */
+
+package service
+
+import "net"
+
+// DefaultOptions is a wireguard service configuration that will be used if no options provided.
+var DefaultOptions = Options{
+	ConnectDelay:   2000,
+	PortMin:        0,
```
soffokl: This sounds like we are using port 0 as a default value.

services/wireguard/service/service_windows.go
tadaskay: The idea was to encapsulate `PortPool` as a `portSupplier` (inline with openvpn implementation). So the service itself doesn't need to care about whether it's using a random or a fixed port (supplier can be either `port.Pool` or `port.Fixed`). WDYT?

services/wireguard/resources/allocator_unix.go
tadaskay: Yeah, that was the plan, since this is only for testing purposes to reduce the number of clashes when running provider nodes from the office. But I'll change to all randomized ports.

services/wireguard/resources/allocator_unix.go
tadaskay: Same as https://github.com/mysteriumnetwork/node/pull/880#discussion_r275735324

services/wireguard/service/service_windows.go
vkuznecovas: That's basicly my thinking, except I'd define an interface here and move the check you're performing even further up. That way the wireguard service itself does not need to know anything about the port package.

tadaskay: On the second thought, the highest in the hierarchy where the supplier can be configured is `Dependencies.bootstrapServiceWireguard`, because openvpn and wireguard use separate flags, thus it must be configured individually for a specific service. And this kind of functionality does belong to the `wireguard` package (wireguard should know how to bootstrap its port supplier). Ideally, we should keep the artificial DI layer thin, and keep things closer to where they are used. Hence I don't think there's anything to gain by moving this `portSupplier` up in the hierarchy, and it may even be wrong, considering the package architecture.

vkuznecovas: But you're doing both now. You're sending in the port supplier which is then overriden by the concrete implementation in the port package. You either need to get rid of the injectable dependency, or inject a proper one instead.

cmd/service_bootstrap_desktop.go
```diff
@@ -1,7 +1,7 @@
 // +build darwin windows linux,!android
 
 /*
- * Copyright (C) 2018 The "MysteriumNetwork/node" Authors.
+ * Copyright (C) 2019 The "MysteriumNetwork/node" Authors.
```
soffokl: I'm not sure, but we probably not need to change this year in the existing files.

core/port/range.go
```diff
@@ -0,0 +1,76 @@
...
+func UnspecifiedRange() *Range {
+	return &Range{0, 0}
+}
+
+// ParseRange parses port range expression, e.g. "1000:1200" into Range struct
+func ParseRange(rangeExpr string) (*Range, error) {
+	if rangeExpr == "" {
+		return UnspecifiedRange(), nil
+	}
+	bounds := strings.FieldsFunc(rangeExpr, func(c rune) bool {
```
soffokl: Maybe [strings.Split()](https://golang.org/pkg/strings/#Split) will be a better choice?

core/port/range.go
```diff
@@ -0,0 +1,76 @@
...
+		return c == ':'
+	})
+	if len(bounds) != 2 {
+		return nil, errors.New("Invalid port range expression: " + rangeExpr)
```
soffokl: [Error strings should not be capitalized](https://github.com/golang/go/wiki/CodeReviewComments#error-strings):
```suggestion
		return nil, errors.New("invalid port range expression: " + rangeExpr)
```

core/port/range.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+	start, err := strconv.Atoi(bounds[0])
+	if err != nil {
+		return nil, errors.New("Invalid start port number: " + bounds[0])
```
soffokl:
```suggestion
		return nil, errors.New("invalid start port number: " + bounds[0])
```

core/port/range.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+	end, err := strconv.Atoi(bounds[1])
+	if err != nil {
+		return nil, errors.New("Invalid end port number: " + bounds[0])
```
soffokl:
```suggestion
		return nil, errors.New("invalid end port number: " + bounds[0])
```

core/port/range.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+	if start > end {
+		return nil, errors.New("Start port cannot be greater than end port: " + rangeExpr)
```
soffokl:
```suggestion
		return nil, errors.New("start port cannot be greater than end port: " + rangeExpr)
```

services/wireguard/resources/allocator_unix.go
```diff
@@ -39,21 +40,19 @@ type Allocator struct {
 	IPAddresses map[int]struct{}
 	Ports       map[int]struct{}
```
soffokl: Do we need to keep the `Ports` map with a new `PortSupplier`?

core/port/range.go
tadaskay: If using `split` you have to deal with empty tokens

tadaskay: Since I'm converting tokens with `Atoi` and checking for err anyway, I think it's fine to let the empty tokens slip here. Changing to `Split()`

services/wireguard/resources/resources.go
```diff
@@ -0,0 +1,25 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package resources
+
+import "github.com/mysteriumnetwork/node/core/port"
+
+// PortSupplier supplies ports to wireguard allocator
+type PortSupplier interface {
```
vkuznecovas: Move the definition of the interface to where it's actually used. No need to keep it in a separate file.

tadaskay: Yeah, this is the closest it's possible :)

tadaskay: Packaging should be revisited, but with current structure, `service` depends on `resources`, not the other way around.

tadaskay: ok

PR #884, opened 2019-04-16
Use real trust oracle in access policies endpoint

donce:
Closes https://github.com/mysteriumnetwork/node/issues/865

### DIFF
--- a/tequilapi/endpoints/access_policies.go
+++ b/tequilapi/endpoints/access_policies.go
@@ -19,8 +19,10 @@ package endpoints
 
 import (
 	"net/http"
+	"time"
 
 	"github.com/julienschmidt/httprouter"
+	"github.com/mysteriumnetwork/node/requests"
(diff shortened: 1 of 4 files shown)


tequilapi/endpoints/access_policies.go
```diff
@@ -42,41 +44,45 @@ type accessRule struct {
...
-	r := accessPolicyCollection{Entries: []accessPolicy{staticAccessPolicy}}
+	req, err := requests.NewGetRequest(ape.accessPolicyEndpointURL, "", nil)
+	if err != nil {
+		utils.SendError(resp, err, http.StatusInternalServerError)
+		return
+	}
+	r := accessPolicyCollection{}
+	err = ape.http.DoRequestAndParseResponse(req, &r)
+	if err != nil {
+		utils.SendError(resp, err, http.StatusInternalServerError)
```
soffokl: We are returning `http.StatusInternalServerError` but there is no description about it in the swagger responses.

Issue #863, opened 2019-04-08
Node integration with LocationOracle


Waldz:
- [x] Mock Tequilapi endpoint `/location` and `/connection/location`
- [x] Refactor CLI arg to have typed location providers in "--location.type"
- [x] Autodetect IP attributes from LocationOracle
- [x] Return IP attribytes in Tequilapi endpoints `/location` and `/connection/location`
- [x] Manual IP attributes configuration (can be done later)
- [x] Refactor CLI arg to have typed IP adapters in "--ip.type" (can be done later)
- [x] Stop using Ipify and migrate to LocationOracle (can be done later)

Node has several location detection providers:
```
--location.type         Location autodetect provider (builtin, mmdb, oracle) default: 'builtin'
--location.address   Address of specific location provider given in '--location.type' e.g. https://testnet-location.mysterium.network/location for --location.type=oracle provider

# Possibility from local binary database of GeoLite2 format e.g. http://dev.maxmind.com/geoip/geoip2/geolite2/
# would suggest localdb, we could eventually modify maxmind db with our own data.
--location.type=localdb

# Possibility to identity IP from Oracle
--location.type=oracle
--location.address="https://testnet-location.mysterium.network/location"

# Possible to have manual provider also:
--location.type=manual
--location.address=""
--location.country="LT"
--location.city="Vilnius"
--location.user_type="datacenter"
```

Node has several IP detection providers:
```
--ip.type         IP autodetect provider (ipify, oracle) default: 'ipify'
--ip.address   Address of specific IP provider given in --ip.type e.g. https://testnet-location.mysterium.network/ip for --ip.type=oracle provider e.g. https://api.ipify.org/ will be used for --ip.type=ipify

# Possibility to identity IP from Oracle
--ip.type=ipify
--ip.address="https://api.ipify.org/"


# Possibility to identity IP from Oracle
--ip.type=oracle
--ip.address="https://testnet-location.mysterium.network/ip"
```

zolia:
@soffokl can you check checkbox'es if they are done with your PR?

soffokl:
@zolia done, checked all of them since it covered.

Waldz:
Moving here from LocationOracle ticket https://github.com/mysteriumnetwork/location-oracle/issues/2

Left overs:
- [ ] Node autodetects IP attributes same ways as "country" and adds to Proposal
- [ ] Implement "node_type" proposals filter in Discovery service (datacenter/residential/..)
[...]

PR #894, opened 2019-04-18
Location resolution through oracle

vkuznecovas:
Updates #863

### DIFF
--- a/cmd/di.go
+++ b/cmd/di.go
@@ -469,22 +469,9 @@ func (di *Dependencies) bootstrapIdentityComponents(options node.Options) {
 func (di *Dependencies) bootstrapLocationComponents(options node.OptionsLocation, configDirectory string) (err error) {
 	di.IPResolver = ip.NewResolver(options.IPDetectorURL)
 
-	switch options.Type {
-	case "builtin":
-		di.LocationResolver, err = location.NewBuiltInResolver(di.IPResolver)
-	case "mmdb":
-		di.LocationResolver, err = location.NewExternalDBResolver(filepath.Join(configDirectory, options.Address), di.IPResolver)
-	case "oracle":
-		di.LocationResolver = location.NewOracleResolver(options.Address)
-	case "manual":
-		di.LocationResolver = location.NewStaticResolver(options.Country, options.City, options.NodeType, di.IPResolver)
-	default:
-		return fmt.Errorf("unknown location detector type: %s", options.Type)
-	}
-
+	di.LocationResolver, err = FallbackResolverFactory(di.IPResolver, options, configDirFlag)
 	if err != nil {
-		log.Error("Failed to load location resolver: ", err)
-		di.LocationResolver = location.NewFailingResolver(err)
+		return err
 	}
 
 	di.LocationOriginal = location.NewLocationCache(di.LocationResolver)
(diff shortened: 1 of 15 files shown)


core/location/interface.go
```diff
@@ -17,14 +17,9 @@
...
 package location
 
-import (
-	"net"
-)
-
 // Resolver allows resolving location by ip
 type Resolver interface {
 	DetectLocation() (Location, error)
-	ResolveLocation(ip net.IP) (Location, error)
```
soffokl: It was kept for cases when a consumer wants to check the provider's IP, whether it's really `residential` or not.
We, probably, don't another place for that now.

cmd/di.go
```diff
@@ -509,3 +496,37 @@ func (di *Dependencies) bootstrapNATComponents(options node.Options) {
 		di.NATPinger = &traversal.NoopPinger{}
 	}
 }
+
+// FallbackResolverFactory creates a fallback resolver given the params
+func FallbackResolverFactory(ipResolver ip.Resolver, options node.OptionsLocation, configDirectory string) (location.Resolver, error) {
```
tadaskay: if `options.Type == "manual"`, it will not create a fallback resolver, but rather a static one. Thus we should change the name to avoid confusion. _Factory_ is a noise word; we should name the function after what it does. WDYT about something similar to`CreateLocationResolver`?

tadaskay: Perhaps we can start working on packaging things by features. I think this func a fairly good candidate to be placed in `location`. Perhaps something like `location/di.go | container.go | config.go` for the configuration that merely assembles things for the app context.

Waldz: Remade type `manual` to be a fallback resolver

tequilapi/endpoints/location.go
```diff
@@ -104,8 +103,7 @@ func (le *LocationEndpoint) GetLocation(writer http.ResponseWriter, request *htt
 //     schema:
 //       "$ref": "#/definitions/ErrorMessageDTO"
 func (le *LocationEndpoint) GetLocationByIP(writer http.ResponseWriter, request *http.Request, params httprouter.Params) {
```
soffokl: Looks like we don't need `GetLocationByIP` if we are removing requests for other's IP.
It's enough to have only `GetLocation`

vkuznecovas: I blame rebase

cmd/di.go
```diff
@@ -468,23 +466,9 @@ func (di *Dependencies) bootstrapIdentityComponents(options node.Options) {
...
-		di.LocationResolver, err = location.NewExternalDBResolver(filepath.Join(configDirectory, options.Address), di.IPResolver)
-	case "oracle":
-		di.LocationResolver = location.NewOracleResolver(options.Address)
-	case "manual":
-		di.LocationResolver = location.NewStaticResolver(options.Country, options.City, options.NodeType, di.IPResolver)
-	default:
-		return fmt.Errorf("unknown location detector type: %s", options.Type)
-	}
-
+	di.LocationResolver, err = location.CreateLocationResolver(di.IPResolver, options.Country, options.City, options.Type, options.NodeType, options.Address, options.ExternalDb, configDirFlag)
```
soffokl: We are passing every single option separately, maybe it's time to pass just `OptionsLocation`?

vkuznecovas: i'll ahve to ref the node package, which in turn add cyclic dependencies. It's either move the model somewhere, or define a new one here and use that instead. Neither seems to be a proper way of doing things :)

PR #898, opened 2019-04-19
sane message is printed if peer does not support NAT punching

zolia:


nat/traversal/config/parser.go
```diff
@@ -43,5 +43,9 @@ func (c *ConsumerConfigParser) Parse(config json.RawMessage) (ip string, port in
 	if err != nil {
 		return "", 0, "", errors.Wrap(err, "parsing consumer address:port failed")
 	}
-	return cfg.IP, cfg.Port, openvpn.ServiceType, nil
+
+	if cfg.IP == nil {
```
soffokl: Shouldn't we keep other parameters if only IP is empty? Or maybe return an error if it's a mandatory argument?

nat/traversal/pinger.go
```diff
@@ -177,7 +180,7 @@ func (p *Pinger) PingProvider(ip string, port int) error {
 	}
 
 	// wait for provider to setup NAT proxy connection
-	time.Sleep(200 * time.Millisecond)
+	time.Sleep(400 * time.Millisecond)
```
soffokl: We are tweaking 3rd time with no explanation. Maybe there could be a better check?

nat/traversal/config/parser.go
soffokl: And another thing that confuses me, why we are using a pointer to string here? Isn't it enough to compare with empty string?
I'd use `net.IP` for every place where it's possible as a best case.

donce: > Or maybe return an error if it's a mandatory argument?

Was thinking the same - it doesn't make much sense to return `nil` error when method failed to execute it's duty.

nat/traversal/config/parser.go
```diff
@@ -43,5 +43,9 @@ func (c *ConsumerConfigParser) Parse(config json.RawMessage) (ip string, port in
...
+		return "", 0, "", nil
```
donce: Why do you return empty data but with no error? I have no idea what is going on here - do you have some specific case in mind? Maybe you can add a test to show us what case are you covering here?

donce: If we return error here instead of empty data, we probably don't need to have `if port == 0 || IP == ""` in the upper layer.
```suggestion
		return "", 0, "", errors.New( "Remote party does not support NAT hole punching, IP:PORT is missing")
```

nat/traversal/config/parser.go
zolia: Parsing fails if there is no IP inside message. Just wanted to make it optional and don't get a parsing error.

nat/traversal/pinger.go
zolia: Theres little bit of tweaking here since we are not sure if proxy has been setup for current session. If consumer starts to send protocol (ssh) messages without proxy set up, connect will fail. 
We might be able to implement another round trip of messaging between consumer - provider confirming that provider is ready to receive protocol connect, but at this point it looks to be excessive still. Later it should be implemented probably. Still I am hoping to find some sane value here, its not nice, but should suffice atm.

nat/traversal/config/parser.go
zolia: good suggestion.

PR #900, opened 2019-04-19
Extend NAT events (ELK and message bus) to support different stages

donce:


nat/mapping/port_mapping.go
```diff
@@ -86,11 +89,11 @@ func addMapping(m portmap.Interface, protocol string, extPort, intPort int, name
...
 		if err := m.AddMapping(protocol, extPort, intPort, name, 0); err != nil {
 			// some gateways support only permanent leases
-			publisher.Publish(traversal.EventTopic, traversal.BuildFailureEvent(err))
+			publisher.Publish(traversal.EventTopic, traversal.BuildFailureEvent(StageName, err))
 			log.Debugf("%s Couldn't add port mapping for port %d: %v", logPrefix, extPort, err)
 			return
 		}
 	}
-	publisher.Publish(traversal.EventTopic, traversal.BuildSuccessEvent())
+	publisher.Publish(traversal.EventTopic, traversal.BuildSuccessEvent(StageName))
```
zolia: I believe we need not only port mapping events, but hole punching events as well, since we are going to implement some '/connection/id/status' endpoint which should report its nat traversal status to application. This status should source data from such events. 
Trick here is that we are starting to deal with session IDs, maybe session manager should update its status with nat-traversal status reported by events.

metrics/sender.go
```diff
@@ -50,8 +49,10 @@ type appInfo struct {
 	Version string `json:"version"`
 }
 
-type natMappingFailContext struct {
-	errorMessage string
+type natMappingContext struct {
+	stage        string
+	successful   bool
```
soffokl: Do we need to have bool `successful` here, or we can just check `if errorMessage != nil`?

metrics/sender.go
```diff
@@ -50,8 +49,10 @@ type appInfo struct {
...
+	errorMessage *string
```
soffokl: Why not keep it as `error` type?

nat/traversal/pinger.go
```diff
@@ -182,6 +184,15 @@ func (p *Pinger) PingProvider(ip string, port int) error {
 	return nil
 }
 
+func (p *Pinger) waitForPreviousStageResult() bool {
+	for {
+		event := p.natEventWaiter.WaitForEvent()
```
vkuznecovas: so what happens when we block here forever? how do we exit this loop?

metrics/sender.go
vkuznecovas: my guess is it makes the querying easier :)

donce: Well we can live without it now, but it is easier to query and understand events. Different NAT failures can have different metadata, i.e. if NAT hole-punching executes without errors, but  client connecting for VPN is unable to connect to us, we might create a failure event with more data, like `ip` of client, `port` which was used, protocol version and so on. So it might be difficult to see in a list of different fields that `error`/`errorMessage` is the key field to see whether it's event for successful or failed event.

metrics/sender.go
donce: I had it as `error` first, but somebody though it's confusing (because type is a `string` and not an `error`), so I fixed a change request :D and now we're going back :D is it really that important?

nat/traversal/pinger.go
donce: We're waiting here for port-mapping event here, just like before. Once we receive a port mapping event, we return from this function (2 lines below).

The idea is that we start pinger, and it starts working only after port-mapping fails. I don't like this myself - we could start pinger only when nat fails, but I didn't change this yet in this PR.

metrics/sender.go
soffokl: Well, with `error` you will not have to convert `err` to string to assign it :)

nat/traversal/pinger.go
vkuznecovas: So why not solve it?

```golang
for {
       // not sure what type of the event we're getting here, so that needs changing
	ec := make(chan nat.Event)
	go func() { ec <- p.natEventWaiter.WaitForEvent()} ()
	select {
	case <-p.stop:
		return
	case event := <- ec 
		if event.Stage == p.previousStage {
			return event.Type == SuccessEventType
		}
	}
}
```

nat/traversal/pinger.go
```diff
@@ -95,9 +97,15 @@ type Params struct {
 func (p *Pinger) Start() {
 	log.Info(prefix, "Starting a NAT pinger")
 
-	// We dont need to run pinger if NAT port auto-configuration is successful
-	if p.natEventWaiter.WaitForEvent().Type == SuccessEventType {
+	resultChannel := make(chan bool)
```
soffokl: We need to have it buffered to finish goroutine even if the stop was invoked early. 
```suggestion
	resultChannel := make(chan bool, 1)
```

PR #902, opened 2019-04-23
Fail service start endpoint with unknown params

donce:
Not sure about other tequilapi requests, but starting service with misinterpreted params is dangerous - i.e. we can expose user to all traffic without his consent, as we did in https://github.com/mysteriumnetwork/node/pull/899.

If we get some params which we don't know, it's likely that request has a mistake, and we shouldn't start service.


donce:
This feature already helped me - just found out that CLI had the old param name :D Added fix in new commit.

Issue #901, opened 2019-04-23
Report router information


tadaskay:
For now, it will be logged in stdout.
Later we can consider reporting it in metrics.

PR #904, opened 2019-04-23
Report UPnP network gateways to stdout

tadaskay:
Fixes: #901

### DIFF
--- a/cmd/di.go
+++ b/cmd/di.go
@@ -53,6 +53,7 @@ import (
 	"github.com/mysteriumnetwork/node/nat/mapping"
 	"github.com/mysteriumnetwork/node/nat/traversal"
 	"github.com/mysteriumnetwork/node/nat/traversal/config"
+	"github.com/mysteriumnetwork/node/nat/upnp"
 	"github.com/mysteriumnetwork/node/services"
 	service_noop "github.com/mysteriumnetwork/node/services/noop"
 	"github.com/mysteriumnetwork/node/services/openvpn"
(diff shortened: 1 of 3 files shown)


nat/upnp/discover.go
```diff
@@ -0,0 +1,108 @@
...
+	defer client.Close()
+	responses, err := ssdp.SSDPRawSearch(client, ssdp.UPNPRootDevice, 2, 3)
+	if err != nil {
+		return nil, err
+	}
+
+	var results []gatewayDevice
+
+	for _, response := range responses {
+		device := gatewayDevice{}
```
vkuznecovas: I'd extract the innards to something like this:
```golang
func SSDPResponseToDevice(response whateverTypeItActuallyIs) (device, error) {
                device := gatewayDevice{
                     server: response.Header.Get("server"),
                }

		loc, err := response.Location()
		if err != nil {
			return err
		}

		device.root, err = goupnp.DeviceByURL(loc)
		return device, err
}
```

then, the loop just becomes: 
```golang
for _, response := range responses {
       device, err := SSDPResponseToDevice(response)
       if err != nil {
           log ....
           continue
       }
        if !isGateway(device.root) {
			log.Debugf("%s not a gateway device: %v", logPrefix, device)
			continue
	}
	results = append(results, device)
}
```

cmd/di.go
```diff
@@ -182,6 +183,8 @@ func (di *Dependencies) Bootstrap(nodeOptions node.Options) error {
 
 	di.PortPool = port.NewPool()
 
+	go upnp.ReportNetworkGateways()
```
vkuznecovas: Do we want to collect data from everyone  even though they are only consumers? Or is this only really useful when dealing with providers? Just asking if we should really call it here or not.

tadaskay: Thanks for pointing that out. I think it's fine to collect from consumers as well. Less logic branching, and anonymous data, so I don't see an issue here.

nat/upnp/discover.go
tadaskay: Good suggestion, thanks

PR #905, opened 2019-04-23
Make oracle default provider

Waldz:


cmd/di.go
```diff
@@ -465,11 +466,39 @@ func (di *Dependencies) bootstrapIdentityComponents(options node.Options) {
 	di.IdentityRegistration = identity_registry.NewRegistrationDataProvider(di.Keystore)
 }
 
-func (di *Dependencies) bootstrapLocationComponents(options node.OptionsLocation, configDirectory string) (err error) {
+func (di *Dependencies) bootstrapLocationComponents(options node.OptionsLocation, configDirectory string) error {
```
tadaskay: @vkuznecovas extracted location di to the relevant package, I don't think it's a good idea to bring this back

soffokl: I agree, at least some explanation for reasons required.

cmd/di.go
```diff
@@ -465,11 +466,39 @@ func (di *Dependencies) bootstrapIdentityComponents(options node.Options) {
...
+		di.LocationResolver = location.NewFallbackResolver([]location.Resolver{manualResolver})
+
+	case node.LocationTypeBuiltin:
+		var builtinResolver location.Resolver
+		builtinResolver, err := location.NewBuiltInResolver(di.IPResolver)
+		if err != nil {
+			log.Error("Failed to load builtin location resolver: ", err)
+			builtinResolver = location.NewFailingResolver(err)
+		}
+		di.LocationResolver = location.NewFallbackResolver([]location.Resolver{builtinResolver, manualResolver})
```
soffokl: Is it a good idea to add `manualResolver` everywhere like the last one? 
Most likely it will contain empty values if user set different type of resolver.
Maybe it's better to return an error then empty values?

cmd/di.go
Waldz: It's not good idea to make it in `location` package, because it's independent package, and making it modular leads us to circular dependencies like `cmd -> node -> location -> node`

I don't see a problem with `di` being high-level entrypoint to glue dependencies. Anyway moved to just 1 factory function  to `location/factory`.

cmd/di.go
Waldz: As I user I should be able to start a node if location auto-detection is not working.
By entering location attributes manually

core/location/factory/factory.go
```diff
@@ -0,0 +1,99 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/*
```
vkuznecovas: wassup with the 3 copyrights?

cmd/di.go
vkuznecovas: But this will result in proposals being announced with no country, as to use the manual resolver - you'll have to provide the flags in order for it to have any values. Is it not better to fail then, and inform the user that he can override the location resolution via flags rather than defaulting to empty values? I'd rather fallback to the builtin DB...

cmd/di.go
```diff
@@ -34,6 +34,7 @@ import (
 	"github.com/mysteriumnetwork/node/core/connection"
 	"github.com/mysteriumnetwork/node/core/ip"
 	"github.com/mysteriumnetwork/node/core/location"
+	location_factory "github.com/mysteriumnetwork/node/core/location/factory"
```
tadaskay: I don't know about this aliasing and 'factory'. Previous version was rather straightforward: `location.CreateLocationResolver`. Perhaps `location.CreateResolver` would be even better? What do we actually gain here from moving DI to a separate sub-package?

cmd/di.go
tadaskay: The issue is that other packages are also not yet well designed, but we got to start somewhere. I wouldn't try to find reasons why current `di` is ok, because it's awful order dependent spaghetti. We should split it feature-wise, and perhaps separate object graph construction from bootstrap, so that construction is not order-dependent, but bootstrap _is_.
I think putting it in a subpackage _for now_ is ok. However, _factory_, aliasing adds just another layer of obfuscation.

cmd/di.go
Waldz: Reverted to have only single type resolving currently, no falbacking until:
- missing city/ASN resolving in `buildin` resolver
- no error handing for `manual` resolver

cmd/di.go
Waldz: As mentioned before, to avoid loops between package dependencies: `di` -> `location` -> `di`

cmd/di.go
Waldz: Modularisation is not scope of this PR

cmd/di.go
```diff
@@ -483,13 +485,25 @@ func (di *Dependencies) bootstrapIdentityComponents(options node.Options) {
...
 	di.IPResolver = ip.NewResolver(options.IPDetectorURL)
-	di.LocationResolver, err = location.CreateLocationResolver(di.IPResolver, options.Country, options.City, options.Type, options.NodeType, options.Address, options.ExternalDb, configDirFlag)
+
+	switch options.Type {
+	case node.LocationTypeManual:
+		di.LocationResolver = location.NewStaticResolver(options.Country, options.City, options.NodeType, di.IPResolver)
+	case node.LocationTypeBuiltin:
+		di.LocationResolver, err = location.NewBuiltInResolver(di.IPResolver)
+	case node.LocationTypeMMDB:
+		di.LocationResolver, err = location.NewExternalDBResolver(filepath.Join(configDirectory, options.Address), di.IPResolver)
```
vkuznecovas: do we really want to use the same param for both oracle address and external db? Looking at `options.Address`

cmd/di.go
```diff
@@ -20,6 +20,9 @@ package cmd
 import (
 	"time"
```
vkuznecovas: redundant whitespace

cmd/di.go
```diff
@@ -483,13 +485,25 @@ func (di *Dependencies) bootstrapIdentityComponents(options node.Options) {
...
+	default:
+		err = fmt.Errorf("unknown location detector type: %s", options.Type)
+	}
 	if err != nil {
 		return err
 	}
 
 	di.LocationOriginal = location.NewLocationCache(di.LocationResolver)
-	return nil
+	return
```
vkuznecovas: keep the returns consistent, either skip the err above or add it here.

cmd/di.go
Waldz: Yes, it's more simple and possible to reuse on all providers, it's by DNS convention https://en.wikipedia.org/wiki/Data_source_name

Maybe we should rename it DSN

PR #906, opened 2019-04-23
Fix NAT events sending to ELK

donce:
I have to admit, I haven't actually tested that proper events are sent in https://github.com/mysteriumnetwork/node/pull/900. And they aren't - no new fields were sent, because they all were private :D

This fixed events sending and improves internal NAT events structure to be similar to ELK events structure - it simplifies stuff.

### DIFF
--- a/nat/traversal/events_sender.go
+++ b/nat/traversal/events_sender.go
@@ -18,8 +18,6 @@
 package traversal
 
 import (
-	"fmt"
-
 	log "github.com/cihub/seelog"
 )
 
@@ -30,7 +28,7 @@ type EventsSender struct {
 	metricsSender metricsSender
 	ipResolver    ipResolver
 	lastIp        string
-	lastEventType string
+	lastEvent     *Event
 }
 
 type metricsSender interface {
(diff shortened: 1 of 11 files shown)


nat/traversal/events_sender.go
```diff
@@ -62,16 +60,25 @@ func (es *EventsSender) ConsumeNATEvent(event Event) {
...
-	case FailureEventType:
-		return es.metricsSender.SendNATMappingFailEvent(event.Stage, event.Error)
-	default:
-		return fmt.Errorf("unknown event type: %v", event.Type)
 	}
+
+	return es.metricsSender.SendNATMappingFailEvent(event.Stage, event.Error)
+}
+
+func (es *EventsSender) isEventRelevant(event Event, ip string) bool {
```
tadaskay: I would name it in a less abstract way, e.g. `matchesLastEvent`

nat/traversal/events_tracker.go
```diff
@@ -62,23 +53,26 @@ func NewEventsTracker() *EventsTracker {
...
 	select {
 	case et.eventChan <- event:
 	case <-time.After(300 * time.Millisecond):
 	}
 }
 
-// LastEvent returns the last known event
-func (et *EventsTracker) LastEvent() Event {
+// LastEvent returns the last known event and boolean flag, indicating if such event exists
+func (et *EventsTracker) LastEvent() (Event, bool) {
```
soffokl: We can just return a pointer to `Event`, in this case `nil` could be a valid value and we don't need a `bool` here.

donce: I did like that at first, but `bool` seemed to clarify more that value can be missing. But I can change this back :)

donce: Fixed

Issue #890, opened 2019-04-17
 Provider sends metrics about NAT traversal local status


Waldz:
As a Mysterium
We want to know of how many Providers fails to configure NAT traversal (node is not accessible)

Tasks:
- [x] Send metrics to ELK stack
- [x] Kibana dashboard

donce:
![Screenshot 2019-04-26 at 12.04.22.png](https://images.zenhubusercontent.com/5ba247ee8c7b5425fc895855/92f730f1-3936-4949-84d7-fab1d992cfce)

PR #908, opened 2019-04-24
Send events to ELK about hole-punching

donce:
https://github.com/mysteriumnetwork/node/issues/890

### DIFF
--- a/nat/traversal/pinger.go
+++ b/nat/traversal/pinger.go
@@ -243,11 +243,7 @@ func (p *Pinger) sendPingRequest(conn *net.UDPConn, ttl int) error {
 	}
 
 	_, err = conn.Write([]byte("continuously pinging to " + conn.RemoteAddr().String()))
-	if err != nil {
-		return err
-	}
-
-	return nil
+	return err
 }
 
 func (p *Pinger) getConnection(ip string, port int, pingerPort int) (*net.UDPConn, error) {
--- a/nat/traversal/pinger.go
+++ b/nat/traversal/pinger.go
@@ -243,7 +243,7 @@ func (p *Pinger) sendPingRequest(conn *net.UDPConn, ttl int) error {
 	}
 
 	_, err = conn.Write([]byte("continuously pinging to " + conn.RemoteAddr().String()))
-	return err
+	return errors.Wrap(err, "pinging request failed")
 }
 
 func (p *Pinger) getConnection(ip string, port int, pingerPort int) (*net.UDPConn, error) {
--- a/nat/traversal/pinger.go
+++ b/nat/traversal/pinger.go
@@ -48,7 +48,7 @@ type Pinger struct {
 	portPool       PortSupplier
 	consumerPort   int
 	previousStage  string
-	publisher      Publisher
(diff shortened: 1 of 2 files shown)


nat/traversal/pinger.go
```diff
@@ -215,21 +223,33 @@ func (p *Pinger) ping(conn *net.UDPConn) error {
...
 }
 
+func (p *Pinger) sendPingRequest(conn *net.UDPConn, ttl int) error {
+	err := ipv4.NewConn(conn).SetTTL(ttl)
+	if err != nil {
+		return errors.Wrap(err, "pinger setting ttl failed")
+	}
+
+	_, err = conn.Write([]byte("continuously pinging to " + conn.RemoteAddr().String()))
+	if err != nil {
```
soffokl: just `return err` will be enough.

nat/traversal/pinger.go
```diff
@@ -47,6 +48,7 @@ type Pinger struct {
 	portPool       PortSupplier
 	consumerPort   int
 	previousStage  string
+	publisher      Publisher
```
soffokl: Can we name it more clear? Just `publisher` does not provide info on what it does.

soffokl: But I was talking about field name, not the interface type. It fine to have `Publisher` as an interface name.

donce: So you're suggesting something like this?
```suggestion
	eventPublisher Publisher
```

soffokl: Correct.

donce: I actually dislike that variable name tells more than the interface, but whatever, changed it :D

nat/traversal/pinger.go
```diff
@@ -215,21 +223,29 @@ func (p *Pinger) ping(conn *net.UDPConn) error {
...
 }
 
+func (p *Pinger) sendPingRequest(conn *net.UDPConn, ttl int) error {
+	err := ipv4.NewConn(conn).SetTTL(ttl)
+	if err != nil {
+		return errors.Wrap(err, "pinger setting ttl failed")
+	}
+
+	_, err = conn.Write([]byte("continuously pinging to " + conn.RemoteAddr().String()))
+	return err
```
vkuznecovas: could wrap this err as well to add context.. conn.Write will be hard to pinpoint without it

nat/traversal/pinger.go
```diff
@@ -215,21 +223,33 @@ func (p *Pinger) ping(conn *net.UDPConn) error {
...
 }
 
+func (p *Pinger) sendPingRequest(conn *net.UDPConn, ttl int) error {
+	err := ipv4.NewConn(conn).SetTTL(ttl)
+	if err != nil {
+		return errors.Wrap(err, "pinger setting ttl failed")
+	}
+
+	_, err = conn.Write([]byte("continuously pinging to " + conn.RemoteAddr().String()))
+	if err != nil {
```
soffokl: It can still be just `return errors.Wrap(err, "pinging request failed")`
if `err` is `nil` `errors.Wrap` will return a nil.

PR #909, opened 2019-04-24
Refactor event bus

donce:
- Decouple from external event bus implementation (github.com/asaskevich/EventBus)
- Simplify interface
- Unify naming

### DIFF
--- a/bus/event_bus.go
+++ b/bus/event_bus.go
@@ -1,0 +1,31 @@
+/*
+ * Copyright (C) 2019 The "MysteriumNetwork/node" Authors.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package bus
+
+import asaskevichEventBus "github.com/asaskevich/EventBus"
+
+// EventBus allows subscribing and publishing data by topic
+type EventBus interface {
+	Subscribe(topic string, fn interface{}) error
+	Publish(topic string, args ...interface{})
+}
+
+// NewEventBus returns implementation of EventBus
+func NewEventBus() EventBus {
+	return asaskevichEventBus.New()
+}
--- a/bus/event_bus.go
+++ b/bus/event_bus.go
@@ -26,19 +26,19 @@ type EventBus interface {
 }
 
 type simplifiedEventBus struct {
-	bus *asaskevichEventBus.Bus
+	bus asaskevichEventBus.Bus
 }
 
 func (bus simplifiedEventBus) Subscribe(topic string, fn interface{}) error {
-	return bus.Subscribe(topic, fn)
+	return bus.bus.Subscribe(topic, fn)
 }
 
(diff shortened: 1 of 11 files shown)


bus/event_bus.go
```diff
@@ -0,0 +1,44 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package bus
+
+import asaskevichEventBus "github.com/asaskevich/EventBus"
```
soffokl: It's better to have some generic name. If we will decide to replace `asaskevich/EventBus` by another package it will be easy to replace only in one place. 
```suggestion
import eventBus "github.com/asaskevich/EventBus"
```

cmd/di.go
```diff
@@ -20,12 +20,12 @@ package cmd
 import (
 	"time"
 
-	"github.com/asaskevich/EventBus"
 	log "github.com/cihub/seelog"
 	"github.com/ethereum/go-ethereum/accounts/keystore"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/ethclient"
 	"github.com/mysteriumnetwork/node/blockchain"
+	"github.com/mysteriumnetwork/node/bus"
```
soffokl: I'm not sure, but just `bus` sounds not clear. Maybe it's better to name package `eventBus`? 
```suggestion
	"github.com/mysteriumnetwork/node/eventBus"
```

core/service/pool_test.go
```diff
@@ -30,12 +30,12 @@ type mockService struct {
 
 type mockPublisher struct {
 	publishedTopic string
-	publishedArgs  []interface{}
+	publishedArg   interface{}
```
soffokl: `Arg` was replaced to data everywhere, probably we should to it here too.
```suggestion
	publishedData   interface{}
```

bus/event_bus.go
```diff
@@ -0,0 +1,44 @@
...
+
+// EventBus allows subscribing and publishing data by topic
+type EventBus interface {
```
vkuznecovas: still not a fan of central declarations of interface

bus/event_bus.go
```diff
@@ -0,0 +1,44 @@
...
+type simplifiedEventBus struct {
+	bus *asaskevichEventBus.Bus
+}
+
+func (bus simplifiedEventBus) Subscribe(topic string, fn interface{}) error {
+	return bus.Subscribe(topic, fn)
+}
+
+func (bus simplifiedEventBus) Publish(topic string, data interface{}) {
+	bus.Publish(topic, data)
```
vkuznecovas: this is recursive

bus/event_bus.go
```diff
@@ -0,0 +1,44 @@
...
+	Subscribe(topic string, fn interface{}) error
+	Publish(topic string, data interface{})
+}
+
+type simplifiedEventBus struct {
+	bus *asaskevichEventBus.Bus
+}
+
+func (bus simplifiedEventBus) Subscribe(topic string, fn interface{}) error {
+	return bus.Subscribe(topic, fn)
```
vkuznecovas: this is recursive

bus/event_bus.go
donce: That's why other components, who needs it, don't import this interface, but declare interface of their own - it's just the reference of what `EventBus` we use when we create it, which is much simpler than the full implementation of asaskevich event bus.

bus/event_bus.go
donce: The idea is that we are using `asaskevich` implementation either way, and it has its own interface. `simpliefiedEventBus` is basically an adapter between `EventBus` (our interface) and `asaskevich/EventBus.BusEventBus`. If we want to use another library, than that other library will have different interface, and this adapter wont work - we will have to modify it. That's why indicating clearly that this is just an adapter makes sense to me.

PR #910, opened 2019-04-25
Turn on NAT punching by default

tadaskay:


e2e/compatibility-matrix/docker-compose.yml
```diff
@@ -35,6 +35,7 @@ services:
       --identity=0xd1a23227bd5ad77f36ba62badcb78a410a1db6c5
       --identity.passphrase=localprovider
       --openvpn.port=3000
+      --openvpn.proto=tcp
```
soffokl: Why we need to explicitly switch to TCP, is it break UDP somehow?

tadaskay: Yeah, yesterday with @Waldz we tested that it didn't work on UDP. But this was just a try to quickfix. I'm closing this PR for now

Issue #794, opened 2019-03-13
Inform user if NAT auto-configuration fails


mjankauskas:
When NAT auto-configuration fails as a user I want to be informed and given link to documentation how to setup it manually.

donce:
Everything is done from the node side, only app side is left :)

PR #911, opened 2019-04-25
Add tequilapi endpoint for NAT traversal status

donce:
https://github.com/mysteriumnetwork/node/issues/794

- Initially, this endpoint returns `not_finished` status.
- If NAT traversal is successful in any stage, `successful` is returned.
- If last stage of NAT traversal fails, then `failure` is returned. If we get failure in stage which is not the last one, then status is still `not_finished`, because other staged will continue traversing the NAT.
- Last stage is either `port_mapping` or `hole_punching`, depending on experimental feature flag. So we can start using this instantly even without enabling flag.


nat/status_tracker.go
```diff
@@ -0,0 +1,67 @@
...
+	statusNotFinished = "not_finished"
+	statusSuccessful  = "successful"
+	statusFailure     = "failure"
+)
+
+// Status represents NAT traversal status (either "not_finished", "successful" or "failure") and an optional error.
+type Status struct {
+	Status string
+	Error  error
+}
```
ignasbernotas: It would be awesome to know which stage failed:) But maybe that will be available through the error message?

cmd/di.go
```diff
@@ -51,6 +51,7 @@ import (
 	"github.com/mysteriumnetwork/node/money"
 	"github.com/mysteriumnetwork/node/nat"
 	"github.com/mysteriumnetwork/node/nat/mapping"
+	"github.com/mysteriumnetwork/node/nat/natevents"
```
soffokl: This package is already in the `nat` directory, we probably should not repeat the name in the subpackage: 
```suggestion
	"github.com/mysteriumnetwork/node/nat/event"
```

And go style guide recommends not to use the plural for the package names: 
> In go, package names are not plural.

tequilapi/endpoints/nat.go
```diff
@@ -0,0 +1,90 @@
...
+	statusNotFinished = "not_finished"
+	statusSuccessful  = "successful"
+	statusFailure     = "failure"
+)
+
+// NATStatusDTO gives information about NAT traversal success or failure
+// swagger:model NATStatusDTO
+type NATStatusDTO struct {
+	Status string  `json:"status"`
+	Error  *string `json:"error,omitempty"`
```
soffokl: `omitempty` should work for empty string too, no need it to be a pointer. 
```suggestion
	Error  string `json:"error,omitempty"`
```

PR #920, opened 2019-04-29
Critical NAT hole punchning fixes

Waldz:
What was changed:
- do not ping provider on upnp success
- certs always a client
- dynamic pool under nat

### DIFF
--- a/cmd/service_bootstrap_desktop.go
+++ b/cmd/service_bootstrap_desktop.go
@@ -134,7 +134,7 @@ func (di *Dependencies) bootstrapServiceOpenvpn(nodeOptions node.Options) {
 		proposal := openvpn_discovery.NewServiceProposalWithLocation(currentLocation, transportOptions.Protocol)
 
 		var portSupplier port.ServicePortSupplier = di.PortPool
-		if transportOptions.Port != 0 {
+		if transportOptions.Port != 0 && locationInfo.PubIP == locationInfo.OutIP {
 			portSupplier = port.NewFixed(transportOptions.Port)
 		}
 
--- a/cmd/service_bootstrap_desktop.go
+++ b/cmd/service_bootstrap_desktop.go
@@ -134,7 +134,7 @@ func (di *Dependencies) bootstrapServiceOpenvpn(nodeOptions node.Options) {
 		proposal := openvpn_discovery.NewServiceProposalWithLocation(currentLocation, transportOptions.Protocol)
 
 		var portSupplier port.ServicePortSupplier = di.PortPool
-		if transportOptions.Port != 0 && locationInfo.PubIP == locationInfo.OutIP {
+		if transportOptions.Port != 0 {
 			portSupplier = port.NewFixed(transportOptions.Port)
 		}
 
--- a/services/openvpn/client.go
+++ b/services/openvpn/client.go
@@ -60,13 +60,12 @@ func (c *Client) Start(options connection.ConnectOptions) error {
 	log.Infof("client config: %v", clientConfig)
 
 	c.natPinger.BindConsumerPort(clientConfig.LocalPort)
-	if clientConfig.vpnConfig.LocalPort > 0 {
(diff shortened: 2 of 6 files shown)


cmd/service_bootstrap_desktop.go
```diff
@@ -134,7 +134,7 @@ func (di *Dependencies) bootstrapServiceOpenvpn(nodeOptions node.Options) {
 		proposal := openvpn_discovery.NewServiceProposalWithLocation(currentLocation, transportOptions.Protocol)
 
 		var portSupplier port.ServicePortSupplier = di.PortPool
-		if transportOptions.Port != 0 {
+		if transportOptions.Port != 0 && locationInfo.PubIP == locationInfo.OutIP {
```
Waldz: @zolia is not sure why we removed this condition. Because it port suplly

soffokl: There was a problem with a static port configuration that @tadaskay discovered.
It was removed to fix the problem. Is it ok to return it back?  
https://github.com/mysteriumnetwork/node/pull/896

tadaskay: @Waldz it was a bug that @zolia himself fixed in https://github.com/mysteriumnetwork/node/pull/896/commits/82f0278834a0cbe0b130f01c7e5e63db81093da0 
With this condition, `--openvpn.port` is not respected for users under NAT (because, obviously, their IP does not match external IP).

Waldz: Would be good to honor Provider's which from CLI flag.
ELSE
Defaults to random port

services/openvpn/client.go
```diff
@@ -60,7 +60,9 @@ func (c *Client) Start(options connection.ConnectOptions) error {
 	log.Infof("client config: %v", clientConfig)
 
 	c.natPinger.BindConsumerPort(clientConfig.LocalPort)
-	err = c.natPinger.PingProvider(clientConfig.vpnConfig.RemoteIP, clientConfig.vpnConfig.RemotePort)
+	if clientConfig.vpnConfig.LocalPort > 0 {
```
donce: So provider sends `LocalPort == 0` when NAT traversal is disabled? Can we make this clear by moving this technical details to config itself with explanatory method name, i.e.
```suggestion
	if clientConfig.vpnConfig.isPingRequested() {
```

cmd/service_bootstrap_desktop.go
vkuznecovas: So this:
```golang
if transportOptions.Port != 0 && locationInfo.PubIP == locationInfo.OutIP {
    portSupplier = port.NewFixed(transportOptions.Port)
}
```

becomes:

```golang
if transportOptions.Port != 0  {
    portSupplier = port.NewFixed(transportOptions.Port)
} else if locationInfo.PubIP == locationInfo.OutIP {
    portSupplier = // default or random, you decide.
}
```

where's the issue?

services/openvpn/service/manager.go
```diff
@@ -160,14 +160,22 @@ func (m *Manager) ProvideConfig(sessionConfig json.RawMessage, pingerPort func(i
 		}
 		if m.isBehindNAT() && m.portMappingFailed() {
 			c.Port = pp.Num()
-			pingerPort(pp.Num())
+			if m.serviceOptions.Port > 0 {
```
vkuznecovas: This if looks fishy. If we provide a specific port, we default to 12234. Why is that?

zolia: It is fishy :) Added TODO notice. Welcome to take :) It might be good first issue.

zolia: Logic basically says that we need to have a separate port for pinger if static port is configured. We cannot run both pinger and service on the same port at the same time.

tadovas:
What really concerns me with these types of "features"/"fixes" etc. is that we don't have any tests or proof that it works. Even if we test it some how by hand - there is no guarantee that any change in the future will break it. And this network related stuff is super critical. We really need to start research how we can automatically test these kinds of things. Maybe we can craft some "virtual" custom networks with various types of nats and verify that our solutions cover most of the cases

PR #922, opened 2019-04-30
Use the unified way to print error logs

soffokl:
Starting a myst service with an incorrect service type does not print logs, it just prints to the `stderr`. 
Running a myst service with systemd swallows these errors and it starting to be very hard to figure out the source of the problem

image

This PR changes formatted output to the unified logs, that we are using everywhere it allows to keep logs in the same form and able to read it.

### DIFF
--- a/utils/stopper.go
+++ b/utils/stopper.go
@@ -18,8 +18,9 @@
 package utils
 
 import (
-	"fmt"
 	"os"
+
+	log "github.com/cihub/seelog"
 )
 
 // Killer kills some resource and performs cleanup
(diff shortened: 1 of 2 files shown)


utils/stopper.go
```diff
@@ -49,12 +50,13 @@ func newStopper(kill Killer, exit exitter) func() {
...
-		fmt.Fprintln(os.Stderr, msg)
+		log.Errorf("Error while killing process: %v\n", err.Error())
+		log.Flush()
 		exit(1)
 		return
 	}
 
-	fmt.Println("Good bye")
+	log.Info("Good bye")
+	log.Flush()
```
Waldz: I guess logs are flushed in `di.Stop()`.  At least they should be in same place there logs are bootstrapped

Waldz: I guess logs are flushed in di.Stop(). At least they should be in same place there logs are bootstrapped

soffokl: Removed it from here.

PR #931, opened 2019-05-02
Fix location cache

### LINKED ISSUE #653: Rework location cache
At the moment the location cache component is not much of a cache. It returns nothing if I call Get without calling the refresh before. 

We need a component to resolve public and outbound IPs as well as location information and, possibly, cache them. I'd imagine something like this:

```golang
type cache interface {
  GetCountry() string
  GetPublicIP() string // or maybe net.IP
  GetOutboundIP() string // or maybe net.IP
}
``` 

#651 introduced lookups for these for all the services independently, this component would allow for  reuse.

vkuznecovas:
This does two:
* Closes #653
* Fixes an unreported bug where location is being actually resolved for each and every ping on nat traversal attempts.

Reworked the old location cache, created a new one that caches location info for 5 minutes by default, and also is invalidated when connection state changes(consumer side). Switched all resolvers to use the cache instead.

### DIFF
--- a/core/location/cache.go
+++ b/core/location/cache.go
@@ -25,30 +25,6 @@ import (
 	"github.com/mysteriumnetwork/node/core/connection"
 )
 
-type cache struct {
-	locationDetector Resolver
-	location         Location
-}
-
-// NewLocationCache constructs Cache
-func NewLocationCache(locationDetector Resolver) Cache {
-	return &cache{
-		locationDetector: locationDetector,
-	}
-}
-
-// Gets location from cache
-func (lc *cache) Get() Location {
-	return lc.location
-}
-
-// Stores location to cache
-func (lc *cache) RefreshAndGet() (Location, error) {
-	location, err := lc.locationDetector.DetectLocation()
-	lc.location = location
-	return lc.location, err
-}
-
 const locationCacheLogPrefix = "[location-cache]"
 
 type ProperCache struct {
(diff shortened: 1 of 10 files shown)


core/location/cache.go
```diff
@@ -17,26 +17,76 @@
...
+
+	return loc, err
 }
 
-// Gets location from cache
-func (lc *cache) Get() Location {
-	return lc.location
+// DetectLocation returns location from cache, or fetches it if needed
+func (c *Cache) DetectLocation() (Location, error) {
+	c.lock.Lock()
```
soffokl: Maybe lock should go to the `fetchAndSave()` and `needsRefresh`?
I see that it's not called outside locks, but one day someone will change it and break it.

core/location/cache.go
```diff
@@ -17,26 +17,76 @@
...
-type cache struct {
+import (
+	"sync"
+	"time"
+
+	log "github.com/cihub/seelog"
+	"github.com/mysteriumnetwork/node/core/connection"
+)
+
+const locationCacheLogPrefix = "[location-cache]"
```
soffokl:
```suggestion
const locationCacheLogPrefix = "[location-cache] "
```

core/location/cache.go
vkuznecovas: I think that the private methods should not be locked. The lock is purposefully put on the public methods, as I want to ensure that no double update occurs. Imagine the lock moved onto the private methods. Two routines call DetectLocation. First one checks if we need to refresh. Finds out we do. The lock is released, we move further. The second routine then kicks in, also performs the same check, finds out we still need to refresh. Both of them proceed to update the state, where only a single update is needed.

PR #995, opened 2019-05-03
Consumer side nat proxy

zolia:
This PR implements NAT proxy on consumer side. This is needed then Provider is behind NAT and auto-port configuration fails. Most NAT routers expects that port will not change during the session, if it detects port change on incoming packet - router blocks it. Thats why we need to  make port on consumer side stable.

Openvpn3 library used by mobile application does not support local port binding. 
Each request from openvpn client comes with some random port which does not work with nat hole punching mechanics. 

What has been done: 
 - complete port randomisation for openvpn
      - randomised pinger port on provider
      - randomised consumer port generated on provider
 - consumer side NAT proxy
 - ProvideConfig remade to support traversal.Params
 - graceful natproxy cleanup after  session disconnect


tadovas:
Could you please provide a short description of exact NAT hole punching case you are trying to resolve?

mobile/mysterium/openvpn_connection_setup.go
```diff
@@ -182,16 +194,18 @@ type OpenvpnConnectionFactory struct {
...
 
 // Create creates a new openvpn connection
-func (ocf *OpenvpnConnectionFactory) Create(stateChannel connection.StateChannel, statisticsChannel connection.StatisticsChannel) (connection.Connection, error) {
-	sessionFactory := func(options connection.ConnectOptions) (*openvpn3.Session, error) {
+func (ocf *OpenvpnConnectionFactory) Create(stateChannel connection.StateChannel, statisticsChannel connection.StatisticsChannel) (con connection.Connection, err error) {
+	sessionFactory := func(options connection.ConnectOptions) (*openvpn3.Session, *openvpn.ClientConfig, error) {
 		vpnClientConfig, err := openvpn.NewClientConfigFromSession(options.SessionConfig, "", "")
 		if err != nil {
-			return nil, err
+			return nil, nil, err
```
Waldz: You dont need explicit return variables as You have named return variables

nat/traversal/nat_proxy.go
```diff
@@ -52,11 +131,29 @@ func (np *NATProxy) handOff(serviceType services.ServiceType, incomingConn *net.
...
+	defer srcConn.Close()
+	totalBytes, err := io.Copy(dstConn, srcConn)
+	if err != nil {
+		//log.Infof("%sdst socket: %v", logPrefix, dstConn.RemoteAddr().String(), dstConn.LocalAddr().String())
+		//log.Infof("%sscr socket: %v", logPrefix, srcConn.RemoteAddr().String(), srcConn.LocalAddr().String())
+		log.Error(logPrefix, "failed to writing / reading a stream to/from NATProxy: ", err)
+	}
+	log.Tracef("%stotal bytes transferred: %d", logPrefix, totalBytes)
+	//log.Tracef("%stotal bytes transferred from %s to %s: %d", logPrefix,
+	//	srcConn.RemoteAddr().String(),
```
Waldz: Maybe remove debugging from code

nat/traversal/pinger.go
```diff
@@ -176,7 +178,7 @@ func (p *Pinger) PingProvider(ip string, port int) error {
 	if err != nil {
 		return errors.Wrap(err, "failed to get connection")
 	}
-	defer conn.Close()
+	//defer conn.Close()
```
Waldz: ?
Maybe resource cleaning should happen as defer of routine `consumerHandOff()`

zolia: yes, its happens inside `consumerHandOff(conn)` since closure may happen only on error or user's intent.

mobile/mysterium/openvpn_connection_setup.go
zolia: its inside lambda

nat/traversal/nat_proxy.go
```diff
@@ -21,22 +21,150 @@ import (
...
+// Read from listener socket and write to remoteConn
+// Read from remoteConn and write to listener socket
+func (np *NATProxy) consumerHandOff(consumerPort int, remoteConn *net.UDPConn) {
+	log.Info(logPrefix, "Inside consumer NAT proxy")
+
+	// reinit a stop channel, since it might be closed by previous connect
+	np.stop = make(chan struct{})
+
+	// outIP, err := np.ipResolver.GetOutboundIP()
+	laddr, err := net.ResolveUDPAddr("udp4", fmt.Sprintf("%s:%d", "127.0.0.1", consumerPort+1))
```
soffokl:
```suggestion
	laddr, err := net.ResolveUDPAddr("udp4", fmt.Sprintf("127.0.0.1:%d", consumerPort+1))
```

nat/traversal/nat_proxy.go
```diff
@@ -21,22 +21,150 @@ import (
...
+	}
+
+	remoteConn.SetReadBuffer(bufferLen)
+	remoteConn.SetWriteBuffer(bufferLen)
+
+	fd, err := remoteConn.File()
+	if err != nil {
+		log.Error(logPrefix, "failed to fetch fd from: ", remoteConn)
+	}
+	defer fd.Close()
```
soffokl: It will panic in case of error. We should return in the `if` block.

nat/traversal/nat_proxy.go
```diff
@@ -21,22 +21,150 @@ import (
...
+		log.Error(logPrefix, "failed to get local address for consumer NAT proxy: ", err)
+		return
+	}
+
+	remoteConn.SetReadBuffer(bufferLen)
+	remoteConn.SetWriteBuffer(bufferLen)
+
+	fd, err := remoteConn.File()
+	if err != nil {
+		log.Error(logPrefix, "failed to fetch fd from: ", remoteConn)
```
soffokl:
```suggestion
		log.Error(logPrefix, "failed to fetch fd from: ", remoteConn)
		return
```

nat/traversal/nat_proxy.go
```diff
@@ -21,22 +21,150 @@ import (
...
+			return
+		}
+
+		select {
+		case <-np.stop:
+			log.Info(logPrefix, "Stopping NATProxy handOff loop")
+			proxyConn.Close()
+			remoteConn.Close()
+			return
+		default:
```
soffokl: We can make nesting level a bit smaller with empty `default` block since the first case ends with return:
```suggestion
		default:
	}
```

nat/traversal/nat_proxy.go
```diff
@@ -21,22 +21,150 @@ import (
...
+}
+
+func (np *NATProxy) masterLoop(conn *net.UDPConn, remoteConn *net.UDPConn) {
+	log.Info(logPrefix, "start copying stream from consumer NAT proxy to remote remoteConn")
+	for {
+		select {
+		case <-np.stop:
+			log.Info(logPrefix, "Stopping NATProxy masterLoop")
+			return
+		default:
```
soffokl: We can make nesting level a bit smaller with empty `default` block since the first case ends with return:
```suggestion
		default:
	}
```

nat/traversal/nat_proxy.go
```diff
@@ -21,22 +21,150 @@ import (
...
+
+			np.masterLoop(proxyConn, remoteConn)
+
+			proxyConn.Close()
+			//			proxyFD.Close()
+		}
+	}
+}
+
+func (np *NATProxy) masterLoop(conn *net.UDPConn, remoteConn *net.UDPConn) {
```
soffokl: `masterLoop` is bad naming, it's not clear what it does.

nat/traversal/nat_proxy.go
```diff
@@ -21,22 +21,150 @@ import (
...
+	}
+}
+
+func (np *NATProxy) readWriteToAddr(conn *net.UDPConn, remoteConn *net.UDPConn, addr *net.UDPAddr) {
+	for {
+		select {
+		case <-np.stop:
+			log.Info(logPrefix, "Stopping NATProxy readWriteToAddr loop")
+			return
+		default:
```
soffokl: We can make nesting level a bit smaller with empty `default` block since the first case ends with return:
```suggestion
		default:
	}
```

nat/traversal/pinger.go
```diff
@@ -100,12 +100,16 @@ type natProxy interface {
...
+	consumerHandOff(consumerPort int, conn *net.UDPConn)
+	setProtectSocketCallback(socketProtect func(socket int) bool)
+	close()
 }
 
 // Params contains session parameters needed to NAT ping remote peer
 type Params struct {
-	RequestConfig json.RawMessage
+	RequestConfig *json.RawMessage
 	Port          int
```
soffokl: Maybe rename it to "ProviderPort" for consistency?

nat/traversal/nat_proxy.go
```diff
@@ -21,22 +21,150 @@ import (
...
+		}
+	}
+}
+
+func (np *NATProxy) masterLoop(conn *net.UDPConn, remoteConn *net.UDPConn) {
+	log.Info(logPrefix, "start copying stream from consumer NAT proxy to remote remoteConn")
+	for {
+		select {
+		case <-np.stop:
+			log.Info(logPrefix, "Stopping NATProxy masterLoop")
```
soffokl: We are naming it `NATProxy` and `NAT proxy` in different places. It will be good to keep it consistent.

nat/traversal/nat_proxy.go
```diff
@@ -21,22 +21,150 @@ import (
...
+			return
+		default:
+			var buf [bufferLen]byte
+			n, addr, err := conn.ReadFromUDP(buf[0:])
+			if err != nil {
+				log.Errorf("%sFailed	 to read local process: %s cause: %s", logPrefix, conn.LocalAddr().String(), err)
```
soffokl:
```suggestion
				log.Errorf("%sFailed to read local process: %s cause: %s", logPrefix, conn.LocalAddr().String(), err)
```

core/port/pool.go
```diff
@@ -29,11 +29,13 @@ import (
 type Pool struct {
 	start, capacity int
 	rand            *rand.Rand
+	openvpnPort     int
 }
 
 // ServicePortSupplier provides port needed to run a service on
 type ServicePortSupplier interface {
 	Acquire() (Port, error)
+	OpenvpnPort() (Port, error)
```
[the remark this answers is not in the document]
zolia: I am planning to have even more methods here. Basically for each service which needs its own static port configuration. 
Not sure if there is a point to spam separate pool for each service. This looks concise enough.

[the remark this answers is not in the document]
zolia: Will make generic `PortForService(serviceType string)` method.

services/wireguard/service/service_unix.go
```diff
@@ -65,39 +66,39 @@ type Manager struct {
 }
 
 // ProvideConfig provides the config for consumer
-func (manager *Manager) ProvideConfig(publicKey json.RawMessage, pingerPort func(int) int) (session.ServiceConfiguration, session.DestroyCallback, error) {
+func (manager *Manager) ProvideConfig(publicKey json.RawMessage, params *traversal.Params) (session.ServiceConfiguration, session.DestroyCallback, *traversal.Params, error) {
```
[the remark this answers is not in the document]
zolia: That is how ProvideConfig is designed. Service interface accepts this method, but at the same time is used as a Manager and as a ConfigNegotiator. Sometimes we need to pass params further into ConfigNegotiator and sometimes return them earlier. 

Will be bundling that into struct as @vkuznecovas suggested, so it will not look that ugly :)

core/service/manager.go
```diff
@@ -43,7 +44,7 @@ var (
 type Service interface {
 	Serve(providerID identity.Identity) error
 	Stop() error
-	ProvideConfig(publicKey json.RawMessage, pingerPort func(int) int) (session.ServiceConfiguration, session.DestroyCallback, error)
+	ProvideConfig(publicKey json.RawMessage, traversalParams *traversal.Params) (session.ServiceConfiguration, session.DestroyCallback, *traversal.Params, error)
```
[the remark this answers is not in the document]
zolia: wrapped.

core/port/pool.go
```diff
@@ -68,6 +70,30 @@ func (pool *Pool) Acquire() (port Port, err error) {
 	return Port(p), errors.Wrap(err, "could not acquire port")
 }
 
+// SetOpenvpnPort sets fixed openvpn service port
+func (pool *Pool) SetOpenvpnPort(port int) {
```
vkuznecovas: is this a leftover?

nat/traversal/nat_proxy.go
```diff
@@ -21,16 +21,134 @@ import (
 	"fmt"
 	"io"
 	"net"
+	"sync"
 
 	log "github.com/cihub/seelog"
+
```
vkuznecovas: redudant whitespace

zolia: thats a grouping :)

core/port/pool.go
```diff
@@ -68,6 +70,36 @@ func (pool *Pool) Acquire() (port Port, err error) {
 	return Port(p), errors.Wrap(err, "could not acquire port")
 }
 
+// SetServicePort sets fixed openvpn service port
+func (pool *Pool) SetServicePort(serviceType string, port int) error {
+	switch serviceType {
+	case "openvpn":
+		pool.openvpnPort = port
+	default:
+		errors.New("service type is not defined for SetServicePort")
```
soffokl: We are creating error, but do nothing with it. Should we return it?

nat/traversal/nat_proxy.go
zolia: lol, change, but in other file, the same style fix.

Issue #926, opened 2019-04-30
Service start fails after start/stop/wifi switch


ignasbernotas:
1. Started service
2. Stopped service
3. Switched wifi
4. Start service
Hangs in starting status and then fails with HTTP 500.

```
2019-04-30T08:22:36.501137 [Debug] [NATS.Receiver] Request '0xe427274f56db8fe8dddba4fc3c696d237cb7f1f7.openvpn.dialog-create' topic has been subscribed to
2019-04-30T08:22:36.5018 [Error] Service serve failed: failed to add NAT forwarding rule: not able to determine outbound ethernet interface
2019-04-30T08:22:36.501897 [Warn] [NATS] Disconnected
2019-04-30T08:22:36.502013 [Info] [discovery] stopping discovery loop..
2019-04-30T08:22:38.183732 [Info] [Mysterium.api] Proposal unregistered for node: 0xe427274f56db8fe8dddba4fc3c696d237cb7f1f7
2019-04-30T08:22:38.183748 [Info] [discovery] Proposal unregistered
2019-04-30T08:22:38.18376 [Info] [Mysterium.api] Proposal registered for node: 0xe427274f56db8fe8dddba4fc3c696d237cb7f1f7 service type: openvpn
2019-04-30T08:22:40.385788 [Info] [Detect Outbound IP] IP detected: 192.168.0.146
// service start called
2019-04-30T08:22:40.683015 [Info] [NATS.DialogWaiter] Connecting to: {nats/v1 {0xe427274f56db8fe8dddba4fc3c696d237cb7f1f7.openvpn [nats://testnet-broker.mysterium.network:4222]}}
2019-04-30T08:22:40.747599 [Info] [Mysterium.api] Total proposals: 25 supported: 25
2019-04-30T08:22:45.202956 [Info] [Mysterium.api] Total proposals: 24 supported: 24
2019-04-30T08:22:55.189689 [Info] [Mysterium.api] Total proposals: 24 supported: 24
2019-04-30T08:23:00.485713 [Info] [Mysterium.api] Total proposals: 24 supported: 24
2019-04-30T08:23:09.975643 [Info] [Detect Outbound IP] IP detected: 97.9.6.1
// service start called
2019-04-30T08:23:10.385993 [Info] [NATS.DialogWaiter] Connecting to: {nats/v1 {0xe427274f56db8fe8dddba4fc3c696d237cb7f1f7.openvpn [nats://testnet-broker.mysterium.network:4222]}}
[...]

PR #1019, opened 2019-05-06
Clean up NAT forwarding rule on OpenVPN service stop

soffokl:
For OpenVPN service, we are only adding new IP forwarding rules, without cleaning old on service stop. 

It prevents new service to start if the node network was changed. 

Updates: #926


ignasbernotas:
@zolia I've tested this locally, it fixes related problems:)

PR #1022, opened 2019-05-06
Log build information

ignasbernotas:
This is very useful for 0.0.0-dev builds, we'll instantly know which commit was used to build this version.

### DIFF
--- a/cmd/di.go
+++ b/cmd/di.go
@@ -173,6 +173,7 @@ func (di *Dependencies) Bootstrap(nodeOptions node.Options) error {
 	nats_discovery.Bootstrap()
 
 	log.Infof("Starting Mysterium Node (%s)", metadata.VersionAsString())
+	log.Infof("Build information (%s)", metadata.BuildAsString())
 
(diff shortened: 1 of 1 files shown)


cmd/di.go
```diff
@@ -173,6 +173,7 @@ func (di *Dependencies) Bootstrap(nodeOptions node.Options) error {
 	nats_discovery.Bootstrap()
 
 	log.Infof("Starting Mysterium Node (%s)", metadata.VersionAsString())
+	log.Infof("Build information (%s)", metadata.BuildAsString())
```
Waldz: Maybe remake `VersionAsString()` function, so that that we would have build info in all places

Waldz: Maybe remake `VersionAsString()` function, so that that we would have build info in all places

ignasbernotas: @Waldz This would require a lot of changes like metrics, tequilapi, etc. If that's needed lets do it in another PR. I need this as this easily solves the current issue at hand.

PR #1036, opened 2019-05-09
Fix default mysterium-node config for Raspberry Pi

soffokl:


bin/package/raspberry/stagemyst/00-install-deb/00-run.sh
```diff
@@ -12,14 +12,15 @@ apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 8B48AD6246925553 7638D0
 apt-get update
 apt-get -y install wireguard openvpn
 
-cat <<'EOF' > /etc/default/mysterium-node
+cat <<'confEOF' > /etc/default/mysterium-node
```
Waldz: what does `confEOF` mean?

soffokl: There was already one `EOF`. `confEOF` is just another label  `configuration-end-of-file` to resolve the problem of nested EOFs.

PR #1037, opened 2019-05-09
added context for nat events

### LINKED ISSUE #693: NAT statistics on provider side
In order to improve service quality we need to start collecting and sending provider statistics. Statistics needed: 
  - node version
  - type of NAT traversal used: public ip, manual port forwarding, upnp, pmp or nat pinhole
  - session statistics
  
Statistics should be covered for both services: openvpn and wireguard.

These needs to be extendible.
 
- [x] Send node version on startup https://github.com/mysteriumnetwork/node/issues/696
- [x] Send statistics about NAT success/failure #843 
- [ ] Add additional info to the events

vkuznecovas:
Wanted to add context to nat events. Issue is that the loader takes around 2sec to complete, so wrapped it in a component to only load it once. The sender is getting a getter instead of the plain value due to the fact that it takes a few seconds to load the gateway info - don't want to block in bootstrap for that.

Screen Shot 2019-05-09 at 11 22 48 AM

Closes #693

### DIFF
--- a/cmd/di.go
+++ b/cmd/di.go
@@ -200,8 +200,6 @@ func (di *Dependencies) Bootstrap(nodeOptions node.Options) error {
 
 	di.PortPool = port.NewPool()
 
-	go upnp.ReportNetworkGateways()
-
 	di.bootstrapNATComponents(nodeOptions)
 	di.bootstrapServices(nodeOptions)
 	di.bootstrapNodeComponents(nodeOptions)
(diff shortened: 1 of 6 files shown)


cmd/di.go
```diff
@@ -525,8 +523,13 @@ func (di *Dependencies) bootstrapLocationComponents(options node.OptionsLocation
 }
 
 func (di *Dependencies) bootstrapMetrics(options node.Options) {
+	loader := &upnp.GatewayLoader{}
+
+	// warm up the loader as the load takes up to a couple of secs
```
soffokl: What will be if we will not warm up it? Since we are starting it in a goroutine app can reach the place where it's needed faster than a couple of seconds.

vkuznecovas: It doesnt matter in that case, the once will be locked there and we'll wait if needed then... But in general this is a "good enough" solution

PR #1038, opened 2019-05-09
Trigger desktop app build and raspberry image build only for tags

soffokl:
We should trigger a desktop app build only on tag 0.0.0-dev. Commits to master do not update artifacts in the releases, it just tags new 0.0.0-dev. 

And we don't really need to build Raspberry Pi image twice to push it to releases.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -103,7 +103,7 @@ jobs:
     - script:
       - bin/build_xgo linux/arm
       - bin/builder_run BINARY=build/myst/myst_linux_arm bin/package_debian $BUILD_VERSION armhf
-      - bin/package_raspberry
+      - if [[ -nz $TRAVIS_TAG ]]; then bin/package_raspberry; fi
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -192,6 +188,10 @@ jobs:
       - gpg --list-keys
       - bin/release_android "$TRAVIS_TAG"
       name: "Publish Android SDK to the Sonatype and Maven Central"
+    - script:
+      - bin/travis_scripts/trigger_desktop_app_build
+      name: "Trigger Desktop APP build"
+      if: tag = env(BUILD_DEV_RELEASE)
```
zolia: why we build just for dev release?

soffokl: we are triggering a build for the master branch only:
https://github.com/mysteriumnetwork/node/blob/master/bin/travis_scripts/trigger_desktop_app_build#L5

I think there is no point to trigger a desktop app master build for other releases.

zolia: We probably should have automated trigger not for master only in the future, but thats for @ignasbernotas to decide. Lets leave it like this for now.

PR #1049, opened 2019-05-10
initialize commands on demand for mac/linux ipforward

vkuznecovas:
We were reusing exec.Command instances in the IPforward, resulting in the following errors:

```
2019-05-09T16:32:00.753637 [Warn] Failed to enable IP forwarding: [/usr/sbin/sysctl -w net.inet.ip.forwarding=1] Returned exit error: exec: Stdout already set Cmd output: 
2019-05-09T16:32:00.776118 [Warn] [nat] Failed to enable IP forwarding: exec: Stdout already set
2019-05-09T16:32:00.776259 [Warn] [service bootstrap] Failed to enable NAT forwarding: exec: Stdout already set
```

We were also using "CommandEnable" for lookup of ip forward flag. Fixed that as well.

### DIFF
--- a/nat/factory_darwin.go
+++ b/nat/factory_darwin.go
@@ -17,10 +17,15 @@
 
 package nat
 
+import "os/exec"
+
 // NewService returns fake nat service since there are no iptables on darwin
 func NewService() NATService {
 	return &servicePFCtl{
 		ipForward: serviceIPForward{
+			CommandFactory: func(name string, arg ...string) Command {
+				return exec.Command(name, arg...)
+			},
 			CommandEnable:  []string{"/usr/sbin/sysctl", "-w", "net.inet.ip.forwarding=1"},
 			CommandDisable: []string{"/usr/sbin/sysctl", "-w", "net.inet.ip.forwarding=0"},
 			CommandRead:    []string{"/usr/sbin/sysctl", "-n", "net.inet.ip.forwarding"},
--- a/nat/factory_darwin.go
+++ b/nat/factory_darwin.go
@@ -17,15 +17,13 @@
 
 package nat
 
-import "os/exec"
-
 // NewService returns fake nat service since there are no iptables on darwin
 func NewService() NATService {
 	return &servicePFCtl{
 		ipForward: serviceIPForward{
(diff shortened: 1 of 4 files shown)


nat/factory_linux.go
```diff
@@ -17,15 +17,13 @@
...
-import "os/exec"
-
 // NewService returns linux os specific nat service based on ip tables
 func NewService() NATService {
 	return &serviceIPTables{
 		ipForward: serviceIPForward{
-			CommandEnable:  exec.Command("sudo", "/sbin/sysctl", "-w", "net.ipv4.ip_forward=1"),
-			CommandDisable: exec.Command("sudo", "/sbin/sysctl", "-w", "net.ipv4.ip_forward=0"),
-			CommandRead:    exec.Command("/sbin/sysctl", "-n", "net.ipv4.ip_forward"),
+			CommandEnable:  []string{"sudo", "/sbin/sysctl", "-w", "net.ipv4.ip_forward=1"},
```
soffokl: The only thing that worries me is that we are doing `sudo` for linux, but not for darwin. 😕

nat/factory_darwin.go
```diff
@@ -17,15 +17,13 @@
...
-import "os/exec"
-
 // NewService returns fake nat service since there are no iptables on darwin
 func NewService() NATService {
 	return &servicePFCtl{
 		ipForward: serviceIPForward{
-			CommandEnable:  exec.Command("/usr/sbin/sysctl", "-w", "net.inet.ip.forwarding=1"),
-			CommandDisable: exec.Command("/usr/sbin/sysctl", "-w", "net.inet.ip.forwarding=0"),
-			CommandRead:    exec.Command("/usr/sbin/sysctl", "-n", "net.inet.ip.forwarding"),
+			CommandEnable:  []string{"/usr/sbin/sysctl", "-w", "net.inet.ip.forwarding=1"},
```
Waldz: Having command factory here would allow to test `servicePFCtl`

tadaskay: Yeah, but is this critical? Perhaps it's about time we stopped abusing 'request changes' feature.

Issue #956, opened 2019-04-30
Bounty Tester: Can't Start the Service (Win10)


AndriusMyst:
**Problem:**
Cannot start the service
`2019-04-30T11:46:04.554Z info: TequilapiError: Request failed with status code 404 (path="services/9acdf042-aed4-41d6-a230-90238ceee48d")`

**User**:
Location: GB
Telegram: `@Sovietsquirrel`

**Node Client**
Build version: 0.8.0(2251)-0.7.0
ID: 0x57daba353514f004c86edf2e3892740ec4d98c4d

**Hardware**
Win10, server2016
Router- "Pfsense", Vodafone - 40mbps

**LOGS**:

```
2019-04-30T11:50:27.863Z info: "[MysteriumVpn] Application launch"
2019-04-30T11:50:27.867Z info: "[MysteriumVpn] Checking terms cache"
2019-04-30T11:50:27.899Z info: "[MysteriumVpn] Opening window"
2019-04-30T11:50:27.914Z info: "[MysteriumVpn] Waiting for renderer boot"
2019-04-30T11:50:29.266Z info: "[MysteriumVpn] Renderer booted"
2019-04-30T11:50:29.267Z info: "[MysteriumVpn] Building tray"
2019-04-30T11:50:29.299Z info: "[needs-install] checksum mismatch"
2019-04-30T11:50:29.300Z info: "[ProcessManager] Installing 'mysterium_client' process"
2019-04-30T11:50:29.301Z info: "[install] config needs to be recreated"
2019-04-30T11:50:29.261Z info: "Renderer was mounted, will initialize"
2019-04-30T11:50:29.267Z info: "Renderer initialized"
2019-04-30T11:50:29.507Z info: "user exec" "sc.exe query \"MysteriumClient\"" "SERVICE_NAME: MysteriumClient \r\n        TYPE               : 10  WIN32_OWN_PROCESS  \r\n        STATE              : 4  RUNNING \r\n                                (STOPPABLE, NOT_PAUSABLE, ACCEPTS_SHUTDOWN)\r\n        WIN32_EXIT_CODE    : 0  (0x0)\r\n        SERVICE_EXIT_CODE  : 0  (0x0)\r\n        CHECKPOINT         : 0x0\r\n        WAIT_HINT          : 0x0"
[...]

PR #1052, opened 2019-05-10
Workaround ICS issues (win 10/win 2016 server)

tadaskay:
Fixes: #956


nat/service_ics.go
```diff
@@ -40,9 +40,14 @@ func getSharingScript(ifaceName, props, action string) string {
 	return `regsvr32 /s hnetcfg.dll;
 		$netShare = New-Object -ComObject HNetCfg.HNetShare;
 		$c = $netShare.EnumEveryConnection |? { $netshare.NetConnectionProps.Invoke($_).Name -eq "` + ifaceName + `" };
-		$config = $netShare.INetSharingConfigurationForINetConnection.Invoke($c);` + action + `
 		$guid = $netShare.NetConnectionProps.Invoke($c).Guid;
```
soffokl: Maybe this line should go to the `Catch` block since it used only there?

PR #1054, opened 2019-05-10
Introduce ProposalID for unique proposal discovery with composite identifier

Waldz:


market/mysterium/mysterium_api.go
```diff
@@ -151,6 +151,19 @@ func (mApi *MysteriumAPI) PingProposal(proposal market.ServiceProposal, signer i
 	return err
 }
 
+// GetProposal fetches service proposal from discovery by exact ID
+func (mApi *MysteriumAPI) GetProposal(id market.ProposalID) (*market.ServiceProposal, error) {
+	proposals, err := mApi.FindProposals(id.ProviderID, id.ServiceType, "", "")
```
tadaskay: When you start passing empty values, it's a sign that it should be converted to a struct `FindProposalsOpts/Query` from positional args

Waldz: Introduced

PR #1059, opened 2019-05-13
desktop consumer connects without natproxy now

zolia:


services/openvpn/client_config.go
```diff
@@ -100,9 +100,11 @@ func NewClientConfigFromSession(sessionConfig []byte, configDir string, runtimeD
 	if vpnConfig.LocalPort > 0 {
 		clientFileConfig.OriginalRemoteIP = vpnConfig.RemoteIP
 		clientFileConfig.OriginalRemotePort = vpnConfig.RemotePort
-		vpnConfig.RemoteIP = "127.0.0.1"
-		// TODO: randomize this too?
-		vpnConfig.RemotePort = vpnConfig.LocalPort + 1
+		if isMobile {
```
tadaskay: Could we document the purpose of this change in the git message body?

nat/traversal/nat_proxy.go
```diff
@@ -41,6 +41,11 @@ type NATProxy struct {
 
 func (np *NATProxy) consumerHandOff(consumerPort int, remoteConn *net.UDPConn) chan struct{} {
 	stop := make(chan struct{})
+	if np.socketProtect == nil {
+		// shutdown pinger session since openvpn client will connect directly (without NATProxy)
```
Waldz: Why session is started at all, if you need to shut it down e.g. DI should launch noopSession

services/openvpn/client_config.go
```diff
@@ -100,9 +100,11 @@ func NewClientConfigFromSession(sessionConfig []byte, configDir string, runtimeD
...
+			vpnConfig.RemoteIP = "127.0.0.1"
```
Waldz: DI should inject these differences instead of conditional programming. Especially inside `openvpn` package, this does not know about NAT Proxy at all

nat/traversal/nat_proxy.go
zolia: pinger session is needed and at this point it already did its work, that why we shutting it down.

services/openvpn/client_config.go
zolia: We are in constructor here, not sure how we can avoid conditional here, we can have 2 different constructors then? Is it better?

services/openvpn/client_config.go
Waldz: Moved mobile `vpnConfig` factoring to same place

services/openvpn/client_config.go
Waldz: Make 2 different constructions of `vpnConfig `

PR #1065, opened 2019-05-15
Update 0.8 release with fixes 

Waldz:


nat/traversal/nat_proxy.go
```diff
@@ -198,7 +198,6 @@ func (np *NATProxy) getConnection(serviceType services.ServiceType) (*net.UDPCon
 	if err != nil {
 		return nil, err
 	}
-
```
tadovas: This one looks as important fix :D

Waldz: important hacks :)

PR #1066, opened 2019-05-15
Make port pool decoupled again

Waldz:
Problems solved:
- `Pool` contained 2 different interfaces in 1 struct
- `Pool` was used globally in DI and it's state was changed randomly in runtime
- Openvpn was aquiring service port 2 times: in `Serve()` and in `ProviderConfig`


core/port/pool_fixed.go
```diff
@@ -0,0 +1,43 @@
...
+	return &PoolFixed{
+		port:       port,
+		randomPool: NewPool(),
+	}
+}
+
+// PoolFixed hands out a fixed port for service use
+type PoolFixed struct {
+	port       Port
+	randomPool *Pool
```
zolia: Whats the point to have Fixed pool, when its giving out random ports? 
Do we really need to create separate structure each time we need additional method?
Basically we duplicate code and scatter code with lots of instances of pools. Changes would mean we need to change them all.

vkuznecovas: I'll also add a question:

How do you solve the possible collisions between the two different instances of port pools in a case when the provider is running both wireguard and openvpn?

Waldz: Collision wasn't solved, I can works towards that problems next.
Basically I am making more more clear signature of port acquire, because now is coupled and sensitive on changes

vkuznecovas: But you're making it worse - there was no collision before, you're introducing it now.

Waldz: I checked:
- Collisions protection is there by trying to open socket on aquired ports
- Before acquired ports was cached to different variables `port` and `openvpnPort`, so it's the same

PR #1067, opened 2019-05-15
bootstrap now checks for presence of node before starting

vkuznecovas:
We're running into an issue where the node is being started multiple times. This causes various issues, namely boltdb lock races. To fix it, introducing an early attempt to bind to the tequilapi port. If the bind suceeds, we'll now that a node isn't running. Well, at least on the same port.

### DIFF
--- a/cmd/di.go
+++ b/cmd/di.go
@@ -19,6 +19,7 @@ package cmd
 
 import (
 	"fmt"
+	"net"
 	"path/filepath"
 	"time"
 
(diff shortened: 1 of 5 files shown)


cmd/di.go
```diff
@@ -180,6 +182,12 @@ func (di *Dependencies) Bootstrap(nodeOptions node.Options) error {
 	log.Infof("Starting Mysterium Node (%s)", metadata.VersionAsString())
 	log.Infof("Build information (%s)", metadata.BuildAsString())
 
+	// check early for presence of an already running node
+	listener, err := net.Listen("tcp", fmt.Sprintf("%s:%d", nodeOptions.TequilapiAddress, nodeOptions.TequilapiPort))
```
Waldz: `Bootstrap()` is not to run things, just to register components.
You can move listener run to beginning of `httpAPIServer.StartServing()` - it's 1st action we run

vkuznecovas: That's not going to work. I need to check for presence of an already running myst BEFORE initializing the storage.

zolia: would suggest `tequilaListener`

PR #1069, opened 2019-05-15
Refactor pinger without timeouts

zolia:


nat/traversal/pinger.go
```diff
@@ -210,14 +160,19 @@ func (p *Pinger) PingProvider(ip string, port int, stop <-chan struct{}) error {
 		return err
 	}
 
+	err = ipv4.NewConn(conn).SetTTL(128)
+	if err != nil {
+		return errors.Wrap(err, "setting ttl failed")
+	}
+	// send one last ping request to end hole punching procedure gracefully
+	err = p.sendPingRequest(conn, 128)
```
soffokl: We are doing nothing with this `err`.

Issue #1057, opened 2019-05-13
Publish DEB packages to the PPA


soffokl:
We need to publish our AMD64 and ARM DEB packages to the https://launchpad.net/~mysteriumnetwork/+archive/ubuntu/node/+packages

PR #1073, opened 2019-05-16
Built DEB source package and upload to PPA

soffokl:
Closes: #1057


.travis.yml
```diff
@@ -181,18 +187,38 @@ jobs:
...
+      - gpg --import bin/package/android/key.gpg
+      - openssl aes-256-cbc -K $encrypted_2c668321e34e_key -iv $encrypted_2c668321e34e_iv -in bin/package/ppa/ppa_key.enc -out ~/.ssh/id_rsa -d
+      - chmod 600 ~/.ssh/id_rsa
+      - bin/release_ppa "$PPA_VERSION" "$BUILD_NUMBER" xenial
+      - bin/release_ppa "$PPA_VERSION" "$BUILD_NUMBER" bionic
+      name: "Publish DEB packages to the PPA"
+      addons:
+        apt:
+          packages:
+          - devscripts
```
Waldz: Did You consider to add dependencies to `builder` container and build inside it?

debian/control
```diff
@@ -0,0 +1,29 @@
...
+XS-Go-Import-Path: github.com/mysteriumnetwork/node
+
+Package: myst
+Architecture: any
+Built-Using: ${misc:Built-Using}
+Depends:
+ ${misc:Depends},
+ ${shlibs:Depends},
+ sudo,
+ resolvconf,
```
Waldz: Looks like duplicated with `bin/package_debian`. 
Aren't we reusing existing .deb? Or are we going to drop .deb file with public PPA releases?

.travis.yml
```diff
@@ -181,18 +187,38 @@ jobs:
...
       - bin/travis_scripts/trigger_desktop_app_build
       name: "Trigger Desktop APP build"
       if: tag = env(BUILD_DEV_RELEASE)
 
+    - script:
+      - openssl aes-256-cbc -K $encrypted_2698a83519b3_key -iv $encrypted_2698a83519b3_iv -in bin/package/android/key.gpg.enc -out bin/package/android/key.gpg -d
+      - gpg --import bin/package/android/key.gpg
+      - openssl aes-256-cbc -K $encrypted_2c668321e34e_key -iv $encrypted_2c668321e34e_iv -in bin/package/ppa/ppa_key.enc -out ~/.ssh/id_rsa -d
+      - chmod 600 ~/.ssh/id_rsa
+      - bin/release_ppa "$PPA_VERSION" "$BUILD_NUMBER" xenial
```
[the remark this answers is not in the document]
soffokl: The default value for `PPA_VERSION` is `0.0.0`, but for `BUILD_VERSION` is `0.0.0-dev` which is not acceptable with a current build process for PPA.

.travis.yml
soffokl: Yeah, but it will make a configuration even more complicated since we need to pass two private keys into the builder. And with a builder container, it takes longer to finish a build. 
I think this way it's more clear what we are doing in this build.

debian/control
soffokl: `bin/package_debian` builds a binary DEB package, but for PPA we have to provide a source DEB package, which will be built on the Launchpad.
And yes, I think it will be possible to drop building DEB packaged and leave it only in PPA.

.travis.yml
```diff
@@ -181,18 +187,38 @@ jobs:
...
+      - bin/release_ppa "$PPA_VERSION" "$BUILD_NUMBER" bionic
```
zolia: Why are we releasing just for ubuntu? IMHO we can release for debian too: `https://launchpad.net/debian/stretch`

debian/rules
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/make -f
+# -*- makefile -*-
+
+export DH_VERBOSE := 1
+export PATH := /usr/lib/go-1.11/bin/:$(PATH)
```
zolia: hardcoded go version? Won't there be a problem?

.travis.yml
soffokl: I don't think we can build PPA for Debian:
image

debian/rules
soffokl: yeah, we have also hardcoded dependency to this version:
https://github.com/mysteriumnetwork/node/pull/1073/files#diff-d837a1c17de0268bcea321239ddbed47R8

.travis.yml
```diff
@@ -181,18 +187,38 @@ jobs:
...
       - bin/release_android "$TRAVIS_TAG"
       name: "Publish Android SDK to the Sonatype and Maven Central"
+
     - script:
       - bin/travis_scripts/trigger_desktop_app_build
       name: "Trigger Desktop APP build"
       if: tag = env(BUILD_DEV_RELEASE)
 
+    - script:
+      - openssl aes-256-cbc -K $encrypted_2698a83519b3_key -iv $encrypted_2698a83519b3_iv -in bin/package/android/key.gpg.enc -out bin/package/android/key.gpg -d
```
vkuznecovas: Maybe it makes sense to take these steps and put them in a script in bin folder with a short explanation of variables needed? If we're moving to another CI would make it a bit less frustrating to move

soffokl: It's totally Travis file encryption process: https://docs.travis-ci.com/user/encrypting-files/
Migration to the new CI will require it to be changed anyway. So keeping it in Travis config.

.travis.yml
Waldz: Builder mounts all project dir as inside volume. Also variables goes to there.
I am bringing this in, because @tadaskay soon will run all jobs in Docker runner, which is based on this builder

soffokl: I have registered issue for that https://github.com/mysteriumnetwork/node/issues/1086
Let's change it a bit later, and now we will have builds like this.

PR #1077, opened 2019-05-16
A UI for node is now available

vkuznecovas:
It can be activated via the ui.enabled flag. The port can also be specified via ui.port flag.

### DIFF
--- a/core/node/node.go
+++ b/core/node/node.go
@@ -36,6 +36,12 @@ type Publisher interface {
 	Publish(topic string, data interface{})
 }
 
+// WebServer represents the web server for our web
+type WebServer interface {
+	Serve() error
+	Stop()
+}
+
 // NewNode function creates new Mysterium node by given options
 func NewNode(
 	connectionManager connection.Manager,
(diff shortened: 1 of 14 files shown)


core/node/node.go
```diff
@@ -36,20 +36,28 @@ type Publisher interface {
...
+}
+
 // NewNode function creates new Mysterium node by given options
 func NewNode(
 	connectionManager connection.Manager,
 	tequilapiServer tequilapi.APIServer,
 	publisher Publisher,
 	metricsSender *metrics.Sender,
 	natPinger NatPinger,
+	webServer WebServer,
```
Waldz: uiServer

web/server.go
```diff
@@ -0,0 +1,77 @@
...
+
+const logPrefix = "[dvpn-web-server] "
+
+// Server represents our web UI server
+type Server struct {
+	srv *http.Server
+}
+
+// NewServer creates a new instance of the server for the given port
+func NewServer(port int) *Server {
```
Waldz: I suggest to keep this server in another repo which has generators of assets.
And include `dep`

web/server.go
```diff
@@ -0,0 +1,77 @@
...
+	gin.SetMode(gin.ReleaseMode)
+	r := gin.New()
+	r.Use(gin.Recovery())
+	r.Use(cors.Default())
+	r.StaticFS("/", godvpnweb.Assets)
```
Waldz: `node` repo hardcodes assets from another repo.
I suggest to keep this server in another repo which has generators of assets and include `dep`

PR #1078, opened 2019-05-16
Tequilapi /identities/current endpoint

### LINKED ISSUE #1093: Add /identities/current endpoint
We need an easier way to fetch and unlock an identity for apps. Currently multiple identities, or even password isn't used, so we can have a single endpoint that creates/finds and unlocks an identity.

ignasbernotas:
Closes #1093 

Finds and unlocks the last used identity, if it fails, unlocks the first identity, if none exist creates one and unlocks it.


tequilapi/endpoints/identities.go
```diff
@@ -91,6 +98,57 @@ func (endpoint *identitiesAPI) List(resp http.ResponseWriter, request *http.Requ
 	utils.WriteAsJSON(idsSerializable, resp)
 }
 
+// swagger:operation PUT /me Identity myIdentity
```
tadovas: It's not very RESTish. A better url would be /identities/current or smth like that. It's subresourse of identities

ignasbernotas: Yeah forgot to mention :D https://github.com/julienschmidt/httprouter/issues/73 couldn't do it due to a limitation in our router package :/

ignasbernotas: basically you can't have `identities/x` if you have `identities/:id`

tadovas: I don't see how its related

tadovas: Ok I see it. Time to migrate to [Gin](https://github.com/gin-gonic/gin)

ignasbernotas: Yeah, that'd be a good choice:)

tadovas: It even can be handled inside identities controller with if identity == "current" { ... } it's ugly hack and needs //TODO at least, but it hides limitations of framework especially in these sensitive parts as external api definition. This way we can have nice API and do framework migration after without external changes.

identity/selector/handler_test.go
```diff
@@ -46,102 +46,43 @@ func TestUseExistingSucceeds(t *testing.T) {
 	assert.Equal(t, "pass", identityManager.LastUnlockPassphrase)
 }
 
-func TestUseExistingFailsWhenUnlockFails(t *testing.T) {
+func TestUUseOrCreateFailsWhenUnlockFails(t *testing.T) {
```
Waldz: TestUUseOrCreateFailsWhenUnlockFails  -> TestUseOrCreateFailsWhenUnlockFails

identity/selector/handler_test.go
```diff
@@ -46,102 +46,43 @@ func TestUseExistingSucceeds(t *testing.T) {
...
-
-	assert.Equal(t, "", registry.registeredIdentity.Address)
-
-	assert.Equal(t, "abc", identityManager.LastUnlockAddress)
-	assert.Equal(t, "pass", identityManager.LastUnlockPassphrase)
-}
-
-func TestUseNewSucceeds(t *testing.T) {
-	identityManager := identity.NewIdentityManagerFake([]identity.Identity{existingIdentity}, newIdentity)
+func TestUseOrCreateReturnsFirstIdentity(t *testing.T) {
```
Waldz: U can test private functions, instead of removing tests

tequilapi/endpoints/identities.go
```diff
@@ -40,6 +41,11 @@ type identityList struct {
 	Identities []identityDto `json:"identities"`
 }
 
+// swagger:model MyIdentityDTO
+type myIdentityDTO struct {
+	Passphrase *string `json:"passphrase"`
```
Waldz: I miss `Identity` as it existed in previous signature

cmd/commands/service/command.go
```diff
@@ -137,8 +133,7 @@ func (sc *serviceCommand) Run(ctx *cli.Context) (err error) {
 }
 
 func (sc *serviceCommand) unlockIdentity(identityOptions service.OptionsIdentity) (identity.Identity, error) {
-	loadIdentity := identity_selector.NewLoader(sc.identityHandler, identityOptions.Identity, identityOptions.Passphrase)
-	return loadIdentity()
+	return sc.identityHandler.UseOrCreate(identityOptions.Identity, identityOptions.Passphrase)
```
Waldz: Map `identityOptions`  to myIDRequest and unlock thru Tequilapi here

tequilapi/endpoints/identities.go
vkuznecovas: Moving to gin will not solve this. Gin is basicly using a custom version of julienschmidt/httprouter. They've just baked it in. Trade off is that you sacrifice speed of routing for the lack of customizability. Gin might not be the web server we need here.

This will panic:

```golang
package main

import (
	"net/http"

	"github.com/gin-gonic/gin"
)

func main() {
	router := gin.Default()
	router.GET("/identity", func(c *gin.Context) {
		c.String(http.StatusOK, "/identity")
	})
	router.GET("/identity/x", func(c *gin.Context) {
		c.String(http.StatusOK, "/identity/x/")
	})
	router.GET("/identity/:id", func(c *gin.Context) {
		c.String(http.StatusOK, "/identity/:id/")
	})
	router.Run(":1234")
}
```

output:
```
panic: wildcard route ':id' conflicts with existing children in path '/identity/:id'
```

ignasbernotas: I have used this not long ago and it worked for this case:
`https://github.com/gorilla/mux`

identity/selector/handler_test.go
ignasbernotas: How? they're not private functions, they're private struct methods:/

ignasbernotas: Is it an OK practice to make things public just for the sake of testing? @Waldz @soffokl

ignasbernotas: Nevermind, Waldz showed me my mistake:)

tequilapi/endpoints/identities.go
ignasbernotas: Changed the endpoint url to /identities/current

cmd/commands/service/command.go
```diff
@@ -83,14 +83,10 @@ func NewCommand(licenseCommandName string) *cli.Command {
...
-				errorChannel: errorChannel,
-				identityHandler: identity_selector.NewHandler(
-					di.IdentityManager,
-					di.MysteriumAPI,
-					identity.NewIdentityCache(nodeOptions.Directories.Keystore, "remember.json"),
-					di.SignerFactory),
-				ap: parseAccessPolicyFlag(ctx),
+				tequilapi:       client.NewClient(nodeOptions.TequilapiAddress, nodeOptions.TequilapiPort),
+				errorChannel:    errorChannel,
+				identityHandler: di.IdentitySelector,
```
Waldz: You dont use this dependency anymore

identity/selector/handler.go
```diff
@@ -50,7 +50,27 @@ func NewHandler(
...
 
-func (h *handler) UseExisting(address, passphrase string) (id identity.Identity, err error) {
+func (h *handler) UseOrCreate(address, passphrase string) (id identity.Identity, err error) {
+	if len(address) > 0 {
+		return h.useExisting(address, passphrase)
+	}
+
+	identities := h.manager.GetIdentities()
+	if len(identities) == 0 {
+		id, err = h.useNew(passphrase)
```
soffokl: We need to do something with this error. 
```suggestion
		return h.useNew(passphrase)
```

cmd/di.go
```diff
@@ -19,6 +19,7 @@ package cmd
 
 import (
 	"fmt"
+
```
soffokl: Extra line, it's not needed here.

tadovas:
@ignasbernotas Could you please describe in few words why we need this functionality anyway? "Current" identity is not needed from consumer side - as we provide it everytime we make connection. The same can be done with provider - when starting new service with API - pass desired identity as a parameter. It also can be used as argument on node startup if needed. This global "current" variable is a bit questionable. A real use case would really help to understand purpose of this.

ignasbernotas:
@tadovas basically app api consumer needs to call 3 endpoints before they're able to start service/connect to vpn. And at the end of the day, all they really need is one endpoint which returns and unlocks their identity. We don't have UI for multiple identities (nor will we need it any time soon) and it just simplifies things. You call one endpoint and that's it.

tadovas:
That is exactly my concern - RESTish APIs usually have single purpose, and there is no need to create new apis for one use case which does everything. Personally I am against "one endpoint does everything" approach, especially as it's so related to one case from million other possibilities. I would like to hear others opinion too.

ignasbernotas:
From an architectural standpoint sure, but there isn't a million possibilities, there's two. One for CLI and one for the apps, and apps need this :) We want our API to be easy to use and dev friendly, this solves an actual case so I don't see why we should be so bothered by it.

tadovas:
Dev friendly api is exactly that - small independent endpoints with single purpose. Doing single endpoint which does 3 jobs is exactly the opposite and as you correctly noticed - not very nice from architecture perspective, just for our two cases. I am not happy with that, therefore as I said - need others opinions.

ignasbernotas:
@zolia @Waldz @vkuznecovas @soffokl for context, we agreed to have this during sprint planning as this simplifies things for app development. Please comment :)

Waldz:
*Option 1.* For this route register anonymous handler, which call your `Current()` handler:
```
func Midleware(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
    if params.id == 'current' {
       return endpoint.Current(w, r, ps)
    }
    return endpoint.Create(w, r, ps)
}
```

*Option 2.* 
Register route `/me` or `/login` and avoid these collision, your route is is anyway not RESTful

PR #1085, opened 2019-05-17
Do not exit if pinger read fails

zolia:
- failed read does not mean a failure. Reproducible on win10 without FW
 - do not start pinging if not required

### DIFF
--- a/nat/traversal/pinger.go
+++ b/nat/traversal/pinger.go
@@ -124,9 +124,15 @@ func (p *Pinger) Start() {
 			log.Info(prefix, "stop pinger called")
 			return
 		case pingParams := <-p.pingTarget:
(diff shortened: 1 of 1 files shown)


nat/traversal/pinger.go
```diff
@@ -124,11 +124,20 @@ func (p *Pinger) Start() {
...
-			go p.pingTargetConsumer(pingParams)
+			if isPunchingRequired(pingParams) {
+				go p.pingTargetConsumer(pingParams)
+			}
 		}
 	}
 }
 
+func isPunchingRequired(params *Params) bool {
+	if params.ConsumerPort > 0 {
```
soffokl: Just return enough here:
```suggestion
	return params.ConsumerPort > 0
```

Issue #1061, opened 2019-05-14
Setup gitlab build


tadaskay:
- PR builds
- Snapshot releases: https://github.com/mysteriumnetwork/node-builds/releases

PR #1096, opened 2019-05-21
Setup gitlab PR build

tadaskay:
Without release/publish functionality, but should be sufficient for PR builds.

Ref #1061

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -42,7 +42,7 @@ install-deps:
       - vendor/
   script:
     - >
-      go get \
+      go get -u \
         github.com/debber/debber-v0.3/cmd/debber \
         golang.org/x/lint/golint \
         golang.org/x/tools/cmd/goimports \
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -44,7 +44,7 @@ install-deps:
     - >
       go get \
         github.com/debber/debber-v0.3/cmd/debber \
-        github.com/golang/lint/golint \
+        golang.org/x/lint/golint \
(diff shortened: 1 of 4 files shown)


.gitlab-ci.yml
```diff
@@ -0,0 +1,204 @@
...
+  tags:
+    - go
+  script:
+    - go get -u github.com/debber/debber-v0.3/cmd/debber
+    - bin/build_xgo linux/arm
+    - BINARY=build/myst/myst_linux_arm bin/package_debian $BUILD_VERSION armhf
+  only:
+    - master
+
+package:linux-debian-arm-raspberry:
```
soffokl:
```suggestion
package:linux-raspberry-image:
```

.gitlab-ci.yml
```diff
@@ -0,0 +1,204 @@
...
+
+package:swagger-redoc-docker:
+  stage: build
+  tags:
+    - go
+  script:
+    - go get -u github.com/go-swagger/go-swagger/cmd/swagger/
+    - bin/swagger_generate
+    - bin/package_docker_docs
+    - docker save tequilapi:${BUILD_VERSION} | gzip > tequilapi_redoc.tgz
```
Waldz: Do we need to export docker images if all tasks are same machine now?

tadaskay: we have 4 concurrent builders for now (including 1 machine from Antanas)

tadaskay: In the next PR i'll change it to artifacts, so that they are properly published. This is interim setup (PR only)

PR #1099, opened 2019-05-22
Close NATS connection correctly.

soffokl:
This commit fixes a memory leak on multiple reconnects to providers.
In normal behaviour memory leak is not really noticeable.
But when network changes for some reasons, all active nats connections tries to reconnect to the broker at the same time and memory usage become significant.

### DIFF
--- a/communication/nats/dialog/dialog.go
+++ b/communication/nats/dialog/dialog.go
@@ -19,16 +19,21 @@ package dialog
 
 import (
 	"github.com/mysteriumnetwork/node/communication"
+	"github.com/mysteriumnetwork/node/communication/nats/discovery"
 	"github.com/mysteriumnetwork/node/identity"
 )
 
 type dialog struct {
 	communication.Sender
 	communication.Receiver
-	peerID identity.Identity
+	peerID      identity.Identity
+	peerAddress *discovery.AddressNATS
 }
 
 func (dialog *dialog) Close() error {
+	if dialog.peerAddress != nil {
+		dialog.peerAddress.Disconnect()
+	}
 	return nil
 }
 
--- a/communication/nats/dialog/dialog_establisher.go
+++ b/communication/nats/dialog/dialog_establisher.go
@@ -68,6 +68,7 @@ func (establisher *dialogEstablisher) EstablishDialog(
 	peerSender := establisher.newSenderToPeer(peerAddress, peerCodec)
 	topic, err := establisher.negotiateTopic(peerSender)
 	if err != nil {
+		peerAddress.Disconnect()
 		return nil, err
 	}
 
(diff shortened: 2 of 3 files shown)


communication/nats/dialog/dialog.go
```diff
@@ -19,16 +19,21 @@ package dialog
...
 type dialog struct {
 	communication.Sender
 	communication.Receiver
-	peerID identity.Identity
+	peerID      identity.Identity
+	peerAddress *discovery.AddressNATS
 }
 
 func (dialog *dialog) Close() error {
+	if dialog.peerAddress != nil {
```
tadovas: In which cases peer address can be nil? Those if not nil guards indicate that there is something wrong with struct init - why we cannot have dialog fully constructed and prepared before usage?

zolia: Usually it is called a defensive programming. Most likely that peerAddress cannot be nil; Not sure if it is compatible with go guidelines, but I used to such coding. Usually e2e tests should expose such nil and it should be fixed in the code that no nil would surface.

If err is produced somewhere in a process of creating such address, then we should terminate early and should not reach Disconnect at all.

tadovas: Defensive programming makes sense in cases where you cannot control your functions input. In here we are constructing private structure with fields. Any unnecessary code lines raise questions

soffokl: The problem here is that we are using this `dialog` struct for both `dialogWaiter` and `dialogEstablisher`. `dialogWaiter` is reusing the single nats connection to all dialogs, but `dialogEstablisher` is creating a new nats connection for every dialog, and this connection should be closed with a dialog. 
I know this is not a perfect architecture, but I don't really want to redo it in the scope of this PR.

Moreover, we are planning to change this communication, so it will be changed. But this PR is a quick fix for a memory leaking.

soffokl: About a defensive part, I prefer to have an extra check, but not an unexpected panic when it will be changed next time.

tadovas: That sounds reasonable enough. I really hope new communication architecture will come soon enough.

tadovas: > About a defensive part, I prefer to have an extra check, but not an unexpected panic when it will be changed next time.

Defensive part is not a single bullet and must be used right, there are some code smells related it. And dont' you think if you find out yourself in dialog with nil that something is terribly wrong anyway long before you reached Close? Do you really want to swallow it and ignore (if not nil does exactly that), fail fast - check in construction that all your params are as expected to be

tadovas:
and BTW @zolia @soffokl defensive programming is about shouting to the world, stopping with panics, throwing errors if your checks or assertions fail. In this case - code is made optional. That rises question - is connection optional to dialog? If connection is optional to dialog, maybe something is wrong with design here?

PR #1101, opened 2019-05-22
Filter proposals locally

Waldz:
*Expected behaviour:*
- Consumer who are interested in proposals fetching now fetches them from local proposal storage
- Proposals are fetched to local storage once just after node starts
- Later proposals are fetched each 1 minute

*Changes made:*
- Introduced `discovery.ProposalFinder` for abstraction of proposals filtering 
- 2 discovery finders introduced: `discover/api/finder.go`- from API and `discovery/finder.go` - from local storage
- Only local storage is used now, discovery finder left if we'll need to restore legacy logic
- 1st local storage fetcher introduced - which fetches from API

*Next:*
- Introduce 2nd storage fetcher - which fetches from Broker messages


Waldz:
E2E issue addressed and PR is ready to read

core/discovery/storage.go
```diff
@@ -0,0 +1,58 @@
...
+
+// Proposals returns table of proposals in storage
+func (storage *ProposalStorage) Proposals() []market.ServiceProposal {
+	storage.mutex.Lock()
+	defer storage.mutex.Unlock()
+	return storage.proposals
+}
+
+// Add puts given proposal to storage
+func (storage *ProposalStorage) Add(proposal market.ServiceProposal) {
```
soffokl: We will not need an `AddMultiple` if `Add` will look like this:
```go
// Add puts given proposal to storage
func (storage *ProposalStorage) Add(proposal ...market.ServiceProposal) {
	storage.mutex.Lock()
	defer storage.mutex.Unlock()
	storage.proposals = append(storage.proposals, proposal...)
}
```

core/discovery/storage.go
```diff
@@ -0,0 +1,58 @@
...
+	storage.mutex.Lock()
+	defer storage.mutex.Unlock()
+	storage.proposals = append(storage.proposals, proposal)
+}
+
+// AddMultiple puts given multiple proposals to storage
+func (storage *ProposalStorage) AddMultiple(proposals []market.ServiceProposal) {
+	storage.mutex.Lock()
+	defer storage.mutex.Unlock()
+	storage.proposals = proposals
```
soffokl: It looks like `Replace` not `Add`.

market/mysterium/mysterium_api.go
```diff
@@ -151,43 +151,37 @@ func (mApi *MysteriumAPI) PingProposal(proposal market.ServiceProposal, signer i
...
 	}
-	if filter.ServiceType != "" {
-		query.Set("service_type", filter.ServiceType)
+	if query.ServiceType != "" {
+		values.Set("service_type", query.ServiceType)
 	}
-	if filter.AccessPolicyID != "" {
-		query.Set("access_policy[id]", filter.AccessPolicyID)
+	if query.AccessPolicyAll {
+		values.Set("access_policy", "*")
```
soffokl: We have 
`"all"` for `ServiceType`
`"*"` for `AccessPolicy`
`bool` as `AccessPolicyAll`

This is very confusing.

core/discovery/api/fetcher.go
```diff
@@ -0,0 +1,106 @@
...
+	return nil
+}
+
+// Stop ends fetching proposals to storage
+func (fetcher *Fetcher) Stop() {
+	fetcher.fetchShutdown <- true
+}
+
+// SubscribeProposals allows to subscribe all fetched proposals
+func (fetcher *Fetcher) SubscribeProposals(proposalsChan chan market.ServiceProposal) {
```
vkuznecovas: you should allow multiple subscribers here. Create an internal storage for channels, and allow for adding/removing from it.

core/discovery/api/fetcher_test.go
```diff
@@ -0,0 +1,138 @@
...
+		assert.Exactly(t, []market.ServiceProposal{proposalFirst, proposalSecond}, storage.Proposals())
+	}()
+}
+
+func Test_Fetcher_StartFetchesNewProposals(t *testing.T) {
+	storage := discovery.NewStorage()
+	fetcher := NewFetcher(storage, proposalsCurrent.Fetch, time.Millisecond)
+
+	proposalsCurrent.Mock(proposalFirst)
+	go func() {
```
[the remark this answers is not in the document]
vkuznecovas: don't think he can do it without a goroutine here, he'll need to routine either start or the mock/assert flow as the start blocks, right? Same applies to most of the tests.

core/discovery/api/fetcher.go
```diff
@@ -0,0 +1,106 @@
...
+		log.Warn(fetcherLogPrefix, fmt.Sprintf("Failed to fetch proposals: %s", err))
+		return err
+	}
+
+	log.Info(fetcherLogPrefix, fmt.Sprintf("Proposals fetched: %d", len(proposals)))
+	fetcher.proposalStorage.AddMultiple(proposals)
+
+	if fetcher.proposalChan != nil {
+		for _, proposal := range proposals {
+			fetcher.proposalChan <- proposal
```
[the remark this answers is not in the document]
vkuznecovas:
```suggestion
			select{
			  case fetcher.proposalChan <- proposal:
			  case <- time.After(time.Millisecond * 50):
			  return nil
			}
```

This would work, right?

core/discovery/api/fetcher_test.go
```diff
@@ -0,0 +1,138 @@
...
+}
+
+func Test_Fetcher_StartFetchesInitialProposals(t *testing.T) {
+	storage := discovery.NewStorage()
+	fetcher := NewFetcher(storage, proposalsCurrent.Fetch, time.Hour)
+
+	proposalsCurrent.Mock(proposalFirst, proposalSecond)
+	go func() {
+		err := fetcher.Start()
+		defer fetcher.Stop()
```
vkuznecovas: If start is blocking, who calls the stop?

core/discovery/api/fetcher_test.go
```diff
@@ -0,0 +1,138 @@
...
+
+	proposalsCurrent.Mock(proposalSecond)
+	assert.Exactly(t, proposalSecond, waitForProposal(t, proposalChan))
+}
+
+func waitForProposal(t *testing.T, proposalsChan chan market.ServiceProposal) market.ServiceProposal {
+	select {
+	case proposal := <-proposalsChan:
+		return proposal
+	case <-time.After(2 * time.Millisecond):
```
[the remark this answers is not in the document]
vkuznecovas: 2ms will be more than enough here. Goroutines are cheap, they have little overhead and in terms of processors 2ms is a TON of time.

core/discovery/finder.go
```diff
@@ -0,0 +1,104 @@
...
+	storage *ProposalStorage
+}
+
+// NewFinder creates instance of local storage finder
+func NewFinder(storage *ProposalStorage) *finder {
+	return &finder{storage: storage}
+}
+
+// GetProposal fetches service proposal from discovery by exact ID
+func (finder *finder) GetProposal(id market.ProposalID) (*market.ServiceProposal, error) {
```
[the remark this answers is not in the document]
vkuznecovas: use the go way here, just return (market.ServiceProposal, bool)

core/discovery/finder.go
```diff
@@ -0,0 +1,104 @@
...
+			return &proposal, nil
+		}
+	}
+
+	return nil, nil
+}
+
+// FindProposals fetches currently active service proposals from discovery
+func (finder *finder) FindProposals(filter market.ProposalFilter) ([]market.ServiceProposal, error) {
+	proposalsFiltered := make([]market.ServiceProposal, 0)
```
vkuznecovas: you're not using the error, return the []market.ServiceProposal

core/discovery/finder.go
```diff
@@ -0,0 +1,104 @@
...
+		proposalsFiltered = append(proposalsFiltered, proposal)
+	}
+	return proposalsFiltered, nil
+}
+
+var (
+	emptyFilterLocation = market.LocationFilter{}
+)
+
+func filterByLocation(proposal market.ServiceProposal, filter market.LocationFilter) bool {
```
vkuznecovas: super confusing naming. I'd probably call it matchesLocation. Maybe even move it to the market.LocationFilter struct, then would call it like so: 
```golang
filter.Matches(proposal)
```

core/discovery/finder.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+
+	return true
+}
+
+var (
+	emptyFilterAccessPolicy = market.AccessPolicyFilter{}
+)
+
+func filterByAccessPolicy(proposal market.ServiceProposal, filter market.AccessPolicyFilter) bool {
```
vkuznecovas: super confusing naming. I'd call it matchesAccessPolicy

core/discovery/interface.go
```diff
@@ -28,3 +28,9 @@ type ProposalRegistry interface {
 	PingProposal(proposal market.ServiceProposal, signer identity.Signer) error
 	UnregisterProposal(proposal market.ServiceProposal, signer identity.Signer) error
 }
+
+// ProposalFinder allows to search proposals by specified params
+type ProposalFinder interface {
```
vkuznecovas: still not a fan of defining interfaces nowhere close to where they are being used.

core/discovery/storage.go
```diff
@@ -0,0 +1,58 @@
...
+func NewStorage() *ProposalStorage {
+	return &ProposalStorage{
+		proposals: make([]market.ServiceProposal, 0),
+	}
+}
+
+// ProposalStorage represents table of currently active proposals in Mysterium Discovery
+type ProposalStorage struct {
+	proposals []market.ServiceProposal
+	mutex     sync.Mutex
```
vkuznecovas: could switch to RWMutex here to allow for less locking time while reading.

core/discovery/api/fetcher_test.go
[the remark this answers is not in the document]
soffokl: But `Start` is not blocking here.

core/discovery/api/fetcher.go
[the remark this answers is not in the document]
Waldz: I can add timeout but why if I have protection `if fetcher.proposalChan != nil`

market/mysterium/mysterium_api.go
Waldz: Yes it is, that's why I do this PR, so that we isolate this behaviour and don't call `MysteriumAPI.FindProposals()` anumore

core/discovery/api/fetcher_test.go
Waldz: Checked. Start is blocking until it gets bunch of proposals successfully

core/discovery/finder.go
Waldz: This implementation doesn't have error, another has. Don't want to change interface signature until local proposal storage is stable

Waldz: Also Discovery down situation should be handled in some other way. Dont want to extend this PR

core/discovery/interface.go
Waldz: Yes, me too, just this interface is used in several places and don't know how to solve that.
Don't want to add ti to `di.go` either, because then everybody includes `cmd` package

core/discovery/api/fetcher.go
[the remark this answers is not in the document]
Waldz: No point to timeout because subscriber must handle his subscription

PR #1104, opened 2019-05-22
Port test/test e2e to mage

tadaskay:
Ref #1061

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -66,14 +66,14 @@ test:
   tags:
     - go
   script:
-    - bin/test
+    - go run mage.go test
 
 test-e2e:
   stage: test
   tags:
     - go
   script:
-    - bin/test_e2e
+    - go run mage.go -v teste2e
 
 package:linux-amd64:
   stage: build
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -263,6 +263,20 @@
   version = "v1.1"
 
 [[projects]]
+  digest = "1:f4e64bb62274202b95403b8ae001d0601cee01afadf60767fb0e1d239f8f1413"
+  name = "github.com/magefile/mage"
+  packages = [
+    "internal",
+    "mage",
+    "mg",
+    "parse",
+    "sh",
+  ]
+  pruneopts = "UT"
+  revision = "aedfce64c122eef47009b7f80c9771044753215d"
+  version = "v1.8.0"
+
+[[projects]]
   digest = "1:c658e84ad3916da105a761660dcaeb01e63416c8ec7bc62256a9b411a05fcd67"
   name = "github.com/mattn/go-colorable"
   packages = ["."]
@@ -733,6 +747,8 @@
     "github.com/huin/goupnp/ssdp",
     "github.com/jackpal/gateway",
     "github.com/julienschmidt/httprouter",
+    "github.com/magefile/mage/mage",
+    "github.com/magefile/mage/sh",
     "github.com/mitchellh/go-homedir",
     "github.com/mysteriumnetwork/go-dvpn-web",
     "github.com/mysteriumnetwork/go-openvpn/openvpn",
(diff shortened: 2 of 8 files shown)


bin/tests/e2e.go
```diff
@@ -0,0 +1,155 @@
+/*
```
soffokl: `bin/tests/e2e.go` 
`bin`/ is not look like a right place for go files.

bin/tests/e2e.go
```diff
@@ -0,0 +1,155 @@
...
+import (
+	"bufio"
+	"fmt"
+	"os"
+
+	"github.com/magefile/mage/sh"
+	"github.com/pkg/errors"
+)
+
+var projectName = "node_e2e_test"
```
soffokl:
```suggestion
const projectName = "node_e2e_test"
```

bin/tests/e2e.go
```diff
@@ -0,0 +1,155 @@
...
+	fmt.Println("starting other services")
+	if err := r.compose("up", "-d", "broker", "geth"); err != nil {
+		return errors.Wrap(err, "starting other services failed!")
+	}
+	fmt.Println("starting DB")
+	if err := r.compose("up", "-d", "db"); err != nil {
+		return errors.Wrap(err, "starting DB failed!")
+	}
+
+	for {
```
soffokl: We can run forever in this loop if it never up.

bin/tests/e2e.go
```diff
@@ -0,0 +1,155 @@
...
+		err := r.compose("exec", "-T", "db", "mysqladmin", "ping", "--protocol=TCP", "--silent")
+		if err != nil {
+			fmt.Print(".")
+		} else {
+			fmt.Println()
+			break
+		}
+	}
+
+	fmt.Println("migrating DB")
```
soffokl: I'd use some logging instead of just prints. At least we will have time marks and levels that will help to read the output.

bin/tests/e2e.go
```diff
@@ -0,0 +1,155 @@
...
+		return errors.Wrap(err, "failed to deploy contracts!")
+	}
+
+	fmt.Println("building app images")
+	if err := r.compose("build"); err != nil {
+		return errors.Wrap(err, "building app images failed!")
+	}
+	fmt.Println("starting app containers")
+	if err := r.compose("up", "-d", "myst-provider", "myst-consumer", "ipify"); err != nil {
+		return errors.Wrap(err, "starting app containers failed!")
```
soffokl: `errors.Wrap()` return nil, if `err == nil`, so you can just do:
```go
err := r.compose("up", "-d", "myst-provider", "myst-consumer", "ipify");
return errors.Wrap(err, "starting app containers failed!")
```

bin/tests/e2e.go
```diff
@@ -0,0 +1,155 @@
...
+		"--geth.url=http://geth:8545",
+	)
+	return errors.Wrap(err, "tests failed!")
+}
+
+func (r *runner) cleanup() {
+	fmt.Println("cleaning up")
+	_ = r.compose("logs")
+	if err := r.compose("down", "--volumes", "--remove-orphans", "--timeout", "30"); err != nil {
+		fmt.Printf("warn: cleanup error: %v\n", err)
```
soffokl: Propper logging will help with `warn` here.

bin/tests/test.go
```diff
@@ -0,0 +1,59 @@
...
+)
+
+// Test runs unit tests
+func Test() error {
+	packages, err := unitTestPackages()
+	if err != nil {
+		return err
+	}
+	args := append([]string{"test", "-race", "-timeout", "3m", "-cover"}, packages...)
+	if err := sh.RunV("go", args...); err != nil {
```
soffokl:
```suggestion
	return sh.RunV("go", args...)
```

bin/tests/e2e.go
tadaskay: Yeah, I can, but it is harder to comprehend, and is inconsistent with all the previous statements. I suppose it could be used for short / one-line functions and such.

bin/tests/e2e.go
vkuznecovas: Agreed. Might make sense to just have a mage package.

tadaskay:
Moved to `ci` package and re-used `logconfig`

PR #1105, opened 2019-05-23
Release snapshots with Gitlab

tadaskay:
Ref: #1061

- Snapshot (dev) release only on master branch (TODO tag release). Sample (done on this branch): https://github.com/mysteriumnetwork/node-builds/releases and corresponding pipeline: https://gitlab.com/mysteriumnetwork/node/pipelines/63307181
- Now all of the packages are built on PR (so no surprises on master), unlike in current travis build
- Port most of existing build jobs to mage
- Mage targets (tasks) check of conditions themselves, avoiding coding logic (e.g. _if release then_ ...) in pipeline script, which is CI platform dependent. The drawback is that tasks which are not applicable to the current pipeline (e.g. PR build won't create a bucket and upload artifacts) will not be marked as skipped. Instead, the build log will contain something like "Skipping...". This is a pending feature in Gitlab. We can use it when they add support to mark jobs as skipped: https://gitlab.com/gitlab-org/gitlab-ce/issues/25738
- `GenerateEnvFile` generates bash script to be sourced in later stages. This is a bit futureproof work, so that `BUILD_VERSION` (including timestamp for snapshot build) can be generated once and used everywhere.
- Mage targets are mostly independent from CI platform specific variables. Mapping from CI specific vars to the ones used by mage targets, should occur in pipeline script `.gitlab-ci.yml` variables section, e.g. `BUILD_COMMIT: $CI_COMMIT_SHORT_SHA`
- Snapshot builds are published to a separate repo, specified by env var `GITHUB_SNAPSHOT_REPO`
- Remove travis dev release task (desktop trigger will be broken for a while, subject to creating a web runner)
- Most of the mage targets can be reused for other projects. Once polished and finalized, I'll move most of the stuff to `go-ci` repo
- The next PR will add proper versioning for snapshot build (for now, it's mosty a port of old behavior, `0.0.0-dev-sha` - only appending a commit sha to the end)


ci/packages/package.go
```diff
@@ -0,0 +1,254 @@
...
+)
+
+// PackageLinuxAmd64 builds and stores linux amd64 package
+func PackageLinuxAmd64() error {
+	logconfig.Bootstrap()
+	defer log.Flush()
+	if err := packageStandalone("build/myst/myst_linux_amd64", "linux", "amd64"); err != nil {
+		return err
+	}
+	return env.IfRelease(storage.UploadArtifacts)
```
Waldz: Quite strange that packaging script start uploading by magic global constants. 
As we put all artifacts to `build/package` I would call upload tasks from CI

tadaskay: Can you elaborate what do you mean by _call upload tasks from CI_?

tadaskay: [Embedding logic into CI configuration scripts isn't a good idea](https://www.thoughtworks.com/radar/techniques/programming-in-your-ci-cd-tool) and that is what I was trying to fix. 
As a result, builds are fully reproducible locally with a few environment variables set.

Environment variables I think are a reasonable way to pass data into build scripts as I haven't found a better way to do that with mage. Basically there are 2 kinds of them: 
1. Build information (version, build type - PR/Tag/Snapshot) 
2. API tokens and other data used with 3rd party services (Github, S3)

Do you see any problems with this approach?

.gitlab-ci.yml
```diff
@@ -63,142 +73,93 @@ checks:
...
+  stage: release
+  tags: [go]
+  script: bin/release_goreport
+  only:
+    - master
+
+remove-bucket:
+  stage: post-release
+  tags: [go]
+  script: go run mage.go -v RemoveBucket
```
soffokl: What if the process fails on the early stage? Will we remove the bucked in any case? 
Just to make sure that we are not leaving not cleaned bucket. We had an `after_fail` job on Travis to clean it up.

ci/packages/package.go
```diff
@@ -0,0 +1,254 @@
...
+	}
+	if err := saveDockerImage("myst:alpine", "build/docker-images/myst_alpine.tgz"); err != nil {
+		return err
+	}
+	return env.IfRelease(storage.UploadDockerImages)
+}
+
+// PackageDockerUbuntu builds and stores docker ubuntu image
+func PackageDockerUbuntu() error {
+	logconfig.Bootstrap()
```
soffokl: This `logconfig.Bootstrap()` in every function looks a bit strange. Maybe it will be enough to call it once in the package `init()` function?

ci/packages/package.go
```diff
@@ -0,0 +1,254 @@
...
+	defer log.Flush()
```
soffokl: The same with `log.Flush()` maybe there is a way to do it once when mage finish his job?

ci/storage/s3.go
```diff
@@ -0,0 +1,124 @@
...
+// MakeBucket creates a bucket in s3 for the build (env.BuildNumber)
+func MakeBucket() error {
+	logconfig.Bootstrap()
+	defer log.Flush()
+	return env.IfRelease(func() error {
+		url, err := bucketUrlForBuild()
+		if err != nil {
+			return err
+		}
+		return sh.RunV("bin/s3", "mb", url)
```
soffokl: There should be some S3 library for golang, we will need to replace bash scripts by it later.

ci/packages/package.go
soffokl: We probably can do both bootstrap and flush in the `main()` of the `mage.go` file

ci/storage/s3.go
tadaskay: Absolutely, this is just work in progress of porting all build scripts.

.gitlab-ci.yml
tadaskay: Not sure about this one. On one hand, if it fails in `package` stage, it will leave a bucket behind. On the other, a failed pipeline can be retried. If we remove bucket `on_failure`, we lose this functionality, because then it will try to package and upload to an already deleted bucket.

soffokl: I'm worried about disk space on S3. We will definitely have a lot of failed jobs, because of it's CI.
After some time we will have a bunch of uncleaned buckets on S3.

I'm not sure about retries, in most cases, you are doing another commit to fix a problem and trigger another build. A retried job with no changes will fail again.

ci/packages/package.go
tadaskay: The problem with this is that `mage.go` is just a wrapper for mage so it can be used without installing. It will not work for users using mage directly, i.e. if you `mage -v TestE2E` you will lose the logging provided by the targets. `init` would probably work for `logconfig.Bootstrap()`, but there is no way to flush before mage exits. Since mage does not provide any hooks for this, I would probably just leave it as is, explicit.

vkuznecovas: I like it explicit. The only way I can think of to make this less repeated, is add a wrapper function for it:

```golang
func WithInit(f func() error) error {
  logconfig.Bootstrap()
  defer log.Flush()
  return  f()
}
```

But then you'd have to nest everything in a function :) So I like it explicit.

ci/packages/package.go
Waldz: Sorry, I just realised it's upload to S3, not a conditional artifact releasing

PR #1109, opened 2019-05-27
Iptables based non tunneled traffic block for consumer

tadovas:
### **Description:**
This PR introduces feature, which blocks all non-tunneled consumer traffic, with allowed exceptions for services to run mysterium network itself (like ethereum node connection, discovery, nats etc.).
The scope of this PR is to track all the places where external connections are made and make function calls to allow those connections (with possibility to remove those rules later when needed).

### **Default behaviour:**
By default this feature should be as least intrusive as possible, so by default Noop firewall is used with simple logging of request to allow service connections.

### **Basic flow:**
On startup consumer node allows some globally used services by default (oracle, discovery, ethereum network etc.).
During connection initiation - contact by provider (connection to nats server) and later provider ip itself are added to list of allowed services. Somewhere in between a hard "block all outogoing consumer traffic" is applied if specified by connection options (currently DisableKillSwitch: false - however I would change it to more descriptive something like "BlockNonTunneledTraffic")

### **Blocking scopes:**
- *Session* - this scope means that traffic is blocked during session and block is removed when session ends.
- *Global* - this scope means that traffic is blocked globally and overrides any session scope block requests (it means that block by session has no effect). This case is useful when consumer wants to make sure that no traffic ever accidently reaches outside world without connection to some provider - this global block can be called on program startup.

### **Design notes:**
All firewall control functions are global functions defined in firewall package, however they are using package scope variable for actual block calls, which might be configurable anytime (usually at program startup), this approach was taken to make these firewall controls as less intrusive as possible (no depedency injecting into any of components) - more like logging system. By default blocker only logs requests to allow some connections.

*Note*: my suggestion is to allways allow traffic to UDP port 53 for DNS, as it would be very complicated to track which exact DNS addreses are used.

*E2e tests output:*
Screenshot 2019-05-29 at 12 39 05
Here we can see that some services are allowed initially, and some are during session (nats, provider ip address, session block itself) and removed when session ends.


cmd/di.go
```diff
@@ -23,6 +23,8 @@ import (
 	"path/filepath"
 	"time"
 
+	"github.com/mysteriumnetwork/node/firewall"
+
```
Waldz: extra line

core/connection/manager.go
```diff
@@ -417,6 +417,22 @@ func (manager *connectionManager) onStateChanged(state State) {
 	}
 }
 
+func (manager *connectionManager) setupTrafficBlock(disableTrafficBlocking bool) error {
```
Waldz: `DisableKillSwitch`, not worth to have 2 different namings

core/connection/manager.go
```diff
@@ -23,11 +23,12 @@ import (
 	"sync"
 	"time"
 
+	"github.com/mysteriumnetwork/node/firewall"
+
```
Waldz: extra line

core/connection/manager.go
```diff
@@ -417,6 +417,22 @@ func (manager *connectionManager) onStateChanged(state State) {
 	}
 }
 
+func (manager *connectionManager) setupTrafficBlock(disableTrafficBlocking bool) error {
+	if disableTrafficBlocking {
+		return nil
+	}
+
+	removeRule, err := firewall.BlockNonTunnelTraffic(firewall.Session)
```
Waldz: Wanna do it as singleton, does it align with started `firewall.iptablesKillSwitch` struct?

firewall/consumer_traffic_blocker_noop.go
```diff
@@ -0,0 +1,31 @@
+package firewall
+
+import "github.com/cihub/seelog"
```
zolia: we usually alias this one

services/openvpn/client.go
```diff
@@ -62,6 +64,11 @@ func (c *Client) Start(options connection.ConnectOptions) error {
 	}
 	c.process = proc
 	log.Infof("client config: %v", clientConfig)
+	removeAllowedIPRule, err := firewall.AllowIPAccess(clientConfig.VpnConfig.RemoteIP)
```
zolia:
```suggestion
	c.removeAllowedIPRule, err := firewall.AllowIPAccess(clientConfig.VpnConfig.RemoteIP)
```
direct assignment to struct member should be sufficient here..

core/connection/manager.go
tadovas: I am note sure about DisableKillSwitch term - its really vague and doesn't tell anything. Didn't want to modify tequilapi yet, but would go for BlockNonTunneledTraffic property coming from API.

core/connection/manager.go
tadovas: Goland autoimports :(

core/connection/manager.go
tadovas: killSwitch* stuff completely removed

firewall/consumer_traffic_blocker_noop.go
tadovas: will do

services/openvpn/client.go
tadovas: Nope :) if error is returned other return values are undefined according to go. So we can asign removal function only if error was nil.

tadovas:
@zolia @waldz thanks for approval, but still is WIP. Re-request review after its final. And yes - my approach is - define api/usage with this PR. Implement with next PR. Lets keep PRs small!

core/connection/manager.go
Waldz: It's ordinary naming in VPN domain

tadovas: Show me

cmd/di.go
```diff
@@ -493,6 +494,18 @@ func (di *Dependencies) bootstrapNetworkComponents(options node.OptionsNetwork)
 	}
 
 	di.NetworkDefinition = network
+	trustedURLs := []string{
+		network.EtherClientRPC,
+		network.MysteriumAPIAddress,
+		network.QualityOracle,
+	}
+	for _, url := range trustedURLs {
+		_, err = firewall.AllowURLAccess(url)
```
Waldz: Can You implement setter for multiple rules to have thin code here
Or `AllowURLAccess(urls ...string)`

firewall/consumer_traffic_blocker_noop.go
```diff
@@ -0,0 +1,53 @@
...
+// BlockNonTunnelTraffic method simply logs block request
+func (nb NoopBlocker) BlockNonTunnelTraffic(scope Scope) (RemoveRule, error) {
+	log.Info(nb.LogPrefix, "Non tunneled traffic block requested. Scope: ", scope)
+	return nb.logRemoval("Block for scope: ", scope, " removed"), nil
+}
+
+// AllowURLAccess method logs url for which access was requested
+func (nb NoopBlocker) AllowURLAccess(url string) (RemoveRule, error) {
+	log.Info(nb.LogPrefix, "Allow ", url, " access")
+	return nb.logRemoval("Rule for ", url, " removed"), nil
```
Waldz: Lets remove error from signature if we don't have errors and simpler callers.
Anyway errors happens then starting firewall with added rules.

cmd/di.go
tadovas: It's the single place. Is it worth it?

firewall/consumer_traffic_blocker_noop.go
tadovas: Errors also happen when adding rule to a firewall. Noop implementation doesn't mean that errors cannot occur on real ones. And it will with iptables

cmd/di.go
tadovas: fixed

core/connection/manager.go
tadovas: anyway left old naming

cmd/di.go
tadovas: OK, challenge accepted

tadaskay:
Thanks for the comprehensive PR description. It makes reading and analyzing so much easier! This should be a template for every PR!

PR #1110, opened 2019-05-27
Proposals can be filtered with custom reducers

Waldz:
*Expected behaviour:*
- Proposal storage filtering now can be done with `ProposalReducer` callbacks
- `ProposalReducer` is simple callback which is run agains list of proposals
- Proposal storage filtering is performed with `discovery.Finder` which has interface: 
```
type ProposalMatcher interface {
	GetProposal(id market.ProposalID) (*market.ServiceProposal, error)
	MatchProposals(match ProposalReducer) ([]market.ServiceProposal, error)
}
discoveryFinder := discovery.NewFinder(discovery.NewStorage())
```
- Most simple noop reducer:
``` 
discoveryFinder.MatchProposals(
  reducer.All(),
)
```
- By proposal field:
``` 
discoveryFinder.MatchProposals(
  reducer.Equal(reducer.ProviderID, "0x1"),
)
```
- By custom callback:
``` 
discoveryFinder.MatchProposals(
  reducer.Equal(reducer.ProviderID, "0x1"),
)
```
- By several conditions:
```
discoveryFinder.MatchProposals(reducer.And(
    reducer.Equal(reducer.ProviderID, "0x1"),
    reducer.In(reducer.ServiceType, "openvpn", "wireguard")),
    reducer.AccessPolicy("mysterium", ""),
))
```
- By several conditions:
```
discoveryFinder.MatchProposals(reducer.Or(
    reducer.Equal(reducer.LocationCountry, "US")),
    reducer.Equal(reducer.LocationCountry, "CA")),
))
```

*Changes made:*
- Depreciated `discovery/api.Finder` all proposals are filtered with `ProposalMatcher.MatchProposals()` now
- Filtering of Tequila[i purposes (dapp) moved to reducer `endpoints.proposalsFilter` which made as reducer
- Depreciated `discovery/api.finderAPI`, it still supports both: API finding and reducing callbacks


core/discovery/api/finder.go
```diff
@@ -50,12 +57,12 @@ func (finder *finderAPI) GetProposal(id market.ProposalID) (*market.ServicePropo
...
-// FindProposals fetches currently active service proposals from discovery
-func (finder *finderAPI) FindProposals(filter market.ProposalFilter) ([]market.ServiceProposal, error) {
-	return finder.mysteriumAPI.QueryProposals(mysterium.ProposalsQuery{
-		NodeKey:            filter.ProviderID,
-		ServiceType:        filter.ServiceType,
-		AccessPolicyID:     filter.AccessPolicy.ID,
-		AccessPolicySource: filter.AccessPolicy.Source,
-	})
+// FindProposals fetches currently active service proposals from discovery by given filter
+func (finder *finderAPI) FindProposals(filter discovery.ProposalFilter) ([]market.ServiceProposal, error) {
```
vkuznecovas: why not take in FilterQueryable interface as a param instead of having to cast it inside?

core/discovery/interface.go
```diff
@@ -29,8 +29,22 @@ type ProposalRegistry interface {
...
+
+// ProposalMatcher defines interface to fetch currently active service proposals from discovery by match function
+type ProposalMatcher interface {
+	GetProposal(id market.ProposalID) (*market.ServiceProposal, error)
+	MatchProposals(match ProposalReducer) ([]market.ServiceProposal, error)
+}
+
+// ProposalFilter defines interface with proposal match function
+type ProposalFilter interface {
+	Matches(proposal market.ServiceProposal) bool
```
vkuznecovas: can we just define this as a func type instead of an interface?

core/discovery/reducer/field.go
```diff
@@ -0,0 +1,81 @@
...
+			return valueTyped == byte(0)
+		case bool:
+			return valueTyped == false
+		case []string:
+		case []int:
+		case []byte:
+		case []bool:
+			return len(valueTyped) == 0
+		default:
+			panic(fmt.Sprintf("unsuported field type: %s", valueTyped))
```
vkuznecovas: do we really want to stop working if a field is unsupported? Log this explicitly and continue?

core/discovery/reducer/proposal.go
```diff
@@ -0,0 +1,84 @@
...
+		// These proposals accepts all access lists
+		if proposal.AccessPolicies == nil {
+			return false
+		}
+
+		var match bool
+		for _, policy := range *proposal.AccessPolicies {
+			match = (id == "" || policy.ID == id) &&
+				(source == "" || policy.Source == source)
+			if match == true {
```
vkuznecovas: if match {}

tequilapi/endpoints/proposals_filter.go
```diff
@@ -0,0 +1,65 @@
...
+package endpoints
+
+import (
+	"github.com/mysteriumnetwork/node/core/discovery/reducer"
+	"github.com/mysteriumnetwork/node/market"
+	"github.com/mysteriumnetwork/node/market/mysterium"
+)
+
+// proposalsFilter defines all flags for proposal filtering in discovery of Mysterium Network
+type proposalsFilter struct {
```
vkuznecovas: This is a weird component. 

I can create it once, but change its state after creation, which will affect the .Matches behaviour. I suggest adding a constructor and making the fields private. That way, I'll know that the behaviour won't change.

core/discovery/api/finder.go
Waldz: Because `finderAPI` and `finderStorage` implements same `ProposalFinder` interface and  just several filters are serialisable for `finderAPI`.
Anyway it's temporarily and we'll drop `finderAPI` soon

core/discovery/interface.go
Waldz: `ProposalReducer` is function, and `ProposalFilter` interface to record which filter was used in tests

Waldz: Anyways, I improved it now by moving interfaces to caller

core/discovery/api/finder.go
Waldz: Anyways, I improved it now by moving interfaces to caller side

tequilapi/endpoints/proposals_filter.go
Waldz: Made attributes private

core/discovery/reducer/field.go
tadovas: What if critical code depends on this? You pass a field which is not supported - that means development bug, I agree with panic.

core/discovery/reducer/all.go
```diff
@@ -0,0 +1,29 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package reducer
```
tadovas: According to functional programming terms, reducer is an operator which takes two or more values and returns a single one. Sum or multiplication are reducers. I.e. list (1 ,2 ,3 ,4).Reduce(Sum) = 10. What your code does here is simply filtering collection values according to some predicates. Maybe it's worth to call that package - predicate/filter?

Issue #897, opened 2019-04-19
Version dev builds


tadaskay:
It's hard to determine node version used from logs when all of them are `0.0.0-dev`. We should version dev builds with timestamp and git hash.

tadovas:
It's already there. /healthcheck endpoint contains all data about version

soffokl:
`myst -v` should work too.

tadaskay:
Yeah, it does print 0.0.0-dev, not exact version

soffokl:
`0.0.0-dev` is a binary from the master branch, there is no release. https://github.com/mysteriumnetwork/node/releases/tag/0.0.0-dev
```
Mysterium Node
  Version: 0.0.0-dev
  Build info: Branch: 0.0.0-dev. Build id: 4539. Commit: 418588165d0e6b3b20cda1e45e9c8864db14abda

```

Released versions include the correct version of the release. 
And both of them includes also commit ID and build number.

tadaskay:
Correct, this ticket is about versioning dev builds so that it's possible to trace exact version from the logs. Not:

```
2019-04-20T14:00:45.466098 [Info] Starting Mysterium Node (0.0.0-dev)
```

But for instance:

```
2019-04-20T14:00:45.466098 [Info] Starting Mysterium Node (0.9.0-dev-20190402-418588165)
```

Where `0.9.0` is the last release version, `dev` is an indicator this is a master branch build, followed by some kind of a timestamp and a git commit sha.

soffokl:
If was fixed in the https://github.com/mysteriumnetwork/node/pull/1022

tadaskay:
It partially fixes the issue, I think we should still version the builds properly and move releases to another repo (e.g. `node-dev-builds`)

PR #1111, opened 2019-05-28
Version master builds (PPA style)

tadaskay:
Fixes: #897 
Ref: #1061


ci/env/generate.go
```diff
@@ -62,6 +61,66 @@ func GenerateEnvFile() error {
...
 
+func isTag() bool {
+	return os.Getenv("BUILD_TAG") != ""
+}
+
+func isSnapshot() bool {
+	return os.Getenv("BUILD_BRANCH") == "master" && !isTag()
+}
+
+func isPr() bool {
```
soffokl: [Words in names that are initialisms or acronyms (e.g. "URL" or "NATO") have a consistent case. For example, "URL" should appear as "URL" or "url" (as in "urlPony", or "URLPony"), never as "Url". ](https://github.com/golang/go/wiki/CodeReviewComments#initialisms)
```suggestion
func isPR() bool {
```

ci/github/release.go
```diff
@@ -35,7 +35,8 @@ type Releaser struct {
 
 // Release represents a Github release
 type Release struct {
-	Id int64
+	Id      int64
```
soffokl: [Words in names that are initialisms or acronyms (e.g. "URL" or "NATO") have a consistent case. For example, "URL" should appear as "URL" or "url" (as in "urlPony", or "URLPony"), never as "Url". ](https://github.com/golang/go/wiki/CodeReviewComments#initialisms)
```suggestion
	ID      int64
```

tadaskay: Thanks, didn't know that

PR #1125, opened 2019-05-29
Feature/websockets

Nordwhale:


websocket/websocket.go
```diff
@@ -0,0 +1,116 @@
...
+	State string `json:"state"`
+	Some  string
+}
+
+func NewWebSocketServer() WebSocket {
+	webSocket := WebSocket{}
+	webSocket.connections = make(map[*net.Conn]bool)
+	webSocket.Actions = make(chan Action)
+	go webSocket.ListenActions()
+	Instance = webSocket
```
ignasbernotas: You should avoid global variables.

websocket/websocket.go
```diff
@@ -0,0 +1,116 @@
...
+func (webSocket WebSocket) ListenActions() {
+	for {
+		action := <-webSocket.Actions
+		if action.Type != "" {
+			webSocket.Dispatch(action)
+		}
+	}
+}
+
+type Test struct {
```
ignasbernotas: Unused?

tadaskay:
It would be great to get rid of those merges and rebase instead.

websocket/websocket.go
ignasbernotas: You need to get rid of the Instance variable. The instance shouldn't be accessed as a global var, but instead, passed around through function arguments.

ignasbernotas: @Nordwhale

tadovas:
Websockets API as all external APIs requires careful design as it will be hard to do changes later. Is there any partical reason why it's needed ASAP?

cmd/di.go
```diff
@@ -23,6 +23,8 @@ import (
 	"path/filepath"
 	"time"
 
+	"github.com/mysteriumnetwork/node/websocket"
+
```
vkuznecovas: redundant whitespace

cmd/di.go
```diff
@@ -188,6 +190,8 @@ type Dependencies struct {
 	BandwidthTracker *bandwidth.Tracker
 
 	UIServer UIServer
+
+	WebSocket websocket.WebSocket
```
vkuznecovas: this should most definitely be a pointer.

cmd/di.go
```diff
@@ -410,6 +414,10 @@ func (di *Dependencies) bootstrapNodeComponents(nodeOptions node.Options, listen
 	tequilapi_endpoints.AddRoutesForPayout(router, di.IdentityManager, di.SignerFactory, di.MysteriumAPI)
 	tequilapi_endpoints.AddRoutesForAccessPolicies(router, nodeOptions.AccessPolicyEndpointAddress)
 	tequilapi_endpoints.AddRoutesForNAT(router, di.NATStatusTracker.Status)
+
+	di.WebSocket = websocket.NewWebSocketServer()
```
vkuznecovas: could wrap these two into a separate boostrapWebsocket method.

core/service/pool.go
```diff
@@ -21,6 +21,8 @@ import (
 	"errors"
 	"sync"
 
+	"github.com/mysteriumnetwork/node/websocket"
+
```
vkuznecovas: redundant whitespace

core/service/pool.go
```diff
@@ -175,3 +177,22 @@ func (i *Instance) Proposal() market.ServiceProposal {
...
+	Status     string `json:"status"`
+	ProviderID string `json:"providerId"`
+	Type       string `json:"type"`
+}
+
+// SetState - Set instance state and send new state to websocket
+func (i *Instance) SetState(state State) {
+	i.state = state
+	proposal := i.Proposal()
+	websocket.Instance.ServiceUpdateStatusAction(instanceSocket{
```
vkuznecovas: This needs to be reworked. Including a global variable here is definitely a non-solution. Hook up the websockets to service start/stop events via the EventBus and avoid adding dependencies where they are not needed. 

Can't remember if the events are already being published or not - either way - add the publishing part if required.

identity/selector/handler_interface.go
```diff
@@ -19,7 +19,7 @@ package selector
 
 import "github.com/mysteriumnetwork/node/identity"
 
-// Handler allows selecting identity to be used
+// Handler interface
```
vkuznecovas: why change this?

websocket/websocket.go
```diff
@@ -0,0 +1,116 @@
...
+	"github.com/gobwas/ws"
+	"github.com/gobwas/ws/wsutil"
+	socketio "github.com/googollee/go-socket.io"
+	"github.com/julienschmidt/httprouter"
+)
+
+const serviceUpdateStatus = "server/SERVICE_UPDATE_STATUS"
+
+// Instance of websocket server
+var Instance WebSocket
```
vkuznecovas: definitely get rid of the global variable.

websocket/websocket.go
```diff
@@ -0,0 +1,116 @@
...
+
+// WebSocket struct
+type WebSocket struct {
+	Server      *socketio.Server
+	connections map[*net.Conn]bool
+	actions     chan action
+}
+
+// ServiceUpdateStatusAction - send new service instance status to websocket
+func (webSocket WebSocket) ServiceUpdateStatusAction(payload interface{}) {
```
vkuznecovas: You'll want to use a pointer receiver here. Otherwise you're pushing to a copy of the channel, not the channel you want to push to, see https://play.golang.org/p/KrNAwzLLWt6

websocket/websocket.go
```diff
@@ -0,0 +1,116 @@
...
+	Type    string      `json:"type"`
+	Payload interface{} `json:"payload"`
+}
+
+// AddRoutesForWebSocket - add websockets to routes
+func AddRoutesForWebSocket(router *httprouter.Router, ws WebSocket) {
+	router.GET("/ws/", ws.handler)
+}
+
+func (webSocket WebSocket) handler(writer http.ResponseWriter, request *http.Request, params httprouter.Params) {
```
vkuznecovas: pointer receiver here as well.

websocket/websocket.go
```diff
@@ -0,0 +1,116 @@
...
+			if header.OpCode == ws.OpClose {
+				_ = log.Error("[Websocket] delete connection. reason: CLOSE")
+				delete(webSocket.connections, &conn)
+			}
+		}
+	}()
+
+}
+
+func (webSocket WebSocket) dispatch(action action) {
```
vkuznecovas: pointer receiver here as well.

websocket/websocket.go
```diff
@@ -0,0 +1,116 @@
...
+		log.Info("[Websocket] send to %v data=%v", *conn, string(bytes))
+		err := wsutil.WriteServerText(*conn, bytes)
+
+		if err != nil {
+			_ = log.Errorf("[Websocket] Error send to %v  data = %v  Error: %v", *conn, string(bytes), err)
+		}
+	}
+}
+
+func (webSocket WebSocket) listenActions() {
```
vkuznecovas: pointer receiver here as well.

websocket/websocket.go
```diff
@@ -0,0 +1,116 @@
...
+	for {
```
vkuznecovas: so this is an infinite loop. You should never start an for without a way of exiting it. Implement a graceful stop.

core/service/manager.go
```diff
@@ -21,6 +21,8 @@ import (
 	"encoding/json"
 	"errors"
 
+	"github.com/mysteriumnetwork/node/websocket"
```
tadovas: Websockets are for exposing external api like rest, it should not leak to other components. We are using internal event publisher/subscriber for events. Websockets should subscribe to interested events and broadcast them to outside world only.

core/service/pool.go
```diff
@@ -175,3 +176,22 @@ func (i *Instance) Proposal() market.ServiceProposal {
 func (i *Instance) State() State {
 	return i.state
 }
+
+type instanceSocket struct {
```
tadovas: Name doesn't reflect purpose of this structure

core/service/pool.go
```diff
@@ -175,3 +176,22 @@ func (i *Instance) Proposal() market.ServiceProposal {
...
+	Id         string `json:"id"`
+	Status     string `json:"status"`
+	ProviderID string `json:"providerId"`
+	Type       string `json:"type"`
+}
+
+// SetState - Set instance state and send new state to websocket
+func (i *Instance) SetState(state State, socket websocket.WebSocket) {
```
tadovas: What is the purpose of this function?

tequilapi/endpoints/service.go
```diff
@@ -59,10 +61,11 @@ type accessPoliciesRequest struct {
 }
 
 // swagger:model ServiceListDTO
-type serviceList []serviceInfo
+type serviceList []ServiceInfo
```
tadovas: serviceInfo is private for a reason. Why export is needed?

tequilapi/http_middlewares.go
```diff
@@ -52,7 +52,8 @@ func allowCorsActions(resp http.ResponseWriter, req *http.Request, corsPolicy Co
 	allowedOrigin := corsPolicy.AllowedOrigin(requestOrigin)
 
 	resp.Header().Set("Access-Control-Allow-Origin", allowedOrigin)
-	resp.Header().Set("Access-Control-Allow-Methods", "POST, GET, OPTIONS, PUT, DELETE")
+	resp.Header().Set("Access-Control-Allow-Methods", "POST, GET, HEAD, OPTIONS, PUT, DELETE")
+	resp.Header().Set("Access-Control-Allow-Credentials", "true")
```
tadovas: Reason for this?

cmd/di.go
```diff
@@ -398,18 +407,22 @@ func (di *Dependencies) bootstrapNodeComponents(nodeOptions node.Options, listen
...
+
 	tequilapi_endpoints.AddRouteForStop(router, utils.SoftKiller(di.Shutdown))
 	tequilapi_endpoints.AddRoutesForIdentities(router, di.IdentityManager, di.IdentitySelector)
 	tequilapi_endpoints.AddRoutesForConnection(router, di.ConnectionManager, di.IPResolver, di.StatisticsTracker, di.DiscoveryFinder)
 	tequilapi_endpoints.AddRoutesForConnectionSessions(router, di.SessionStorage)
 	tequilapi_endpoints.AddRoutesForConnectionLocation(router, di.ConnectionManager, di.LocationResolver)
 	tequilapi_endpoints.AddRoutesForLocation(router, di.LocationResolver)
 	tequilapi_endpoints.AddRoutesForProposals(router, di.DiscoveryFinder, di.MysteriumMorqaClient)
-	tequilapi_endpoints.AddRoutesForService(router, di.ServicesManager, serviceTypesRequestParser, nodeOptions.AccessPolicyEndpointAddress)
+	tequilapi_endpoints.AddRoutesForService(router, di.ServicesManager, serviceTypesRequestParser, nodeOptions.AccessPolicyEndpointAddress, di.WebSocket)
```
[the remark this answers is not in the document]
Nordwhale: because in this way we no need another separate port. websockets works on the same port as tequila api. and also websockets are only needs for the Client App as a tequila api too. that's why I've integrated it in tequila api router.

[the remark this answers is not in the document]
tadovas: My point was: tequilapi_endpoints.AddRoutesForWebsocket(di.Websocket) - configures the same router but separate function only

core/service/pool.go
Nordwhale: Set instance state and send new state to websocket

core/service/pool.go
Nordwhale: how you will name it ?

PR #1127, opened 2019-06-03
More proposal filters

Waldz:


core/discovery/reducer/in.go
```diff
@@ -0,0 +1,58 @@
...
+package reducer
+
+import (
+	"github.com/mysteriumnetwork/node/market"
+)
+
+// In matches proposal if field value exists in array
+func In(field FieldSelector, valuesExpected ...interface{}) func(market.ServiceProposal) bool {
+	if len(valuesExpected) == 0 {
+		return All()
```
soffokl: Shouldn't we return none if there are no expected values?

core/discovery/reducer/in.go
```diff
@@ -0,0 +1,58 @@
...
+			if value == valueExpected {
+				return true
+			}
+		}
+		return false
+	})
+}
+
+// InString matches proposal if string value exists in array
+func InString(field FieldSelector, valuesExpected ...string) func(market.ServiceProposal) bool {
```
soffokl: `In` function already receiving `valuesExpected ...interface{}`, we can pass string and ints with it. No need to have extra functions for `int` and `string`.

core/discovery/reducer/in.go
```diff
@@ -0,0 +1,58 @@
...
+	valuesExpectedTyped := make([]interface{}, len(valuesExpected))
+	for i, value := range valuesExpected {
+		valuesExpectedTyped[i] = value
+	}
+
+	return In(field, valuesExpectedTyped...)
+}
+
+// InInt matches proposal if integer value exists in array
+func InInt(field FieldSelector, valuesExpected ...int) func(market.ServiceProposal) bool {
```
soffokl: `In` function already receiving `valuesExpected ...interface{}`, we can pass string and ints with it. No need to have extra functions for `int` and `string`.

core/discovery/reducer/mocks.go
```diff
@@ -80,3 +80,11 @@ func conditionIsProvider1(provider market.ServiceProposal) bool {
 func conditionIsStreaming(provider market.ServiceProposal) bool {
 	return provider.ServiceType == "streaming"
 }
+
+func fieldID(proposal market.ServiceProposal) interface{} {
```
soffokl: Why do we returning `interface{}` here? Don't we have a predefined type for `ID`?

core/discovery/reducer/mocks.go
```diff
@@ -80,3 +80,11 @@ func conditionIsProvider1(provider market.ServiceProposal) bool {
...
+	return proposal.ID
+}
+
+func fieldProviderID(proposal market.ServiceProposal) interface{} {
```
soffokl: Why do we returning `interface{}` here? Don't we have a predefined type for `ProviderID`?

core/discovery/reducer/in.go
Waldz: No, it means filter is not filled. Later need to introduce different factories e.g. `In(field, value, isFilled)`

core/discovery/reducer/in.go
Waldz: No, no generics in Golang. This is why I created this PR - because noticed bug in live

core/discovery/reducer/mocks.go
Waldz: Basically because no generics in Golang

core/discovery/reducer/in.go
soffokl: If this is a bug, we can have a test that reproduces it. 
Because now I can change the tests to `In` only, and it keeps working:
image

core/discovery/reducer/in.go
Waldz: Removed this. So that fields would be checked caller side

core/discovery/reducer/in.go
Waldz: It's impossible then I have array `providerIDs []string` -> `In(fieldProviderID, providerIDs...)`

core/discovery/reducer/in.go
```diff
@@ -0,0 +1,54 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package reducer
+
+import (
+	"github.com/mysteriumnetwork/node/market"
+)
+
+// In matches proposal if field value exists in array
```
zolia: Such documentation makes it hard to understand what these functions do. Since most of them does nothing, just returns other function, which does actual job. I would suggest fixing docs as follows: 
```suggestion
// In returns a matcher for checking if proposal's field value exists in given array of values
```

core/discovery/reducer/in.go
```diff
@@ -0,0 +1,54 @@
...
+}
+
+// InString matches proposal if string value exists in array
+func InString(field FieldSelector, valuesExpected ...string) func(market.ServiceProposal) bool {
+	valuesExpectedTyped := make([]interface{}, len(valuesExpected))
+	for i, value := range valuesExpected {
+		valuesExpectedTyped[i] = value
+	}
+
+	return In(field, valuesExpectedTyped...)
```
zolia: In what cases will we use `In` method as a public method? Maybe we can make it private?

core/discovery/reducer/in_test.go
```diff
@@ -0,0 +1,51 @@
...
+
+package reducer
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func Test_In(t *testing.T) {
```
zolia: basically it's the same test as `Test_InString` do we really need it public?

Waldz: Well `In()` is for any typed fields, also for complex structs

tadaskay:
Please provide a PR description (perhaps the bug that you are attempting to fix).

core/discovery/reducer/in_test.go
zolia: This test then should reflect complex struct.

core/discovery/reducer/in.go
Waldz: It's for more generic cases. Added documentation

core/discovery/reducer/in_test.go
Waldz: They do now

Waldz:
Feedback addressed

PR #1128, opened 2019-06-03
e2e traversal env simulation

zolia:
Env to execute e2e tests for provider / consumer behind NAT. Actually working setup.

Reusing most of the localnet code. The same basic connect sequence tests works with this env too. 

Future work: 
 - reuse the same network env to spawn different consumers / providers ( on public net, android, windows etc )
 - remove outdated bash scripts
 - integrate compatibility matrix into the same env and rewrite to mage
 - get rid of ipify and use location oracle for real simulation. Change 172.30.x.x fake public IPs to real ones to work with public resolve service


core/discovery/discovery.go
```diff
@@ -22,6 +22,7 @@ import (
 	"time"
 
 	log "github.com/cihub/seelog"
+
```
tadovas: Extra space?

e2e/config/named.conf
```diff
@@ -0,0 +1,47 @@
+options {
```
tadovas: maybe simple dnsmasq daemon would be enough?

e2e/traversal/deployer/deployer.go
```diff
@@ -0,0 +1,94 @@
+/*
```
tadovas: A copy of localnet file?

e2e/traversal/deployer/keystore/UTC--2018-07-25T13-08-10.138246914Z--a754f0d31411d88e46aed455fa79b9fced122497
```diff
@@ -0,0 +1 @@
+{"address":"a754f0d31411d88e46aed455fa79b9fced122497","crypto":{"cipher":"aes-128-ctr","ciphertext":"56c3a9b187b9acfe64d9d9ab2f0236f990f533f99ddb05cc0d5be9340f7fb755","cipherparams":{"iv":"fa76b9c835859593db14cd4820078b88"},"kdf":"scrypt","kdfparams":{"dklen":32,"n":262144,"p":1,"r":8,"salt":"3dfdd74c4e7eb4d6ef3b449b836438f3082a34bd0c6dc82a85bde4f30421e1b9"},"mac":"97c1323a22cddb1dac9652a1b3b6b07706193a12cf056256dee963647686cde7"},"id":"965998e2-2a62-4aae-9b0d-b3a723b8d521","version":3}
```
tadovas: Another copy of a file of localnet?

e2e/traversal/deployer/local_acc_password.txt
```diff
@@ -0,0 +1 @@
+localaccount
```
tadovas: Another copy of a file of localnet?

e2e/traversal/docker-compose.yml
```diff
@@ -0,0 +1,271 @@
+version: '2.4'
```
tadovas: Too much copying

e2e/traversal/functions.sh
```diff
@@ -0,0 +1,93 @@
+#!/usr/bin/env bash
```
tadovas: I have seen this somewhere else

tadovas:
Ok label says its work in progress :) lets wait and see how it will be polished ;)

bin/docker/alpine/Dockerfile
```diff
@@ -15,11 +15,10 @@ ADD . .
 RUN bin/build
 
 
-
 FROM alpine:3.6
 
 # Install packages
-RUN apk add --update --no-cache iptables ca-certificates openvpn bash sudo \
+RUN apk add --update --no-cache iptables ca-certificates openvpn bash sudo curl \
```
soffokl: I'm not sure that we need to add extra packages to the main image. If curl required somewhere else maybe create a new image for it?

core/discovery/discovery.go
```diff
@@ -93,6 +94,7 @@ func NewService(
 
 // Start launches discovery service
 func (d *Discovery) Start(ownIdentity identity.Identity, proposal market.ServiceProposal) {
+	log.Info(logPrefix, "starting discovery..")
```
soffokl:
```suggestion
	log.Info(logPrefix, "starting discovery...")
```

e2e/config/named.conf
zolia: yea, tested, it works with dnsmasq too. Since its smaller footprint, will switch to it. Thanx.

bin/helpers/prepare-run-env.sh
```diff
@@ -1,5 +1,27 @@
 #!/usr/bin/env bash
 
+setDefaultRoute() {
```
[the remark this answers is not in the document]
zolia: Good question, while doing it I tried to construct those more as a configurable features, not as e2e only variables. IMHO you can try to think of a situation as user might want it to use it as I do in e2e env (for example user might want to route all traffic through its own shaper). Also we avoid cooking separate images.

core/discovery/discovery.go
zolia: thats import grouping..

e2e/traversal/deployer/deployer.go
zolia: made to reuse one from localnet

bin/docker/alpine/Dockerfile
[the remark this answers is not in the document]
zolia: removed, used it for testing.

bin/docker/alpine/Dockerfile
```diff
@@ -1,7 +1,7 @@
 FROM golang:1.11-alpine AS builder
 
 # Install packages
-RUN apk add --update --no-cache bash gcc musl-dev make \
+RUN apk add --update --no-cache bash gcc musl-dev make conntrack-tools tcpdump \
```
Waldz: Is it needed here?

zolia: removed

ci/test/e2e.go
```diff
@@ -118,16 +141,29 @@ func (r *runner) setup() error {
...
 }
 
 func (r *runner) test() error {
-	log.Info("running tests")
+	log.Info("running tests for env: ", r.testEnv)
+
+	var consumerServices string
+
+	if r.testEnv == "basic" {
+		consumerServices = "openvpn,noop,wireguard"
```
vkuznecovas: pass these as a constructor arg to the runner itself? no need to check it here then.

bin/localnet/functions.sh
```diff
@@ -27,8 +24,16 @@ setupInfra () {
         exit 1
     fi
 
+    if [[ ${NET_ENV} == "localnet" ]]; then
+        SERVICES="broker geth"
+    fi
+
+    if [[ ${NET_ENV} == "traversal" ]]; then
+        SERVICES="broker geth ipify"
```
Waldz: ipify is so much needed everythere

ci/test/e2e.go
```diff
@@ -118,16 +141,29 @@ func (r *runner) setup() error {
 	if err := r.compose("build"); err != nil {
 		return errors.Wrap(err, "building app images failed!")
 	}
+
+	return nil
+}
+
+func (r *runner) setupBasicEnv() error {
```
tadaskay: What does _basic_ mean? What about `startAppContainers` or `startProviderConsumerNodes`?

ci/test/e2e.go
```diff
@@ -24,43 +24,66 @@ import (
...
+	runner, err := prepareTestRunner(composeFiles, "node_e2e_nat_test")
+	if err != nil {
+		return err
+	}
+	defer runner.cleanup()
+	return runner.test()
 }
 
-func (r *runner) setup() error {
+func (r *runner) setupCoreEnv() error {
```
tadaskay: What does _coreEnv_ mean? What about `startDependencies`/`setupNodeDependencies`?

ci/test/e2e.go
```diff
@@ -118,16 +141,29 @@ func (r *runner) setup() error {
 	if err := r.compose("build"); err != nil {
```
tadaskay: This should probably go to the other setup method where we start app containers.

bin/localnet/functions.sh
zolia: true, tried removing it, but locationOracle depends on maxmind and it does not resolve internal IPs. Probably I can remake e2e testing using public IPs for containers, but thats in separate PR if really required.

PR #1129, opened 2019-06-04
sse implemented for service status and nat events

### LINKED ISSUE #735: SSE service for pushing updates to node consumer apps
We need a way to receive status changes, statistics, other data without polling via TequilApi since it's ineffective and requires a lot of boilerplate code.

AC:
  * [ ] Start WS server in node
  * [ ] Send service status
  * [ ] Send NAT status

vkuznecovas:
Closes #735 

This is ready to be merged, will improve testing in a separate PR since this is most likely blocking UI development.

### DIFF
--- a/core/service/cleaner.go
+++ b/core/service/cleaner.go
@@ -28,6 +28,8 @@ type SessionStorage interface {
 }
 
 // Cleanup removes sessions of stopped service
-func (cleaner *Cleaner) Cleanup(instance EventPayload) {
-	cleaner.SessionStorage.RemoveForService(instance.ID)
+func (cleaner *Cleaner) Cleanup(event EventPayload) {
+	if event.Status == string(NotRunning) {
+		cleaner.SessionStorage.RemoveForService(event.ID)
+	}
 }
--- a/core/service/cleaner.go
+++ b/core/service/cleaner.go
@@ -27,8 +27,8 @@ type SessionStorage interface {
 	RemoveForService(serviceID string)
 }
 
-// Cleanup removes sessions of stopped service
-func (cleaner *Cleaner) Cleanup(event EventPayload) {
+// HandleServiceStatus removes sessions of stopped service
+func (cleaner *Cleaner) HandleServiceStatus(event EventPayload) {
 	if event.Status == string(NotRunning) {
 		cleaner.SessionStorage.RemoveForService(event.ID)
 	}
--- a/core/service/manager.go
+++ b/core/service/manager.go
@@ -30,9 +30,6 @@ import (
 	"github.com/mysteriumnetwork/node/session"
 )
 
-// StopTopic is used in event bus to announce that service was stopped
-const StopTopic = "Service stop"
-
 var (
 	// ErrorLocation error indicates that action (i.e. disconnect)
 	ErrorLocation = errors.New("failed to detect service location")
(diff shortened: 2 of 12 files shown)


core/service/pool.go
```diff
@@ -173,5 +178,25 @@ func (i *Instance) Proposal() market.ServiceProposal {
 
 // State returns the service instance state.
 func (i *Instance) State() State {
+	i.stateLock.RLock()
+	defer i.stateLock.RUnlock()
 	return i.state
 }
+
+func (i *Instance) setState(newState State) {
```
tadovas: It's a bit weird that state switch is not related to event publishing. Because this is exact place where events happen.  I would expect some kind of SetState or SwitchState which will publish even immediately after state switch.

tequilapi/endpoints/sse.go
```diff
@@ -0,0 +1,34 @@
...
+)
+
+// SSEHandler represents the sse handler
+type SSEHandler interface {
+	Sub(resp http.ResponseWriter, req *http.Request, params httprouter.Params)
+}
+
+// AddRoutesForSSE adds route for sse
+func AddRoutesForSSE(router *httprouter.Router, handler SSEHandler) {
+	router.GET("/sse", handler.Sub)
```
tadovas: SSE is content type, I would prefer more descriptive url like "/events", "/service-states" or something similar

core/service/manager.go
```diff
@@ -145,14 +144,15 @@ func (manager *Manager) Start(providerID identity.Identity, serviceType string,
 	manager.servicePool.Add(&instance)
 
 	go func() {
-		instance.state = Running
+		instance.setState(Running)
+
+		manager.eventPublisher.Publish(StartTopic, instance.ToEvent())
```
Waldz: Shouldn't event publish happen inside state change. It's also with lock

core/service/pool.go
```diff
@@ -173,5 +178,25 @@ func (i *Instance) Proposal() market.ServiceProposal {
...
+	i.stateLock.Lock()
+	defer i.stateLock.Unlock()
+	i.state = newState
+}
+
+// ToEvent returns an event representation of the instance
+func (i *Instance) ToEvent() EventPayload {
+	i.stateLock.RLock()
+	defer i.stateLock.RUnlock()
+	return EventPayload{
```
Waldz: I think we can send session instance, anyways it's for internal usage

core/service/pool.go
vkuznecovas: moved publish to setState

tequilapi/endpoints/sse.go
vkuznecovas: renamed to /events

core/service/cleaner.go
```diff
@@ -24,10 +24,12 @@ type Cleaner struct {
...
 // SessionStorage keeps sessions and allows removing them by proposal id
 type SessionStorage interface {
-	RemoveForService(serviceId string)
+	RemoveForService(serviceID string)
 }
 
 // Cleanup removes sessions of stopped service
-func (cleaner *Cleaner) Cleanup(instance *Instance) {
-	cleaner.SessionStorage.RemoveForService(string(instance.id))
+func (cleaner *Cleaner) Cleanup(event EventPayload) {
```
zolia: Its probably not a `Cleanup` then, since imho its should be called only once and we don't care about event's state.

PR #1131, opened 2019-06-06
Introduce QualityOracle options in 'myst'

Waldz:
*Expected behaviour:*
- All nodes who are interested to filter proposals by quality metrics can use their chosen QualityOracle
- Also nodes their quality metrics to chosen QualityOracle 
- In this way MysteriumDiscovery proposals can be extended with any custom metrics (throuput stats, connect fails, location stats, etc.)

*Changes made:*
- Quality metrics sender is now configurable with myst flag `--quality.type`
- 2 different QualityOracle providers introduced: `morqa` and `none`
- QualityOracle provider can be chosen by `myst` arguments `--quality.type` and `--quality.address`
- `--quality.type=morqa` enables reporting metrics to Mysterium MORQA. This option will also allow to filter proposal quality metrics from MORQA
- `--quality.type=none` opt-outs from sending usage metrics

Screen Shot 2019-06-06 at 15 25 56


cmd/flags_quality.go
```diff
@@ -0,0 +1,59 @@
...
+)
+
+var (
+	qualityTypeFlag = cli.StringFlag{
+		Name: "quality.type",
+		Usage: fmt.Sprintf(
+			"Quality Oracle provider (%s, %s - %s)",
+			node.QualityTypeMORQA,
+			node.QualityTypeNone,
+			"opt-out from sending usage metrics",
```
tadaskay: Quality service is being referred to in 3 different ways:
- Quality type
- Quality Oracle provide
- Usage (usage metrics)

We should align and use unified wording for this service.

Furthermore, I don't think `quality.type` conveys the meaning well enough. It's more like `quality.provider`.

cmd/flags_quality.go
```diff
@@ -0,0 +1,59 @@
...
+
+	"github.com/mysteriumnetwork/node/core/node"
+	"github.com/urfave/cli"
+)
+
+var (
+	qualityTypeFlag = cli.StringFlag{
+		Name: "quality.type",
+		Usage: fmt.Sprintf(
+			"Quality Oracle provider (%s, %s - %s)",
```
tadaskay: Let's try to keep the same option enumeration format as in e.g. `services/openvpn/service/options.go:36`

cmd/di.go
```diff
@@ -610,9 +610,8 @@ func (di *Dependencies) bootstrapQualityComponents(options node.OptionsQuality)
 }
 
 func (di *Dependencies) bootstrapLocationComponents(options node.OptionsLocation, configDirectory string) (err error) {
-	_, err = firewall.AllowURLAccess(options.IPDetectorURL)
-	if err != nil {
-		return err
+	if _, err = firewall.AllowURLAccess(options.IPDetectorURL); err != nil {
+		return
```
tadaskay: I thought we agreed to avoid naked returns. It's bad for readability. Now I have to go back to the function signature to check what's needs to be returned and then scan all the lines to determine what's actually being returned.

cmd/di.go
```diff
@@ -605,24 +624,27 @@ func (di *Dependencies) bootstrapLocationComponents(options node.OptionsLocation
...
 
 	err = di.EventBus.SubscribeAsync(connection.StateEventTopic, di.LocationResolver.HandleConnectionEvent)
 	if err != nil {
-		return err
+		return
 	}
 
 	err = di.EventBus.SubscribeAsync(nodevent.Topic, di.LocationResolver.HandleNodeEvent)
 	if err != nil {
-		return err
```
soffokl: My linter is not happy about it :(
> naked return in func `bootstrapLocationComponents` with 39 lines of code (nakedret)

cmd/flags_quality.go
Waldz: - Renamed to `Quality Oracle adapter`
- removed `usage metrics`

Issue #1113, opened 2019-05-29
Improve raspberry image build time


tadaskay:
It should be possible to cache first stages and not rebuild raspbian from scratch every time

PR #1134, opened 2019-06-09
Configure raspberrypi image using existing raspbian-stretch-lite

tadaskay:
Fixes: #1113 
Ref: #1061

Use an existing raspbian image and install package onto it instead of rebuilding raspbian from scratch with https://github.com/RPi-Distro/pi-gen

## Design notes

- Raspbian image is fetched from `raspbian` bucket and cache is size-based
- `systemd-nspawn` is used to create a lightweight container without booting it, then `bin/package/raspberry/files/setup.sh` is executed inside the container to set it up. Also tried https://github.com/proot-me/proot which didn't work.
- `ci/util/shell/cmd.go` small wrapper around mage's `sh` created to be able to pass the whole command string instead of splitting it. This is much easier to troubleshoot with shell: just copy cmd and execute (and replace `%s` with actual value in case of `Cmdf`)
- added a dependency on `aws-sdk-go-v2`, should still refactor the rest of build scripts to use it. For now it's just raspberry task that utilizes it.

## TODO

- pass env vars to systemd container
- change default username to `mystberry`

### DIFF
--- a/bin/package/raspberry/files/default-myst-conf
+++ b/bin/package/raspberry/files/default-myst-conf
@@ -1,0 +1,6 @@
+# Define additional args for `myst` service (see `myst --help` for full list)
+CONF_DIR="--config-dir=/etc/mysterium-node"
+RUN_DIR="--runtime-dir=/var/run/mysterium-node"
+DATA_DIR="--data-dir=/var/lib/mysterium-node"
+DAEMON_OPTS="--tequilapi.address=0.0.0.0"
+SERVICE_OPTS="openvpn --access-policy.list mysterium"
--- a/ci/packages/raspberry.go
+++ b/ci/packages/raspberry.go
@@ -1,0 +1,183 @@
+/*
+ * Copyright (C) 2019 The "MysteriumNetwork/node" Authors.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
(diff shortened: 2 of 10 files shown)


bin/package/raspberry/files/default-myst-conf
```diff
@@ -0,0 +1,6 @@
+# Define additional args for `myst` service (see `myst --help` for full list)
```
soffokl: If we will delete `bin/package/raspberry/myststage`, GIT will track these files as renamed, but not new ones. 
We probably not need them anymore.

ci/packages/raspberry.go
```diff
@@ -0,0 +1,183 @@
...
+	if err != nil {
+		return err
+	}
+	defer func() {
+		if err := device.DetachLoop(loopDevice); err != nil {
+			log.Warn(err)
+		}
+	}()
+
+	_ = os.MkdirAll(raspbianMountPoint, 755)
```
soffokl: > SA9002: file mode '755' evaluates to 01363; did you mean '0755'? (staticcheck)

ci/packages/raspberry.go
```diff
@@ -0,0 +1,183 @@
...
+	if err := shell.NewCmdf("sudo mount --options rw %s %s", loopDevice+"p2", raspbianMountPoint).Run(); err != nil {
+		return err
+	}
+	_ = os.MkdirAll(raspbianMountPoint+"/boot", 755)
```
soffokl: > SA9002: file mode '755' evaluates to 01363; did you mean '0755'? (staticcheck)

bin/package/raspberry/files/default-myst-conf
tadaskay: Release build is not using gitlab yet, so I'm leaving it not to break compatibility.

PR #1136, opened 2019-06-10
Feature/sse with full state

### LINKED ISSUE #1130: Emit state change events through sse
So we've finished the initial sse implementation with nat/service-status events.

The issue is that end consumers still need to load the initial state via rest, which then introduced the problem of race conditions between rest requests and sse events.

So @ignasbernotas @tadaskay @zolia and me had a short discussion how to alleviate the issue.

The things we've considered:

1) Emitting a "current" state event when a new consumer arrives. This was discarded as we'd introduce additional complexity by having to store the state and handle the events on top of it.
1) Building an internal list of events, which would then be replayed on new consumers arrival. This was discarded due to the fact that our services are long-running, which means we'd have to store and replay insane amounts of events in the long run.

However, we've dreamed up a solution where many of us seemed to agree. Which is to create a "State" object in memory on node side and emit it via sse and add a REST endpoint for it. On connect, the user would get a snapshot of the whole state, which would then get emitted once a change occurs. This way, the consumer would consistently know about the node's state.

vkuznecovas:
We're now keeping track of whole state and emitting it through sse.

Closes #1130

### DIFF
--- a/cmd/di.go
+++ b/cmd/di.go
@@ -250,6 +250,11 @@ func (di *Dependencies) Bootstrap(nodeOptions node.Options) error {
 	di.bootstrapNATComponents(nodeOptions)
 	di.bootstrapServices(nodeOptions)
 	di.bootstrapStateKeeper(nodeOptions)
+
+	if err := di.bootstrapSSEHandler(); err != nil {
+		return err
+	}
+
 	di.bootstrapNodeComponents(nodeOptions, tequilaListener)
 
 	di.registerConnections(nodeOptions)
@@ -309,7 +314,7 @@ func (di *Dependencies) registerNoopConnection() {
 
 // bootstrapSSEHandler bootstraps the SSEHandler and all of its dependencies
 func (di *Dependencies) bootstrapSSEHandler() error {
-	di.SSEHandler = sse.NewHandler()
+	di.SSEHandler = sse.NewHandler(di.StateKeeper)
 	err := di.EventBus.Subscribe(nodevent.Topic, di.SSEHandler.ConsumeNodeEvent)
 	if err != nil {
 		return err
--- a/cmd/di.go
+++ b/cmd/di.go
@@ -275,7 +275,7 @@ func (di *Dependencies) bootstrapStateKeeper(options node.Options) error {
 
 	tracker := nat.NewStatusTracker(lastStageName)
 
-	di.StateKeeper = state.NewKeeper(tracker, di.EventBus, di.ServicesManager, di.ServiceSessionStorage)
+	di.StateKeeper = state.NewKeeper(tracker, di.EventBus, di.ServicesManager, di.ServiceSessionStorage, state.DefaultDebounceDuration)
 
 	err := di.EventBus.SubscribeAsync(service.StatusTopic, di.StateKeeper.ConsumeServiceStateEvent)
 	if err != nil {
--- a/cmd/di.go
+++ b/cmd/di.go
@@ -124,12 +124,6 @@ type NatEventSender interface {
 	ConsumeNATEvent(event event.Event)
 }
 
-// NATStatusTracker tracks status of NAT traversal by consuming NAT events
-type NATStatusTracker interface {
-	Status() nat.Status
-	ConsumeNATEvent(event event.Event)
-}
-
 // CacheResolver caches the location resolution results
 type CacheResolver interface {
 	location.Resolver
@@ -185,10 +179,9 @@ type Dependencies struct {
 	ServiceRegistry       *service.Registry
 	ServiceSessionStorage *session.StorageMemory
 
-	NATPinger        NatPinger
-	NATTracker       NatEventTracker
-	NATEventSender   NatEventSender
-	NATStatusTracker NATStatusTracker
+	NATPinger      NatPinger
+	NATTracker     NatEventTracker
+	NATEventSender NatEventSender
 
 	BandwidthTracker *bandwidth.Tracker
 
@@ -279,9 +272,11 @@ func (di *Dependencies) bootstrapStateKeeper(options node.Options) error {
 	} else {
 		lastStageName = mapping.StageName
 	}
+
 	tracker := nat.NewStatusTracker(lastStageName)
 
 	di.StateKeeper = state.NewKeeper(tracker, di.EventBus, di.ServicesManager, di.ServiceSessionStorage)
+
 	err := di.EventBus.SubscribeAsync(service.StatusTopic, di.StateKeeper.ConsumeServiceStateEvent)
 	if err != nil {
 		return err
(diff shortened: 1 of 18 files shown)


cmd/di.go
```diff
@@ -236,13 +234,15 @@ func (di *Dependencies) Bootstrap(nodeOptions node.Options) error {
 		return err
 	}
 
+	di.bootstrapMetrics(nodeOptions)
+	di.bootstrapNATComponents(nodeOptions)
+	di.bootstrapServices(nodeOptions)
+	di.bootstrapStateKeeper(nodeOptions)
```
soffokl: `bootstrapStateKeeper` return error, we need to check it.

core/state/state_test.go
```diff
@@ -0,0 +1,244 @@
...
+func Test_Debounce_CallsOnceInInterval(t *testing.T) {
+	dt := &debounceTester{}
+	duration := time.Microsecond * 1000
+	f := debounce(dt.do, duration)
+	for i := 1; i < 10; i++ {
+		f(struct{}{})
+	}
+
+	select {
+	case <-time.After(duration * 5):
```
soffokl: Select is not needed here:
```go
time.Sleep(duration * 5)
assert.Equal(t, 1, dt.get())
```

core/state/state_test.go
```diff
@@ -0,0 +1,244 @@
...
+	for i := 0; i < 5; i++ {
+		// shoot a few events to see if we'll debounce
+		keeper.ConsumeNATEvent(natEvent.Event{
+			Stage:      "booster separation",
+			Successful: false,
+			Error:      errors.New("explosive bolts failed"),
+		})
+	}
+
+	select {
```
soffokl: Select is not needed here:
```go
time.Sleep(duration * 3)
assert.Equal(t, 1, natProvider.getTimesCalled())
```

core/state/state_test.go
```diff
@@ -0,0 +1,244 @@
...
+	duration := time.Millisecond * 3
+	keeper := NewKeeper(natProvider, publisher, sl, sessionStorage, duration)
+
+	for i := 0; i < 5; i++ {
+		// shoot a few events to see if we'll debounce
+		keeper.ConsumeSessionEvent(sessionEvent.Payload{})
+	}
+
+	select {
+	case <-time.After(duration * 3):
```
soffokl: Select is not needed here:
```go
time.Sleep(duration * 3)
assert.Equal(t, 1, sessionStorage.getTimesCalled())
```

core/state/state_test.go
```diff
@@ -0,0 +1,244 @@
...
+}
+
+func (mp *mockPublisher) Publish(topic string, data interface{}) {
+	mp.lock.Lock()
+	defer mp.lock.Unlock()
+	mp.publishedData = data
+	mp.publishedTopic = topic
+}
+
+func (mp *mockPublisher) getLast() (string, interface{}) {
```
soffokl: `getLast` is unused.

session/event/event.go
```diff
@@ -0,0 +1,37 @@
...
+
+package event
+
+// Topic represents the session change topic
+const Topic = "Session change"
+
+// Action represents the different actions that might happen on a session
+type Action string
+
+var (
```
soffokl:
```suggestion
const (
```

session/storage_memory_test.go
```diff
@@ -111,9 +113,78 @@ func TestStorage_Remove_Does_Not_Panic(t *testing.T) {
...
+	mp := &mockPublisher{}
+	sessionStore := NewStorageMemory(mp)
+
+	sessionStore.Add(instance)
+
+	// since we're shooting the event in an asynchronous fashion, try every millisecond to see if we already have it
+	attempts := 0
+	// functions are really just easier to exit from than nested blocks
+	func() {
+		for {
```
soffokl: This can be a `for` with `range`: 
```go
	for range time.After(time.Microsecond) {
		if attempts > 1000 {
			assert.Fail(t, "no change after a 1000 attempts")
			break
		}
		attempts++
		if mp.getLast().Action == sessionEvent.Created {
			break
		}
	}
```

session/storage_memory_test.go
```diff
@@ -111,9 +113,78 @@ func TestStorage_Remove_Does_Not_Panic(t *testing.T) {
...
+	time.Sleep(time.Millisecond * 5)
+
+	sessionStore.Remove(instance.ID)
+
+	// since we're shooting the event in an asynchronous fashion, try every microsecond to see if we already have it
+	attempts := 0
+
+	// functions are really just easier to exit from than nested blocks
+	func() {
+		for {
```
soffokl: The same here with `range`

tequilapi/sse/sse_test.go
```diff
@@ -84,3 +77,64 @@ func TestHandler_ConsumeNodeEvent_Starts(t *testing.T) {
...
+	go func() {
+		err = http.Serve(listener, router)
+		serveExit <- err
+	}()
+
+	time.Sleep(time.Millisecond * 50)
+	w := fmt.Sprintf("http://127.0.0.1:%v/whatever", port)
+	req, _ := http.NewRequest("GET", w, nil)
+	ctx, cancel := context.WithCancel(context.Background())
+	req.WithContext(ctx)
```
soffokl: > SA4017: WithContext is a pure function but its return value is ignored (staticcheck)
```suggestion
	req = req.WithContext(ctx)
```

vkuznecovas: good catch!

session/storage_memory_test.go
```diff
@@ -111,9 +113,72 @@ func TestStorage_Remove_Does_Not_Panic(t *testing.T) {
...
+	instance := expectedSession
+	mp := &mockPublisher{}
+	sessionStore := NewStorageMemory(mp)
+
+	sessionStore.Add(instance)
+
+	// since we're shooting the event in an asynchronous fashion, try every millisecond to see if we already have it
+	attempts := 0
+	// functions are really just easier to exit from than nested blocks
+	func() {
```
soffokl: This `func` is not needed with `for`, you can just `break` it.

session/storage_memory_test.go
```diff
@@ -111,9 +113,72 @@ func TestStorage_Remove_Does_Not_Panic(t *testing.T) {
...
+
+	time.Sleep(time.Millisecond * 5)
+
+	sessionStore.Remove(instance.ID)
+
+	// since we're shooting the event in an asynchronous fashion, try every microsecond to see if we already have it
+	attempts := 0
+
+	// functions are really just easier to exit from than nested blocks
+	func() {
```
soffokl: This `func` is not needed with `for`, you can just `break` it.

core/state/state.go
```diff
@@ -0,0 +1,187 @@
...
+		serviceLister:         serviceLister,
+		serviceSessionStorage: serviceSessionStorage,
+	}
+	k.ConsumeServiceStateEvent = debounce(k.updateServiceState, debounceDuration)
+	k.ConsumeNATEvent = debounce(k.updateNatStatus, debounceDuration)
+	k.ConsumeSessionEvent = debounce(k.updateSessionState, debounceDuration)
+	return k
+}
+
+func (k *Keeper) updateServiceState(e interface{}) {
```
zolia:
```suggestion
func (k *Keeper) updateServiceState(_ interface{}) {
```

PR #1142, opened 2019-06-11
Iptables based firewall for non tunneled traffic blocking

tadovas:
#### Description
This PR enables to effectively control (block if needed) outgoing consumer traffic. It's currently implemented on linux only, by using iptables. However it can be easily extended by adding simple vendor interface implementations for other OSes.
Iptables blocking is realised the following way:

- New chain CONSUMER_KILLSWITCH is created. By default it rejects all the traffic with ICMP port unreachable. Why not DROP? Because since its outgoing traffic, a better errors are returned for process if its outgoing send is failing (and its important too, to detect which services from our side needed to be added to exceptions)
```bash
-A CONSUMER_KILL_SWITCH -m conntrack --ctstate NEW -j REJECT
```
Only NEW connections are blocked since we need incoming connections (like tequila api) to work as expected
- Any exceptions if needed are added to CONSUMER_KILLSWITCH with ACCEPT target.
- Traffic blocking itself is enabled/disabled by adding/removing rule to OUTPUT chain by 
```bash
-A OUTPUT -s <detected outbound ip> -j CONSUMER_KILL_SWITCH
```

#### Basic flow:
Myst daemon started with flags:
--firewall.killSwitch - enable firewall control of outgoing traffic
--firewall.killSwitch.always - always block non-tunneled outgoind traffic (firewall.BlockNonTunneledTraffic(firewall.Global)
First - no connection made to provider:
Screenshot 2019-06-12 at 10 05 27
Then - succesfull connection was established:
Screenshot 2019-06-12 at 10 06 59
_Note_: DNS is always passed through at the moment, since our services heavily depend on it. If it becomes a problem - firewall functionality can be extended to allow exception by port, not by ip only.
### Design notes:
- **trackingBlocker** is a firewall object which provides functionality to all other components. It tracks which ips were added as exceptions and how many times to avoid "overlapping" addtions/removals, and calls underlying firewall vendor when needed.
- **vnd** package is name is ugly, but I could not use vendor since this name is reserverd. It contains compile conditional code to provide different versions of "SetupVendor" for different OSes to initialize actual firewall vendor.

### DIFF
--- a/cmd/di.go
+++ b/cmd/di.go
@@ -375,7 +375,7 @@ func (di *Dependencies) Shutdown() (err error) {
 			errs = append(errs, err)
 		}
 	}
-
+	firewall.Reset()
 	log.Flush()
 	return nil
 }
--- a/cmd/di.go
+++ b/cmd/di.go
@@ -23,8 +23,6 @@ import (
 	"path/filepath"
 	"time"
 
-	"github.com/mysteriumnetwork/node/firewall/vnd"
-
 	log "github.com/cihub/seelog"
 	"github.com/ethereum/go-ethereum/accounts/keystore"
 	"github.com/ethereum/go-ethereum/common"
@@ -54,6 +52,7 @@ import (
 	"github.com/mysteriumnetwork/node/core/storage/boltdb/migrations/history"
 	"github.com/mysteriumnetwork/node/eventbus"
 	"github.com/mysteriumnetwork/node/firewall"
+	"github.com/mysteriumnetwork/node/firewall/vnd"
 	"github.com/mysteriumnetwork/node/identity"
 	identity_registry "github.com/mysteriumnetwork/node/identity/registry"
 	identity_selector "github.com/mysteriumnetwork/node/identity/selector"
@@ -740,9 +739,7 @@ func (di *Dependencies) bootstrapFirewall(options node.OptionsFirewall) error {
 	}
 	if options.BlockAlways {
 		_, err := firewall.BlockNonTunnelTraffic(firewall.Global)
-		if err != nil {
-			return err
-		}
+		return err
 	}
 	return nil
 }
--- a/cmd/di.go
+++ b/cmd/di.go
@@ -23,6 +23,8 @@ import (
 	"path/filepath"
 	"time"
 
+	"github.com/mysteriumnetwork/node/firewall/vnd"
+
 	log "github.com/cihub/seelog"
 	"github.com/ethereum/go-ethereum/accounts/keystore"
 	"github.com/ethereum/go-ethereum/common"
@@ -52,7 +54,6 @@ import (
 	"github.com/mysteriumnetwork/node/core/storage/boltdb/migrations/history"
 	"github.com/mysteriumnetwork/node/eventbus"
 	"github.com/mysteriumnetwork/node/firewall"
-	"github.com/mysteriumnetwork/node/firewall/iptables"
 	"github.com/mysteriumnetwork/node/identity"
 	identity_registry "github.com/mysteriumnetwork/node/identity/registry"
 	identity_selector "github.com/mysteriumnetwork/node/identity/selector"
@@ -668,7 +669,7 @@ func (di *Dependencies) bootstrapLocationComponents(options node.OptionsLocation
 	case node.LocationTypeMMDB:
 		resolver, err = location.NewExternalDBResolver(filepath.Join(configDirectory, options.Address), di.IPResolver)
 	case node.LocationTypeOracle:
-		if _, err = firewall.AllowURLAccess(options.Address); err != nil {
+		if _, err := firewall.AllowURLAccess(options.Address); err != nil {
 			return err
 		}
 		resolver, err = location.NewOracleResolver(options.Address), nil
@@ -731,15 +732,11 @@ func (di *Dependencies) bootstrapNATComponents(options node.Options) {
 
 func (di *Dependencies) bootstrapFirewall(options node.OptionsFirewall) error {
 	if options.EnableKillSwitch {
-		ip, err := ip.GetOutbound()
+		fwVendor, err := vnd.SetupVendor()
 		if err != nil {
 			return err
 		}
-		killSwitchVendor := iptables.New(ip)
-		if err := killSwitchVendor.Setup(); err != nil {
-			return err
-		}
-		firewall.Configure(killSwitchVendor)
+		firewall.Configure(fwVendor)
 	}
 	if options.BlockAlways {
 		_, err := firewall.BlockNonTunnelTraffic(firewall.Global)
--- a/cmd/di.go
+++ b/cmd/di.go
@@ -52,6 +52,7 @@ import (
 	"github.com/mysteriumnetwork/node/core/storage/boltdb/migrations/history"
 	"github.com/mysteriumnetwork/node/eventbus"
 	"github.com/mysteriumnetwork/node/firewall"
+	"github.com/mysteriumnetwork/node/firewall/iptables"
 	"github.com/mysteriumnetwork/node/identity"
 	identity_registry "github.com/mysteriumnetwork/node/identity/registry"
 	identity_selector "github.com/mysteriumnetwork/node/identity/selector"
@@ -222,6 +223,10 @@ func (di *Dependencies) Bootstrap(nodeOptions node.Options) error {
 	}
 
 	if err := nodeOptions.Directories.Check(); err != nil {
+		return err
+	}
+
+	if err := di.bootstrapFirewall(nodeOptions.Firewall); err != nil {
 		return err
 	}
 
@@ -722,4 +727,25 @@ func (di *Dependencies) bootstrapNATComponents(options node.Options) {
 	loader := &upnp.GatewayLoader{}
 	go loader.Get()
 	di.NATEventSender = event.NewSender(di.QualityMetricsSender, di.IPResolver.GetPublicIP, loader.HumanReadable)
+}
+
+func (di *Dependencies) bootstrapFirewall(options node.OptionsFirewall) error {
+	if options.EnableKillSwitch {
+		ip, err := ip.GetOutbound()
+		if err != nil {
+			return err
+		}
+		killSwitchVendor := iptables.New(ip)
+		if err := killSwitchVendor.Setup(); err != nil {
+			return err
+		}
+		firewall.Configure(killSwitchVendor)
+	}
+	if options.BlockAlways {
+		_, err := firewall.BlockNonTunnelTraffic(firewall.Global)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
 }
--- a/cmd/di.go
+++ b/cmd/di.go
@@ -730,13 +730,11 @@ func (di *Dependencies) bootstrapNATComponents(options node.Options) {
 }
 
 func (di *Dependencies) bootstrapFirewall(options node.OptionsFirewall) error {
-	if options.EnableKillSwitch {
-		fwVendor, err := vnd.SetupVendor()
-		if err != nil {
-			return err
-		}
-		firewall.Configure(fwVendor)
+	fwVendor, err := vnd.SetupVendor()
+	if err != nil {
+		return err
 	}
+	firewall.Configure(fwVendor)
 	if options.BlockAlways {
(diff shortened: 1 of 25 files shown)


firewall/iptables/iptables.go
```diff
@@ -0,0 +1,197 @@
...
+}
+
+// New initializes and returns Iptables with defined outboundIP
+func New(outboundIP string) *Iptables {
+	return &Iptables{
+		outboundIP: outboundIP,
+	}
+}
+
+// BlockOutgoingTraffic starts blocking outgoing traffic and returns functiont to remove block
```
soffokl:
```suggestion
// BlockOutgoingTraffic starts blocking outgoing traffic and returns function to remove the block
```

firewall/iptables/iptables.go
```diff
@@ -0,0 +1,197 @@
...
+	destinationIP = "-d"
+	module        = "-m"
+
+	protocol        = "-p"
+	tcp             = "tcp"
+	udp             = "udp"
+	destinationPort = "--dport"
+
+	conntrack    = "conntrack"
+	ct_state     = "--ctstate"
```
soffokl: > don't use underscores in Go names; const ct_state should be ctState (golint)

firewall/iptables/iptables.go
```diff
@@ -0,0 +1,197 @@
...
+	ct_state_new = "NEW"
```
soffokl: > don't use underscores in Go names; const ct_state_new should be ctStateNew (golint)

firewall/iptables/iptables.go
```diff
@@ -0,0 +1,197 @@
...
+
+	reject = "REJECT"
+	accept = "ACCEPT"
+
+	version = "--version"
+)
+
+var iptablesExec = func(args ...string) ([]string, error) {
```
soffokl: Why not to use just named function? 
```suggestion
func iptablesExec(args ...string) ([]string, error) {
```

firewall/iptables/iptables_test.go
```diff
@@ -0,0 +1,181 @@
+/*
+ * Copyright (C) 2018 The "MysteriumNetwork/node" Authors.
```
soffokl:
```suggestion
 * Copyright (C) 2019 The "MysteriumNetwork/node" Authors.
```

firewall/iptables/iptables_test.go
```diff
@@ -0,0 +1,181 @@
...
+import (
+	"fmt"
+	"net"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestBlockerSetupIsSuccesful(t *testing.T) {
```
soffokl:
```suggestion
func TestBlockerSetupIsSuccessful(t *testing.T) {
```

firewall/tracking_blocker_test.go
```diff
@@ -0,0 +1,134 @@
+/*
+ * Copyright (C) 2018 The "MysteriumNetwork/node" Authors.
```
soffokl:
```suggestion
 * Copyright (C) 2019 The "MysteriumNetwork/node" Authors.
```

firewall/tracking_blocker_test.go
```diff
@@ -0,0 +1,134 @@
...
+func (mv *mockedVendor) AllowIPAccess(ip string) (RemoveRule, error) {
+	return mv.increaseRef("allow:" + ip)
+}
+
+func (mockedVendor) Reset() {
+
+}
+
+func (mv *mockedVendor) increaseRef(ref string) (RemoveRule, error) {
+	mv.requests[ref] = mv.requests[ref] + 1
```
soffokl:
```suggestion
	mv.requests[ref] += 1
```

firewall/tracking_blocker_test.go
```diff
@@ -0,0 +1,134 @@
...
+	return mv.decreaseRef(ref), nil
+}
+
+func (mv *mockedVendor) decreaseRef(ref string) RemoveRule {
+	return func() {
+		mv.requests[ref] = mv.requests[ref] - 1
```
soffokl:
```suggestion
		mv.requests[ref] += - 1
```

firewall/vnd/setup_vendor.go
```diff
@@ -0,0 +1,31 @@
+//+build !linux
+
+/*
+ * Copyright (C) 2018 The "MysteriumNetwork/node" Authors.
```
soffokl:
```suggestion
 * Copyright (C) 2019 The "MysteriumNetwork/node" Authors.
```

nat/upnp/discover.go
```diff
@@ -67,6 +69,12 @@ func discoverGateways() ([]GatewayDevice, error) {
 		return nil, err
 	}
 	defer client.Close()
+
+	// this IP address used to discover gateways - allow it
+	if _, err := firewall.AllowIPAccess("239.255.255.250"); err != nil {
```
soffokl: Maybe allow all multicast addresses `224.0.0.0/4`? UI discovery that we are using uses it too.

cmd/flags_firewall.go
```diff
@@ -0,0 +1,47 @@
...
+
+import (
+	"github.com/mysteriumnetwork/node/core/node"
+	"github.com/urfave/cli"
+)
+
+var (
+	enableKillSwitch = cli.BoolFlag{
+		Name:  "firewall.killSwitch",
+		Usage: "Enable consumer outgoing non tunneled traffic blocking during connections",
```
zolia: Do we really need as a flag here? IMHO it is a sensible out of the box behaviour. You launch VPN - everything goes through it. This flag would only puzzle our users.

cmd/di.go
```diff
@@ -691,3 +697,20 @@ func (di *Dependencies) bootstrapNATComponents(options node.Options) {
...
+	if options.EnableKillSwitch {
+		fwVendor, err := vnd.SetupVendor()
+		if err != nil {
+			return err
+		}
+		firewall.Configure(fwVendor)
+	}
+	if options.BlockAlways {
+		_, err := firewall.BlockNonTunnelTraffic(firewall.Global)
+		if err != nil {
```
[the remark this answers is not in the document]
tadovas: 👍

cmd/di.go
```diff
@@ -23,6 +23,8 @@ import (
 	"path/filepath"
 	"time"
 
+	"github.com/mysteriumnetwork/node/firewall/vnd"
+
```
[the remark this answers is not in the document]
tadovas: Shredded to non-existense

nat/upnp/discover.go
tadovas: UI discovery needs to do extra call itself where its needed

firewall/consumer_traffic_blocker.go
```diff
@@ -17,62 +17,36 @@
 
 package firewall
 
-// RemoveRule type defines function for removal of created rule
-type RemoveRule func()
+var trackingBlocker = newTrackingBlocker()
```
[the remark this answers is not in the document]
tadovas: Design notes: trackingBlocker

firewall/consumer_traffic_blocker.go
```diff
@@ -17,62 +17,36 @@
...
-	for _, url := range urls {
-		remover, err := currentBlocker.AllowURLAccess(url)
-		if err != nil {
-			removeAll()
-			return nil, err
-		}
-		ruleRemovers = append(ruleRemovers, remover)
-	}
-	return removeAll, nil
+func AllowURLAccess(URLs ...string) (RemoveRule, error) {
```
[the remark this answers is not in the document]
tadovas: Fixed

firewall/iptables/iptables.go
tadovas: To be able to replace it in Tests with mocked one.

core/ip/resolver.go
```diff
@@ -25,9 +25,12 @@ type Resolver interface {
 	GetOutboundIP() (string, error)
 }
 
+// delcared as var for override in test
```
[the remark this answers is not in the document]
tadovas: Fixed

firewall/consumer_traffic_blocker.go
```diff
@@ -17,62 +17,36 @@
...
+	return trackingBlocker.AllowURLAccess(URLs...)
 }
 
 // AllowIPAccess adds IP based exception to underlying blocker implementation
 func AllowIPAccess(ip string) (RemoveRule, error) {
```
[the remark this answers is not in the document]
tadovas: Explained in #1109 Design notes, first section

firewall/iptables/iptables.go
```diff
@@ -0,0 +1,197 @@
...
+	args = append([]string{"/sbin/iptables"}, args...)
```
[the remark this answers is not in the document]
tadovas: there is no such wrapper. What do you mean?

cmd/flags_firewall.go
tadovas: Its still experimental. I wrote args for you and @Waldz, actually I am ok to remove it ant leave Noop for other OSes.

firewall/iptables/iptables.go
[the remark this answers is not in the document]
tadovas: Ok it seems that on docker we don't need sudo at all. On debian package /sbin/iptables are already on sudoers list. So "sudo" can safely be removed

firewall/iptables/iptables.go
tadovas: Fixed

firewall/iptables/iptables.go
tadovas: Fixed

firewall/tracking_blocker_test.go
tadovas: 👍

firewall/tracking_blocker_test.go
tadovas: 👍

cmd/flags_firewall.go
Waldz: Lets remove experimental flag

firewall/consumer_traffic_blocker.go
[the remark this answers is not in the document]
Waldz: As OS does not changes, you configure it once and why not to call vendor implementation directly?
I mean absolutely all the functions is proxying and doing nothing

firewall/iptables/iptables.go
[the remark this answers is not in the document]
Waldz: @zolia You implemented `sudoers` feature. It will not allow this command in .deb package, correct?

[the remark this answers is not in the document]
zolia: @Waldz even if you list command in sudoers file you still need to prefix it with sudo prefix.. you just don't need a root password anymore.

firewall/consumer_traffic_blocker.go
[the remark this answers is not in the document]
tadovas: You mean, remove tracking blocker layer?

firewall/iptables/iptables.go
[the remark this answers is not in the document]
zolia: we created `bin/package/config/linux/nonpriv-ip` because openvpn does expect a single script as  a substitution for ip routing functionality executable. There is no point to create more wrappers here if we can wrap it directly in a go code.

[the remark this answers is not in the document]
tadovas: A side note. /sbin/iptables where used previously in code (for masquarading from provider site) in exactly the same way. I don't see any reason why it should work now (btw it works - tested on docker and linux machines).

firewall/consumer_traffic_blocker.go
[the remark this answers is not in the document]
tadovas: As per our offchain discussion:

> All firewall control functions are global functions defined in firewall package, however they are using package scope variable for actual block calls, which might be configurable anytime (usually at program startup), this approach was taken to make these firewall controls as less intrusive as possible (no depedency injecting into any of components) - more like logging system. By default blocker only logs requests to allow some connections.

cmd/flags_firewall.go
tadovas: Removed

tadovas:
All comments addressed/ issues fixed.

Issue #1117, opened 2019-05-29
Enable basic auth for Web UI


ignasbernotas:
We need a way to secure the raspberry UI

PR #1144, opened 2019-06-12
Add basic auth support to the builtin UI

soffokl:
Feature: authenticate users to the builtin web UI at `localhost:4449` via http basic. 
Ref: #1117

Decided to ditch system user and simply store credentials programmatically. It would be weird if our app changed system user credentials as well. We might add an option to do so in web UI (might be useful for raspberry users).
Also I was a bit naughty with this one and tried out what DDD would look like in go. It's a small piece of code, but still something towards that direction. I like it better than the manager and factory plague.

## What's done

1. Basic auth for web UI (Thanks @soffokl )
2. Store credentials in `myst.db` and use them for verification

## Still left to do (separate PR)

1. Add an endpoint to change pass

Note to self: after review, do the squash, so there's no trace of PAM libs in gopkg lock.


ui/auth/auth_linux.go
```diff
@@ -0,0 +1,44 @@
...
+	"github.com/msteinert/pam"
+)
+
+type pamAuth struct{}
+
+func newAuth() *pamAuth {
+	return &pamAuth{}
+}
+
+func (pa *pamAuth) Auth(header string) error {
```
soffokl: @tadaskay we just need to replace this method with any approach and it will work.

ui/server.go
```diff
@@ -39,13 +40,29 @@ type Server struct {
 	discovery discovery.LANDiscovery
 }
 
+// Authenticator handles web UI auth
+type Authenticator interface {
+	AuthenticateHttpBasic(header string) error
```
soffokl:
```suggestion
	AuthenticateHTTPBasic(header string) error
```

ui/auth/credentials.go
```diff
@@ -0,0 +1,105 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package auth
+
+import (
+	"github.com/asdine/storm"
```
soffokl: Can we avoid depending on the internal storage `storm` in the UI auth package? We have an interface for it, but still checking internal errors.

ui/auth/credentials.go
```diff
@@ -0,0 +1,105 @@
...
+
+// Credentials verifies/sets user credentials for web UI
+type Credentials struct {
+	username, password string
+	db                 Storage
+}
+
+// storeCredentials - struct stored in DB
+type storeCredentials struct {
+	ID           int
```
soffokl: Do we really need an `ID`, maybe it's enough to use a username as a key?

ui/auth/credentials.go
```diff
@@ -0,0 +1,105 @@
...
+	if err != nil {
+		return errors.Wrap(err, "unable to generate password hash")
+	}
+	sc := &storeCredentials{
+		ID:           credentialsStorageID,
+		Username:     credentials.username,
+		PasswordHash: string(passwordHash),
+	}
+	err = credentials.db.StoreOrUpdate(credentialsStorageBucket, sc)
+	if err != nil {
```
soffokl: `if` block is not needed, it's enough to have only:
```go
return errors.Wrap(err, "unable to set credentials")
```

ui/auth/credentials.go
tadaskay: As we have only one set of credentials, I found it the easiest to always store/update record with `ID=1`. Do you have other thoughts?

soffokl: Hm, for me keeping it as `username:passwordHash` looks simpler, you don't need extra `storeCredentials` struct, and username is a const, so we have only a single user always.

ui/auth/credentials.go
tadaskay: It's not necessary, but it helps readability

soffokl: You have a different approach in the same file: https://github.com/mysteriumnetwork/node/pull/1144/files#diff-aedec7358933457d9e43934d625860a4R72

tadaskay: I fixed that :)
I think it's ok if it's a single return, but when there are more ifs, explicit is nice. That was probably a leftover.

ui/auth/credentials.go
tadaskay: Changed to key value store as discussed

core/storage/boltdb/storage.go
```diff
@@ -40,6 +40,16 @@ func openDB(name string) (*Bolt, error) {
 	return &Bolt{db}, errors.Wrap(err, "failed to open boltDB")
 }
 
+// GetKey gets key value
```
ignasbernotas: GetValue

core/storage/boltdb/storage.go
```diff
@@ -40,6 +40,16 @@ func openDB(name string) (*Bolt, error) {
...
+func (b *Bolt) GetValue(bucket string, key interface{}, to interface{}) error {
+	return b.db.Get(bucket, key, to)
+}
+
+// SetKey sets key value
```
ignasbernotas: SetValue

PR #1145, opened 2019-06-12
Cleanup containers in case of startup failure

zolia:


zolia:
@tadaskay true most likely. Will refactor in separate PR.

PR #1158, opened 2019-06-14
Openvpn server now reports session data transfer statistics

vkuznecovas:
The openvpn service will now publish events on statistics data change. These will trigger an event in the session storage, which will in turn update the required session. It will notify all the subscribers of a session change after that. This will in turn update the state and push the change through the events endpoint via sse.

### DIFF
--- a/core/state/event/event.go
+++ b/core/state/event/event.go
@@ -62,7 +62,7 @@ type ServiceSession struct {
 	// example: 2019-06-06T11:04:43.910035Z
 	CreatedAt time.Time `json:"createdAt"`
 	// example: 12345
-	BytesUp int64 `json:"bytesUp"`
+	BytesOut int64 `json:"bytesOut"`
 	// example: 23451
-	BytesDown int64 `json:"bytesDown"`
+	BytesIn int64 `json:"bytesIn"`
 }
--- a/core/state/event/event.go
+++ b/core/state/event/event.go
@@ -61,4 +61,8 @@ type ServiceSession struct {
 	ConsumerID string `json:"consumerId"`
 	// example: 2019-06-06T11:04:43.910035Z
 	CreatedAt time.Time `json:"createdAt"`
+	// example: 12345
+	BytesUp int64 `json:"bytesUp"`
+	// example: 23451
+	BytesDown int64 `json:"bytesDown"`
 }
--- a/services/openvpn/service/factory.go
+++ b/services/openvpn/service/factory.go
@@ -24,6 +24,7 @@ import (
 	log "github.com/cihub/seelog"
 	"github.com/mysteriumnetwork/go-openvpn/openvpn"
 	"github.com/mysteriumnetwork/go-openvpn/openvpn/middlewares/server/auth"
+	"github.com/mysteriumnetwork/go-openvpn/openvpn/middlewares/server/bytecount"
 	"github.com/mysteriumnetwork/go-openvpn/openvpn/middlewares/state"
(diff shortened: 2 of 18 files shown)


core/state/event/event.go
```diff
@@ -61,4 +61,8 @@ type ServiceSession struct {
 	ConsumerID string `json:"consumerId"`
 	// example: 2019-06-06T11:04:43.910035Z
 	CreatedAt time.Time `json:"createdAt"`
+	// example: 12345
+	BytesUp int64 `json:"bytesUp"`
```
ignasbernotas: I'd call it IncomingBandwidth/OutgoingBandwith or IncomingBytes/OutgoingBytes

session/storage_memory.go
```diff
@@ -23,24 +23,28 @@ import (
...
-		lock:      sync.Mutex{},
-		publisher: publisher,
+func NewStorageMemory(bus bus) *StorageMemory {
+	sm := &StorageMemory{
+		sessions: make(map[ID]Session),
+		lock:     sync.Mutex{},
+		bus:      bus,
 	}
+	// intentionally ignoring the error here, it will only error in case ConsumeDataTransferedEvent is not a function
+	_ = bus.SubscribeAsync(event.DataTransfered, sm.ConsumeDataTransferedEvent)
```
Waldz: Very strange then constructor does something

session/storage_memory.go
```diff
@@ -87,12 +91,34 @@ func (storage *StorageMemory) Find(id ID) (Session, bool) {
 	return Session{}, false
 }
 
+// ConsumeDataTransferedEvent consumes the data transfer event
+func (storage *StorageMemory) ConsumeDataTransferedEvent(e event.DataTransferEventPayload) {
```
Waldz: Can we separate event logic from storage itself like we do for all `Consume..` events

services/openvpn/service/factory.go
```diff
@@ -47,10 +55,24 @@ func NewManager(nodeOptions node.Options,
...
-	sessionValidator := openvpn_session.NewValidator(sessionMap, identity.NewExtractor())
+	clientMap := openvpn_session.NewClientMap(sessionMap)
+
+	sessionValidator := openvpn_session.NewValidator(clientMap, identity.NewExtractor())
+
+	callback := func(sbc bytecount.SessionByteCount) {
+		sessions := clientMap.GetClientSessions(sbc.ClientID)
+		if len(sessions) > 0 {
+			publisher.Publish(event.DataTransfered, event.DataTransferEventPayload{
+				ID:   string(sessions[0]),
```
zolia: Bit awkward here, I understand that in our env single clientID should have at most 1 active session, but we get a first session only here, thus logic here implies that we might have more then one.  Actually i don't know if there might be any situation when this same clientID would have more than 1 session. We probably should fail early if we see that we are adding another session for the same clientID in GetClientSessions.

core/state/event/event.go
soffokl: bytesIn/bytesOut

services/openvpn/service/factory.go
vkuznecovas: i've added a log here, don't think we need to fail here as this situation should not be possible under normal use. Either way - the whole client map thing could use a second look at it when we have the time.

session/storage_memory.go
Waldz: ???

session/storage_memory.go
Waldz: ???

PR #1160, opened 2019-06-14
Metric transport to map events to new QualityOracle

Waldz:
*Changes made:*
- 3 different QualityOracle providers are now supported: `elastic`, `morqa`, and `none`
- `--quality.type=morqa` allows to configure metrics sending to new QualityOracle
- Anyways `--quality.type=elastic` is still a default behavior

### DIFF
--- a/core/quality/morqa_transport.go
+++ b/core/quality/morqa_transport.go
@@ -1,0 +1,66 @@
+/*
+ * Copyright (C) 2019 The "MysteriumNetwork/node" Authors.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package quality
+
+import (
+	"github.com/mysteriumnetwork/metrics"
+)
+
+var (
+	emptyMetric = &metrics.Event{}
+)
+
+// NewMORQATransport creates transport allowing to send events to Mysterium Quality Oracle - MORQA
+func NewMORQATransport(morqaClient *MysteriumMORQA, backupTransport Transport) *morqaTransport {
+	return &morqaTransport{
+		morqaClient:     morqaClient,
+		backupTransport: backupTransport,
+	}
+}
+
+type morqaTransport struct {
+	morqaClient     *MysteriumMORQA
+	backupTransport Transport
+}
+
+func (transport *morqaTransport) SendEvent(event Event) error {
+	if metric := mapEventToMetric(event); metric != emptyMetric {
+		err := transport.morqaClient.SendMetric(metric)
+		if err != nil {
+			return err
+		}
+	}
+
+	return transport.backupTransport.SendEvent(event)
+}
+
+func mapEventToMetric(event Event) *metrics.Event {
+	switch event.EventName {
+	case startupEventName:
+		return &metrics.Event{
+			IsProvider: false,
+			TargetId:   "0x1",
+			Metric: &metrics.Event_VersionPayload{
+				VersionPayload: &metrics.VersionPayload{
+					Version: event.Application.Version,
+				},
+			},
+		}
+	}
+	return emptyMetric
+}
--- a/core/quality/morqa_transport.go
+++ b/core/quality/morqa_transport.go
@@ -18,6 +18,8 @@
 package quality
 
 import (
+	"fmt"
+
 	"github.com/mysteriumnetwork/metrics"
 )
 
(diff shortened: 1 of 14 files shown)


core/quality/morqa_transport.go
```diff
@@ -0,0 +1,63 @@
...
+func (transport *morqaTransport) SendEvent(event Event) error {
+	if metric := mapEventToMetric(event); metric != emptyMetric {
+		return transport.morqaClient.SendMetric(metric)
+	}
+
+	return fmt.Errorf("event not implemented")
+}
+
+func mapEventToMetric(event Event) *metrics.Event {
+	switch event.EventName {
```
soffokl: > singleCaseSwitch: should rewrite switch statement to if statement (gocritic)

core/quality/morqa_transport.go
```diff
@@ -0,0 +1,63 @@
...
+	case startupEventName:
+		return &metrics.Event{
+			IsProvider: false,
```
soffokl: The default value is `false`, we can remove this line.

core/quality/morqa_transport_test.go
```diff
@@ -0,0 +1,140 @@
...
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/golang/protobuf/proto"
+	"github.com/mysteriumnetwork/metrics"
+	"github.com/stretchr/testify/assert"
+)
+
+// func mockMORQATransport() *morqaTransport {
```
soffokl: Maybe remove commented code?

core/quality/mysterium_morqa.go
```diff
@@ -18,45 +18,145 @@
...
 		log.Warn(mysteriumMorqaLogPrefix, "Failed to create proposals metrics request: ", err)
 		return nil
 	}
 
-	var metricsResponse ServiceMetricsResponse
-	err = m.http.DoRequestAndParseResponse(req, &metricsResponse)
+	response, err := m.http.Do(request)
 	if err != nil {
-		log.Warn(mysteriumMorqaLogPrefix, "Failed to request or parse proposals metrics: ", err)
+		log.Warn(mysteriumMorqaLogPrefix, err)
```
soffokl: Let's add some context to the logging, it could be hard to read logs without it.

core/quality/mysterium_morqa.go
```diff
@@ -18,45 +18,145 @@
...
+		return nil
+	}
+	defer response.Body.Close()
+
+	var metricsResponse ServiceMetricsResponse
+	if err = parseResponseJSON(response, &metricsResponse); err != nil {
+		log.Warn(mysteriumMorqaLogPrefix, err)
```
soffokl: Let's add some context to the logging, it could be hard to read logs without it.

core/quality/mysterium_morqa.go
```diff
@@ -18,45 +18,145 @@
...
+		return err
+	}
+
+	response, err := m.http.Do(request)
+	if err != nil {
+		return err
+	}
+	defer response.Body.Close()
+
+	if err = parseResponseError(response); err != nil {
```
soffokl:
```suggestion
	return parseResponseError(response)
```

core/quality/mysterium_morqa.go
```diff
@@ -18,45 +18,145 @@
...
+	"github.com/mysteriumnetwork/metrics"
 )
 
 const (
 	mysteriumMorqaLogPrefix = "[Mysterium.morqa] "
+	mysteriumMorqaAgentName = "goclient-v0.1"
 )
 
+// HttpClient sends actual HTTP requests
+type HttpClient interface {
```
soffokl:
```suggestion
type HTTPClient interface {
```

core/quality/mysterium_morqa.go
```diff
@@ -18,45 +18,145 @@
...
+
+func (m *MysteriumMORQA) newRequestBinary(method, path string, payload proto.Message) (*http.Request, error) {
+	payloadBody, err := proto.Marshal(payload)
+	if err != nil {
+		return nil, err
+	}
+
+	request, err := m.newRequest(method, path, payloadBody)
+	request.Header.Set("Content-Type", "application/octet-stream")
+	return request, nil
```
soffokl:
```suggestion
	return request, err
```

core/quality/morqa_transport.go
zolia: I think for readability is a good think to have it, but for me still question remains - why it is 'false'? Don't we need to send provider startup events?

core/quality/mysterium_morqa.go
```diff
@@ -18,45 +18,145 @@
...
+// HttpClient sends actual HTTP requests
+type HttpClient interface {
+	Do(*http.Request) (*http.Response, error)
+}
+
 // MysteriumMORQA HTTP client for Mysterium QualityOracle - MORQA
 type MysteriumMORQA struct {
-	http                 requests.HTTPTransport
-	qualityOracleAddress string
+	http    HttpClient
```
zolia: why did we need redefine generic requests.HTTPTransport with HTTPClient?

core/quality/mysterium_morqa.go
```diff
@@ -18,45 +18,145 @@
...
+	}
+	return json.Unmarshal(responseJSON, dto)
+}
+
+type errorDTO struct {
+	Message string `json:"message"`
+}
+
+// Sometimes we can get json message with single "message" field which represents error - try to get that
+func parseResponseError(response *http.Response) error {
```
zolia: we just duplicating code here from http_client. Maybe make it more generic or merge with requests package and reuse here?

core/quality/morqa_transport.go
Waldz: Will have more events soon

core/quality/morqa_transport.go
Waldz: Still general startup now, and we dont event know identity at that time. Solving it in next PR

core/quality/mysterium_morqa.go
Waldz: I have little different contract with this API for error messages, no pot reuse it.

Also I remade it with: :
- less code, means much simpler
- less dependencies, this client can be separate lib
- transport is mocked in tests

core/quality/mysterium_morqa.go
Waldz: https://github.com/mysteriumnetwork/node/pull/1160/files/463785708a6afc8344cd0940a41bc69434adf91b#r294191008

Waldz: Yes it has this function duplicated, but this API could have totally different error codes and DTOs also

zolia:
need rebasing.

core/quality/mysterium_morqa.go
Waldz: Logging shouldn't happen lowest level library , I am moving then out to upper layer

PR #1165, opened 2019-06-17
Separate referral Tequilapi endpoint

ignasbernotas:


tequilapi/endpoints/payout.go
```diff
@@ -123,6 +130,57 @@ func (endpoint *payoutEndpoint) UpdatePayoutInfo(resp http.ResponseWriter, reque
...
+//   schema:
+//     $ref: "#/definitions/ReferralInfoDTO"
+// responses:
+//   200:
+//     description: Referral info registered
+//   400:
+//     description: Bad request
+//     schema:
+//       "$ref": "#/definitions/ErrorMessageDTO"
+//   422:
```
soffokl: We are sending only `http.StatusBadRequest` in the function, no 422 and 500.

tequilapi/endpoints/payout.go
```diff
@@ -34,7 +34,13 @@ type payoutInfo struct {
 	// in Ethereum address format
 	// required: true
 	// example: 0x000000000000000000000000000000000000000a
-	EthAddress   string `json:"ethAddress"`
+	EthAddress string `json:"ethAddress"`
```
soffokl: We are using a different style for JSON tags.
`ethAddress` - camalCase
`referral_code` - snake_case

tequilapi/endpoints/payout.go
ignasbernotas: it can return 500 in line https://github.com/mysteriumnetwork/node/pull/1165/files#diff-76a2d4645eb783ed4db8078ac22056fbR184

ignasbernotas: Removed the 422 though :)

PR #1166, opened 2019-06-18
Automatically prefix package name to logger

tadaskay:
Motivation: error-prone playing around with log prefixes, inconsistent formatting

How to use: simply declare `var log = logconfig.NewLogger()` and use it in the package.

To avoid runtime cost, call information is resolved once at a declare time,
thus there is a need to declare a variable. If overhead (around 25k ns) is
acceptable, we can remove package `log` variables.

Migrated core package for starters.

### DIFF
--- a/cmd/di.go
+++ b/cmd/di.go
@@ -23,7 +23,6 @@ import (
 	"path/filepath"
 	"time"
 
-	log "github.com/cihub/seelog"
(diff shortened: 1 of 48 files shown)


tadaskay:
I put in separate file so that there’s no unnecessary dependency between files just to use the logger. Also, it unifies the layout - easy to find in any package.

PR #1168, opened 2019-06-18
Connection attempts are now counted on provider side

vkuznecovas:
After a successful connection, the consumer will now send an ACK to the provider. On provider side, this ack is then published as an event.

The state keeper will also listen to session creation event to count the attempts.

### DIFF
--- a/core/state/state_test.go
+++ b/core/state/state_test.go
@@ -24,6 +24,7 @@ import (
 	"time"
 
 	"github.com/mysteriumnetwork/node/core/service"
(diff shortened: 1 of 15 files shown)


core/state/state_test.go
```diff
@@ -23,6 +23,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/mysteriumnetwork/node/core/state/event"
+
```
soffokl: Extra line.

PR #1169, opened 2019-06-18
Wait for valid registered identity before starting services

### LINKED ISSUE #1163: Node crashes/cli hangs if network unavailable on startup

tadaskay:
Fixes #1163

Functional changes:

- If initial proposal fetch fails, ignore it and continue (it will continuously attempt to fetch in a separate goroutine)
- Use [new discovery API endpoint](https://github.com/mysteriumnetwork/api/pull/138) to check if an existing identity is registered
- Proceed with node services startup only when the identity is valid and registered
  This also kind of solves 'app starting before the network is up' problem, since it will retry to get identity every 10s.
  To properly address network state issue, we should make separate node components react to network state events, but that's a 
  separate huge epic.

Additionally:

- Migrate "errors" to "github.com/pkg/errors" to retain stacktraces
- More error wraps
- Cut HTTP timeouts. _Note: leaving tequilapi timeout rather high, because of e2e test issues. Need to investigate further, perhaps containers are not yet healthy before running tests?_

### DIFF
--- a/cmd/commands/daemon/command.go
+++ b/cmd/commands/daemon/command.go
@@ -18,6 +18,7 @@
 package daemon
 
 import (
+	"github.com/cihub/seelog"
 	"github.com/mysteriumnetwork/node/cmd"
 	"github.com/urfave/cli"
 )
@@ -31,18 +32,27 @@ func NewCommand() *cli.Command {
 		Usage:     "Starts Mysterium Tequilapi service",
 		ArgsUsage: " ",
 		Action: func(ctx *cli.Context) error {
-			errorChannel := make(chan error, 2)
+			quit := make(chan error, 2)
 			if err := di.Bootstrap(cmd.ParseFlagsNode(ctx)); err != nil {
 				return err
 			}
-			go func() { errorChannel <- di.Node.Wait() }()
+			go func() { quit <- di.Node.Wait() }()
 
-			cmd.RegisterSignalCallback(func() { errorChannel <- nil })
+			cmd.RegisterSignalCallback(func() { quit <- nil })
 
-			return <-errorChannel
+			return describeQuit(<-quit)
 		},
 		After: func(ctx *cli.Context) error {
 			return di.Shutdown()
 		},
 	}
+}
+
+func describeQuit(err error) error {
+	if err == nil {
+		seelog.Info("stopping application")
+	} else {
+		seelog.Errorf("terminating application due to error: %+v\n", err)
+	}
+	return err
 }
--- a/cmd/commands/daemon/command.go
+++ b/cmd/commands/daemon/command.go
@@ -18,10 +18,12 @@
 package daemon
 
 import (
-	"github.com/cihub/seelog"
 	"github.com/mysteriumnetwork/node/cmd"
+	"github.com/mysteriumnetwork/node/logconfig"
 	"github.com/urfave/cli"
 )
+
+var log = logconfig.NewLogger()
 
 // NewCommand function creates run command
 func NewCommand() *cli.Command {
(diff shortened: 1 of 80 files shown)


cmd/commands/daemon/command.go
```diff
@@ -18,6 +18,7 @@
 package daemon
 
 import (
+	"github.com/cihub/seelog"
```
soffokl: We probably should use a new approach for logging that was merged yesterday.

cmd/commands/service/command.go
```diff
@@ -131,13 +143,18 @@ func (sc *serviceCommand) Run(ctx *cli.Context) (err error) {
 	return <-sc.errorChannel
 }
 
-func (sc *serviceCommand) unlockIdentity(identityOptions service.OptionsIdentity) (*identity.Identity, error) {
-	id, err := sc.tequilapi.CurrentIdentity(identityOptions.Identity, identityOptions.Passphrase)
-	if err != nil {
-		return nil, err
-	}
+const retryRate = 10 * time.Second
```
soffokl: Can we put the const on top of the file?

services/openvpn/core/options_node.go
```diff
@@ -20,14 +20,15 @@ package core
 import (
 	"bufio"
 	"bytes"
-	"errors"
+
```
soffokl: Extra line.

cmd/commands/service/command.go
tadaskay: Having it next to the code using it is easier to understand/discover, but I can move it.

tadaskay: Moved it to the inside of the function. There's no need to expose it to package.

services/openvpn/discovery/dto/bandwidth_test.go
```diff
@@ -19,10 +19,10 @@ package dto
 
 import (
 	"encoding/json"
-	"errors"
 	"testing"
 
 	"github.com/mysteriumnetwork/node/datasize"
+	assert2 "github.com/mysteriumnetwork/node/testkit/assert"
```
vkuznecovas: could we rename this to kitassert or something like this instead of the number two?

tadaskay: I'm not sure about this one. Is this critical?

identity/selector/handler.go
```diff
@@ -71,15 +77,26 @@ func (h *handler) UseOrCreate(address, passphrase string) (id identity.Identity,
...
+		return id, errors.Wrap(err, "failed to unlock identity")
+	}
+
+	registered, err := h.registry.IdentityExists(id, h.signerFactory(id))
+	if err != nil {
+		return id, errors.Wrap(err, "failed to verify registration status of local identity")
+	}
+	if !registered {
+		log.Info("existing identity is not registered, attempting to register")
+		if err = h.registry.RegisterIdentity(id, h.signerFactory(id)); err != nil {
```
Waldz: @zolia maybe it's time to drop identity registration at all - now it's a feature  identity stats collecting.
For Ethereum flow we can't just register identity, it's user flow, calling. For this flow user he will need to transfer money.

services/openvpn/discovery/dto/bandwidth_test.go
tadaskay: Renamed `assert` package to `assertkit` to avoid clashes and import aliasing.

identity/selector/handler.go
Waldz: Actually we a calling `registry.RegisterIdentity()` just to make `SetPayoutInfo()` to be successful, then user calls it later.

Lets remake SetPayoutInfo(), so that it fixes it's DB records

PR #1175, opened 2019-06-20
upload docker snapshot images to separate docker hub repo

tadovas:


ci/release/docker.go
```diff
@@ -0,0 +1,169 @@
...
+	"github.com/mysteriumnetwork/node/ci/env"
+	"github.com/mysteriumnetwork/node/ci/storage"
+	"github.com/mysteriumnetwork/node/logconfig"
+)
+
+const dockerImagesDir = "build/docker-images"
+const mystSnapshotsRepo = "mysteriumnetwork/myst-snapshots"
+
+// ReleaseDockerSnapshot uploads docker snapshot images to myst snapshots repo
+func ReleaseDockerSnapshot() error {
```
soffokl: > func name will be used as release.ReleaseDockerSnapshot by other packages, and that stutters; consider calling this DockerSnapshot (golint)

tadovas: Couldn't agree more from package perspective. BUT :)
```go
go run mage.go ReleaseDockerSnapshot
```
sounds more expressive as its not a normal function but mage action :|

PR #1176, opened 2019-06-20
UIServer Tequilapi proxy

ignasbernotas:
Closes: #1172

Works without CORS, but has BasicAuth instead

### DIFF
--- a/ui/server.go
+++ b/ui/server.go
@@ -20,8 +20,11 @@ package ui
 import (
 	"context"
 	"fmt"
+	"net"
 	"net/http"
+	"net/http/httputil"
 	"strconv"
+	"strings"
 	"time"
 
 	log "github.com/cihub/seelog"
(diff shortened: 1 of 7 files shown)


ui/server.go
```diff
@@ -20,45 +20,98 @@ package ui
...
+	`strings`
 	"time"
 
 	log "github.com/cihub/seelog"
 	"github.com/gin-contrib/cors"
 	"github.com/gin-gonic/gin"
 	godvpnweb "github.com/mysteriumnetwork/go-dvpn-web"
-	"github.com/mysteriumnetwork/node/ui/discovery"
 	"github.com/pkg/errors"
+
```
vkuznecovas: redundant space

ui/server.go
```diff
@@ -20,45 +20,98 @@ package ui
...
+				Timeout:   24 * time.Hour,
+				KeepAlive: 24 * time.Hour,
+			}).DialContext,
+		}
+
+		// suppress SSE streaming errors
+		// there's nothing we can do about them
+		// and they don't cause any issues AFAIK
+		defer func() {
+			recover()
```
vkuznecovas: this should handle the specific error instead.

Maybe something like this? Needs testing though.


```golang
defer func() {
			if err := recover(); err != nil {
				if err == http.ErrAbortHandler {
					// ignore
				} else {
					panic(err)
				}
			}
		}()
```

PR #1182, opened 2019-06-25
moving blockchain package to mysteriumnetwork/payment-bindings

vkuznecovas:
Moving binding generation to a new package, mysteriumnetwork/payment-bindings since this will be used by transactor/accountant and node.


tadovas:
This one is ok, but I have a suggestion - rename payment-bindings to go-payments, and package inside it to "payments" as bindings (generated code) will be just a part of that repo.

vkuznecovas:
I'll shuffle it around in the future.

Issue #1191, opened 2019-06-28
Too many open files


ignasbernotas:
UIServer reverse proxy doesn't close connections.
```
Jun 28 08:17:07 raspberrypi myst[1163]: 2019/06/28 08:17:07 http: Accept error: accept tcp [::]:4449: accept4: too many open files; retrying in 1s
Jun 28 08:17:08 raspberrypi myst[1163]: 2019/06/28 08:17:08 http: proxy error: dial tcp 127.0.0.1:4050: socket: too many open files
Jun 28 08:17:08 raspberrypi myst[1163]: 2019/06/28 08:17:08 http: Accept error: accept tcp [::]:4449: accept4: too many open files; retrying in 1s
Jun 28 08:17:08 raspberrypi myst[1163]: 2019-06-28T07:17:08.463956589 [Info] [core/discovery/api] proposals fetched: 102
Jun 28 08:17:11 raspberrypi myst[1163]: 2019/06/28 08:17:11 http: Accept error: accept tcp 127.0.0.1:4050: accept4: too many open files; retrying in 5ms
Jun 28 08:17:11 raspberrypi myst[1163]: 2019/06/28 08:17:11 http: Accept error: accept tcp 127.0.0.1:4050: accept4: too many open files; retrying in 10ms
Jun 28 08:17:11 raspberrypi myst[1163]: 2019/06/28 08:17:11 http: Accept error: accept tcp 127.0.0.1:4050: accept4: too many open files; retrying in 20ms
Jun 28 08:17:11 raspberrypi myst[1163]: 2019/06/28 08:17:11 http: Accept error: accept tcp 127.0.0.1:4050: accept4: too many open files; retrying in 40ms
Jun 28 08:17:11 raspberrypi myst[1163]: 2019/06/28 08:17:11 http: Accept error: accept tcp 127.0.0.1:4050: accept4: too many open files; retrying in 80ms
Jun 28 08:17:11 raspberrypi myst[1163]: 2019/06/28 08:17:11 http: Accept error: accept tcp 127.0.0.1:4050: accept4: too many open files; retrying in 160ms
Jun 28 08:17:11 raspberrypi myst[1163]: 2019/06/28 08:17:11 http: Accept error: accept tcp 127.0.0.1:4050: accept4: too many open files; retrying in 320ms
[...]

PR #1194, opened 2019-06-28
Re-use http transport for proxying requests

tadaskay:
Fixes: #1191 

Previously, new http transport was being created with every request,
leaking open connections and hitting max open FDs. This seems to help
(at least on OSX).

Gratis:
- added idle conn limitations just in case
- cut down timeouts/keepalive from 1 day to 20secs


tadaskay:
@ignasbernotas 🍖

Issue #1189, opened 2019-06-28
Wireguard installation in RaspberryPI error


ignasbernotas:
Installation fails following instructions from `https://github.com/adrianmihalko/raspberrypiwireguard`
First issue was related to this: [`https://github.com/adrianmihalko/raspberrypiwireguard/pull/21`](https://github.com/adrianmihalko/raspberrypiwireguard/pull/21)

Then later on:
```
pi@raspberrypi:~ $ sudo apt-get install wireguard
Reading package lists... Done
Building dependency tree
Reading state information... Done
The following NEW packages will be installed:
  wireguard
0 upgraded, 1 newly installed, 0 to remove and 0 not upgraded.
1 not fully installed or removed.
Need to get 0 B/21.5 kB of archives.
After this operation, 38.9 kB of additional disk space will be used.
perl: warning: Setting locale failed.
perl: warning: Please check that your locale settings:
	LANGUAGE = (unset),
	LC_ALL = (unset),
	LC_CTYPE = "UTF-8",
	LANG = "en_GB.UTF-8"
    are supported and installed on your system.
perl: warning: Falling back to a fallback locale ("en_GB.UTF-8").
locale: Cannot set LC_CTYPE to default locale: No such file or directory
locale: Cannot set LC_ALL to default locale: No such file or directory
Selecting previously unselected package wireguard.
(Reading database ... 88664 files and directories currently installed.)
Preparing to unpack .../wireguard_0.0.20190601-1_all.deb ...
Unpacking wireguard (0.0.20190601-1) ...
Setting up wireguard-dkms (0.0.20190601-1) ...
Removing old wireguard-0.0.20190601 DKMS files...

-------- Uninstall Beginning --------
Module:  wireguard
Version: 0.0.20190601
Kernel:  4.19.50+ (armv7l)
-------------------------------------

Status: Before uninstall, this module version was ACTIVE on this kernel.

wireguard.ko:
 - Uninstallation
   - Deleting from: /lib/modules/4.19.50+/kernel/net/
rmdir: failed to remove 'kernel/net': Directory not empty
 - Original module
   - No original module was found for this module on this kernel.
   - Use the dkms install command to reinstall any previous module version.

depmod...

DKMS: uninstall completed.

-------- Uninstall Beginning --------
Module:  wireguard
Version: 0.0.20190601
Kernel:  4.19.50-v7+ (armv7l)
-------------------------------------

Status: Before uninstall, this module version was ACTIVE on this kernel.

[...]

Issue #1198, opened 2019-07-02
Easy installation script


tadaskay:
It should be easy to install mysterium node at least on linux systems. A single script should be able to set up required dependencies and myst service so that no manual work is needed. This will also save us time to update myst when testing raspberries build.

```bash
curl https://raw.githubusercontent.com/mysteriumnetwork/node/master/install.sh | bash
```

- [x] Debian https://github.com/mysteriumnetwork/node/pull/1199
- [x] Ubuntu (wireguard installation slightly differs on Ubuntu)
- [x] Other distros (todo list)

Related: #1189 #1190 #1192

zolia:
covered since 0.15.1

Issue #1192, opened 2019-06-28
Make a bash script for RPI Wireguard installation


ignasbernotas:
Repo [https://github.com/adrianmihalko/raspberrypiwireguard](https://github.com/adrianmihalko/raspberrypiwireguard) is no longer maintained and the installation fails without unmerged PR fixes in that repo. We should create and host our own bash script which would fix this as well as https://github.com/mysteriumnetwork/node/issues/1189.

We could reuse some of the script from: https://github.com/LiveChief/wireguard-install

Issue #1190, opened 2019-06-28
Installing Raspberry DEB fails due to wireguard dependency


ignasbernotas:
```
wget https://github.com/mysteriumnetwork/node-builds/releases/download/0.9.0-1snapshot-20190627T0751-28afaa32/myst_linux_armhf.deb
```

```
pi@raspberrypi:~ $ sudo dpkg -i myst_linux_armhf.deb
Selecting previously unselected package myst.
(Reading database ... 88670 files and directories currently installed.)
Preparing to unpack myst_linux_armhf.deb ...
locale: Cannot set LC_CTYPE to default locale: No such file or directory
locale: Cannot set LC_ALL to default locale: No such file or directory
Unpacking myst (0.9.0-1snapshot-20190627T0751-28afaa32) ...
dpkg: dependency problems prevent configuration of myst:
 myst depends on wireguard; however:
  Package wireguard is not configured yet.

dpkg: error processing package myst (--install):
 dependency problems - leaving unconfigured
Errors were encountered while processing:
 myst
```

Workaround:
```
sudo dpkg --ignore-depends=wireguard -i myst_linux_armhf.deb
```

Should the installation fail without WG?

PR #1199, opened 2019-07-02
Provide debian/raspbian installation script

tadaskay:
Tested on raspbian

Ref: #1198 

Fixes: #1189 #1190 #1192

Leaving `INSTALL.md` for now, issue for cleaning it up and moving to docs: https://github.com/mysteriumnetwork/docs/issues/15


tadaskay:
Project wise - yep, it makes sense to keep it in `bin`. But I wanted the url to be as short as possible, because this is what users need to type in on raspberries.

Issue #1188, opened 2019-06-26
Replace UIServer basic auth with JWT based auth


ignasbernotas:
Current basic auth UX is horrible. Changing the password forces the browser to start spamming the credentials dialog (due to background XHR calls) so we need to create a login page which generates a JWT token and sets it as a httpOnly cookie.

PR #1207, opened 2019-07-05
JWT authorization and authentication for UIServer reverse proxy

ignasbernotas:
Closes: #1188


cmd/di.go
```diff
@@ -26,6 +26,9 @@ import (
 	"github.com/ethereum/go-ethereum/accounts/keystore"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/ethclient"
+	payment_bindings "github.com/mysteriumnetwork/payment-bindings"
+	"github.com/pkg/errors"
+
```
soffokl: Extra line

core/auth/jwt.go
```diff
@@ -0,0 +1,145 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package auth
+
+import (
+	"math/rand"
```
soffokl:
```suggestion
	"crypto/rand"
```

core/auth/jwt.go
```diff
@@ -0,0 +1,145 @@
...
+	"time"
+
+	"github.com/dgrijalva/jwt-go"
+	"github.com/pkg/errors"
+)
+
+// JwtAuthenticator contains JWT handling methods
+type JwtAuthenticator struct {
```
soffokl:
```suggestion
type JWTAuthenticator struct {
```

core/auth/jwt.go
```diff
@@ -0,0 +1,145 @@
...
+		return false, err
+	}
+	claims := &jwtClaims{}
+
+	tkn, err := jwt.ParseWithClaims(token, claims, func(token *jwt.Token) (interface{}, error) {
+		return jwtAuth.encryptionKey, nil
+	})
+
+	if tkn == nil || !tkn.Valid {
+		return false, errors.New("Invalid JWT token")
```
soffokl:
```suggestion
		return false, errors.New("invalid JWT token")
```

core/auth/jwt.go
```diff
@@ -0,0 +1,145 @@
...
+}
+
+func (jwtAuth *JwtAuthenticator) setupEncryptionKey() error {
+	key := jwtKey{}
+	err := jwtAuth.storage.GetValue(encryptionKeyBucket, encryptionKeyName, &key)
+	if err != nil {
+		key = generateRandomKey(64)
+		err := jwtAuth.storage.SetValue(encryptionKeyBucket, encryptionKeyName, key)
+		if err != nil {
+			return errors.Wrap(err, "Failed to store JWT encryption key")
```
soffokl:
```suggestion
			return errors.Wrap(err, "failed to store JWT encryption key")
```

ui/server.go
```diff
@@ -20,19 +20,16 @@ package ui
...
-	"strings"
 	"time"
 
 	log "github.com/cihub/seelog"
 	"github.com/gin-contrib/cors"
 	"github.com/gin-gonic/gin"
 	godvpnweb "github.com/mysteriumnetwork/go-dvpn-web"
-	"github.com/mysteriumnetwork/node/ui/discovery"
 	"github.com/pkg/errors"
+
```
soffokl: Extra line.

core/auth/jwt.go
```diff
@@ -0,0 +1,145 @@
...
+)
+
+// JwtAuthenticator contains JWT handling methods
+type JwtAuthenticator struct {
+	storage       Storage
+	encryptionKey []byte
+}
+
+// JWTToken contains token details
+type JWTToken struct {
```
vkuznecovas: if expanded, this will be jsonwebtokentoken, rename it to JWT.

core/auth/jwt.go
```diff
@@ -0,0 +1,145 @@
...
+}
+
+type jwtKey []byte
+
+const expiresIn = 48 * time.Hour
+const encryptionKeyBucket = "jwt"
+const encryptionKeyName = "jwt-encryption-key"
+
+// NewJWTAuthenticator creates a new JWT authentication instance
+func NewJWTAuthenticator(storage Storage) *JwtAuthenticator {
```
vkuznecovas: I'd redesign this component a bit. You're injecting storage just to store the key - why not actually fetch(or generate and store if it does not exist) the key and pass it as an argument here? No need to inject storage then, and you're not really. You won't need the ensureEncryptionKeyIsSet() method which is a racing anti-pattern anyway. I could now call create/validate token with multiple goroutines and the state would be modified in an upredictable way.

core/auth/jwt.go
ignasbernotas: math/crypto have a different API and in this case pseudo-random works for us.

cmd/di.go
```diff
@@ -22,10 +22,11 @@ import (
 	"net"
 	"path/filepath"
 	"time"
-
```
vkuznecovas: there should be a space here :)

core/auth/jwt.go
```diff
@@ -0,0 +1,145 @@
...
+	claims := &jwtClaims{}
+
+	tkn, err := jwt.ParseWithClaims(token, claims, func(token *jwt.Token) (interface{}, error) {
+		return jwtAuth.encryptionKey, nil
+	})
+
+	if tkn == nil || !tkn.Valid {
+		return false, errors.New("invalid JWT token")
+	}
+	if err != nil {
```
vkuznecovas: move this up to where the error is being generated. Also - in the current implementation, if the error is not nil, and is not equal to jwt.ErrSignatureInvalid - we return that the token is valid. Is this the intended behaviour?

core/auth/jwt.go
vkuznecovas: does it though? i'd still switch to crypto

ignasbernotas:
@soffokl @vkuznecovas switched to crypto/rand, refactored key generator :)

PR #1222, opened 2019-07-16
RPI: Wait for network up before starting service

tadaskay:


bin/package/installation/systemd.service
```diff
@@ -3,6 +3,7 @@
 [Unit]
 Description=Server for decentralised VPN built on blockchain
 Documentation=https://mysterium.network/
+Requires=network-online.target
```
zolia: Do you know what it actually does? Tests for presence of gateway or simply senses ethernet port being ip?

tadaskay: network-online.target
Units that strictly require a configured network connection should pull in network-online.target (via a Wants= type dependency) and order themselves after it. This target unit is intended to pull in a service that delays further execution until the network is sufficiently set up. What precisely this requires is left to the implementation of the network managing service.

Note the distinction between this unit and network.target. This unit is an active unit (i.e. pulled in by the consumer rather than the provider of this functionality) and pulls in a service which possibly adds substantial delays to further execution. In contrast, network.target is a passive unit (i.e. pulled in by the provider of the functionality, rather than the consumer) that usually does not delay execution much. Usually, network.target is part of the boot of most systems, while network-online.target is not, except when at least one unit requires it. Also see Running Services After the Network is up for more information.

All mount units for remote network file systems automatically pull in this unit, and order themselves after it. Note that networking daemons that simply provide functionality to other hosts generally do not need to pull this in.

systemd automatically adds dependencies of type Wants= and After= for this target unit to all SysV init script service units with an LSB header referring to the "$network" facility.

Note that this unit is only useful during the original system start-up logic. After the system has completed booting up, it will not track the online state of the system anymore. Due to this it cannot be used as a network connection monitor concept, it is purely a one-time system start-up concept.

tadaskay: https://www.freedesktop.org/software/systemd/man/systemd.special.html

PR #1237, opened 2019-07-30
Fix RPI build

zolia:
Fix RPI build by fetching proper keys


soffokl:
`altsrc` was just added by @tadaskay in the https://github.com/mysteriumnetwork/node/pull/1232. We don't need it anymore?

zolia:
@soffokl we should not need that altsrc in most places, only when we use config loading functions.  Also these improper cli dependencies gets spread across the code. Wanted to stop that.

tadaskay:
Removed altsrc removal as it broke config

PR #1257, opened 2019-08-02
Fix WG connect to server behind upnp router

zolia:
- trace logs added
 - ConnectDelay works now
 - do not block on session acknowledgement

TODO: investigate why session ack takes so long. Does it timeout? @vkuznecovas ?


vkuznecovas:
> * trace logs added
> * ConnectDelay works now
> * do not block on session acknowledgement
> 
> TODO: investigate why session ack takes so long. Does it timeout? @vkuznecovas ?

I do think it's not the the ack that takes a long time. It's the reconnect to nats that takes a while. On a network change, it waits for nats connection to timeout and then reconnect and only then it sends the message. But I'd need more context to know for sure.

PR #1264, opened 2019-08-09
OpenVPN user subnet parameter

zolia:


firewall/subnet_calculator.go
```diff
@@ -0,0 +1,31 @@
...
+
+package firewall
+
+import (
+	"fmt"
+	"net"
+)
+
+// SimplifiedSubnet returns IP subnet in x.x.x.x/yy format
+func SimplifiedSubnet(subnet, netmask string) string {
```
soffokl: If we will use `net.IPNet` instead of strings, we will not need to do our own conversion:
https://golang.org/pkg/net/#IPNet.String

zolia: True, but then we will need to do IPNet construction. I think passing around string is more universal and reduces dependencies. I would just convert to IPNet where its really needed.

PR #1281, opened 2019-08-27
Provider serves DNS proxy on Openvpn's virtual IP

Waldz:
Screen Shot 2019-08-27 at 15 19 31


services/openvpn/service/manager.go
```diff
@@ -130,28 +134,40 @@ func (m *Manager) Serve(providerID identity.Identity) (err error) {
 		}
 	}()
 
-	if err = m.vpnServer.Start(); err != nil {
-		return
+	if err := m.startServer(m.vpnServer, stateChannel); err != nil {
+		return errors.Wrap(err, "failed to start Openvpn server")
+	}
+
+	m.dnsServer = dns.NewServer("10.8.0.1:53", dns.ResolveViaConfigured())
```
soffokl: "10.8.0.1/24" is just a default value, this subnet could be changed, we should respect it for DNS too.

Waldz: Improved

services/openvpn/service/manager.go
```diff
@@ -115,12 +124,13 @@ func (m *Manager) Serve(providerID identity.Identity) (err error) {
...
 	vpnServerConfig := m.vpnServerConfigFactory(primitives, m.vpnServerPort)
-	m.vpnServer = m.vpnServerFactory(vpnServerConfig)
+	stateChannel := make(chan openvpn.State, 10)
+	m.vpnServer = m.vpnServerFactory(vpnServerConfig, stateChannel)
 
 	// register service port to which NATProxy will forward connects attempts to
 	m.natPinger.BindServicePort(openvpn_service.ServiceType, m.vpnServerPort)
 
-	log.Info(logPrefix, "starting openvpn server on port: ", m.vpnServerPort)
+	log.Info(logPrefix, "Starting openvpn server on port: ", m.vpnServerPort)
```
tadaskay: Case should be consistent, i.e. lowercase

Waldz: So logs most places we do with `log.Info("Uppercase ..")`.
Maybe You're mixing with `errors.New("lowercase ..")` which is Golang convention.

tadaskay: No @Waldz, I'm talking about the logging messages. They are not consistent, thus no, in most places, we don't do capitalization of the log messages. Quite often, the error is being logged as-is, resulting in a lowercase message as well. Perhaps it's a good addition for our devbook if we can agree on the choice?

PR #1282, opened 2019-08-28
Persist UI changes in config.toml

tadaskay:
`GET /config/user` returns current user configuration
`POST /config/user` changes and saves user configuration


config/config_test.go
```diff
@@ -0,0 +1,87 @@
...
+	assert.Equal(t, 40000, cfg.GetInt("openvpn.port"))
+
+	// when: user configuration is saved
+	err = cfg.SaveUserConfig()
+	// then: only user configuration values are stored
+	assert.NoError(t, err)
+	tomlContent, err := ioutil.ReadFile(configFileName)
+	assert.NoError(t, err)
+	assert.Contains(t, string(tomlContent), "port = 22822")
+	assert.NotContains(t, string(tomlContent), "proto = \"tcp\"")
```
vkuznecovas:
```suggestion
	assert.NotContains(t, string(tomlContent), `proto = "tcp"`)
```

config/urfavecli/clicontext/user.go
```diff
@@ -0,0 +1,96 @@
...
+		return err
+	}
+
+	return config.Current.LoadUserConfig(configFilePath)
+}
+
+// LoadUserConfigQuietly like LoadUserConfig, but instead of returning an error,
+// it logs it on a `warn` level.
+// `error` is specified as a return to adhere to `cli.BeforeFunc` for convenience.
+func LoadUserConfigQuietly(ctx *cli.Context) error {
```
vkuznecovas: don't need the return value here

config/urfavecli/clicontext/user.go
```diff
@@ -0,0 +1,96 @@
...
+	return nil
+}
+
+func dirExists(dir string) error {
+	fileStat, err := os.Stat(dir)
+	if err != nil {
+		return err
+	}
+	if isDir := fileStat.IsDir(); !isDir {
+		return errors.New("directory expected")
```
vkuznecovas: would be nice to specify the dir in the error

utils/jsonutil/strings.go
```diff
@@ -0,0 +1,26 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package jsonutil
+
+import "encoding/json"
+
+// ToJson marshals v to a JSON string
+func ToJson(v interface{}) string {
+	bytes, _ := json.Marshal(&v)
```
vkuznecovas: Would be nice to at least log it here and not ignore it. What if I pass a channel here? :)

config/urfavecli/clicontext/user.go
tadaskay: Need it to implement `BeforeFunc` type from `urfave/cli`:

```go
type BeforeFunc func(*Context) error
```

This is to simplified usage:

```go
return &cli.Command{
		Name:   cliCommandName,
		Usage:  "Starts a CLI client with a Tequilapi",
		Before: clicontext.LoadUserConfigQuietly,
		...
}
```

config/urfavecli/clicontext/user.go
tadaskay: Point taken

utils/jsonutil/strings.go
tadaskay: Point taken

utils/jsonutil/strings.go
```diff
@@ -0,0 +1,33 @@
...
+
+	"github.com/mysteriumnetwork/node/logconfig"
+)
+
+var log = logconfig.NewLogger()
+
+// ToJson marshals v to a JSON string
+func ToJson(v interface{}) string {
+	bytes, err := json.Marshal(&v)
+	log.Warnf("failed to marshal to JSON string: %v", err)
```
soffokl: We probably lacking if block here to check the `err`.

tadaskay: This is intended. I'd like it to use in simple one-line situations, where error checking is not mandatory, e.g.:

```go
log.Info("user configuration loaded: \n", jsonutil.ToJson(cfg))
```

soffokl: But you are logging every time record `failed to marshal to JSON string` even if no error happened. Or I missing something?

tadaskay: Ah I misunderstood you. Thanks, changed it. :)

PR #1285, opened 2019-09-02
Use provider's DNS for consumer connection

soffokl:


services/openvpn/service/factory.go
```diff
@@ -135,6 +136,10 @@ func newSessionConfigNegotiatorFactory(networkOptions node.OptionsNetwork, servi
 		return &OpenvpnConfigNegotiator{
 			natEventGetter: natEventGetter,
 			vpnConfig: &openvpn_service.VPNConfig{
+				DNS: providerIP(net.IPNet{
```
Waldz: Should be just DNS endpoint `DNS: "10.0.0.1:53"` or `DNS: ["10.0.0.1:53", "10.0.0.2:53"]`
What's the point to communicate virtual network information?

soffokl: It's just an IP `10.8.0.1`, nothing more.

services/openvpn/client_config.go
```diff
@@ -71,23 +72,26 @@ func defaultClientConfig(runtimeDir string, scriptSearchPath string) *ClientConf
...
 	// TODO Rename `vpnConfig` to `sessionConfig`
 	err := NewDefaultValidator().IsValid(vpnConfig)
 	if err != nil {
 		return nil, err
 	}
 
-	clientFileConfig := newClientConfig(runtimeDir, configDir)
+	clientFileConfig := newClientConfig(runtimeDir, configDir, enableDNS)
+	if enableDNS && len(vpnConfig.DNS) > 0 {
+		for _, dns := range strings.Split(vpnConfig.DNS, ",") {
```
Waldz: Just a single address, if we want list we'll extend structure to have array

services/openvpn/service/factory.go
Waldz: Oh, You're calculating it here. Lets just inject it instead of duplicated calculation

cmd/flags_network.go
```diff
@@ -85,6 +85,11 @@ var (
 		Name:  "experiment-natpunching",
 		Usage: "Enables experimental NAT hole punching",
 	})
+
+	enableDNSFlag = altsrc.NewBoolTFlag(cli.BoolTFlag{
```
Waldz: As it's connection choice, the best would be to have it as option passed to connection creation, instead global flag

services/openvpn/service/factory.go
soffokl: Changed

services/openvpn/client_config.go
soffokl: changed

cmd/flags_network.go
soffokl: Moved to connection options.

services/openvpn/service/manager.go
```diff
@@ -50,7 +49,7 @@ type ServerFactory func(*openvpn_service.ServerConfig, chan openvpn.State) openv
 type ProposalFactory func(currentLocation market.Location) market.ServiceProposal
 
 // SessionConfigNegotiatorFactory initiates ConfigProvider instance during runtime
-type SessionConfigNegotiatorFactory func(secPrimitives *tls.Primitives, outboundIP, publicIP string, port int) session.ConfigNegotiator
+type SessionConfigNegotiatorFactory func(secPrimitives *tls.Primitives, serverIP, outboundIP, publicIP string, port int) session.ConfigNegotiator
```
Waldz: `serverIP`  -> `dnsIP` probably

core/node/options.go
```diff
@@ -47,8 +47,9 @@ type Options struct {
 	Location   OptionsLocation
 	Transactor OptionsTransactor
 
-	Openvpn  Openvpn
-	Firewall OptionsFirewall
+	EnableDNS bool
```
Waldz: Not needed anymore

core/node/options_network.go
```diff
@@ -32,5 +32,7 @@ type OptionsNetwork struct {
 	EtherClientRPC       string
 	EtherPaymentsAddress string
 
+	EnableDNS bool
```
Waldz: Not needed anymore

services/openvpn/service/manager.go
soffokl: Right, changed it.

core/node/options.go
soffokl: Thanks, remove it.

core/node/options_network.go
soffokl: Thanks, remove it.

services/wireguard/service/service_unix.go
```diff
@@ -89,6 +91,19 @@ func (manager *Manager) ProvideConfig(sessionConfig json.RawMessage, traversalPa
 		return nil, err
 	}
 
+	config.Consumer.DNS = providerIP(config.Consumer.IPAddress).String()
+	dnsServer := dns.NewServer(
+		net.JoinHostPort(config.Consumer.DNS, "53"),
+		dns.ResolveViaConfigured(),
+	)
+
+	go func() {
```
Waldz: I dont think it's a good idea to not to care if DNS wasn't started

services/wireguard/service/service_unix.go
```diff
@@ -100,11 +115,14 @@ func (manager *Manager) ProvideConfig(sessionConfig json.RawMessage, traversalPa
 	}
 
 	destroy := func() {
+		if err := dnsServer.Stop(); err != nil {
+			_ = log.Error("failed to stop DNS server", err)
+		}
 		if err := manager.natService.Del(natRule); err != nil {
-			log.Error("failed to delete NAT forwarding rule: ", err)
+			_ = log.Error("failed to delete NAT forwarding rule: ", err)
```
Waldz: Is this change really needed here? If not using variable, lets not define it as `_` at all

services/wireguard/service/service_unix.go
```diff
@@ -127,3 +145,12 @@ func (manager *Manager) Stop() error {
 	log.Info("wireguard service stopped")
 	return nil
 }
+
+func providerIP(subnet net.IPNet) net.IP {
```
Waldz: Looks very complicated function, I suppose we already had one in this package

services/wireguard/service/service_unix.go
soffokl: `dnsServer.Run()` is a blocking function, it returns nothing until fails. We cannot wait it forever, added logging to be able to track any problem with DNS.

services/wireguard/service/service_unix.go
soffokl: `log.Error` return and error and linter complain about it. Explicitly marking that we don't care about this error.

services/wireguard/service/service_unix.go
soffokl: We have similar functions in a number of places, but all of them a bit different. We probably need some general and complex function that will work for all cases. I didn't want to add it right now.

soffokl: Added this function to the utils package as a single place.

services/wireguard/service/service_unix.go
Waldz: Ups, then need to do the same in Openvpn DNS server

services/wireguard/service/service_unix.go
Waldz: I would say, lets change our abstraction `log.Logger` instead

services/wireguard/service/service_unix.go
soffokl: Changed for OpenVPN too.

services/wireguard/service/service_unix.go
soffokl: Changed.

PR #1300, opened 2019-09-12
Register to my.mysterium.network

ignasbernotas:


mmn/http_client.go
```diff
@@ -0,0 +1,157 @@
...
+
+	return nil
+}
+
+// Client is able perform remote requests to Tequilapi server
+type Client struct {
+	http httpClientInterface
+}
+
+func newHTTPClient(baseURL string, logPrefix string, ua string) *httpClient {
```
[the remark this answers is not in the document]
zolia: Yes, package is called 'requests'. Lets reuse that. We just eliminated a bunch for client duplicates. Lets not get back to it..

mmn/node_information.go
```diff
@@ -0,0 +1,125 @@
...
+	Arch        string `json:"arch"`
+	NodeVersion string `json:"node_version"`
+	Identity    string `json:"identity"`
+}
+
+func GetNodeInformation() (*NodeInformation, error) {
+	var mac, ip string
+	ip, err := getLocalNetworkIP()
+
+	if err == nil {
```
zolia: usually its vice versa, fail early by checking if err != nil and returning. More readable and no excessive execution.

mmn/node_information.go
```diff
@@ -0,0 +1,125 @@
...
+		mac, err = getMACAddress(ip)
+		if err != nil {
+			mac = ""
```
zolia: we should not continue if theres an error. We should be returning immediately and caller should not interpret returned data if theres an error.

mmn/node_information.go
```diff
@@ -0,0 +1,125 @@
...
+	LocalIP     string `json:"local_ip"`
+	OS          string `json:"os"`
+	Arch        string `json:"arch"`
+	NodeVersion string `json:"node_version"`
+	Identity    string `json:"identity"`
+}
+
+func GetNodeInformation() (*NodeInformation, error) {
+	var mac, ip string
+	ip, err := getLocalNetworkIP()
```
zolia: You probably need GetOutboundIPAsString() in core ip package. Your getLocalNetworkIP is inefficient and probably incorrect since node might be not binded to all IPs, just specific one.

mmn/node_information.go
```diff
@@ -0,0 +1,125 @@
...
+			if ipnet.IP.To4() != nil {
+				ip = ipnet.IP.String()
+			}
+		}
+	}
+
+	return
+}
+
+func getMACAddress(ip string) (string, error) {
```
zolia: It would be good to have this function as utility function or as part of Resolver, since we might / will need it in other places.

cmd/di_desktop.go
```diff
@@ -278,3 +280,19 @@ func (di *Dependencies) bootstrapUIServer(options node.Options) {
 
 	di.UIServer = uinoop.NewServer()
 }
+
+func (di *Dependencies) bootstrapMMN(options node.Options) {
+	err := di.EventBus.SubscribeAsync("identity-unlocked", func(param string) {
+		info, err := mmn.GetNodeInformation()
+		if err != nil {
+			log.Error("Failed to get NodeInformation for MMN", err.Error())
```
soffokl:
```suggestion
			log.Error("Failed to get NodeInformation for MMN", err)
```

cmd/di_desktop.go
```diff
@@ -278,3 +280,19 @@ func (di *Dependencies) bootstrapUIServer(options node.Options) {
...
+			return
+		}
+		info.Identity = param
+		client := mmn.NewMMNClient(options.MMN.Address)
+		client.RegisterNode(*info)
+	})
+	if err != nil {
+		log.Error("Failed to get register to mmn event", err.Error())
```
soffokl:
```suggestion
		log.Error("Failed to get register to mmn event", err)
```

mmn/http_client.go
```diff
@@ -0,0 +1,157 @@
+package mmn
```
soffokl: `Copyright` header is missing here.

mmn/node_information.go
```diff
@@ -0,0 +1,125 @@
...
+	"os/exec"
+	"runtime"
+	"strings"
+
+	log "github.com/cihub/seelog"
+
+	"github.com/mysteriumnetwork/node/metadata"
+)
+
+type NodeInformation struct {
```
soffokl: We have almost all of this data as a quality metric: https://github.com/mysteriumnetwork/node/blob/master/core/quality/sender.go#L74
Maybe we should reuse it?

zolia: What bothers we is WHAT data we collect. Ignas collects more detailed info. Such as Darwin version. `sw_vers -productVersion`. Which I suppose is needed. Maybe we can extend quality metrics with that info and then reuse that?

soffokl: My biggest worry here is that some providers probably will not want to report some information like OS details and MAC addresses because of privacy.

ignasbernotas: We can hash it. MAC address is the only unique identifier we have for RPI nodes. We don't need it in plain text, but we do need a persistent property:)

ignasbernotas:
@zolia @soffokl please rereview :)

cmd/di_desktop.go
soffokl: My suggestion is removing explicit converting to string, it enough to just pass `err` here.

core/ip/mac.go
```diff
@@ -0,0 +1,56 @@
...
+	"net"
+	"strings"
+
+	"github.com/pkg/errors"
+)
+
+// GetMACAddressForIP finds MAC address
+func GetMACAddressForIP(ip string) (string, error) {
+	var currentNetworkHardwareName string
+	interfaces, _ := net.Interfaces()
```
soffokl: Let's not ignore this error, It could happen if it exists. 
```suggestion
	interfaces, _ := net.Interfaces()
```

mmn/client.go
```diff
@@ -0,0 +1,50 @@
...
+	mmnAddress string
+}
+
+func (m *client) RegisterNode(information *NodeInformation) error {
+	req, err := requests.NewPostRequest(m.mmnAddress, "node", information)
+	if err != nil {
+		return err
+	}
+
+	if err = m.http.DoRequest(req); err != nil {
```
soffokl: No need to do this check here.
```suggestion
	return m.http.DoRequest(req)
```

mmn/node_information.go
soffokl: So, what we are going to do with it?

cmd/flags_network.go
```diff
@@ -58,6 +58,12 @@ var (
 		Value: metadata.DefaultNetwork.AccessPolicyOracleAddress,
 	})
 
+	mmnAddressFlag = altsrc.NewStringFlag(cli.StringFlag{
+		Name:  "mmn-address",
```
tadaskay: This looks a bit cryptic to me, and also it contains a dash. How about `mymysterium.url`?

mmn/node_information.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+
+	return ""
+}
+
+func getOSByCommand(os string, command string, args ...string) string {
+	if runtime.GOOS == os {
+		output, err := exec.Command(command, args...).Output()
+		if err != nil {
+			log.Error(errors.Wrap(err, "failed to get OS information for "+os+" using "+command))
```
tadaskay: I suppose `errors.Errorf()` would fit better here.

mmn/node_information.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+
+	if output := getOSByCommand("linux", "lsb_release", "-d"); len(output) > 0 {
+		return strings.TrimSpace(strings.Replace(string(output), "Description:", "", 1))
+	}
+
+	return ""
+}
+
+func getOSByCommand(os string, command string, args ...string) string {
```
tadaskay: This func looks rather odd to me, without a clear purpose; more like a piece of `getOS` extracted for the sake of extraction. I would rewrite `getOS()` using a switch:

```go
	switch runtime.GOOS {
	case "darwin":
		output, err := shell.NewCmd("sw_vers -productVersion").Output()
		// trim and do stuff
	case "linux":
		output, err := shell.NewCmd("lsb_release -d").Output()
		// trim and do stuff
	}
```

Or as an alternative option, could use smth like https://github.com/matishsiao/goInfo - that has windows implementation as well.

mmn/node_information.go
tadaskay: I say hash it, or provide an ability to opt-out, e.g. `mymysterium.enabled=false`

ignasbernotas:
@tadaskay @soffokl @zolia added `mymysterium.enabled` flag, renamed `mmn-address` to `mymysterium.address`, hashed MAC address to sha256 and other cleanups :)

mmn/node_information.go
ignasbernotas: Solved

cmd/flags_mmn.go
```diff
@@ -0,0 +1,53 @@
...
+)
+
+var (
+	mmnAddressFlag = altsrc.NewStringFlag(cli.StringFlag{
+		Name:  "mymysterium.url",
+		Usage: "URL of my.mysterium.network API",
+		Value: metadata.DefaultNetwork.MMNAddress,
+	})
+
+	mmnEnabledFlag = altsrc.NewBoolTFlag(cli.BoolTFlag{
```
zolia: Does it have to be 'True' by default? Maybe user can decide?

core/ip/mac.go
```diff
@@ -0,0 +1,59 @@
...
+	if err != nil {
+		return "", errors.Wrap(err, "failed to retrieve network interfaces")
+	}
+
+	for _, i := range interfaces {
+		if addresses, err := i.Addrs(); err == nil {
+			for _, addr := range addresses {
+				// only interested in the name with current IP address
+				if strings.Contains(addr.String(), ip) {
+					currentNetworkHardwareName = i.Name
```
zolia: As soon as we get a matching IP we should terminate the loop. There cannot be more than one interface with the same IP address on a system.

zolia: Also would suggest naming:
```suggestion
					matchedInterface = i.Name
```

mmn/node_information.go
```diff
@@ -0,0 +1,106 @@
...
+	MACAddress  string `json:"mac_address"` // SHA256 hash
+	LocalIP     string `json:"local_ip"`
+	OS          string `json:"os"`
+	Arch        string `json:"arch"`
+	NodeVersion string `json:"node_version"`
+	Identity    string `json:"identity"`
+}
+
+// OnIdentityUnlockCallback sends node information to MMN on identity unlock
+func OnIdentityUnlockCallback(client *client, resolver ip.Resolver) func(string) {
```
zolia: Function name does not say anything on what it does. 
```suggestion
func CollectNodeData(client *client, resolver ip.Resolver) func(string) {
```

cmd/flags_mmn.go
ignasbernotas: It does for now at least :)

Issue #399, opened 2018-09-20
Conform to godoc


vkuznecovas:
We're pretty close to conforming to godoc but not there yet.

Things that need doing:
- [ ] Add missing package comments
- [ ] Add examples where appropriate
- [ ] Add godoc to readme

PR #1301, opened 2019-09-12
New payment pingpong scaffold

vkuznecovas:
Initial scaffold of the new pingpong between the consumer and provider for the upcoming payments.

Should be backwards compatible, as I've tested conencting to an older node, as well as had this session while running in provider mode: https://testnet.mysterium.network/session/2f822bd5-c5ff-4ccb-a921-c5ee7255790a

### DIFF
--- a/session/manager.go
+++ b/session/manager.go
@@ -83,7 +83,8 @@ type Storage interface {
 }
 
 // BalanceTrackerFactory returns a new instance of balance tracker
-type BalanceTrackerFactory func(consumer, provider, issuer identity.Identity) (BalanceTracker, error)
+type BalanceTrackerFactory func(consumer, provider, issuer identity.Identity) (PaymentEngine, error)
+type PaymentEngineFactory func() (PaymentEngine, error)
 
 // NATEventGetter lets us access the last known traversal event
 type NATEventGetter interface {
@@ -96,6 +97,7 @@ func NewManager(
 	idGenerator IDGenerator,
 	sessionStorage Storage,
 	balanceTrackerFactory BalanceTrackerFactory,
+	paymentEngineFactory PaymentEngineFactory,
 	natPingerChan func(*traversal.Params),
 	natEventGetter NATEventGetter,
 	serviceId string,
@@ -110,8 +112,8 @@ func NewManager(
 		natEventGetter:        natEventGetter,
 		serviceId:             serviceId,
 		publisher:             publisher,
-
-		creationLock: sync.Mutex{},
+		paymentEngineFactory:  paymentEngineFactory,
+		creationLock:          sync.Mutex{},
 	}
 }
 
@@ -121,6 +123,7 @@ type Manager struct {
 	generateID            IDGenerator
 	sessionStorage        Storage
 	balanceTrackerFactory BalanceTrackerFactory
+	paymentEngineFactory  PaymentEngineFactory
 	provideConfig         ConfigProvider
 	natPingerChan         func(*traversal.Params)
 	natEventGetter        NATEventGetter
@@ -131,7 +134,7 @@ type Manager struct {
 }
 
 // Create creates session instance. Multiple sessions per peerID is possible in case different services are used
-func (manager *Manager) Create(consumerID identity.Identity, issuerID identity.Identity, proposalID int, config ServiceConfiguration, pingerParams *traversal.Params) (sessionInstance Session, err error) {
+func (manager *Manager) Create(consumerID identity.Identity, consumerInfo ConsumerInfo, proposalID int, config ServiceConfiguration, pingerParams *traversal.Params) (sessionInstance Session, err error) {
 	manager.creationLock.Lock()
 	defer manager.creationLock.Unlock()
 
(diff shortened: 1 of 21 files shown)


session/pingpong/invoice_tracker.go
```diff
@@ -0,0 +1,177 @@
...
+	}
+}
+
+func calculateMaxNotReceivedExchangeMessageCount(chargeLeeway, chargePeriod time.Duration) uint64 {
+	return uint64(math.Round(float64(chargeLeeway) / float64(chargePeriod)))
+}
+
+// Start stars the invoice tracker
+func (it *InvoiceTracker) Start() error {
+	log.Debug("starting...")
```
tadaskay: Let's use uppercase for logging messages

session/pingpong/invoice_tracker.go
```diff
@@ -0,0 +1,177 @@
...
+	log.Debug(it.maxNotReceivedExchangeMessages)
```
tadaskay: This should have some context in addition to logging an `int`.

session/pingpong/invoice_messaging.go
```diff
@@ -0,0 +1,112 @@
...
+	"github.com/mysteriumnetwork/payments/crypto"
+)
+
+// InvoiceRequest structure represents the invoice message that the provider sends to the consumer
+type InvoiceRequest struct {
+	Invoice crypto.Invoice `json:"invoice"`
+}
+
+const endpointInvoice = "session-invoice"
+const messageEndpointInvoice = communication.MessageEndpoint(endpointInvoice)
```
tadaskay: I see where this is coming from, but let's use normalized order for naming, it makes it easier to comprehend:
- `endpointInvoice` → `invoiceEndpoint`
- `messageEndpointInvoice` → `invoiceMessageEndpoint`

session/pingpong/invoice_tracker.go
```diff
@@ -0,0 +1,176 @@
...
+	peerInvoiceSender               PeerInvoiceSender
+	exchangeMessageChan             chan crypto.ExchangeMessage
+	chargePeriod                    time.Duration
+	exchangeMessageWaitTimeout      time.Duration
+	notReceivedExchangeMessageCount uint64
+	maxNotReceivedExchangeMessages  uint64
+	once                            sync.Once
+}
+
+// NewInvoiceTracker creates a new instancec of invoice tracker
```
soffokl:
```suggestion
// NewInvoiceTracker creates a new instance of invoice tracker
```

session/pingpong/invoice_tracker.go
```diff
@@ -0,0 +1,176 @@
...
+}
+
+// ErrExchangeWaitTimeout indicates that we did not get an exchange message in time
+var ErrExchangeWaitTimeout = errors.New("did not get a new exchange message")
+
+// ErrExchangeValidationFailed indicates that there was an error with the exchange signature
+var ErrExchangeValidationFailed = errors.New("exchange validation failed")
+
+// errBoltNotFound indicates that bolt did not find a record
+var errBoltNotFound = errors.New("not found")
```
soffokl: `errBoltNotFound` is unused (varcheck)

session/pingpong/exchange_message_tracker.go
```diff
@@ -0,0 +1,96 @@
...
+	if err != nil {
+		return errors.Wrap(err, "could not create exchange message")
+	}
+	log.Debug("Sending message", *msg)
+	err = emt.peerExchangeMessageSender.Send(*msg)
+	if err != nil {
+		log.Warn("Failed to send exchange message: ", err)
+	}
+
+	return nil
```
soffokl: Shouldn't we return `err` here?
```suggestion
	return err
```

session/promise/promise_messaging.go
```diff
@@ -37,6 +37,7 @@ type Message struct {
 type PaymentInfo struct {
 	LastPromise LastPromise `json:"lastPromise"`
 	FreeCredit  uint64      `json:"freeCredit"`
+	Supports    string      `json:"supported"`
```
soffokl: `Supports` and `supported` here, probably should be only one.

session/manager.go
```diff
@@ -83,7 +83,10 @@ type Storage interface {
 }
 
 // BalanceTrackerFactory returns a new instance of balance tracker
-type BalanceTrackerFactory func(consumer, provider, issuer identity.Identity) (BalanceTracker, error)
+type BalanceTrackerFactory func(consumer, provider, issuer identity.Identity) (PaymentEngine, error)
+
+// PaymentEngineFactory creates a new instance of payment engine
```
anjmao: Minor suggestion. End each comment sentence with a dot at the end. Godoc page generates methods documentation based on these comments.

session/pingpong/exchange_message_tracker.go
vkuznecovas: No, the consumer is allowed to miss a few messages therefore returning nil here. THe provider will kill the session if it exceed an allowed amount.

session/pingpong/exchange_message_tracker.go
```diff
@@ -0,0 +1,96 @@
...
+				return err
+			}
+		}
+	}
+}
+
+func (emt *ExchangeMessageTracker) issueExchangeMessage(invoice crypto.Invoice) error {
+	msg, err := crypto.CreateExchangeMessage(invoice, 10, "", emt.keystore, common.HexToAddress(emt.identity.Address))
+	if err != nil {
+		return errors.Wrap(err, "could not create exchange message")
```
anjmao: Just to let you know. Go 1.13 introduced errors wrapping in standard library  https://golang.org/doc/go1.13#error_wrapping

session/pingpong/factory.go
```diff
@@ -0,0 +1,57 @@
...
+	"github.com/ethereum/go-ethereum/accounts/keystore"
+	"github.com/mysteriumnetwork/node/communication"
+	"github.com/mysteriumnetwork/node/core/connection"
+	"github.com/mysteriumnetwork/node/identity"
+	"github.com/mysteriumnetwork/node/session"
+	"github.com/mysteriumnetwork/payments/crypto"
+)
+
+// InvoiceFactoryCreator returns a payment engine factory
+func InvoiceFactoryCreator(dialog communication.Dialog, balanceSendPeriod, promiseTimeout time.Duration) func() (session.PaymentEngine, error) {
```
anjmao: It's a bit strange to see invoice and payments related logic in pingpong package.

session/pingpong/invoice_tracker.go
```diff
@@ -0,0 +1,176 @@
...
+	// TODO: fill the fields
+	return it.peerInvoiceSender.Send(crypto.Invoice{AgreementID: 1234})
+}
+
+func (it *InvoiceTracker) receiveExchangeMessageOrTimeout() error {
+	select {
+	case pm := <-it.exchangeMessageChan:
+		if res := pm.ValidateExchangeMessage(common.HexToAddress(it.peer.Address)); !res {
+			return ErrExchangeValidationFailed
+		}
```
anjmao: If ValidateExchangeMessage returns boolean when it's common to name variable ok so you know what this is a boolean result.

```ts
if ok := pm.ValidateExchangeMessage(common.HexToAddress(it.peer.Address)); !ok {
    return ErrExchangeValidationFailed
}
```

session/manager.go
vkuznecovas: makes sense, although we'll need to refactor the whole codebase for it. #399

session/pingpong/invoice_tracker.go
vkuznecovas: https://github.com/mysteriumnetwork/payments/issues/34

session/pingpong/factory.go
vkuznecovas: I'll add in docs in a separate PR documenting why this is the case.

PR #1303, opened 2019-09-13
Add go modules support

Waldz:
This PR migrates dep vendor to native go modules support. 

There is one problem which need to be fixed in separate PR. Go mobile doesn't support modules yet so I did a temporary hack with vendoring. After go modules are installed it is possible to vendor whem and then go mobile can use vendor folder and GOPATH to build bindings. Also gomobile uses xgo which itself uses older go1.11 version which should be updated to xgo 1.13.1 too.


anjmao:
@Waldz I looked at the CI logs. In gitlab-ci.yaml before_script runs `source build/env.sh`  which is generated prepare:env. Isn't it that before_script block runs before prepare:env? If you don't know quick answer don't spend time on it, I will try to revert changes line by line and check what is wrong.

CONTRIBUTING.md
```diff
@@ -121,21 +120,23 @@ bin/swagger_generate
 
 ## Dependency management
 
+Dependencies are managed by go modules. Make sure you have at least go1.11 installed.
+
 * Install project's frozen packages
 ```bash
-dep ensure
+go mod download
```
tadovas: Do we really need to make this step explicit? All go commands respect modules and will download any missing deps.

tadaskay:
> @Waldz I looked at the CI logs. In gitlab-ci.yaml before_script runs `source build/env.sh` which is generated prepare:env. Isn't it that before_script block runs before prepare:env? If you don't know quick answer don't spend time on it, I will try to revert changes line by line and check what is wrong. @anjmao 

`before_script` applies to every job in pipeline definition, effectively sourcing env variables before every job script(s). `prepare:env` overrides the `before_script, skipping the source, because it generates that file.

tadaskay:
I'm seeing some irrelevant commits in this PR, such as https://github.com/mysteriumnetwork/node/pull/1303/commits/a24700c2649f1aa9bb92afedb0cfb191b267eb41

Rebase is preferred. Check out our [devbook](https://mysteriumnetwork.github.io/devbook/git/)

.gitlab-ci.yml
```diff
@@ -33,18 +34,18 @@ after_script:
 env:
   stage: prepare
   tags: [go]
+  before_script:
+    - go mod download
```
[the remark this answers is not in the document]
anjmao: Yes, I will remove it later, as you mentioned I found the issue with before_script block too.

.gitlab-ci.yml
```diff
@@ -20,9 +20,10 @@ variables:
   GO_PACKAGE: github.com/mysteriumnetwork/node
   GIT_CLONE_PATH: /home/gitlab-runner/go/src/$GO_PACKAGE
   GOFLAGS: "-count=1" # Supersedes GOCACHE=off, see: https://github.com/golang/go/issues/29378#issuecomment-449383809
+  GO111MODULE: "on"
+  GOPROXY: "https://proxy.golang.org"
```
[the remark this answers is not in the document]
anjmao: GO111MODULE=on and GOPROXY=https://proxy.golang.org are temporal and will not be needed when using go1.13 as it uses GOPROXY be default. This proxy is a google hosted databases + cdn of all modules so you get faster download speed and better security.

.gitlab-ci.yml
```diff
@@ -33,18 +34,18 @@ after_script:
...
-    - dep ensure -v
   script: go run mage.go -v GenerateEnvFile
 
 checks:
   stage: test
   tags: [go]
   script:
     - |
-      go get -u \
+      GO111MODULE=off go get -u \
```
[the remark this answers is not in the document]
anjmao: Since the source code is under GOPATH and Go modules are turned on go get will update local go.mod and go.sum file with all indirect dependencies which are used for these tools. By addding GO111MODULE=off it will preserve current behaviour of installing  tools to global GOPATH/bin.

anjmao:
Few notes: https://github.com/karalabe/xgo  doesn't support go modules, but there is active fork https://github.com/techknowlogick/xgo with added support, but it doesn't look to be working on mac. I'm going to fix issues related to xgo in a few upcoming days and then continue here.

tadaskay:
@anjmao That sounds great! I was in fact mostly worried about xgo issues when upgrading. If you can solve that, that’ll make things easier. You’re right about go runners having 1.11 pre-installed. The update would have to take in a private repo, but you need to have the NDA signed beforehand. Other projects don’t use docker for E2E tests so they can use docker runner, which is easy to upgrade by changing docker image tag in `.gitlab-ci.yml`. In case of `node`, that’s a bit more complicated. Please follow up with HR regarding NDA. Thanks again!

anjmao:
Found working solution which requires minimum changes. Using tool https://github.com/nomad-software/vend it is possible to vendor all dependencies including c files. Vendor step need to be done only during CI pipeline.

.gitlab-ci.yml
```diff
@@ -100,11 +99,6 @@ package:windows-amd64:
   tags: [go]
   script: go run mage.go -v PackageWindowsAmd64
 
-package:ios:
```
anjmao: As we don't have ios for now there is no point to package it.

bin/build
```diff
@@ -16,12 +16,13 @@
 #> brew install readelf
 #> readelf -d build/myst/myst
 
-. ./bin/helpers/functions.sh
-. ./bin/helpers/output.sh
+source bin/helpers/functions.sh
```
anjmao: In other shell files helpers are linked using source. Changed to be consistent.

bin/build_xgo
```diff
@@ -23,21 +23,23 @@ if [ -z "$XGO_TARGETS" ]; then
...
+    -v "$PWD":/source \
+    -e PACK=./cmd/mysterium_node \
+    -e OUT=myst \
+    -e FLAG_V=false \
+    -e FLAG_X=false \
+    -e FLAG_RACE=false \
+    -e FLAG_LDFLAGS="-w -s $(get_linker_ldflags)" \
+    -e FLAG_BUILDMODE=default \
+    -e TARGETS="$XGO_TARGETS" \
+    anjmao/xgo:go-1.13.1
```
anjmao: 1. I removed dependency on xgo tool as all it does is generates docker run command string.
2. This image **anjmao/xgo:go-1.13.1** is my custom fork of xgo with added go modules support. Ideally it should land to our mysteriumnetwork/xgo repo.

bin/builder_package
```diff
@@ -13,7 +13,7 @@ set -e
 # Build image and push to docker hub:
 #> bin/builder_package latest
 
-DOCKER_IMAGE="mysteriumnetwork/mysterium-node-builder"
+DOCKER_IMAGE="anjmao/mysterium-node-builder"
```
anjmao: This is temporary. After I get push access to mysteriumnetwork I revert this change is separate PR.

bin/package_android
```diff
@@ -14,17 +14,21 @@ set -e
...
+    -v "$GOPATH"/src:/ext-go/1/src:ro \
+    -e OUT=Mysterium \
+    -e FLAG_V=false \
+    -e FLAG_X=false \
+    -e FLAG_RACE=false \
+    -e FLAG_LDFLAGS="-w -s $(get_linker_ldflags)" \
+    -e FLAG_BUILDMODE=default \
+    -e TARGETS=android/. \
+    -e EXT_GOPATH=/ext-go/1 \
+    mysteriumnetwork/xgo:1.11 github.com/mysteriumnetwork/node/mobile/mysterium
```
anjmao: Here you can see that android build uses mysteriumnetwork/xgo:1.11 with xgo:1.11. I will fix it in separate PR as it will require to update this image to go1.13 and find gomobile commit which was still compatible with this android.

bin/builder_package
tadaskay: I've added you to the owners in our org

bin/package_android
tadaskay: Does this mean we can't use 1.13 features without breaking android build?

.gitlab-ci.yml
mdomasevicius: :rage:

PR #1308, opened 2019-09-17
Payment pingpong storage

vkuznecovas:
Added three types of storage:

1) invoice storage for consumer and provider
2) accountant promise storage for provider
3) promised totals storage for consumer

These are going to be used for the new payment flow.

### DIFF
--- a/session/pingpong/accountant_promise_storage.go
+++ b/session/pingpong/accountant_promise_storage.go
@@ -1,0 +1,64 @@
+/*
+ * Copyright (C) 2019 The "MysteriumNetwork/node" Authors.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package pingpong
+
+import (
+	"sync"
+
+	"github.com/mysteriumnetwork/node/identity"
+	"github.com/mysteriumnetwork/payments/crypto"
+	"github.com/pkg/errors"
+)
+
+const accountantPromiseBucketName = "accountant_promises"
+
+// AccountantPromiseStorage allows for storing of accountant promises
+type AccountantPromiseStorage struct {
+	lock sync.Mutex
+	bolt persistentStorage
+}
+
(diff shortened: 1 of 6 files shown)


session/pingpong/invoice_storage.go
```diff
@@ -0,0 +1,127 @@
...
+	SetValue(bucket string, key interface{}, to interface{}) error
+}
+
+// InvoiceStorage allows to store promises
+type InvoiceStorage struct {
+	bolt persistentStorage
+	lock sync.Mutex
+}
+
+// ErrNotFound represents an error that indicates that there's no such invoice
```
soffokl: I'd move it to the top of the file.

session/pingpong/consumer_totals_storage.go
```diff
@@ -0,0 +1,73 @@
...
+	"github.com/mysteriumnetwork/payments/crypto"
+	"github.com/pkg/errors"
+)
+
+const consumerTotalStorageBucketName = "consumer_promised_totals"
+
+// ConsumerTotalsStorage allows to store total promised amounts for each channel
+type ConsumerTotalsStorage struct {
+	bolt persistentStorage
+	lock sync.Mutex
```
tadaskay: Since there is only one lock, what do you think about simply embedding it, so you can `cts.Lock()` and `cts.Unlock()`?

session/pingpong/consumer_totals_storage.go
```diff
@@ -0,0 +1,73 @@
...
+
+// Store stores the given amount as promised for the given channel
+func (cts *ConsumerTotalsStorage) Store(channelAddress string, amount uint64) error {
+	cts.lock.Lock()
+	defer cts.lock.Unlock()
+	return cts.bolt.SetValue(consumerTotalStorageBucketName, channelAddress, amount)
+}
+
+// GenerateAndStore generates the channel address and stores the given amount as promised
+func (cts *ConsumerTotalsStorage) GenerateAndStore(identity, registry, channelImplementation string, amount uint64) error {
```
tadaskay: Perhaps extract params to a struct? Lots of positional params of the same type.

session/pingpong/accountant_promise_storage.go
```diff
@@ -0,0 +1,64 @@
...
+	"github.com/mysteriumnetwork/node/identity"
+	"github.com/mysteriumnetwork/payments/crypto"
+	"github.com/pkg/errors"
+)
+
+const accountantPromiseBucketName = "accountant_promises"
+
+// AccountantPromiseStorage allows for storing of accountant promises
+type AccountantPromiseStorage struct {
+	lock sync.Mutex
```
tadaskay: Same question as for consumer storage mutex.

session/pingpong/consumer_totals_storage.go
soffokl: Embedding it will make `Lock()` and `Unlock()` publically available, and someone will use it one day outside and will break everything.

tadaskay: Good catch

PR #1318, opened 2019-10-02
Report node type and country in proposal events

soffokl:


core/quality/morqa_transport.go
```diff
@@ -109,11 +109,14 @@ func sessionDataToMetricsEvent(context sessionDataContext) *metrics.Event {
 }
 
 func proposalEventToMetricsEvent(context market.ServiceProposal, info appInfo) *metrics.Event {
+	location := context.ServiceDefinition.GetLocation()
 	return &metrics.Event{
```
anjmao: Just a question. Have you guys thought about using prometheus instrumentations for metrics?

soffokl: It's Prometheus at the end of the chain. On the node level, we need some abstract push-based metrics, which can be compatible with any solution.

PR #1320, opened 2019-10-03
Option to limit openvpn bandwidth

### LINKED ISSUE #1273: Ability to limit node bandwidth to 5Mbps

tadaskay:
Fixes #1273 

* Based on https://github.com/magnific0/wondershaper ; it'll work on linux machines, specifically on raspbian based mystberry image.
* Enabled by config flag `shaper.enabled` (bool); off by default
* Limits to 5000Kbps
* Shaper starts when openvpn service is started (on the tunnel interface)
* It reacts to configuration updates live, thus API for setting user configuration can be re-used

Still some final testing needed on a physical raspberry (todo tomorrow)

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -22,6 +22,7 @@ import (
 	"strings"
 
 	"github.com/BurntSushi/toml"
+	"github.com/mysteriumnetwork/node/eventbus"
 	"github.com/mysteriumnetwork/node/logconfig"
 	"github.com/mysteriumnetwork/node/utils/jsonutil"
 	"github.com/pkg/errors"
@@ -30,12 +31,18 @@ import (
 
 var log = logconfig.NewLogger()
 
+// Topic returns event bus topic for the given config key to listen for its updates
+func Topic(configKey string) string {
+	return "config:" + configKey
+}
+
 // Config stores app configuration: default values + user configuration + CLI flags
 type Config struct {
 	userConfigLocation string
 	defaults           map[string]interface{}
 	user               map[string]interface{}
 	cli                map[string]interface{}
+	eventBus           eventbus.EventBus
 }
 
 // Current global configuration instance
@@ -57,6 +64,11 @@ func NewConfig() *Config {
 
 func (cfg *Config) userConfigLoaded() bool {
 	return cfg.userConfigLocation != ""
+}
+
+// EnableEventPublishing enables config event publishing to the event bus
+func (cfg *Config) EnableEventPublishing(eb eventbus.EventBus) {
+	cfg.eventBus = eb
 }
 
 // LoadUserConfig loads and remembers user config location
@@ -110,6 +122,9 @@ func (cfg *Config) SetDefault(key string, value interface{}) {
 
 // SetUser sets user configuration value for key
 func (cfg *Config) SetUser(key string, value interface{}) {
+	if cfg.eventBus != nil {
+		cfg.eventBus.Publish(Topic(key), value)
+	}
 	cfg.set(&cfg.user, key, value)
 }
 
--- a/core/shaper/shaper.go
+++ b/core/shaper/shaper.go
@@ -1,0 +1,125 @@
+/*
+ * Copyright (C) 2019 The "MysteriumNetwork/node" Authors.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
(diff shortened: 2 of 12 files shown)


core/shaper/shaper.go
```diff
@@ -0,0 +1,131 @@
...
+	"github.com/mysteriumnetwork/node/eventbus"
+	"github.com/mysteriumnetwork/node/logconfig"
+	"github.com/mysteriumnetwork/node/services/shared"
+	"github.com/pkg/errors"
+)
+
+var log = logconfig.NewLogger()
+
+const (
+	limitKbps = 5000
```
Waldz: I remember, we talked about default of 6Mbps

services/shared/options.go
```diff
@@ -50,11 +59,13 @@ func Configure(ctx *cli.Context) {
 }
 
 func configureDefaults() {
-	config.Current.SetDefault("access-policy.list", "")
+	config.Current.SetDefault(accessPoliciesFlag.Name, "")
+	config.Current.SetDefault(ShaperEnabledFlag.Name, false)
```
Waldz: Why aren't we enabling it by default?

core/shaper/shaper.go
```diff
@@ -0,0 +1,131 @@
...
+ */
+
+package shaper
+
+import (
+	"net"
+	"os/exec"
+	"strconv"
+
+	"github.com/magefile/mage/sh"
```
soffokl: Do we really want to have this `magefile/mage/sh` here? We have our own tools for calling external commands.

soffokl:
Just an idea for the future. I'd like to see some scheduler for the shaper to disable it at night and enable for working hours. 
The best option is integration with https://ifttt.com, to enable custom scenario to node manipulation. i.e. stop the shaper when I'm leaving home, and start it when I'm back.

services/shared/options.go
tadaskay: Why should we?

core/shaper/shaper.go
tadaskay: After reconsidering it, 5Mbps should be enough for Netflix.

services/openvpn/service/manager.go
```diff
@@ -149,6 +151,11 @@ func (m *Manager) Serve(providerID identity.Identity) (err error) {
 		}
 	}()
 
+	err = m.shaper.Start(m.vpnServer.DeviceName())
+	if err != nil {
+		log.Error("Could not start traffic shaper: ", err)
```
anjmao: Is it safe to call m.vpnServer.Wait() if you have error here?

core/shaper/shaper.go
```diff
@@ -0,0 +1,131 @@
...
+	if enabled {
+		log.Info("Shaper enabled, limiting bandwidth")
+		return s.limitBandwidth()
+	}
+	log.Info("Shaper disabled, removing bandwidth limit")
+	return s.unlimitBandwidth()
+}
+
+func (s *wonderShaper) limitBandwidth() error {
+	iface := "tun0"
```
anjmao: Pass inferface name.

core/shaper/shaper.go
```diff
@@ -0,0 +1,131 @@
...
+	err := s.runCmd(iface, strconv.Itoa(limitKbps), strconv.Itoa(limitKbps))
+	return errors.Wrap(err, "could not limit bandwidth on "+iface)
+}
+
+func (s *wonderShaper) unlimitBandwidth() error {
+	iface := "tun0"
```
anjmao: Pass interface name.

services/openvpn/service/manager.go
tadaskay: Yes

core/shaper/shaper.go
tadaskay: We don't have equivalent tools for creating a command

tadaskay: I'll copy over from the `go-ci`

config/config.go
```diff
@@ -59,6 +66,11 @@ func (cfg *Config) userConfigLoaded() bool {
 	return cfg.userConfigLocation != ""
 }
 
+// EnableEventPublishing enables config event publishing to the event bus
+func (cfg *Config) EnableEventPublishing(eb eventbus.EventBus) {
+	cfg.eventBus = eb
```
vkuznecovas: why though? why not just inject this as a dependency and no have the nil check in set user?

tadaskay: Not nil check is just a safeguard. This essentially is the dependency injection, because `config.Current` is a one global instance.

Issue #1319, opened 2019-10-03
Cache discovery API proposals http response


anjmao:
Currently we use quite a lot of bandwidth for fetching proposals from discovery api. Possible solution is to return 304 status code (Not Modified) if proposals was not changed so client do not need to refetch them again.

PR #1321, opened 2019-10-03
Send etag while pulling proposals

anjmao:
Fixes https://github.com/mysteriumnetwork/node/issues/1319

### DIFF
--- a/market/mysterium/mysterium_api.go
+++ b/market/mysterium/mysterium_api.go
@@ -43,9 +43,10 @@ type MysteriumAPI struct {
 	http                requests.HTTPTransport
 	discoveryAPIAddress string
 
-	latestProposalsMux  sync.Mutex
-	latestProposalsEtag string
-	latestProposals     []market.ServiceProposal
+	latestProposalsEtagMux sync.RWMutex
+	latestProposalsEtag    string
+	latestProposalsMux     sync.RWMutex
+	latestProposals        []market.ServiceProposal
 }
 
 // NewClient creates Mysterium centralized api instance with real communication
@@ -244,10 +245,7 @@ func (mApi *MysteriumAPI) QueryProposals(query ProposalsQuery) ([]market.Service
 	if err != nil {
 		return nil, err
 	}
-
-	mApi.latestProposalsMux.Lock()
-	defer mApi.latestProposalsMux.Unlock()
-	req.Header.Add("If-None-Match", mApi.latestProposalsEtag)
+	req.Header.Add("If-None-Match", mApi.getLatestProposalsEtag())
 
 	res, err := mApi.http.Do(req)
 	if err != nil {
@@ -256,7 +254,7 @@ func (mApi *MysteriumAPI) QueryProposals(query ProposalsQuery) ([]market.Service
 	defer res.Body.Close()
 
 	if res.StatusCode == http.StatusNotModified {
-		return mApi.latestProposals, nil
+		return mApi.getLatestProposals(), nil
 	}
 
 	if err := requests.ParseResponseError(res); err != nil {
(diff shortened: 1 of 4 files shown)


market/mysterium/mysterium_api.go
```diff
@@ -235,12 +238,28 @@ func (mApi *MysteriumAPI) QueryProposals(query ProposalsQuery) ([]market.Service
...
+	if err := requests.ParseResponseError(res); err != nil {
+		return nil, err
+	}
+
+	var proposalsResponse ProposalsResponse
+	if err := requests.ParseResponseJSON(res, &proposalsResponse); err != nil {
+		return nil, fmt.Errorf("cannot parse proposals response: %v", err)
+	}
+
+	mApi.latestProposalsEtag = res.Header.Get("ETag")
```
soffokl: This `latestProposalsEtag` does not thread-safe. It can be called from different places at the same time.

market/mysterium/mysterium_api.go
```diff
@@ -235,12 +238,28 @@ func (mApi *MysteriumAPI) QueryProposals(query ProposalsQuery) ([]market.Service
...
+		return nil, errors.New("node already contains newest proposals")
+	}
+
+	if err := requests.ParseResponseError(res); err != nil {
+		return nil, err
+	}
+
+	var proposalsResponse ProposalsResponse
+	if err := requests.ParseResponseJSON(res, &proposalsResponse); err != nil {
+		return nil, fmt.Errorf("cannot parse proposals response: %v", err)
```
soffokl: Maybe it's better to `errors.Wrap` the error?

anjmao: Yep, will do.

market/mysterium/mysterium_api.go
anjmao: Thanks. Initially I thought that only fetcher can pull proposals but now I see that there are more places. My current solution is not going to work. I propose that I modify proposals [fetcher.go](https://github.com/mysteriumnetwork/node/blob/master/core/discovery/api/fetcher.go) and hold etag state inside this struct. @soffokl Does that make sense?

soffokl: Fetcher is using mysteriumAPI `QueryProposals` too. So it's enough to keep it only here. My suggestion is just to protect this field by the mutex to avoid unexpected panic.

market/mysterium/mysterium_api.go
```diff
@@ -235,12 +238,28 @@ func (mApi *MysteriumAPI) QueryProposals(query ProposalsQuery) ([]market.Service
...
-	err = mApi.http.DoRequestAndParseResponse(req, &proposalsResponse)
+	res, err := mApi.http.Do(req)
 	if err != nil {
 		return nil, err
 	}
+	defer res.Body.Close()
+
+	if res.StatusCode == http.StatusNotModified {
+		return nil, errors.New("node already contains newest proposals")
+	}
```
tadaskay: This doesn't look like an error case to me. It should simply encapsulate the Not Modified case, but to the outside, it would appear as a simple proposal list.

market/mysterium/mysterium_api.go
```diff
@@ -236,13 +245,33 @@ func (mApi *MysteriumAPI) QueryProposals(query ProposalsQuery) ([]market.Service
...
+	if err := requests.ParseResponseError(res); err != nil {
 		return nil, err
 	}
+
+	var proposalsResponse ProposalsResponse
+	if err := requests.ParseResponseJSON(res, &proposalsResponse); err != nil {
+		return nil, errors.Wrap(err, "cannot parse proposals response")
+	}
+
+	mApi.latestProposalsEtag = res.Header.Get("ETag")
```
vkuznecovas: I'm not sure about the way you're locking here. You're potentially keeping another routine waiting for 20 seconds(see timeout). The way i'd do it, is extract 2 methods - getLatestEtag and setLatestEtag. Both of those internally would lock the underlying field before getting/setting it. The methods would only be called when needed, preventing excess waiting time for other routines. But otherwise - does look good.

zolia:
Do I understand correctly that StatusNotModified would be returned from our Discovery and that it has to be implemented there still? Do we have a task for it? This etag will be some sort of hash?

anjmao:
@zolia I implemented changes in discovery api already https://github.com/mysteriumnetwork/api/blob/master/api/proposals.py#L151

zolia:
Cool, now I see. Thanx.

anjmao:
@zolia It will be backwards compatible.

PR #1325, opened 2019-10-07
Payment v2 messaging

vkuznecovas:
Added the off chain communications for the v2 version of payments. They are backwards compatible with the old pingpong mechanism, meaning that there should be no breaking changes for existing consumers.


core/node/options_transactor.go
```diff
@@ -21,5 +21,5 @@ package node
 type OptionsTransactor struct {
 	TransactorEndpointAddress string
 	RegistryAddress           string
-	AccountantID              string
+	ChannelImplementation     string
```
anjmao: ChannelImplementation field naming is not clear, is it some kind of standard? Maybe it can be named ChannelID?

session/pingpong/accountant_caller.go
```diff
@@ -0,0 +1,69 @@
...
+	}
+}
+
+// RequestPromise requests a promise from accountant
+func (ac *AccountantCaller) RequestPromise(em crypto.ExchangeMessage) (crypto.Promise, error) {
+	req, err := requests.NewPostRequest(ac.accountantBaseURI, "/request_promise", em)
+	if err != nil {
+		return crypto.Promise{}, errors.Wrap(err, "could not form request_promise request")
+	}
+	var resp crypto.Promise
```
anjmao: Can it be changed to return pointer `*crypto.Promise`? I saw there is crypto.CreatePromise which returns pointer so maybe we should keep it the same.

session/pingpong/accountant_caller.go
```diff
@@ -0,0 +1,69 @@
...
+
+// RevealObject represents the reveal request object
+type RevealObject struct {
+	R           string
+	Provider    string
+	AgreementID uint64
+}
+
+// RevealR reveals the R for the accountant
+func (ac *AccountantCaller) RevealR(r string, provider string, agreementID uint64) error {
```
anjmao: You can join arguments with the same type `r, provider string, agreementID uint64`. Examples from go std https://github.com/golang/go/blob/master/src/bytes/bytes.go#L18

session/pingpong/exchange_message_tracker.go
```diff
@@ -27,11 +33,35 @@ import (
 	"github.com/pkg/errors"
 )
 
+// ErrWrongProvider represents an issue where the wrong provider is supplied
```
anjmao: Minor thing. I suggest to end each comment with a dot to end sentence because you can have more than one sentence it comment block. 

Example from Godoc https://blog.golang.org/godoc-documenting-go-code
Example from go std https://github.com/golang/go/blob/master/src/bytes/bytes.go#L439

session/pingpong/exchange_message_tracker.go
```diff
@@ -59,32 +117,135 @@ var ErrInvoiceMissmatch = errors.New("invoice mismatch")
...
+
+func (emt *ExchangeMessageTracker) getGrandTotalPromised() (uint64, error) {
+	res, err := emt.consumerTotalsStorage.Get(grandTotalKey)
+	if err != nil {
+		if err == ErrNotFound {
+			log.Debug("no previous invoice grand total, assuming zero")
+			return 0, nil
+		}
+	}
+	return res, errors.Wrap(err, "could not get previous grand total")
```
anjmao: It would be more clear if you put this error inside if err != nil

```go
if err != nil {
   if err == ErrNotFound {
      log.Debug("no previous invoice grand total, assuming zero")
      return 0, nil
   }
   return 0, errors.Wrap(err, "could not get previous grand total")
}
return res, nil
```

session/pingpong/exchange_message_tracker.go
```diff
@@ -59,32 +117,135 @@ var ErrInvoiceMissmatch = errors.New("invoice mismatch")
...
+}
+
+func (emt *ExchangeMessageTracker) incrementGrandTotalPromised(amount uint64) error {
+	res, err := emt.consumerTotalsStorage.Get(grandTotalKey)
+	if err != nil {
+		if err == ErrNotFound {
+			log.Debug("no previous invoice grand total, assuming zero")
```
anjmao: In this case if err ==  ErrNotFound  you will still call `emt.consumerTotalsStorage.Store(grandTotalKey, res+amount)`  Maybe you need to return here?

session/pingpong/factory.go
```diff
@@ -78,13 +112,29 @@ func BackwardsCompatibleExchangeFactoryFunc(keystore *keystore.KeyStore, options
...
+				InvoiceChan:               invoices,
+				PeerExchangeMessageSender: NewExchangeSender(dialog),
+				ConsumerInvoiceStorage:    invoiceStorage,
+				ConsumerTotalsStorage:     totalStorage,
+				TimeTracker:               &timeTracker,
+				Ks:                        keystore,
+				Identity:                  consumer,
+				Peer:                      dialog.PeerID(),
+				PaymentInfo: dto.PaymentPerTime{
+					Price:    money.NewMoney(1, money.CurrencyMyst),
```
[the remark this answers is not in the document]
zolia: Thats just for the moment. In reality this parameter will be tunable by provider. 
As a first step we will use strictly defined value to calculate the amount of service that was supplied by concrete provider and figure out payouts from this during testnet period..

core/node/options_transactor.go
chompomonim: ChannelImplementation address is an address in blockchain where channel implementation code is hosted.

All deployed channels are pointing into this smart contract. So it's definitely not ChannelID.

session/pingpong/invoice_tracker.go
```diff
@@ -143,18 +242,74 @@ func (it *InvoiceTracker) handleExchangeMessageReceiveError(err error) error {
...
+		it.resetAccountantFailureCount()
+		err = it.accountantPromiseStorage.Store(it.accountantID, promise)
+		if err != nil {
+			return errors.Wrap(err, "could not store accountant promise")
+		}
+		log.Debug("accountant promise stored")
+		hexR := hex.EncodeToString(it.lastInvoice.r)
+		err = it.accountantCaller.RevealR(hexR, it.providerID.Address, it.lastInvoice.invoice.AgreementID)
+		if err != nil {
+			// TODO: need to think about handling this a bit better
```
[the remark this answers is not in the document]
chompomonim: Provider should check failure message and act accordingly. Possible scenarios:

 - 500 of any kind or server is not responding --> retry in one minute
 - 400 "R have to be provided as proper hex" --> ??? don't know how it could happen  here
 - 400 of any other kind --> try to `request_promise` (with same exchange message) once again and try revealing R again.

P.S. should we have all that messages somewhere in one package as constant and reuse in node and in accountant?

session/pingpong/exchange_message_tracker.go
```diff
@@ -59,32 +117,135 @@ var ErrInvoiceMissmatch = errors.New("invoice mismatch")
...
+
+	return nil
+}
+
+func (emt *ExchangeMessageTracker) calculateAmountToPromise(invoice crypto.Invoice) (toPromise uint64, diff uint64, err error) {
+	previous, err := emt.consumerInvoiceStorage.Get(emt.peer)
+	if err != nil {
+		if err == ErrNotFound {
+			// do nothing, really
+			log.Debug("no previous invoice found, assuming zero")
```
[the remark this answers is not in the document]
vkuznecovas: yup, we're just starting something new here, so no need to do anything really.

session/pingpong/invoice_storage.go
```diff
@@ -109,6 +133,52 @@ func (is *InvoiceStorage) StoreInvoice(bucket string, identity identity.Identity
...
+// GetNewAgreementID generates a new agreement id
+func (is *InvoiceStorage) GetNewAgreementID() (uint64, error) {
+	is.lock.Lock()
+	defer is.lock.Unlock()
+
+	var res uint64
+	err := is.bolt.GetValue(string(agreementIDCounter), "provider_agreement_id", &res)
+	if err != nil {
+		if err.Error() == errBoltNotFound {
+			res = 1
```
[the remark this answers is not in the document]
vkuznecovas: it is, they are supposed to be incremented one by one.

session/pingpong/invoice_tracker.go
```diff
@@ -50,36 +79,95 @@ type InvoiceTracker struct {
...
 
+func (it *InvoiceTracker) generateInitialInvoice() error {
+	agreementID, err := it.invoiceStorage.GetNewAgreementID()
+	if err != nil {
+		return errors.Wrap(err, "could not get new agreement id")
+	}
+	// TODO: set fee
+	r := make([]byte, 64)
+	rand.Read(r)
+	invoice := crypto.CreateInvoice(agreementID, it.paymentInfo.GetPrice().Amount, 0, r)
```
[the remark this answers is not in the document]
vkuznecovas: it is

session/pingpong/invoice_tracker.go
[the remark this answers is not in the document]
vkuznecovas: #1346

session/create_consumer.go
```diff
@@ -69,7 +69,7 @@ func (consumer *createConsumer) Consume(requestPtr interface{}) (response interf
 	issuerID := consumer.peerID
 	if request.ConsumerInfo != nil {
 		issuerID = request.ConsumerInfo.IssuerID
-		if request.ConsumerInfo.PaymentVersion == PaymentVersionV2 {
+		if request.ConsumerInfo.PaymentVersion == PaymentVersionV3 {
```
chompomonim: Why it is now called `PaymentVersionV3`?

vkuznecovas: because we've already introduced version 2 into master a while ago, since this is incompatible, we need to switch to 3

Issue #1326, opened 2019-10-07
Random pipeline failures with not possible to fast-forward, aborting error


anjmao:
Sometimes jobs are failing due to git error when fetching golint tool which depends on golang.org/x/tools.

```
Runtime platform                                    arch=amd64 os=linux pid=3714 revision=d0b76032 version=12.0.2
Downloading artifacts from coordinator... ok        id=313589953 responseStatus=200 OK token=G4uSyQU6
WARNING: build/env.sh: lchown build/env.sh: operation not permitted (suppressing repeats) 
$ dep ensure -v
(1/94) Wrote github.com/gin-contrib/sse@master
(2/94) Wrote github.com/jbenet/go-context@master
(3/94) Wrote github.com/google/go-github@v26.1.3
(4/94) Wrote github.com/json-iterator/go@v1.1.6
(5/94) Wrote github.com/jackpal/gateway@v1.0.5
(6/94) Wrote github.com/jmespath/go-jmespath@c2b33e84
(7/94) Wrote github.com/hashicorp/go-cleanhttp@v0.5.1
(8/94) Wrote github.com/gin-gonic/gin@v1.4.0
(9/94) Wrote github.com/go-stack/stack@v1.7.0
(10/94) Wrote github.com/gofrs/uuid@v3.1.1
(11/94) Wrote github.com/golang/snappy@master
(12/94) Wrote github.com/huin/goupnp@v1.0.0
(13/94) Wrote github.com/jackpal/go-nat-pmp@v1.0.1
(14/94) Wrote github.com/golang/protobuf@v1.3.1
(15/94) Wrote github.com/google/go-querystring@v1.0.0
(16/94) Wrote gopkg.in/yaml.v2@v2.2.2
(17/94) Wrote github.com/hashicorp/go-retryablehttp@v0.5.4
(18/94) Wrote github.com/julienschmidt/httprouter@v1.1
(19/94) Wrote github.com/kevinburke/ssh_config@0.5
(20/94) Wrote github.com/koron/go-ssdp@4a0ed625a78b6858dc8d3a55fb7728968b712122
(21/94) Wrote github.com/mattn/go-runewidth@v0.0.4
(22/94) Wrote github.com/mattn/go-isatty@v0.0.3
(23/94) Wrote github.com/mdlayher/genetlink@master
(24/94) Wrote github.com/mdlayher/netlink@master
(25/94) Wrote github.com/mholt/archiver@v3.1.1
[...]

PR #1329, opened 2019-10-07
Convert common checks to go-ci and mage

anjmao:
Fixes https://github.com/mysteriumnetwork/node/issues/1326


ci/check/check.go
```diff
@@ -0,0 +1,85 @@
...
+func CheckCopyright() error {
+	return commands.Copyright("./...", "docs")
+}
+
+// CheckGoLint reports linting errors in the solution.
+func CheckGoLint() error {
+	return commands.GoLint("./...", "docs")
+}
+
+// CheckGoVet Checks that the source is compliant with go vet.
```
soffokl:
```suggestion
// CheckGoVet checks that the source is compliant with go vet.
```

Issue #1297, opened 2019-09-10
Consumer reports connection status to provider


zolia:
Node runner needs to see if his node is reachable from outside.

In order to reliably determine whether connection succeeded or failed we need consumer to inform provider on connection initiation result. 

Simple attempt to determine public IP change (with ipify or other means) from consumer side should be enough to state that connection works. Send connection status through communication channel.

PR #1337, opened 2019-10-10
Report conn status

anjmao:
Closes https://github.com/mysteriumnetwork/node/issues/1297

- This PR adds ability to send session connectivity status via established dialog (message broker).
- When peer receives status it is inserted to in memory storage so it can be fetched via tequila API.

### DIFF
--- a/communication/dialog.go
+++ b/communication/dialog.go
@@ -1,0 +1,54 @@
+/*
+ * Copyright (C) 2017 The "MysteriumNetwork/node" Authors.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package communication
+
+import (
+	"github.com/mysteriumnetwork/node/identity"
+	"github.com/mysteriumnetwork/node/market"
+)
+
+// Dialog represent established connection between 2 peers in network.
+// Enables bidirectional communication with another peer.
+type Dialog interface {
+	PeerID() identity.Identity
+	Sender
+	Receiver
+	Close() error
+}
+
+// DialogWaiter defines server which:
+//   - waits and serves incoming dialog requests
+//   - negotiates with Dialog initiator
+//   - finally creates Dialog, when it is accepted
+type DialogWaiter interface {
+	Start() (market.Contact, error)
+	Stop() error
+	ServeDialogs(DialogHandler) error
+}
+
+// DialogHandler defines how to handle incoming Dialog
+type DialogHandler interface {
+	Handle(Dialog) error
+}
+
+// DialogEstablisher interface defines client which:
+//   - initiates Dialog requests to network
+//   - creates Dialog, when it is negotiated
+type DialogEstablisher interface {
+	EstablishDialog(peerID identity.Identity, peerContact market.Contact) (Dialog, error)
+}
--- a/communication/nats/dialog/codec_secured.go
+++ b/communication/nats/dialog/codec_secured.go
@@ -50,12 +50,12 @@ type codecSecured struct {
 func (codec *codecSecured) Pack(payloadPtr interface{}) ([]byte, error) {
 	payloadData, err := codec.codecPacker.Pack(payloadPtr)
 	if err != nil {
-		return []byte{}, err
+		return []byte{}, errors.Wrap(err, "could not pack payload data")
 	}
 
 	signature, err := codec.signer.Sign(payloadData)
 	if err != nil {
-		return []byte{}, err
+		return []byte{}, errors.Wrap(err, "could not sign payload data")
 	}
 
 	return codec.codecPacker.Pack(&messageEnvelope{
--- a/communication/nats/dialog/dialog_waiter.go
+++ b/communication/nats/dialog/dialog_waiter.go
@@ -119,8 +119,7 @@ func (waiter *dialogWaiter) ServeDialogs(dialogHandler communication.DialogHandl
 	}
 	codec := NewCodecSecured(communication.NewCodecJSON(), waiter.signer, identity.NewVerifierSigned())
 	receiver := nats.NewReceiver(waiter.address.GetConnection(), codec, waiter.address.GetTopic())
-
-	return receiver.Respond(&dialogCreateConsumer{createDialog})
+	return receiver.Respond(&dialogCreateConsumer{Callback: createDialog})
 }
 
 func (waiter *dialogWaiter) newCodecForPeer(peerID identity.Identity) *codecSecured {
--- a/core/connection/manager.go
+++ b/core/connection/manager.go
@@ -154,7 +154,7 @@ func (manager *connectionManager) Connect(consumerID identity.Identity, proposal
 		return err
 	}
 
-	currentPublicIP := manager.getPublicIP()
+	originalPublicIP := manager.getPublicIP()
 	// Try to establish connection with peer.
 	err = manager.startConnection(connection, consumerID, proposal, params, sessionDTO, stateChannel, statisticsChannel)
 	if err != nil {
@@ -165,15 +165,15 @@ func (manager *connectionManager) Connect(consumerID identity.Identity, proposal
 		return err
 	}
 
-	manager.checkSessionIP(dialog, sessionDTO.ID, currentPublicIP)
+	manager.checkSessionIP(dialog, sessionDTO.ID, originalPublicIP)
 	return nil
 }
 
 // checkSessionIP checks if ip was changed after connection is established and notify peer.
-func (manager *connectionManager) checkSessionIP(dialog communication.Dialog, sessionID session.ID, currentPublicIP string) {
+func (manager *connectionManager) checkSessionIP(dialog communication.Dialog, sessionID session.ID, originalPublicIP string) {
 	newPublicIP := manager.getPublicIP()
-	if currentPublicIP == newPublicIP {
-		manager.sendSessionStatus(dialog, sessionID, connectivity.StatusSessionIpNotChanged, nil)
+	if originalPublicIP == newPublicIP {
+		manager.sendSessionStatus(dialog, sessionID, connectivity.StatusSessionIPNotChanged, nil)
 		return
 	}
 
--- a/core/connection/manager.go
+++ b/core/connection/manager.go
@@ -23,10 +23,12 @@ import (
 
 	"github.com/mysteriumnetwork/node/communication"
 	"github.com/mysteriumnetwork/node/consumer"
+	"github.com/mysteriumnetwork/node/core/ip"
 	"github.com/mysteriumnetwork/node/firewall"
 	"github.com/mysteriumnetwork/node/identity"
 	"github.com/mysteriumnetwork/node/market"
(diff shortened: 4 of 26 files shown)


communication/dialog.go
```diff
@@ -0,0 +1,38 @@
...
+// Enables bidirectional communication with another peer.
+type Dialog interface {
+	PeerID() identity.Identity
+	Sender
+	Receiver
+	Close() error
+}
+
+// DialogWaiter defines server which:
+//   - waits and servers incoming dialog requests
```
zolia:
```suggestion
//   - waits and serves incoming dialog requests
```

communication/nats/dialog/codec_secured.go
```diff
@@ -50,12 +50,12 @@ type codecSecured struct {
 func (codec *codecSecured) Pack(payloadPtr interface{}) ([]byte, error) {
 	payloadData, err := codec.codecPacker.Pack(payloadPtr)
 	if err != nil {
-		return []byte{}, err
+		return []byte{}, errors.Wrap(err, "could not pack payload data")
```
zolia: 👍

communication/nats/dialog/dialog_waiter.go
```diff
@@ -119,8 +120,20 @@ func (waiter *dialogWaiter) ServeDialogs(dialogHandler communication.DialogHandl
 	}
 	codec := NewCodecSecured(communication.NewCodecJSON(), waiter.signer, identity.NewVerifierSigned())
 	receiver := nats.NewReceiver(waiter.address.GetConnection(), codec, waiter.address.GetTopic())
+	return receiver.Respond(&dialogCreateConsumer{Callback:createDialog})
+}
 
-	return receiver.Respond(&dialogCreateConsumer{createDialog})
+// handleDialogStatus listens to peer dialog communication status and stores it local storage.
```
zolia:
```suggestion
// handleDialogStatus listens to peer dialog communication status and stores it to local storage.
```

communication/nats/dialog/dialog_waiter.go
```diff
@@ -119,8 +120,20 @@ func (waiter *dialogWaiter) ServeDialogs(dialogHandler communication.DialogHandl
...
+func (waiter *dialogWaiter) handleDialogStatus(dialog communication.Dialog, peerID string) {
+	cb := func(r *communication.DialogStatusRequest) error {
+		msg := fmt.Sprintf("Peer '%s' dialog status code: %d, message: %s", peerID, r.Code, r.Message)
+		log.Info(waiterLogPrefix, msg)
+		// TODO: Log dialog status to local storage so it can be displayed in UI later.
```
zolia: Not sure if we need it. Most likely we would need to store some last xxx sessions, just to troubleshoot things if needed. That can be saved in-memory only. Our policy is no-logs policy, such additional storage complicates issues with state storage and is also against our no-logs policy.

core/connection/manager.go
```diff
@@ -132,24 +132,46 @@ func (manager *connectionManager) Connect(consumerID identity.Identity, proposal
...
-		return ErrConnectionCancelled
+	if err != nil {
+		if err == context.Canceled {
+			return ErrConnectionCancelled
+		}
+		// TODO: Here we need to extract mode detailed error.
+		manager.sendDialogStatus(dialog, communication.StatusConnectionFailed, err.Error())
+		return err
+	}
+	manager.sendDialogStatus(dialog, communication.StatusConnectionOk, "")
```
zolia: At this point, when dialog establishes we still do not know anything about actual consumer connectivity status (check for ipify ip change). This dialog status logic you implemented can go within dialog establishment, but for actual consumer connectivity status you need to produce a separate message, i.e. as @vkuznecovas done in pingpong package.

communication/interface.go
```diff
@@ -17,55 +17,4 @@
...
-}
-
-// Sender represents interface for:
-//   - sending asynchronous messages
-//   - sending and HTTP-like request and waiting for response
-type Sender interface {
-	Send(producer MessageProducer) error
-	Request(producer RequestProducer) (responsePtr interface{}, err error)
-}
+// TODO: remove this file when done.
```
soffokl: Looks like this file should be removed.

core/connection/manager.go
```diff
@@ -137,19 +144,62 @@ func (manager *connectionManager) Connect(consumerID identity.Identity, proposal
...
+	currentPublicIP := manager.getPublicIP()
+	// Try to establish connection with peer.
 	err = manager.startConnection(connection, consumerID, proposal, params, sessionDTO, stateChannel, statisticsChannel)
-	if err == context.Canceled {
-		return ErrConnectionCancelled
+	if err != nil {
+		if err == context.Canceled {
+			return ErrConnectionCancelled
+		}
+		manager.sendSessionStatus(dialog, sessionDTO.ID, connectivity.StatusConnectionFailed, err)
```
anjmao: @zolia  Here we will get underlying connection error. As I remember we talked that in case of openvpn conn failure sometimes error is not clear. In order to get openvpn errors I will need to dig into our go-openvpn library. Probably it would be better to work on that in a separate task.

session/connectivity/status_storage.go
```diff
@@ -0,0 +1,64 @@
...
+	defer s.entriesMux.RUnlock()
+
+	return s.entries
+}
+
+func (s *statusStorage) AddStatusEntry(msg *StatusEntry) {
+	s.entriesMux.Lock()
+	defer s.entriesMux.Unlock()
+
+	s.entries = append(s.entries, msg)
```
anjmao: Running node for a long time entries eventually may consume more memory as this is in memory storage. As I have CreatedAtUTC field I could remove older entires (1 month old etc.) when calling AddStatusEntry.

core/connection/manager.go
```diff
@@ -137,19 +144,62 @@ func (manager *connectionManager) Connect(consumerID identity.Identity, proposal
...
+		return err
+	}
+
+	manager.checkSessionIP(dialog, sessionDTO.ID, currentPublicIP)
+	return nil
+}
+
+// checkSessionIP checks if ip was changed after connection is established and notify peer.
+func (manager *connectionManager) checkSessionIP(dialog communication.Dialog, sessionID session.ID, currentPublicIP string) {
```
soffokl: I'd call `currentPublicIP` something like `originalPublicIP`. 
It's is not really `current` here, and `if currentPublicIP == newPublicIP {` looks wierd. 
```suggestion
func (manager *connectionManager) checkSessionIP(dialog communication.Dialog, sessionID session.ID, originalPublicIP string) {
```

session/connectivity/status_message.go
```diff
@@ -0,0 +1,83 @@
...
+	StatusConnectionOk StatusCode = 1000
+
+	// StatusSessionEstablishmentFailed indicates that session is failed to establish.
+	StatusSessionEstablishmentFailed StatusCode = 2000
+
+	// StatusSessionPaymentsFailed indicates that session payments failed.
+	StatusSessionPaymentsFailed StatusCode = 2001
+
+	// StatusSessionIpNotChanged indicates that session is established but ip is not changed.
+	StatusSessionIpNotChanged StatusCode = 2002
```
soffokl: > const `StatusSessionIpNotChanged` should be `StatusSessionIPNotChanged` (golint)
```suggestion
	StatusSessionIPNotChanged StatusCode = 2002
```

session/connectivity/status_storage.go
```diff
@@ -0,0 +1,64 @@
...
+func NewStatusStorage() StatusStorage {
+	return &statusStorage{}
+}
+
+type statusStorage struct {
+	entries    []*StatusEntry
+	entriesMux sync.RWMutex
+}
+
+func (s *statusStorage) GetAllStatusEntries() []*StatusEntry {
```
soffokl: You are returning a slice of the pointers here under the lock, but after that, any element can be changed from outside by the pointer without a lock, which can break something.

tequilapi/endpoints/session_connectivity.go
```diff
@@ -0,0 +1,86 @@
...
+// ---
+// summary: Returns session connectivity status
+// description: Returns list of session connectivity status
+// responses:
+//   200:
+//     description: List of connectivity statuses
+//     schema:
+//       "$ref": "#/definitions/ConnectivityStatus"
+//   500:
+//     description: Internal server error
```
soffokl: Is it possible to really get 500 error here?

session/connectivity/status_storage.go
vkuznecovas: as far as i'm aware, session storage emits events when a session is removed - you could subscribe to those and remove the relevant entries here as well.

session/connectivity/status_storage.go
anjmao: Yes, that is true. I could change it to return []StatusEntry and return new slice so both slice and it's elements are copied.

session/connectivity/status_storage.go
anjmao: I think node runner may want to see if consumers can connect successfully thats why I also send success event.

communication/interface.go
anjmao: Yes, for now I'm keeping it until this is PR is ready to merge because in case there are merge conflicts it would be easer to solve.

core/connection/manager.go
anjmao: Good point, updated.

session/connectivity/status_storage.go
```diff
@@ -0,0 +1,68 @@
...
+func NewStatusStorage() StatusStorage {
+	return &statusStorage{}
+}
+
+type statusStorage struct {
+	entries    []StatusEntry
+	entriesMux sync.RWMutex
+}
+
+func (s *statusStorage) GetAllStatusEntries() []StatusEntry {
```
anjmao: Now this endpoint returns unsorted data, but we may want to sort by newest entries when displaying in UI.

core/connection/manager.go
```diff
@@ -137,19 +144,62 @@ func (manager *connectionManager) Connect(consumerID identity.Identity, proposal
...
+		}
+		manager.sendSessionStatus(dialog, sessionDTO.ID, connectivity.StatusConnectionFailed, err)
+		return err
+	}
+
+	manager.checkSessionIP(dialog, sessionDTO.ID, originalPublicIP)
+	return nil
+}
+
+// checkSessionIP checks if ip was changed after connection is established and notify peer.
```
zolia:
```suggestion
// checkSessionIP checks if IP has changed after connection was established. Notifies peer about change status.
```

logconfig/log.go
```diff
@@ -64,18 +64,18 @@ func (l Logger) Infof(format string, params ...interface{}) {
 }
 
 // Warnf warn log fmt
-func (l Logger) Warnf(format string, params ...interface{}) error {
-	return seelog.Warnf(l.prefix+format, params...)
+func (l Logger) Warnf(format string, params ...interface{}) {
+	_ = seelog.Warnf(l.prefix+format, params...)
```
zolia:
```suggestion
	seelog.Warnf(l.prefix+format, params...)
```

zolia: do we really need explicit ignore?

tequilapi/endpoints/service.go
```diff
@@ -388,7 +388,7 @@ func validateServiceRequest(sr serviceRequest) *validation.FieldErrorMap {
 	return errors
 }
 
-// ServiceManager represents service manager that will be used for manipulation node services.
+// ServiceManager represents service manager that is used for manipulation node services.
```
zolia:
```suggestion
// ServiceManager represents service manager that is used for services management.
```

session/connectivity/status_subscriber_test.go
```diff
@@ -0,0 +1,91 @@
...
+	peerAddress string
+	msg         *StatusMessage
+}
+
+func (m *mockDialog) PeerID() identity.Identity {
+	return identity.Identity{Address: m.peerAddress}
+}
+
+func (m *mockDialog) Send(producer communication.MessageProducer) error {
+	panic("implement me")
```
soffokl: I'd call it "Mock method should never be called", `implement me` sounds like it should be implemented one day. :)

Issue #1296, opened 2019-09-10
Implement faucet button


zolia:
- [x] transactor endpoint to top-up Mysts to given channel
 - [x] throttle top-up requests on transactor
 - [x] transparent endpoint on node to top-up consumer channel
 - [x] top-up button on mobile client

PR #1347, opened 2019-10-15
Adds top up endpoint to tequila

vkuznecovas:
This endpoint will be used with the new payment system on testnet for minting myst to the given identity.

MYST will be required for full simulation of the payment protocol.

Updates #1296

### DIFF
--- a/core/transactor/transactor.go
+++ b/core/transactor/transactor.go
@@ -122,7 +122,7 @@ func (t *Transactor) TopUp(id string) error {
 
 	req, err := requests.NewPostRequest(t.endpointAddress, "fee/topup", payload)
 	if err != nil {
-		return errors.Wrap(err, "identity request to Transactor failed")
+		return errors.Wrap(err, "failed to create TopUp request")
 	}
 	return t.http.DoRequest(req)
 }
@@ -141,7 +141,7 @@ func (t *Transactor) RegisterIdentity(id string, regReqDTO *IdentityRegistration
 
 	req, err := requests.NewPostRequest(t.endpointAddress, "identity/register", regReq)
 	if err != nil {
-		return errors.Wrap(err, "identity request to Transactor failed")
+		return errors.Wrap(err, "failed to create RegisterIdentity request")
 	}
 
 	return t.http.DoRequest(req)
--- a/core/transactor/transactor.go
+++ b/core/transactor/transactor.go
@@ -29,27 +29,28 @@ import (
 	"github.com/pkg/errors"
 
 	"github.com/mysteriumnetwork/node/identity"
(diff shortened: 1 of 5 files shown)


tequilapi/endpoints/transactor.go
```diff
@@ -67,6 +68,47 @@ func (te *transactorEndpoint) TransactorFees(resp http.ResponseWriter, _ *http.R
...
+//     schema:
+//       "$ref": "#/definitions/ErrorMessageDTO"
+//   400:
+//     description: Bad request
+//     schema:
+//       "$ref": "#/definitions/ErrorMessageDTO"
+func (te *transactorEndpoint) TopUp(resp http.ResponseWriter, request *http.Request, _ httprouter.Params) {
+	topUpDTO := &transactor.TopUpRequest{}
+
+	if request.ContentLength > 0 {
```
anjmao: Not sure about this check. If would suggest to remove it and simply allow to decode body to TopUpRequest and let if fail because now I will be able to pass empty content and call transactor.TopUp

tequilapi/endpoints/transactor_test.go
```diff
@@ -82,6 +86,29 @@ func Test_Get_TransactorFees(t *testing.T) {
...
+
+	router := httprouter.New()
+
+	tr := transactor.NewTransactor(server.URL, server.URL, "0x241F6e1d0bB17f45767DC60A6Bd3D21Cdb543a0c", "0x241F6e1d0bB17f45767DC60A6Bd3D21Cdb543a0c", "0x241F6e1d0bB17f45767DC60A6Bd3D21Cdb543a0c", fakeSignerFactory)
+	AddRoutesForTransactor(router, tr)
+
+	req, err := http.NewRequest(
+		http.MethodPost,
+		"/transactor/topup",
+		bytes.NewBufferString(topUpData),
```
anjmao: As topUpData variable is used only for this test I suggest to keep it as a local variable inside this func.

core/transactor/transactor.go
```diff
@@ -101,6 +109,24 @@ func (t *Transactor) FetchFees() (Fees, error) {
...
+		return errors.Wrap(err, "failed to calculate channel address")
+	}
+
+	payload := TopUpRequest{
+		Identity: channelAddress,
+	}
+
+	req, err := requests.NewPostRequest(t.endpointAddress, "fee/topup", payload)
+	if err != nil {
+		return errors.Wrap(err, "identity request to Transactor failed")
```
anjmao: This error message sounds like it failed to do http request, but it only failed to create request object. Maybe

```go
return errors.Wrap(err, "failed to construct TopUp request")
```

Issue #1336, opened 2019-10-10
Enable additional dependencies during unattended upgrades


zolia:
Due to excessive dependency unattendant upgrade fails to update a node. 

We need to find a way to do upgrade. Investigate if current debian packaging mechanic would allow somehow to include additional dependency. If not - device a way to install such dependencies. 

One possible solution would be to inject script with updated node version and node's post-install script would execute such script installing missing dependencies. If post-install script would not allow this, we might consider developing some autoupdating deamon for a node.

PR #1348, opened 2019-10-15
Fix unattended upgrade origin for mystberry

soffokl:
The wrong origins were preventing packages to be installed. 
Closes https://github.com/mysteriumnetwork/node/issues/1336


bin/package/raspberry/files/unattended-upgrades
```diff
@@ -1,6 +1,6 @@
 Unattended-Upgrade::Origins-Pattern {
-        "origin=Debian,codename=${distro_codename},label=Debian";
-        "origin=Debian,codename=${distro_codename},label=Debian-Security";
+        "origin=Raspbian,codename=${distro_codename},label=Raspbian";
```
zolia: 💯 
just that?

soffokl: yeah, no `origin=Raspbian` packages were allowed, so no upgrades allowed as well.

soffokl: Unfortunately, it will help for only new installations, all raspberries with `origin=Debian` need manual changes.

PR #1358, opened 2019-10-16
Automatic top up after registration

### LINKED ISSUE #1294: New consumer channel auto top-up
All new consumers will auto-register and will get an airdrop to given channel:
 - consumer requests transactor to top-up Mysts into given channel

vkuznecovas:
We'll now top up the channel after a succesful registration

Closes #1294

### DIFF
--- a/core/discovery/discovery.go
+++ b/core/discovery/discovery.go
@@ -45,6 +45,8 @@ const (
 const (
 	// ProposalEventTopic represent proposal events topic.
 	ProposalEventTopic = "proposalEvent"
+	// IdentityRegistrationTopic represent topics for identity registration events.
+	IdentityRegistrationTopic = "identityRegistrationOccurred"
 )
 
 // Publisher is responsible for publishing given events.
@@ -180,6 +182,7 @@ func (d *Discovery) registerIdentity() {
 		switch registerEvent {
 		case identity_registry.Registered:
 			log.Info("identity registered, proceeding with proposal registration")
+			go d.eventPublisher.Publish(IdentityRegistrationTopic, d.ownIdentity)
 			d.changeStatus(RegisterProposal)
 		case identity_registry.Cancelled:
 			log.Info("cancelled identity registration")
--- a/core/transactor/top_upper.go
+++ b/core/transactor/top_upper.go
@@ -1,0 +1,90 @@
+/*
+ * Copyright (C) 2019 The "MysteriumNetwork/node" Authors.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package transactor
+
+import (
+	"sync"
+	"time"
+
+	"github.com/mysteriumnetwork/node/core/discovery"
+	"github.com/mysteriumnetwork/node/identity"
+)
+
+type transactorCaller interface {
+	TopUp(id string) error
+}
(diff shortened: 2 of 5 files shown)


core/discovery/discovery.go
```diff
@@ -45,6 +45,8 @@ const (
 const (
 	// ProposalEventTopic represent proposal events topic.
 	ProposalEventTopic = "proposalEvent"
+	// IdentityRegistrationTopic represent topics for identity registration events.
+	IdentityRegistrationTopic = "identityRegistrationOccured"
```
soffokl:
```suggestion
	IdentityRegistrationTopic = "identityRegistrationOccurred"
```

core/transactor/topper_upper.go
```diff
@@ -0,0 +1,91 @@
...
+
+// NewTopperUpper returns a new instance of topper upper
+func NewTopperUpper(transactor transactorCaller, retries int, delay time.Duration) *TopperUpper {
+	return &TopperUpper{
+		tc:                transactor,
+		retryAttempts:     retries,
+		delayBetweenRetry: delay,
+		stop:              make(chan struct{}),
+	}
+
```
soffokl: > unnecessary trailing newline (whitespace) go-lint

core/transactor/topper_upper.go
```diff
@@ -0,0 +1,91 @@
...
+// DefaultDelayBetweenRetries represents the delay between retries
+var DefaultDelayBetweenRetries = time.Minute * 1
+
+type eventBus interface {
+	SubscribeAsync(topic string, fn interface{}) error
+}
+
+// TopperUpper tops up accounts on relevant identity related events
+type TopperUpper struct {
```
soffokl: Why `TopperUpper` instead of `TopUpper`?

core/transactor/topper_upper_test.go
```diff
@@ -0,0 +1,81 @@
...
+import (
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/mysteriumnetwork/node/identity"
+
+	"github.com/pkg/errors"
```
soffokl: Something wrong with grouping.

core/transactor/topper_upper.go
vkuznecovas: I dunno, just think it sounds better.

core/discovery/discovery.go
anjmao: Just for the feature. As go doesn't have  enums usually you name related const starting with a type.

```go
const (
   TopicProposalEvent = ".."
   TopicIdentityRegistration = ".."
   TopicAnotherOne = ".."
)
```
Some examples from go std https://github.com/golang/net/blob/master/http2/frame.go#L30

core/transactor/topper_upper.go
anjmao: Hm, it's strange that CI is not caching this style issue as we have golint checker.

core/transactor/topper_upper_test.go
```diff
@@ -0,0 +1,79 @@
...
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/mysteriumnetwork/node/identity"
+	"github.com/pkg/errors"
+	"github.com/stretchr/testify/assert"
+)
+
+type mockTransactor struct {
```
anjmao: I would prefer putting mock objects to the bottom so I can open this file and start reading tests.

core/transactor/top_upper.go
```diff
@@ -0,0 +1,90 @@
...
+	})
+}
+
+func (tu *TopUpper) handleRegistrationEvent(id identity.Identity) {
+	for i := 0; i < tu.retryAttempts; i++ {
+		err := tu.tc.TopUp(id.Address)
+		if err != nil {
+			log.Warnf("could not top up newly registered identity channel %v. Will retry", err)
+			if i+1 == tu.retryAttempts {
+				log.Errorf("top up failed after multiple attempts, aborting", err)
```
zolia: We probably should monitor such metric. This would probably mean that transactor is either: 
  - very loaded
  - unreachable
  - down
  - malfunctioning 

Report spikes from such events would signal us to take actions immediately. 
We could report such metric to quality oracle.  We should probably have a ticket for this and for similar cases.

vkuznecovas: This should probably go under transactor monitoring itself.

PR #1360, opened 2019-10-17
Replace sonatype with bintray for maven releases

### LINKED ISSUE #1343: Publish to github package registry instead of maven central
Moving to Github package registry (beta) will address the following:

1. Sonatype is rather unstable (random 5xx errors when uploading, which are addressed with retries)
2. A proxy is needed to be able to upload artifacts. We can get rid of the tinyproxy completely.
3. Supposedly, simpler configuration and release: only Github token is needed to do the works with the package registry

The drawback is that the maven repository will have to be manually added to the dependent projects. However, in java ecosystem, this is common as there are a lot of projects using non-central repositories such as `jcenter` or `google`.

Extra tasks:

- [ ] Document adding `mobile-node` from the new repository
- [ ] Add repository in https://github.com/mysteriumnetwork/mysterium-vpn-mobile

This will also address #1334

tadaskay:
Fixes #1343 

Initial idea was to use Github package registry (beta). However, further digging showed that:

1. It forces strange groupId and artifactId (with slashes) which are alien to maven ecosystem
2. Publishing works, but it's not possible to upload extra content besides `aar`+`pom` (e.g. source JAR, checksums, etc.)
3. I was not able to download the artifact as a consumer (tried lots of different combinations)

It seems to me that it's still premature to use the feature as a maven repository.
So I went back and took a look into the other options, and chose Bintray (which is a popular choice in java ecosystem). We initially discarded it, but the problem was that the organization account was created as an enterprise, instead of the default OSS choice.

Publishing/consuming artifacts from bintray is quite straightforward and worked out of the box.
Until `mysteriumnetwork/node` is synced to `jcenter/maven central`, clients will need to add:

```
maven {
        url 'https://dl.bintray.com/mysteriumnetwork/maven'
}
```

And dependency usage:
```
    implementation 'network.mysterium:mobile-node:0.13.0'
```

Repo web URL: https://bintray.com/mysteriumnetwork/maven/network.mysterium%3Amobile-node

TODO Docs update


ci/release/android_sdk.go
```diff
@@ -48,28 +55,148 @@ func ReleaseAndroidSDK() error {
...
+	defer file.Close()
+
+	requestURL := uploadURL(*up.releaseOpts, *up.bintrayOpts, filename)
+	log.Info("Upload URL: ", requestURL)
+
+	req, err := http.NewRequest(http.MethodPut, requestURL.String(), file)
+	if err != nil {
+		return err
+	}
+	req.SetBasicAuth(up.bintrayOpts.username, up.bintrayOpts.password)
```
vkuznecovas: could extract the auth + header setting into a separate func since you're doing it more than once

ci/packages/package.go
```diff
@@ -119,6 +120,26 @@ func PackageAndroid() error {
...
 	}
+	err := env.EnsureEnvVars(env.BuildVersion)
+	if err != nil {
+		return err
+	}
+	pomTemplate, err := template.ParseFiles("bin/package/android/mvn.pom")
+	if err != nil {
+		return err
+	}
+	pomFileOut, err := os.Create("build/package/mvn.pom")
```
anjmao: Don'f forget to close the file
```go
defer pomFileOut.Close()
```

Issue #1290, opened 2019-09-06
Improve logging classification


zolia:
Much of needed debug logs falls under TRACE level. In order to improve analysis of logs we need more information under DEBUG level. Review and change code where appropriate.

PR #1362, opened 2019-10-21
Replace seelog with zerolog

tadaskay:
Ref: #1290

As a part of improving logging classifications, this aims to improve log readability by using a structured logger. 

- Plain output is used, but errors are structured, thus will be highlighted in console
- Seelog is long dead, so this was a choice between zap/logrus/zerolog
- Later we can use this to report to ELK in JSON format which is natively supported 
- API is slightly different, see diff for the change
- Migrated log capture (used in a couple tests)
- For now full caller info is displayed (`file:line` - out of the box), later we can consider logging package instead
 
As a bonus:
- Finally removed all log prefixes
- Polished logging message formatting

Sample screen:
Screenshot 2019-10-21 at 16 44 08

⚠️ This breaks log file rotation, which I'll implement with the next PR. But I'd like to merge this, because it conflicts with every incoming PR 😿


logconfig/config.go
```diff
@@ -18,69 +18,77 @@
...
+	if len(s) >= length {
+		return s
+	}
+	return s + strings.Repeat(" ", length-len(s))
+}
+
+func configureZerolog(opts *LogOptions) {
+	zerolog.TimeFieldFormat = time.RFC3339Nano
+	zerolog.CallerMarshalFunc = func(file string, line int) string {
+		pkg := "github.com/mysteriumnetwork/node"
```
anjmao: As pkg variable is not changing it could be allocated once

```go
const rootPkg = "github.com/mysteriumnetwork/node"
func configureZerolog() {
 // ...
}
```
```

Comment (tadaskay) [reply logconfig/config.go]: It’s easier to read when it’s colocated. Besides, it’s only called once.

Comment (anjmao) [reply logconfig/config.go]: I think it can't be called once as you get caller name from different files there actual logging is done during runtime.

File: mobile/mysterium/wireguard_connection_setup.go [hunk 2]
```diff
@@ -187,20 +188,20 @@ func newTunnDevice(wgTunnSetup WireguardTunnelSetup, config *wireguard.ServiceCo
...
+		log.Info().Msgf("Delaying tunnel creation for %v milliseconds", config.Consumer.ConnectDelay)
 		time.Sleep(time.Duration(config.Consumer.ConnectDelay) * time.Millisecond)
 	}
 
 	fd, err := wgTunnSetup.Establish()
 	if err != nil {
 		return nil, err
 	}
-	log.Info(logPrefix, "Tun value is: ", fd)
+	log.Info().Msg("Tun value is: " + strconv.Itoa(fd))
```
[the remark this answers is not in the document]
zolia: Probably its not less efficient and looks better.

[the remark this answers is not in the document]
anjmao: Yes, you win about 100 nanoseconds which in this context is not important. I prefer Msgf as it is simpler, in this case performance is not a problem and you don't need to import extra library strconv.

ci/packages/raspberry.go
```diff
@@ -100,7 +98,7 @@ func configureRaspbianImage(raspbianImagePath string) error {
 	}
 	defer func() {
 		if err := device.DetachLoop(loopDevice); err != nil {
-			log.Warn(err)
+			log.Warn().Err(err).Msg("")
```
zolia: This looks a lot more verbose and less performant.. 3 functions instead of 1?

anjmao:
@zolia I think new logging is much easer to read and faster for debugging as you get full filename and line number out of the box. See Tadas attached sample screenshot.

mobile/mysterium/wireguard_connection_setup.go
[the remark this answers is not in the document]
tadaskay: It’s a package from stdlib, not a library. In this particular context there’s not any difference and “looks better” is subjective, thus I’d rather keep it as is.

ci/packages/raspberry.go
tadaskay: Have you tried benchmarking it?

tadaskay: I have considered making a facade for it to keep simple log calls simple, but with the added ability to log typed structured fields, it just does not fit, thus we would have use both the facade and zerolog/log. So I decided to simply go with zerolog directly.

logconfig/config.go
tadaskay: You are right about this one, forgot that it's invoked every time the marshal is made. I'll extract it.

PR #1368, opened 2019-10-25
Add go 1.13.1 for ios and android builds

anjmao:
This PR adds updated [xgo](https://github.com/mysteriumnetwork/xgo/pull/13) with go.1.13.1 support so iOS and Android.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -18,7 +18,6 @@ require (
 	github.com/chzyer/logex v1.1.10 // indirect
 	github.com/chzyer/readline v0.0.0-20180603132655-2972be24d48e
 	github.com/chzyer/test v0.0.0-20180213035817-a1ea475d72b1 // indirect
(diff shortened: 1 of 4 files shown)


go.mod
```diff
@@ -18,7 +18,6 @@ require (
 	github.com/chzyer/logex v1.1.10 // indirect
 	github.com/chzyer/readline v0.0.0-20180603132655-2972be24d48e
 	github.com/chzyer/test v0.0.0-20180213035817-a1ea475d72b1 // indirect
-	github.com/cihub/seelog v0.0.0-20170130134532-f561c5e57575
```
anjmao: @tadaskay I forgot to mention when was doing code review. When working with modules it is advisable to use `go mod tidy` after modules update. This command removes old unused packages.

Issue #1298, opened 2019-09-10
Report connection status to QualityOracle


zolia:
after #1297 is implemented report connection status to QualityOracle. We could do reporting from both sides (from consumer and from provider). QualityOracle should update Provider quality metrics by matching reported status from both sides.

Currently consumer reports on failed connects, but not all failed cases are accounted for. After Consumer gets connected event he still needs to be sure that actual traffic gets transferred. 

Sent Connection status from Provider could contain and determined NAT traversal state for given connection moment. These can be two separate messages tho.

soffokl:
Is this a duplicate of https://github.com/mysteriumnetwork/node/issues/891?

anjmao:
Yes they looks similar, let's continue here as this task is a follow up of #1297.

anjmao:
Currently we have these metrics for session in prometheus.

```
session_event{event="Connecting"}
session_event{event="Connected"}
session_event{event="Created"}
session_event{event="Disconnecting"}
session_event{event="Ended"}
session_event{event="NotConnected"}
```

In node session_event metric is aggregated from two different events (EventTopicState = "State" and EventTopicSession = "Session") which makes it confusing to understand the difference in Connected or Created and Disconnecting and Ended statuses. To be more clear in node code only Created and Ended are actual session events and all others are connection state events.

[Here](https://github.com/mysteriumnetwork/node/blob/master/core/connection/status.go) we have connection status state:

```go
// State represents list of possible connection states
type State string

const (
	// NotConnected means no connection exists
	NotConnected = State("NotConnected")
	// Connecting means that connection is startCalled but not yet fully established
	Connecting = State("Connecting")
	// Connected means that fully established connection exists
	Connected = State("Connected")
	// Disconnecting means that connection close is in progress
	Disconnecting = State("Disconnecting")
	// Reconnecting means that connection is lost but underlying service is trying to reestablish it
	Reconnecting = State("Reconnecting")
	// Unknown means that we could not map the underlying transport state to our state
	Unknown = State("Unknown")
	// Canceled means that connection initialization was started, but failed never reaching Connected state
	Canceled = State("Canceled")
)
```

And separate connection session status [here](https://github.com/mysteriumnetwork/node/blob/master/core/connection/event.go#L38)
```go
const (
	// SessionCreatedStatus represents a session creation event
	SessionCreatedStatus = "Created"
	// SessionEndedStatus represents a session end
	SessionEndedStatus = "Ended"
)
```
[...]

PR #1369, opened 2019-10-29
Send consumer connection metrics

anjmao:
This PR implements consumer session IP change check and sends status to another peer and quality oracle. IP check is done in separate go routine with default 3 retries and 2 seconds delay between each try.

Part of https://github.com/mysteriumnetwork/node/issues/1298

### DIFF
--- a/core/connection/manager.go
+++ b/core/connection/manager.go
@@ -20,6 +20,7 @@ package connection
 import (
 	"context"
 	"sync"
+	"time"
 
 	"github.com/mysteriumnetwork/node/communication"
 	"github.com/mysteriumnetwork/node/consumer"
@@ -46,6 +47,22 @@ var (
 	// ErrUnsupportedServiceType indicates that target proposal contains unsupported service type
 	ErrUnsupportedServiceType = errors.New("unsupported service type in proposal")
 )
+
+// IPCheckParams contains common params for connection ip check.
+type IPCheckParams struct {
+	MaxAttempts             int
+	SleepDurationAfterCheck time.Duration
+	Done                    chan struct{}
+}
+
+// DefaultIPCheckParams returns default params.
+func DefaultIPCheckParams() IPCheckParams {
+	return IPCheckParams{
+		MaxAttempts:             3,
+		SleepDurationAfterCheck: 1 * time.Second,
+		Done:                    make(chan struct{}, 1),
+	}
+}
 
 // Creator creates new connection by given options and uses state channel to report state changes
 type Creator func(serviceType string, stateChannel StateChannel, statisticsChannel StatisticsChannel) (Connection, error)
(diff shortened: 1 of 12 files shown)


core/connection/manager.go
```diff
@@ -468,6 +504,27 @@ func (manager *connectionManager) setupTrafficBlock(disableKillSwitch bool) erro
...
+}
+
+func (manager *connectionManager) setCurrentSession(info SessionInfo) {
+	manager.sessionInfoMux.Lock()
+	defer manager.sessionInfoMux.Unlock()
+
+	manager.sessionInfo = info
+}
+
+func (manager *connectionManager) currentSession() SessionInfo {
```
vkuznecovas: i'd call this a getCurrentSession, since we already have a setter.

core/connection/manager_test.go
```diff
@@ -431,14 +517,13 @@ func (mpm *MockPaymentIssuer) Stop() {
...
-	return &mockStatusSender{}
-}
-
 type mockStatusSender struct {
 	sentMsg *connectivity.StatusMessage
+	sync.Mutex
 }
 
-func (s mockStatusSender) Send(dialog communication.Sender, msg *connectivity.StatusMessage) {
+func (s *mockStatusSender) Send(dialog communication.Sender, msg *connectivity.StatusMessage) {
```
vkuznecovas: you might want to implement a getter under lock here, as races can still happen if you directly try to access the sentMsg variable.

anjmao: sentMsg is accessed in unit tests asserts which are running in a single thread so there is no way for race condition to happen. If it happens test will fail which will indicate that something is wrong with unit tests logic.

core/connection/manager.go
```diff
@@ -82,14 +99,16 @@ type connectionManager struct {
...
-	status      Status
-	statusLock  sync.RWMutex
-	sessionInfo SessionInfo
-	cleanup     []func() error
-	cancel      func()
+	ctx            context.Context
+	status         Status
+	statusLock     sync.RWMutex
+	sessionInfo    SessionInfo
+	sessionInfoMux sync.Mutex
```
soffokl: I'd call it `sessionInfoLock`, `mux` is not associated with a mutex.

core/ip/fake_resolver.go
```diff
@@ -17,39 +17,59 @@
...
 
 func (client *mockResolver) GetOutboundIPAsString() (string, error) {
-	return client.ipAddress, client.error
+	return client.getNextIP(), client.error
+}
+
+func (client *mockResolver) getNextIP() string {
+	// Return first address if only one provided.
+	if len(client.ipAddresses) == 1 {
+		fmt.Println("get first ip")
```
soffokl: Do we need the output here?

PR #1375, opened 2019-10-30
Provider will now check for user registration prior to serving the se…

### LINKED ISSUE #1373: Provider should reject not registered consumer from session establishment
Should be done as separate feature toggle. When it is enabled, not registered identities (all old client apps) will be not supported.

vkuznecovas:
…ssion

The providers side of pingpong will now check wether the user is registered with the registry.
In case that user is  not registered, the session will be cancelled and the consumer will not be served.

Closes #1373

Opening this as a separate PR so that payment-v2-messaging does not get super duper huge and hard to review.

### DIFF
--- a/session/pingpong/blockchain.go
+++ b/session/pingpong/blockchain.go
@@ -61,3 +61,19 @@ func (bc *Blockchain) GetAccountantFee(accountantAddress common.Address) (uint16
 
 	return res.Value, err
 }
+
+// IsRegistered checks wether the given identity is registered or not
+func (bc *Blockchain) IsRegistered(registryAddress, addressToCheck common.Address) (bool, error) {
+	caller, err := bindings.NewRegistryCaller(registryAddress, bc.client)
+	if err != nil {
+		return false, errors.Wrap(err, "could not create registry caller")
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), bc.bcTimeout)
+	defer cancel()
+
+	res, err := caller.IsRegistered(&bind.CallOpts{
+		Context: ctx,
+	}, addressToCheck)
+	return res, errors.Wrap(err, "could not check registration status")
+}
--- a/session/pingpong/invoice_tracker.go
+++ b/session/pingpong/invoice_tracker.go
@@ -48,6 +48,7 @@ type PeerInvoiceSender interface {
 
 type bcHelper interface {
 	GetAccountantFee(accountantAddress common.Address) (uint16, error)
+	IsRegistered(registryAddress, addressToCheck common.Address) (bool, error)
 }
 
 type providerInvoiceStorage interface {
@@ -72,6 +73,9 @@ var ErrExchangeWaitTimeout = errors.New("did not get a new exchange message")
 
 // ErrExchangeValidationFailed indicates that there was an error with the exchange signature.
 var ErrExchangeValidationFailed = errors.New("exchange validation failed")
+
+// ErrConsumerNotRegistered represents the error that the consumer is not registered
+var ErrConsumerNotRegistered = errors.New("consumer not registered")
 
 const chargePeriodLeeway = time.Hour * 2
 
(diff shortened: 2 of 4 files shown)


vkuznecovas:
Ignore the broken windows build for now :)

session/pingpong/invoice_tracker_test.go
```diff
@@ -109,6 +109,101 @@ func Test_InvoiceTracker_Start_Stop(t *testing.T) {
...
+		PaymentInfo:                dto.PaymentPerTime{Price: money.NewMoney(10, money.CurrencyMyst), Duration: time.Minute},
+		ProviderID:                 identity.FromAddress(acc.Address.Hex()),
+		AccountantID:               identity.FromAddress(acc.Address.Hex()),
+		AccountantCaller:           &mockAccountantCaller{},
+		AccountantPromiseStorage:   accountantPromiseStorage,
+		BlockchainHelper:           &mockBlockchainHelper{isRegistered: false},
+	}
+	invoiceTracker := NewInvoiceTracker(deps)
+
+	go func() {
```
anjmao: 1. What is the purpose of time.Sleep?
2. If you need to stop invoice tracker maybe you can use `defer invoiceTracker.Stop()`?

chompomonim:
What will happen with old consumer apps? They will have not registered identities. Can node choose not to check if consumer is registered (at least for now)?

session/pingpong/blockchain.go
```diff
@@ -61,3 +61,19 @@ func (bc *Blockchain) GetAccountantFee(accountantAddress common.Address) (uint16
 
 	return res.Value, err
 }
+
+// IsRegistered checks wether the given identity is registered or not
+func (bc *Blockchain) IsRegistered(registryAddress, addressToCheck common.Address) (bool, error) {
```
zolia: why are we passing 'registryAddress'  here? We probably should construct it together with bc. Registry is not something that change?

session/pingpong/invoice_tracker.go
```diff
@@ -181,6 +185,15 @@ func (it *InvoiceTracker) Start() error {
 	log.Debug().Msg("Starting...")
 	it.timeTracker.StartTracking()
 
+	isConsumerRegistered, err := it.bcHelper.IsRegistered(common.HexToAddress(it.registryAddress), it.peer.ToCommonAddress())
+	if err != nil {
+		return err
```
zolia: we need wrap here.

session/pingpong/invoice_tracker_test.go
```diff
@@ -109,6 +109,101 @@ func Test_InvoiceTracker_Start_Stop(t *testing.T) {
...
+	deps := InvoiceTrackerDeps{
+		Peer:                       identity.FromAddress("some peer"),
+		PeerInvoiceSender:          mockSender,
+		InvoiceStorage:             invoiceStorage,
+		TimeTracker:                &tracker,
+		ChargePeriod:               time.Nanosecond,
+		ExchangeMessageChan:        exchangeMessageChan,
+		ExchangeMessageWaitTimeout: time.Second,
+		PaymentInfo:                dto.PaymentPerTime{Price: money.NewMoney(10, money.CurrencyMyst), Duration: time.Minute},
+		ProviderID:                 identity.FromAddress(acc.Address.Hex()),
```
zolia: The same identity as acc?

session/pingpong/blockchain.go
vkuznecovas: this component is just a caller for blockchain, it is up to the consumer to know which registry to call.

session/pingpong/invoice_tracker_test.go
vkuznecovas: start is blocking, henceforth the stop needs to be invoked before calling start.

session/pingpong/invoice_tracker_test.go
vkuznecovas: yes - in this case the mocks are fine with any addresses, really.

PR #1382, opened 2019-10-31
Proper pricing for payments

### LINKED ISSUE #1380: Ensure correct pricing for payment alpha
ensure that 1 minute price is 0.00100000 MYSTT

vkuznecovas:
Ensuring that payments are now actually charged at a rate of 0.001Myst/minute.

Closes #1380


session/pingpong/factory.go
```diff
@@ -38,6 +38,12 @@ import (
 // DefaultAccountantFailureCount defines how many times we're allowed to fail to reach accountant in a row before announcing the failure.
 const DefaultAccountantFailureCount uint64 = 3
 
+// DefaultPaymentInfo represents the default payment info for the alpha release
+var DefaultPaymentInfo = dto.PaymentPerTime{
```
tadaskay: I think `Rate` is a more precise and concise term for `PaymentPerTime`.

session/pingpong/factory.go
```diff
@@ -38,6 +38,12 @@ import (
...
+	Price:    money.NewMoney(0.001, money.CurrencyMyst),
```
tadaskay: My suggestion is to use decimal for money to avoid the unexpected in float to int conversions. Accountant, on the other hand, may use ints (which - per our discussion - I understood is used by blockchain).

vkuznecovas: As talked - we'll just use uints here as we're aligning to what the blockchain prefers

session/pingpong/factory.go
```diff
@@ -38,12 +38,18 @@ import (
 // DefaultAccountantFailureCount defines how many times we're allowed to fail to reach accountant in a row before announcing the failure.
 const DefaultAccountantFailureCount uint64 = 3
 
+// DefaultPaymentInfo represents the default payment info for the alpha release
+var DefaultPaymentInfo = dto.PaymentRate{
+	Price:    money.NewMoney(100000, money.CurrencyMyst),
+	Duration: 1 * time.Minute,
+}
```
anjmao: The same DefaultPaymentInfo is described in services/openvpn/discovery/dto/payment_method_per_time.go
I would expect to have this payment configurations in one package like `payment` etc.

PR #1389, opened 2019-11-03
Use dir based checks to operate outside gopath

tadaskay:
After migration to gomodules, project sources can be placed outside of gopath. However, the previous version of checks (`mage check`) was incompatible. Switching to the dir based checks here.


ci/check/check.go
```diff
@@ -34,22 +34,22 @@ func Check() {
...
 // CheckGoLint reports linting errors in the solution.
 func CheckGoLint() error {
-	return commands.GoLint("./...", "docs")
+	return commands.GoLintD(".", "docs")
 }
 
 // CheckGoVet checks that the source is compliant with go vet.
 func CheckGoVet() error {
-	return commands.GoVet("./...")
+	return commands.GoVet(".")
```
anjmao: Probably should be `commands.GoVetD(".")`

tadaskay: There is no `GoVetD`, `go vet` does not play nicely with directories :(

tadaskay:
`go vet` reports the following error on macOS Catalina 10.15.1 (19B88):

```
▷ ~/mygo/node [tk/checks-gomodules-use-dirs]
└─ $ go vet ./...
# github.com/mysteriumnetwork/wireguard-go/rwcancel
../../go/pkg/mod/github.com/mysteriumnetwork/wireguard-go@v0.0.0-20190318074909-625d21fc42f3/rwcancel/select_default.go:13:2: too many arguments to return
	have (int, error)
	want (error)
```

Issue #1345, opened 2019-10-15
RFC: collecting node deployments data from 3rd parties


zolia:
# Partner tokens

In order to collect statistics how many nodes different 3rd parties managed to deploy we need some way to identify them. 

Currently we have referral model which lets you generate your own referral code and pass it to registering node owner in hope that he will enter this referral code. This is not a robust solution for partners that a willing to account 100% of their nodes and let us know their contribution to a network. 

We assume that Partner is able to pre-configure nodes with required parameters. 

Suggestion is that we generate unique token for each partner willing to report their nodes to us and request them to submit { node_identity, partner_token } info through their nodes.

We cannot allow them to submit such info directly since they might abuse such system and report nodes that are not theirs. 

Implementation details: 
  - add node parameter / cli command and optionally tequilapi endpoint accepting partner_token 
  - if parameter was set - trigger Partner data submit to mmn portal
  - save  such submit state - we should not allow to re-submit such data. Its one time binding.
  - service receiving Parner data (mmn) should not accept re-submits and should always store the same and only 1 { node_identity, partner_token } binding.

In the future partner token can be used as a way to identify which nodes belongs to Parner and enable him centralized management place through mmn.

# Setting up payout wallet

Currently users or partners willing to receive payments according to bounty rules have to register payout (wallet) address. This can be done either through web-ui of a node, either through CLI interface (using payout command) or using tequilapi REST call through a node: 

https://tequilapi.mysterium.network/#operation/updatePayoutInfo

tadaskay:
> save such submit state - we should not allow to re-submit such data. Its one time binding.
> only 1 { node_identity, partner_token } binding

This will forbid users from switching devices. Identity is not unique to a device. It may be exported by copying data files to another device (some users even requested an easier way to do it).

> tequilapi endpoint accepting partner_token
Is this really necessary? We could simply add a configuration flag (CLI/config) which is easy to configure either at runtime (by call configuration API) or preconfiguration (providing configuration file or node options). We could then append this flag to the metrics/events we already report to grafana or MMN.

zolia:
>>save such submit state - we should not allow to re-submit such data. Its one time binding.
>>only 1 { node_identity, partner_token } binding

>This will forbid users from switching devices. Identity is not unique to a device. It may be >exported by copying data files to another device (some users even requested an easier way to >do it).

[...]

PR #1390, opened 2019-11-03
Provide a way for 3rd party vendor to identify their nodes

tadaskay:
`vendor.id` flag (can be set in `config.toml`) will allow 3rd party vendors to track their nodes.

Fixes: #1345 

_Note: configuration flags refactoring is needed to unify the approach and/or improve current config implementation, that I will do with a separate PR_

### DIFF
--- a/mmn/node_information.go
+++ b/mmn/node_information.go
@@ -24,6 +24,7 @@ import (
 	"strings"
 
 	"github.com/mysteriumnetwork/go-ci/shell"
+	"github.com/mysteriumnetwork/node/config"
 	"github.com/rs/zerolog/log"
 
 	"github.com/mysteriumnetwork/node/core/ip"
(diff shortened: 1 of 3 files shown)


mmn/node_information.go
```diff
@@ -64,6 +66,7 @@ func CollectNodeData(client *client, resolver ip.Resolver) func(string) {
 		info.MACAddress = hex.EncodeToString(sha256Bytes[:])
 		info.LocalIP = outboundIp
 		info.Identity = identity
+		info.VendorID = config.Current.GetString(config.VendorIDFlag.Name)
```
anjmao: Depending on global state makes code harder to test. I would suggest have app Config struct which can be passed via params.

```go
func MyFunc(cfg config.Config, ...) {
    cfg.GetString(config.VendorIDFlag.Name)
}
```

PR #1394, opened 2019-11-05
Decrease location cache to 30 seconds

anjmao:
Mobile app is fetching location cache each 10 seconds but as IP may not change instantly it will update IP in UI only after 5 minutes. As a short term solution decreasing to cache to 30 seconds will make UI experience a bit better.

### DIFF
--- a/cmd/di.go
+++ b/cmd/di.go
@@ -724,7 +724,7 @@ func (di *Dependencies) bootstrapLocationComponents(options node.Options) (err e
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


anjmao:
@vkuznecovas Cache is updated correctly on events. The problem was that IP may not yet change after Connected event is fired, but I see there logic to not update cache if IP fetch is failing.

@zolia I'm going to close this as it looks that decreasing cache will not help as we discussed.

PR #1399, opened 2019-11-05
register provider automatically

### LINKED ISSUE #1385: register provider on service start
On service start, we should check if the provider is registered as a provider, if not - we should register the provider

vkuznecovas:
Implemented a provider registrar, which will register the provider with stake if provider is not registered yet.

Closes #1385

### DIFF
--- a/session/pingpong/provider_registrar.go
+++ b/session/pingpong/provider_registrar.go
@@ -1,0 +1,212 @@
+/*
+ * Copyright (C) 2019 The "MysteriumNetwork/node" Authors.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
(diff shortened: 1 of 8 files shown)


session/pingpong/provider_registrar.go
```diff
@@ -0,0 +1,212 @@
...
+	RegistryAddress     common.Address
+}
+
+// NewProviderRegistrar creates a new instance of provider registrar
+func NewProviderRegistrar(transactor txer, bcHelper bc, prc ProviderRegistrarConfig) *ProviderRegistrar {
+	return &ProviderRegistrar{
+		stopChan:             make(chan struct{}),
+		bc:                   bcHelper,
+		queue:                make(chan queuedEvent),
+		registeredIdentities: make(map[string]struct{}, 0),
```
soffokl: > S1019: should use make(map[string]struct{}) instead (gosimple)
```suggestion
		registeredIdentities: make(map[string]struct{}),
```

session/pingpong/provider_registrar.go
```diff
@@ -0,0 +1,212 @@
...
+}
+
+func (pr *ProviderRegistrar) handleEvent(qe queuedEvent) error {
+	registered, err := pr.bc.IsRegisteredAsProvider(pr.cfg.AccountantAddress, pr.cfg.RegistryAddress, common.HexToAddress(qe.event.ProviderID))
+	if err != nil {
+		return errors.Wrap(err, "could not check registration status on BC")
+	}
+
+	if registered {
+		log.Info().Msgf("provider %q already registered on bc, skipping", qe.event.ProviderID)
```
tadaskay: Use sentence case for log messages. @vkuznecovas

vkuznecovas: k

PR #1410, opened 2019-11-14
Update to latest wireguard

tadaskay:
Rebased our fork on latest https://github.com/WireGuard/wireguard-go head and reworked our patches.
Branch name is [sync-fork](https://github.com/mysteriumnetwork/wireguard-go/tree/sync-fork) (not pushing to master for now to keep original patch commits)

### DIFF
--- a/ci/packages/package.go
+++ b/ci/packages/package.go
@@ -213,7 +213,7 @@ func PackageDockerSwaggerRedoc() error {
 // does not support go modules yet and go mod vendor does not include c dependencies.
 func vendordModules() error {
 	mg.Deps(checkVend)
-	return sh.RunV("vend")
+	return sh.RunV("vend", "-replace", "golang.zx2c4.com/wireguard=github.com/mysteriumnetwork/wireguard-go")
 }
 
 func checkVend() error {
@@ -222,7 +222,7 @@ func checkVend() error {
 		fmt.Println("Tool 'vend' already installed")
 		return nil
 	}
-	err := goGet("github.com/nomad-software/vend")
+	err := goGet("github.com/mysteriumnetwork/vend")
 	if err != nil {
 		fmt.Println("could not go get vend")
 		return err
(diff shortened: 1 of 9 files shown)


ci/packages/package.go
```diff
@@ -222,7 +222,7 @@ func checkVend() error {
 		fmt.Println("Tool 'vend' already installed")
 		return nil
 	}
-	err := goGet("github.com/nomad-software/vend")
+	err := goGet("github.com/mysteriumnetwork/vend")
```
soffokl: We need a patched vendor tool to vendor our patched wireguard. It feels like we doing something wrong. :D

mobile/mysterium/wireguard_connection_setup.go
```diff
@@ -28,10 +28,10 @@ import (
 	"github.com/mysteriumnetwork/node/core/connection"
 	"github.com/mysteriumnetwork/node/services/wireguard"
 	"github.com/mysteriumnetwork/node/services/wireguard/key"
-	"github.com/mysteriumnetwork/wireguard-go/device"
```
soffokl: From this, it was clear, that we are using our version of wireguard-go. And now it looks like we are using original packages which is confusing.

ci/packages/package.go
tadaskay: This is only because gomobile does not support gomodules yet

mobile/mysterium/wireguard_connection_setup.go
tadaskay: The goal is to integrate sooner into upstream. It is only logical that we keep original packages.

ci/packages/package.go
soffokl: Let's patch gomobile to support gomodules. 💃

tadaskay: We could contribute but that will take a while.

PR #1411, opened 2019-11-15
Protect networks from VPN access on macOS and Linux

zolia:
- Prohibit access form VPN users to providers local networks. 
- Added flag 'firewall.protected.networks' to override/extend defaults

### DIFF
--- a/nat/service_iptables.go
+++ b/nat/service_iptables.go
@@ -103,7 +103,10 @@ func iptables(action string, rule RuleForwarding) error {
 
 func dropToLocal(action, sourceSubnet string) error {
 	destinations := config.GetString(config.FlagFirewallProtectedNetworks)
-
+	if destinations == "" {
+		log.Info().Msgf("no protected networks set")
+		return nil
+	}
 	command := fmt.Sprintf("/sbin/iptables --%s FORWARD --source %s --destination %s -j DROP",
 		action, sourceSubnet, destinations)
 	cmd := utils.SplitCommand("sudo", command)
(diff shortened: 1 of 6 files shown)


Waldz:
Usually configuration goes like this: `-A FORWARD -s 192.168.255.0/24 -d 192.168.255.0/24 -i tun0 -o tun0 -j DROP`
There `192.168.255.0/24` is Openvpn's subnet, so that connected clients don't see each other (by default).
Will we have such default now?

nat/service_iptables.go
```diff
@@ -92,3 +100,21 @@ func iptables(action string, rule RuleForwarding) error {
 	log.Info().Msgf("Action %q applied for forwarding packets from %s to IP: %s", action, rule.SourceSubnet, rule.TargetIP)
 	return nil
 }
+
+func dropToLocal(action, sourceSubnet string) error {
+	destinations := config.GetString(config.FlagFirewallProtectedNetworks)
+	if destinations == "" {
+		log.Info().Msgf("no protected networks set")
```
tadaskay: Log messages should start with capital case.

3esmit:
Nice workaround!

Would be possible instead of blocking access to LAN ip ranges, configuring OpenVPN to not route traffic to LAN, but just forward the traffic to gateway?

For the clients, a protection is also needed, I am not sure if the LAN is exposed to the VPN provider, but the client connecting it certainly is and a firewall there would be nice, as we would be connecting to "unknown" vpn providers.

PR #1420, opened 2019-11-22
Enable OpenVPN packet filtering

soffokl:
Another part of the feature is here https://github.com/mysteriumnetwork/go-openvpn/pull/49

### DIFF
--- a/services/openvpn/service/factory.go
+++ b/services/openvpn/service/factory.go
@@ -20,12 +20,15 @@ package service
 import (
 	"crypto/x509/pkix"
 	"encoding/json"
+	"strings"
 
 	"github.com/mysteriumnetwork/go-openvpn/openvpn"
 	"github.com/mysteriumnetwork/go-openvpn/openvpn/middlewares/server/auth"
 	"github.com/mysteriumnetwork/go-openvpn/openvpn/middlewares/server/bytecount"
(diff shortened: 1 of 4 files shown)


services/openvpn/service/factory.go
```diff
@@ -121,9 +124,11 @@ func newServerFactory(nodeOptions node.Options, sessionValidator *openvpn_sessio
 			}
 		}
 
+		protectedNetworks := strings.Split(node_config.GetString(node_config.FlagFirewallProtectedNetworks), ",")
 		return openvpn.CreateNewProcess(
 			nodeOptions.Openvpn.BinaryPath(),
 			config.GenericConfig,
+			filter.NewMiddleware(nil, protectedNetworks),
```
Waldz: `nil`?

soffokl: It's a list of addresses that should be excluded from the block list. We are going to use it in the future when we will need to allow some hosts, like DNS servers. 

It will be looking like this:

`filter.NewMiddleware([]string{"10.10.10.1/32"}, []string{"10.0.0.0/8"})`

Waldz: anyway would be to pass array instead of nil-able array

tadaskay: nil slice is actually idiomatic in go

PR #1429, opened 2019-11-26
Expose mobile api directly via mobile node bindings

anjmao:
- Disable tequilapi for android mobile app.
- Disable proposals auto fetcher for android mobile app. Mobile app will load proposals once and allow user to refresh via pull to refresh.
- Expose needed API for android directly via mobile node.
- Fix few disconnect issues due to race conditions.


tequilapi/noop_http_api_server.go
```diff
@@ -0,0 +1,39 @@
...
+	return &noopAPIServer{}
+}
+
+type noopAPIServer struct{}
+
+func (n noopAPIServer) Wait() error {
+	return nil
+}
+
+func (n noopAPIServer) StartServing() {
```
tadaskay: Since this is noop, remove the following from `core/node/node.go`:
```
log.Info().Msg("API started on: " + address)
```
Otherwise it will say `API started on: noop`. Add it to the HTTP API server instead.

tadaskay:
Try to avoid making single commit PRs. It's easier to review on a commit-by-commit basis, just like you outlined changes in your PR description.

Issue #1408, opened 2019-11-11
RFC: detect and push system DNS IPs to consumers


zolia:
Currently consumers willing to use providers DNS have to dns-proxy traffic through provider. 
This poses some issues since provider has to run with lifted permissions to run DNS service on port 53. 

Also this rises questions with transparency since consumer is not able to see which actual DNS server is being used.

Proposed model would be to implement parameters for DNS configuration with following defaults:

<pre>
--dns="provider" (default)
--dns="8.8.8.8,1.1.1.1"
--dns="system"
</pre>

For unix type systems we could try getting system DNS'es using: 

github.com/miekg/dns

In case node fails to get DNS'es using this method. We should fall back to some default (i.e.  openDNSes)

Consumer node should check if received DNS IPs are not overlapping with local IPs. In case they do we should fall back to system DNS servers.

## dns setup for traffic slicing 

In order to facilitate traffic slicing DNS requests need to go through node itself. In case node will want to enable traffic slicing it will have to start DNS as a proxy or redirect DNS requests through itself.

To make it more portable it makes sense to start node with DNS server inside on port 53. This can also be done with iptables rules, but limited to Linux only.  

Be default, node should not start DNS server on 53 port, but push its own DNS servers to consumer side. If slicing will be turned on it should start pushing itself or start using iptables redirect rules.

tadaskay:
If `dns=provider` fails, should we try to fallback to `system` and then to `public` or just `public`?
If `dns=system` fails, should we fallback to `public` or error?

* `public` = cf/google/opendns

soffokl:
I don't think we need to have a fallback policy here. 
If the consumer wants to use `provider` DNS, he doesn't want to leak any requests to other servers. We cannot do it transparently for the consumer. It should be explicitly defined that we are passing DNS requests to other servers.

zolia:
@soffokl is right, we should not fall back since that would unexpected for consumer and would result in DNS leaks. 

As for the case we discussed when DNS server for provider is a router or some other LAN device (such as 192.168.0.1). We should not proxy request through node itself as I suggested at first. It would be better to set up an exception to firewall rule to access just that IP in LAN to 53 udp / tcp ports from VPN subnet. That would involve NATting just to that LAN address on macOS and DROP exception on Linux.

zolia:
[...]

PR #1431, opened 2019-11-27
Use provider's DNS proxy in consumer side (with selection strategies)

tadaskay:
Fixes: #1408 

- macOS and linux providers will always start DNS proxies (proxying to DNS servers configured on the provider's system) on ports 11153 (openvpn) and 11253 (wireguard). Windows provider will not provider a DNS server (though client's will fallback if using the default `auto` strategy)
- Clients can choose dns selection strategy: `auto|provider|system|exact` (exact = exact IPs); default is `auto` which will try provider first but fallback to the system DNS if provider DNS is unavailable


tadaskay:
@soffokl Just wanted you to take a quick look into the last 3-4 commits if you're ok with the refactoring done.

services/openvpn/service/manager.go
```diff
@@ -102,10 +100,20 @@ func (m *Manager) Serve(providerID identity.Identity) (err error) {
 		IP:   net.ParseIP(m.serviceOptions.Subnet),
 		Mask: net.IPMask(net.ParseIP(m.serviceOptions.Netmask).To4()),
 	}
-	err = m.natService.Add(nat.RuleForwarding{
+
+	defer func() {
+		err = m.natService.Clear()
+	}()
+	protectNetworksRule := nat.ProtectNetworksRule(m.vpnNetwork.String(), m.outboundIP)
```
vkuznecovas:
```suggestion
	if protectNetworksRule := nat.ProtectNetworksRule(m.vpnNetwork.String(), m.outboundIP); protectNetworksRule != nil {
	if err := m.natService.Add(*protectNetworksRule); err != nil {
			return errors.Wrap(err, "failed to add NAT block rule for protected networks")
		}
	}
```

services/wireguard/service/service_unix.go
```diff
@@ -94,34 +93,26 @@ func (manager *Manager) ProvideConfig(sessionConfig json.RawMessage, traversalPa
...
-	}()
 
 	outIP, err := manager.ipResolver.GetOutboundIPAsString()
 	if err != nil {
 		return nil, err
 	}
 
-	natRule := nat.RuleForwarding{SourceSubnet: config.Consumer.IPAddress.String(), TargetIP: outIP}
+	protectNetworksRule := nat.ProtectNetworksRule(config.Consumer.IPAddress.String(), outIP)
+	if protectNetworksRule != nil {
```
vkuznecovas: same as above

nat/interface.go
```diff
@@ -20,13 +20,77 @@ package nat
...
-// RuleForwarding describes fake nat rule
-type RuleForwarding struct {
-	SourceSubnet string
-	TargetIP     string
+// RuleType specifies what happens to the packets matching the rule
+type RuleType string
+
+const (
+	// Forward matching packets will be accepted and forwarded
+	Forward = RuleType("forward")
```
anjmao: Add RuleType  prefix for each value otherwise nat package will have public nat.Forward, nat.Block which sound like structs or functions.  

```go
RuleTypeForward = RuleType("forward")
RuleTypeBlock = RuleType("block")
```

nat/rules.go
```diff
@@ -0,0 +1,41 @@
...
+
+import (
+	"strings"
+
+	"github.com/mysteriumnetwork/node/config"
+	"github.com/rs/zerolog/log"
+)
+
+// ProtectNetworksRule creates a NAT rule to block packets to protected networks if set.
+func ProtectNetworksRule(sourceSubnet, targetIP string) *Rule {
```
anjmao: ProtectNetworksRule sounds like action. Maybe call it `NewProtectedNetworksRule`.

nat/rules.go
```diff
@@ -0,0 +1,41 @@
...
+	s := config.GetString(config.FlagFirewallProtectedNetworks)
```
anjmao: Try to not depend on global state here. This method can be pure function which creates nat rule from given params. It would be more clear if you check if FlagFirewallProtectedNetworks is set before calling this method and pass servers string directly.

tadaskay: The point of extracting this to a separate function was to encapsulate the logic and improve readability of the `openvpn/manager.Serve`. The boilerplate part of this is string parsing so if it is left out, there's no point to extract it anymore. Passing server list as a single comma separated string to the function does not seem right as it says little about the contract: it would make more sense to pass already parsed IPs.

core/connection/connect_options.go
```diff
@@ -18,16 +18,74 @@
...
+
 	"github.com/mysteriumnetwork/node/identity"
 	"github.com/mysteriumnetwork/node/market"
 	"github.com/mysteriumnetwork/node/session"
+	"github.com/pkg/errors"
 )
 
+// DNSOption defines DNS server selection strategy for consumer
+type DNSOption string
+
```
anjmao: Add unit tests for NewDNSOption, UnmarshalJSON, Exact

nat/service_iptables.go
```diff
@@ -18,111 +18,172 @@
...
+	chainPreRouting  = "PREROUTING"
+	chainPostRouting = "POSTROUTING"
+)
 
-	if _, ok := service.rules[rule]; ok {
-		return errors.New("rule already exists")
-	}
-	service.rules[rule] = struct{}{}
+// Setup sets NAT/Firewall rules for the given NATOptions.
+func (svc *serviceIPTables) Setup(opts Options) (appliedRules []interface{}, err error) {
```
anjmao: Add unit test

nat/service_iptables.go
```diff
@@ -18,111 +18,172 @@
...
+	}
+	log.Info().Msg("Setting up NAT/Firewall rules... done")
+	return untypedIptRules(applied), nil
 }
 
-func (service *serviceIPTables) Del(rule RuleForwarding) error {
-	if err := iptables("delete", rule); err != nil {
-		return err
+// Del removes given NAT/Firewall rules that were previously set up.
+func (svc *serviceIPTables) Del(rules []interface{}) (err error) {
```
anjmao: Maybe unit test

services/openvpn/service/manager.go
```diff
@@ -103,12 +102,16 @@ func (m *Manager) Serve(providerID identity.Identity) (err error) {
...
-		SourceSubnet: m.vpnNetwork.String(),
-		TargetIP:     m.outboundIP,
-	})
-	if err != nil {
-		return errors.Wrap(err, "failed to add NAT forwarding rule")
+
+	var dnsOK bool
+	var dnsIP net.IP
+	var dnsPort = 11153
+	m.dnsProxy = dns.NewProxy("", dnsPort)
```
anjmao: Would it be possible to run one DNS proxy for both OpenVPN and Wireguard?

services/openvpn/service/manager.go
```diff
@@ -103,12 +102,16 @@ func (m *Manager) Serve(providerID identity.Identity) (err error) {
...
-	err = m.natService.Add(nat.RuleForwarding{
-		SourceSubnet: m.vpnNetwork.String(),
-		TargetIP:     m.outboundIP,
-	})
-	if err != nil {
-		return errors.Wrap(err, "failed to add NAT forwarding rule")
+
+	var dnsOK bool
+	var dnsIP net.IP
+	var dnsPort = 11153
```
anjmao: Would it be possible to allow user to set dnsPort = 53? For example If I run in datacenter I can run provider node with dns on 53 port.

services/openvpn/service/manager.go
tadaskay: Yes, but it's more complicated if we want to keep services isolated. If the node starts up, it may be a consumer node, so we don't want to always start a DNS proxy. Just for provider nodes.

nat/service_iptables.go
tadaskay: Unit test would require to at least run actual iptables and see if the packets are routed/redirected properly. It would make more sense to actually do an end to end test case for this instead.

services/openvpn/client_config.go
```diff
@@ -113,22 +113,34 @@ func selectDNSServers(vpnConfig *VPNConfig, dnsOption connection.DNSOption) ([]s
...
+		log.Debug().Msg("Attempting to use provider DNS")
+		if providerDNS, err := selectDNSFromProvider(vpnConfig); err == nil {
+			return providerDNS, nil
 		}
+		log.Debug().Msg("Attempting to use system DNS")
 		if systemDNS, err := dns.ConfiguredServers(); err == nil {
 			return systemDNS, nil
 		}
 	}
+	log.Debug().Msg("Falling back to public DNS")
```
anjmao: What is public DNS in this case? Maybe it should be `Falling back to system DNS` ?

services/openvpn/client.go
```diff
@@ -131,7 +131,7 @@ type VPNConfig struct {
 	OriginalRemoteIP   string
 	OriginalRemotePort int
 
-	DNS             string `json:"dns"`
+	DNS             string `json:"dns_servers"`
```
anjmao: Maybe rename this field to DNSServers too to match json field.  It would be easer to understand because usually you expect json structure to be the same as your fields names.

PR #1432, opened 2019-11-28
Wireguard fixes

tadaskay:
Extracted from #1431 

Fixes a few consumer side errors with wireguard providers.


services/wireguard/endpoint/userspace/tun_windows.go
```diff
@@ -89,7 +89,7 @@ func (tun *nativeTun) Close() error {
 }
 
 func (tun *nativeTun) Flush() error {
-	return tun.Flush()
```
soffokl: We don't need to flush any more?

tadaskay: There's no flush on the delegate. This was an error causing stack overflow (calling flush on self)

Issue #1436, opened 2019-11-29
Better handling of HTTP requests after connected to VPN


anjmao:
**Problem**
Currently in our [HTTP clients](https://github.com/mysteriumnetwork/node/blob/master/requests/client.go#L50) with have 20 seconds timeouts and disabled keep alive. The problem is that if consumer connects to slow VPN server it can still get timeouts or connection close errors. Another problem is that we disable keep alive for all cases.

**Proposal**
I propose to implement custom HTTP transport which can close idle connections when connection state changed. We can get this state change by subscribing to our EvenBus.

**Example**
Here is pseudo  implementation which works on my local machine. In my local tests I got much quicker responses and no timeouts or closed connections issues.

```go
// NodeHTTPTransport is a wrapper around http.Transport which can close
// idle connections and do http retries in case of a VPN connection change.
type NodeHTTPTransport struct {
	rt http.RoundTripper
	eventBus EventBus
}

// Subscribe subscribes to connection state event
// and closes idle connections if connection changed.
func (t *NodeHTTPTransport) Subscribe() {
	type closeIdler interface {
		CloseIdleConnections()
	}

	t.Subscribe(StateEventTopic, func(e StateEvent) {
		if connChanged {
			if tr, ok := t.rt.(closeIdler); ok {
				tr.CloseIdleConnections()
			}
		}	
	})
}

// RoundTrip does HTTP request with 1 extra when got timeout or close error.
func (t *NodeHTTPTransport) RoundTrip(req *http.Request) (*http.Response, error) {
	// Handle err here and retry in case it was canceled.
	return t.rt.RoundTrip(req)
}

func main() {
	client := http.Client{
		Timeout:   20 * time.Second,
		Transport: NewNodeHTTPTransport(eventBus),
	}
}
```

vkuznecovas:
I'm all up for this, however one thing you'll need to take into account in the implementation is the state change timing. It might not occur as rapidly as you'd like in some cases(for instance, on connection establishing) as some time is passed before a "connected" event is emitted. Need to check if the events would be sufficient. In case they are not, the timings might need adjustment introduction of completely new ones.

soffokl:
There is a comment in the code, which is almost TODO regarding this:
https://github.com/mysteriumnetwork/node/blob/master/requests/client.go#L48-L49

We just need to make sure it works, and I fully support these changes.

zolia:
This is nice workaround. My concerns are also with this dependency on event. But otherwise - good thinking!

PR #1438, opened 2019-12-01
Close idle http requests on connection state change

anjmao:
Use single HTTP client which can be recreated. This allows to close idle connections once connection state changed (Connected -> NotConnected or NotConnected -> Connected) and create new underlying HTTP client.

Fixes https://github.com/mysteriumnetwork/node/issues/1436


cmd/di.go
```diff
@@ -465,8 +476,9 @@ func (di *Dependencies) bootstrapNodeComponents(nodeOptions node.Options, tequil
 	// TODO: switch this to channel implementation from params after #1325 is merged
 	channelImplementation := metadata.TestnetDefinition.ChannelImplAddress
 
+	client := requests.NewHTTPClient(nodeOptions.BindAddress, requests.DefaultTimeout)
```
soffokl: Can't we just use `di.HTTPClient` here?

cmd/di.go
```diff
@@ -809,3 +821,24 @@ func (di *Dependencies) bootstrapFirewall(options node.OptionsFirewall) error {
 	}
 	return nil
 }
+
+func (di *Dependencies) handleHTTPClientConnections() error {
+	if di.HTTPClient == nil {
+		return errors.New("HTTPClient is not initialized")
+	}
+
+	latestState := connection.NotConnected
```
soffokl: Is it thread-safe to have a kind of global variable here with the state. Can we receive several events that will try to change single variable at the same time?

core/location/oracle_resolver.go
```diff
@@ -18,23 +18,21 @@
...
 }
 
 // NewOracleResolver returns new db resolver initialized from Location Oracle service
-func NewOracleResolver(srcIP, address string) *oracleResolver {
+func NewOracleResolver(httpClient *requests.HTTPClient, address string) *oracleResolver {
 	return &oracleResolver{
-		requests.NewHTTPClient(srcIP, 20*time.Second),
-		address,
+		http:                  httpClient,
+		oracleResolverAddress: address,
```
soffokl: I know it's not related to the PR, but the this `oracleResolver.oracleResolverAddress` could be better like `oracleResolver.Address`.

core/transactor/transactor.go
```diff
@@ -42,9 +41,9 @@ type Transactor struct {
 }
 
 // NewTransactor creates and returns new Transactor instance
-func NewTransactor(bindAddress, endpointAddress, registryAddress, accountantID, channelImplementation string, signerFactory identity.SignerFactory) *Transactor {
+func NewTransactor(httpClient *requests.HTTPClient, endpointAddress, registryAddress, accountantID, channelImplementation string, signerFactory identity.SignerFactory) *Transactor {
 	return &Transactor{
-		http:                  requests.NewHTTPClient(bindAddress, 20*time.Second),
+		httpClient:            httpClient,
```
soffokl: In [some](https://github.com/mysteriumnetwork/node/pull/1438/files#diff-4801e8de631b36d8ef7e26fabad34ae2R83) places you are calling it  `http` in other it's `httpClient`. It should be consistent everywhere.

requests/transport.go
```diff
@@ -0,0 +1,64 @@
...
+	"github.com/rs/zerolog/log"
+)
+
+const (
+	maxHTTPRetries  = 3
+	delayAfterRetry = 50 * time.Millisecond
+)
+
+func newTransport(localIPAddress *net.TCPAddr) *transport {
+	t := &transport{
```
soffokl:
```suggestion
	return &transport{
```

requests/transport.go
```diff
@@ -0,0 +1,64 @@
...
+// RoundTrip does HTTP requests with retries. Retries are needed since
+// error can happen when HTTP client closes idle connections.
+func (t *transport) RoundTrip(req *http.Request) (*http.Response, error) {
+	for i := 1; i <= maxHTTPRetries; i++ {
+		res, err := t.rt.RoundTrip(req)
+		if err == nil {
+			return res, nil
+		}
+
+		log.Error().Err(err).Msgf("Failed to call %q. Retrying.", req.URL.String())
```
soffokl: I'd say it's warning or info/debug not error, if we can recover it by retrying.  We can log error at the end of the function if we failed to succeed it.

requests/transport.go
```diff
@@ -0,0 +1,64 @@
...
+	"net"
+	"net/http"
+	"time"
+
+	"github.com/rs/zerolog/log"
+)
+
+const (
+	maxHTTPRetries  = 3
+	delayAfterRetry = 50 * time.Millisecond
```
soffokl: Does it enough to have only 150ms(3*50) for retries?

cmd/di.go
vkuznecovas: depends if you care about races, as this will race. Subscribe async means that the handler will be called in a goroutine, so if many events get flooded this will have a racy behaviour. 

1) add a lock in the subscribe async
2) consider subscribing in a non async fashion

requests/transport.go
```diff
@@ -0,0 +1,64 @@
...
+// error can happen when HTTP client closes idle connections.
+func (t *transport) RoundTrip(req *http.Request) (*http.Response, error) {
+	for i := 1; i <= maxHTTPRetries; i++ {
+		res, err := t.rt.RoundTrip(req)
+		if err == nil {
+			return res, nil
+		}
+
+		log.Error().Err(err).Msgf("Failed to call %q. Retrying.", req.URL.String())
+		time.Sleep(delayAfterRetry)
```
vkuznecovas: this prevents graceful shutdowns. Say I stop the node, the connection state changes to cancelled. This cancels the request which then causes this request to be retries 2 more times. If the requests are slow, it will hang for a long imte. It will cause the node to hang. Suggest subscribing to a stop event, which will then cause a channel to be closed. This part can then become something like this:

```golang
select {
    case <- time.After(delayAfterRetry):
                 if i == maxHTTPRetries {
			return res, err
		}
   case <-t.stop:
               return nil, nil
} 
```

anjmao: I tested in with this simple program.

```go
func runHTTP() {
	c := http.Client{
		Timeout:   20 * time.Second,
		Transport: newTransport(),
	}

	go func() {
		panic(http.ListenAndServe(":8005", nil))
	}()

	for {
		req, _ := http.NewRequest("GET", "http://localhost:8007", nil)
		res, err := c.Do(req)
		if err != nil {
			fmt.Printf("ERR: %v\n", err)
			continue
		}
		defer res.Body.Close()
		b, _ := ioutil.ReadAll(res.Body)
		fmt.Println("OK", time.Now().String(), len(b))
	}
}

func newTransport() {
   // Uses the same logic as is now.
}
```

- Client calls some HTTP server on localhost:8007. Server exposes slow HTTP endpoint which returns response only after 10 seconds.
- Client also starts HTTP on localhost:8005. After I stop client I can run it again immediately otherwise I should get error saying that HTTP port 8005 is in use.

requests/transport.go
zolia: I would say, in case we need a retry, ideally it should be exponential, but for the sake of simplicity we just should make it acceptable to the user. Lets make this (3*350).

cmd/di.go
anjmao: Yes, missed this one.

cmd/di.go
anjmao: Since I care only about Connected or NotConnected events it would be strange If I get race condition here, but yes, in theory it could happen. Changed to Subscribe().

requests/transport.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+type transport struct {
+	rt http.RoundTripper
+}
+
+// RoundTrip does HTTP requests with retries. Retries are needed since
+// error can happen when HTTP client closes idle connections.
+func (t *transport) RoundTrip(req *http.Request) (*http.Response, error) {
+	for i := 1; i <= maxHTTPRetries; i++ {
```
zolia: why not? :)

```suggestion
	for i := 0; i < maxHTTPRetries; i++ {
```

requests/transport.go
anjmao: Added changes as suggested.

PR #1439, opened 2019-12-02
Refactor MMN code, send node type to MMN

ignasbernotas:


mmn/mmn.go
```diff
@@ -0,0 +1,77 @@
...
+	collector *Collector
+	client    *client
+	eventBus  eventbus.EventBus
+}
+
+func NewMMN(collector *Collector, client *client, eventBus eventbus.EventBus) *MMN {
+	return &MMN{collector, client, eventBus}
+}
+
+func (m *MMN) SubscribeToEvents() {
```
vkuznecovas: accept eventBus as a param here, remove from MMN struct as you're already calling a method explicitly for subscription there is no benefit of having it attached to the struct

mmn/mmn.go
```diff
@@ -0,0 +1,77 @@
...
+
+func (m *MMN) subscribeToProviderEvent() {
+	m.subscribeToNodeTypeEvent(service.StatusTopic, true)
+}
+
+func (m *MMN) subscribeToNodeTypeEvent(event string, isProvider bool) {
+	m.eventBus.SubscribeAsync(
+		event,
+		func(i interface{}) {
+			log.Debug().Msgf("Node type (provider: %v) event", isProvider)
```
vkuznecovas: can a node not be both a provider and a consumer at the same time?

mmn/mmn.go
```diff
@@ -0,0 +1,77 @@
...
+	m.subscribeToConsumerEvent()
+}
+
+func (m *MMN) Register(identity string) error {
+	m.collector.SetIdentity(identity)
+
+	return m.client.RegisterNode(m.collector.GetCollectedInformation())
+}
+
+func (m *MMN) UpdateNodeType(isProvider bool) error {
```
vkuznecovas: might want to put this under lock as this will have tendency to race since you're subscribing in async fashion

mmn/mmn.go
```diff
@@ -0,0 +1,77 @@
...
+		return nil
+	}
+
+	m.collector.SetNodeType(isProvider)
+
+	return m.client.UpdateNodeType(m.collector.GetCollectedInformation())
+}
+
+func (m *MMN) subscribeToRegistrationEvent() {
+	m.eventBus.SubscribeAsync(
```
vkuznecovas: subscribes do return an error, don't ignore it

mmn/mmn.go
```diff
@@ -0,0 +1,77 @@
...
+func (m *MMN) subscribeToConsumerEvent() {
+	m.subscribeToNodeTypeEvent(connection.SessionEventTopic, false)
+}
+
+func (m *MMN) subscribeToProviderEvent() {
+	m.subscribeToNodeTypeEvent(service.StatusTopic, true)
+}
+
+func (m *MMN) subscribeToNodeTypeEvent(event string, isProvider bool) {
+	m.eventBus.SubscribeAsync(
```
vkuznecovas: subscribes do return an error, don't ignore it

mmn/mmn.go
```diff
@@ -0,0 +1,77 @@
...
+}
+
+func (m *MMN) subscribeToRegistrationEvent() {
+	m.eventBus.SubscribeAsync(
+		identity.IdentityUnlockTopic,
+		func(identity string) {
+			log.Debug().Msg("Registration event")
+
+			if err := m.Register(identity); err != nil {
+				log.Error().Msg("Failed to register to MMN: " + err.Error())
```
anjmao: Use Msgf which formats message error for you

```go
log.Error().Msgf("Failed to register to MMN: %v",  err)
```

mmn/mmn.go
```diff
@@ -0,0 +1,77 @@
...
+}
+
+func (m *MMN) subscribeToNodeTypeEvent(event string, isProvider bool) {
+	m.eventBus.SubscribeAsync(
+		event,
+		func(i interface{}) {
+			log.Debug().Msgf("Node type (provider: %v) event", isProvider)
+
+			if err := m.UpdateNodeType(isProvider); err != nil {
+				log.Error().Msg("Failed to send node type to MMN: " + err.Error())
```
anjmao: Same here

mmn/mmn.go
```diff
@@ -0,0 +1,77 @@
...
+
+			if err := m.Register(identity); err != nil {
+				log.Error().Msg("Failed to register to MMN: " + err.Error())
+			}
+		},
+	)
+}
+
+func (m *MMN) subscribeToConsumerEvent() {
+	m.subscribeToNodeTypeEvent(connection.SessionEventTopic, false)
```
anjmao: Not sure about this logic. if I understand it correctly you will update node info each time session event is fired, but in my understanding you need to do it only once when node is started.

mmn/collector.go
```diff
@@ -24,64 +24,82 @@ import (
...
-	OS          string `json:"os"`
-	Arch        string `json:"arch"`
-	NodeVersion string `json:"node_version"`
-	Identity    string `json:"identity"`
-	VendorID    string `json:"vendor_id"`
+// Collector collects environment data
+type Collector struct {
+	ipResolver      ip.Resolver
+	node            *NodeInformationDto
+	nodeTypeUpdated bool
```
anjmao: This field seems to be unused.

mmn/mmn.go
```diff
@@ -0,0 +1,118 @@
...
+
+	err = eventBus.SubscribeAsync(
+		service.StatusTopic,
+		m.handleProvider,
+	)
+	if err != nil {
+		return err
+	}
+
+	err = eventBus.SubscribeAsync(
```
vkuznecovas: can just return here

mmn/mmn.go
tadaskay: For errors, it should be: `log.Error().Err(err).Msg("Failed to register to MMN")` @ignasbernotas @anjmao

PR #1440, opened 2019-12-03
Expose user feedback reporter endpoint for mobile entry point

anjmao:
Previously mobile was using crashlytics to send user feedback, now we can use feedback reporter which also attaches node logs.

### DIFF
--- a/mobile/mysterium/entrypoint.go
+++ b/mobile/mysterium/entrypoint.go
@@ -18,6 +18,7 @@
 package mysterium
 
 import (
+	"errors"
 	"fmt"
 	"path/filepath"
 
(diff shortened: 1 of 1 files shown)


mobile/mysterium/entrypoint.go
```diff
@@ -294,6 +297,27 @@ func (mb *MobileNode) UnlockIdentity() (string, error) {
...
+}
+
+// SendFeedback sends user feedback via feedback reported.
+func (mb *MobileNode) SendFeedback(req *SendFeedbackRequest) error {
+	report := feedback.UserReport{
+		Description: req.Description,
+	}
+	result, err := mb.feedbackReporter.NewIssue(report)
+	if err != nil {
+		return err
```
zolia: We should wrap error here.

PR #1454, opened 2019-12-10
Add godoc and coverage report

anjmao:
Add godoc and coverate via https://codecov.io so we are compliant with https://github.com/avelino/awesome-go requirements.


tadaskay:
Nice!

PR #1456, opened 2019-12-11
More accurate descriptions [skip ci]

chompomonim:
We're using blockchain for payments but we're not building
VPN Network on top of it.


tadaskay:
@anjmao Note we are talking about the node as a whole package here, not just the VPN.

> build on sth
> to use a success or achievement as a base from which to achieve more success:

https://dictionary.cambridge.org/dictionary/english/build-on-sth

🤷‍♂

PR #1464, opened 2019-12-18
Topup via identity

chompomonim:


vkuznecovas:
also, can we just postpone this? have 2 separate endpoints a ne wone with channel and this one with id? that way no issues for compatibility

Issue #1460, opened 2019-12-17
Old node logs are not removed


anjmao:
- Rolling logs writer is missing MaxRemain field to keep only last 7 days. Config location https://github.com/mysteriumnetwork/node/blob/master/logconfig/config.go#L92
- Logs directory is set to DataDir as default location for logs https://github.com/mysteriumnetwork/node/blob/master/config/flags_directory.go#L65 but running `myst --help` shows that default logs directory is $HOME/.mysterium. Fix could be to update help doc saying that by default logs would be in the same directory as data.

PR #1467, opened 2019-12-19
Fix log file rolling & cleanup

tadaskay:
Fixes: #1460

### DIFF
--- a/bin/package/installation/post-install.sh
+++ b/bin/package/installation/post-install.sh
@@ -91,5 +91,9 @@ if [[ ! -f $DAEMON_DEFAULT ]]; then
     cp $OS_DIR_INSTALLATION/default $DAEMON_DEFAULT
 fi
 
+# Cleanup old log files (before log file rolling has been fixed)
+printf "Cleaning legacy log files...\n" \
+    && rm -rf /var/lib/mysterium-node/.mysterium.log*
+
 printf "\nInstallation successfully finished.\n" \
     && printf "Usage: service mysterium-node restart\n"
--- a/config/flags_directory.go
+++ b/config/flags_directory.go
@@ -38,7 +38,7 @@ var (
 	// FlagLogDir is a directory for storing log files.
 	FlagLogDir = cli.StringFlag{
 		Name:  "log-dir",
-		Usage: "Log directory for storing log files",
+		Usage: "Log directory for storing log files. data-dir/logs is used if not specified.",
 	}
 	// FlagRuntimeDir runtime writable directory for temporary files.
 	FlagRuntimeDir = cli.StringFlag{
@@ -61,7 +61,7 @@ func RegisterFlagsDirectory(flags *[]cli.Flag) error {
 
 	FlagConfigDir.Value = filepath.Join(currentDir, "config")
 	FlagDataDir.Value = filepath.Join(userHomeDir, ".mysterium")
-	FlagLogDir.Value = FlagDataDir.Value
+	FlagLogDir.Value = filepath.Join(FlagDataDir.Value, "logs")
 	FlagRuntimeDir.Value = currentDir
 
 	*flags = append(*flags,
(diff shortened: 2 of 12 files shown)


bin/package/installation/post-install.sh
```diff
@@ -91,5 +91,9 @@ if [[ ! -f $DAEMON_DEFAULT ]]; then
     cp $OS_DIR_INSTALLATION/default $DAEMON_DEFAULT
 fi
 
+# Cleanup old log files (before log file rolling has been fixed)
+printf "Cleaning legacy log files...\n" \
+    && rm -rf /var/lib/mysterium-node/mysterium.log*
```
anjmao: I believe old logs names starts with dot like `.mysterium.log`.

config/flags_directory.go
```diff
@@ -62,7 +61,7 @@ func RegisterFlagsDirectory(flags *[]cli.Flag) error {
 
 	FlagConfigDir.Value = filepath.Join(currentDir, "config")
 	FlagDataDir.Value = filepath.Join(userHomeDir, ".mysterium")
-	FlagLogDir.Value = FlagDataDir.Value
+	FlagLogDir.Value = filepath.Join(FlagDataDir.Value, "logs")
```
anjmao: Also update cli docs mentioning that default path would be data dir + logs.

core/node/options.go
```diff
@@ -138,7 +140,10 @@ func GetOptions() *Options {
 
 // GetLogOptions retrieves logger options from the app configuration.
 func GetLogOptions() *logconfig.LogOptions {
-	logDir := config.GetString(config.FlagLogDir)
+	filepath := ""
+	if logDir := config.GetString(config.FlagLogDir); logDir != "" {
```
anjmao: Not sure if we need to append with custom path if user sets his own logDir. I would expect logs to be written to my specified directory.

debian/postinst
```diff
@@ -91,6 +91,10 @@ if [[ ! -f $DAEMON_DEFAULT ]]; then
     cp $OS_DIR_INSTALLATION/default $DAEMON_DEFAULT
 fi
 
+# Cleanup old log files (before log file rolling has been fixed)
+printf "Cleaning legacy log files...\n" \
+    && rm -rf /var/lib/mysterium-node/mysterium.log*
```
anjmao: Same here, check if logs starts with `.mysterium*`.

logconfig/collector.go
```diff
@@ -38,6 +38,9 @@ func NewCollector(options *LogOptions) *Collector {
 
 // Archive creates ZIP archive containing all node log files.
 func (c *Collector) Archive() (outputFilepath string, err error) {
+	if c.options.Filepath == "" {
+		return "", errors.New("file logging is disabled, can't retrieve logs")
+	}
```
anjmao: This method is still possible insecure as it may archive not only logs files. I suggest to filter file names to ensure that they start with `.mysterium-log*`.

core/node/options.go
tadaskay: This builds the filepath using given log directory. Filepath is directory + filename.

logconfig/collector.go
tadaskay: I believe this comment points at the wrong line. This fragment specifically checks for disabled file logging.

Sure, I'll check what we can do.

config/flags_directory.go
tadaskay: Actually, it's kind of mentioned, because we set flag's `Value`. Here's an example (`node --help`):
```
--log-dir value                      Log directory for storing log files (default: "/Users/tadas/.mysterium/logs")
```

logconfig/collector.go
tadaskay: It should be `mysterium-node.log`, not `mysterium-log`.

`logFilepaths` actually checks for this. The problem was that the filename was not being set (so there was nothing to check against). Everything should work now as originally intended.

Issue #1330, opened 2019-10-08
e2e tests: add accountant, test payment flow


vkuznecovas:
need to add an accountant to the e2e testing, add tests if needed. Work should be based on payment-v2-messaging branch.

PR #1485, opened 2020-01-06
Invoice tracker will now work independent of the order of exchange me…

### LINKED ISSUE #1466: invoice_tracker.go - separate wait from send
at the moment we send a message and then wait for a response. This is not ideal as in some situations an invoice might get sent after the timeout - meaning that we'll have a backlog of invoices to go through, causing hashlock missmatches

vkuznecovas:
…ssages

Invoice tracker now does a better job of keeping track of invoices.

No matter what orders the correspoding exchange messages arrive, the invoice tracker will be able to detect if it was correctly issued.

This fixes #1466

### DIFF
--- a/session/pingpong/invoice_tracker.go
+++ b/session/pingpong/invoice_tracker.go
@@ -43,6 +43,18 @@ var ErrConsumerPromiseValidationFailed = errors.New("consumer failed to issue pr
 // ErrAccountantFeeTooLarge indicates that we do not allow accountants with such high fees
 var ErrAccountantFeeTooLarge = errors.New("accountants fee exceeds")
 
+// ErrInvoiceExpired shows that the given invoice has already expired
+var ErrInvoiceExpired = errors.New("invoice expired")
+
+// ErrExchangeWaitTimeout indicates that we did not get an exchange message in time.
+var ErrExchangeWaitTimeout = errors.New("did not get a new exchange message")
+
+// ErrExchangeValidationFailed indicates that there was an error with the exchange signature.
+var ErrExchangeValidationFailed = errors.New("exchange validation failed")
+
+// ErrConsumerNotRegistered represents the error that the consumer is not registered
+var ErrConsumerNotRegistered = errors.New("consumer not registered")
+
 // PeerInvoiceSender allows to send invoices.
 type PeerInvoiceSender interface {
 	Send(crypto.Invoice) error
@@ -75,55 +87,26 @@ type accountantCaller interface {
 	RevealR(r string, provider string, agreementID uint64) error
 }
 
-// ErrExchangeWaitTimeout indicates that we did not get an exchange message in time.
-var ErrExchangeWaitTimeout = errors.New("did not get a new exchange message")
-
-// ErrMissmatchingHashlock represents an error where a consumer sends a hashlock we're not waiting for
-var ErrMissmatchingHashlock = errors.New("hashlock missmatch")
-
-// ErrExchangeValidationFailed indicates that there was an error with the exchange signature.
-var ErrExchangeValidationFailed = errors.New("exchange validation failed")
-
-// ErrConsumerNotRegistered represents the error that the consumer is not registered
-var ErrConsumerNotRegistered = errors.New("consumer not registered")
-
 const chargePeriodLeeway = time.Hour * 2
 
-type lastInvoice struct {
+type sentInvoice struct {
 	invoice crypto.Invoice
 	r       []byte
 }
 
 // InvoiceTracker keeps tab of invoices and sends them to the consumer.
 type InvoiceTracker struct {
-	peer                            identity.Identity
 	stop                            chan struct{}
-	peerInvoiceSender               PeerInvoiceSender
-	exchangeMessageChan             chan crypto.ExchangeMessage
-	chargePeriod                    time.Duration
-	exchangeMessageWaitTimeout      time.Duration
 	accountantFailureCount          uint64
 	notReceivedExchangeMessageCount uint64
 	maxNotReceivedExchangeMessages  uint64
 	once                            sync.Once
-	invoiceStorage                  providerInvoiceStorage
-	accountantPromiseStorage        accountantPromiseStorage
-	timeTracker                     timeTracker
-	paymentInfo                     dto.PaymentRate
-	providerID                      identity.Identity
-	accountantID                    identity.Identity
-	lastInvoice                     lastInvoice
+	agreementID                     uint64
 	lastExchangeMessage             crypto.ExchangeMessage
-	accountantCaller                accountantCaller
-	registryAddress                 string
-	maxAccountantFailureCount       uint64
-	maxAllowedAccountantFee         uint16
-	bcHelper                        bcHelper
-	publisher                       eventbus.Publisher
-	feeProvider                     feeProvider
 	transactorFee                   uint64
-	maxRRecoveryLength              uint64
-	channelAddressCalculator        channelAddressCalculator
+	invoicesSent                    map[string]sentInvoice
+	invoiceLock                     sync.Mutex
+	deps                            InvoiceTrackerDeps
 }
 
 // InvoiceTrackerDeps contains all the deps needed for invoice tracker.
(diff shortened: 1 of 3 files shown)


session/pingpong/invoice_tracker.go
```diff
@@ -154,57 +137,155 @@ type InvoiceTrackerDeps struct {
...
 		maxNotReceivedExchangeMessages: calculateMaxNotReceivedExchangeMessageCount(chargePeriodLeeway, itd.ChargePeriod),
-		maxAccountantFailureCount:      itd.MaxAccountantFailureCount,
-		maxAllowedAccountantFee:        itd.MaxAllowedAccountantFee,
-		bcHelper:                       itd.BlockchainHelper,
-		publisher:                      itd.Publisher,
-		registryAddress:                itd.Registry,
-		feeProvider:                    itd.FeeProvider,
-		channelAddressCalculator:       itd.ChannelAddressCalculator,
-		maxRRecoveryLength:             itd.MaxRRecoveryLength,
+		invoicesSent:                   make(map[string]sentInvoice, 0),
```
soffokl:
```suggestion
		invoicesSent:                   make(map[string]sentInvoice),
```

session/pingpong/invoice_tracker.go
```diff
@@ -154,57 +137,155 @@ type InvoiceTrackerDeps struct {
...
+	it.lastExchangeMessage = pm
+	it.markInvoicePaid(pm.Promise.Hashlock)
+	it.resetNotReceivedExchangeMessageCount()
+
+	needsRevealing := false
+	accountantPromise, err := it.deps.AccountantPromiseStorage.Get(it.deps.ProviderID, it.deps.AccountantID)
+	switch err {
+	case nil:
+		needsRevealing = !accountantPromise.Revealed
+		break
```
soffokl: > redundant break statement (gosimple)

session/pingpong/invoice_tracker.go
```diff
@@ -154,57 +137,155 @@ type InvoiceTrackerDeps struct {
...
+	case ErrNotFound:
+		needsRevealing = false
+		break
```
soffokl: > redundant break statement (gosimple)

session/pingpong/invoice_tracker.go
```diff
@@ -154,57 +137,153 @@ type InvoiceTrackerDeps struct {
...
+		err = it.deps.AccountantPromiseStorage.Store(it.deps.ProviderID, it.deps.AccountantID, accountantPromise)
+		if err != nil {
+			return errors.Wrap(err, "could not store accountant promise")
+		}
+		log.Debug().Msg("Accountant promise stored")
+	}
+
+	err = it.deps.InvoiceStorage.StoreR(it.deps.ProviderID, it.agreementID, hex.EncodeToString(invoice.r))
+	if err != nil {
+		return errors.Wrap(err, "could not store r")
```
zolia: would be good to log 'r'.

session/pingpong/invoice_tracker.go
```diff
@@ -154,57 +137,153 @@ type InvoiceTrackerDeps struct {
...
+			if err != nil && err != ErrInvoiceExpired {
+				return err
+			}
+		case <-it.stop:
+			return nil
+		}
+	}
+}
+
+func (it *InvoiceTracker) handleExchangeMessage(pm crypto.ExchangeMessage) error {
```
zolia: too long function. Can we split it to smaller ones for readability?

session/pingpong/invoice_tracker.go
```diff
@@ -154,57 +137,153 @@ type InvoiceTrackerDeps struct {
...
+		if strings.Contains(err.Error(), "400 Bad Request") {
+			recoveryError := it.initiateRRecovery()
+			if recoveryError != nil {
+				return errors.Wrap(err, "could not recover R")
+			}
+		}
+
+		it.incrementAccountantFailureCount()
+		if it.getAccountantFailureCount() > it.deps.MaxAccountantFailureCount {
+			return errors.Wrap(err, "could not call accountant")
```
zolia:
```suggestion
			return errors.Wrap(err, "Max failure count calling accountant reached")
```

session/pingpong/invoice_tracker.go
```diff
@@ -154,57 +137,153 @@ type InvoiceTrackerDeps struct {
...
+	}
+	log.Debug().Msg("Accountant promise stored")
+
+	promise.R = invoice.r
+	it.deps.Publisher.Publish(AccountantPromiseTopic, AccountantPromiseEventPayload{
+		Promise:      promise,
+		AccountantID: it.deps.AccountantID,
+		ProviderID:   it.deps.ProviderID,
+	})
+	it.resetAccountantFailureCount()
```
zolia: we already reset count above. Count did not chance since.

session/pingpong/invoice_tracker.go
```diff
@@ -213,43 +292,54 @@ func (it *InvoiceTracker) Start() error {
 		return ErrConsumerNotRegistered
 	}
 
-	fees, err := it.feeProvider.FetchSettleFees()
+	fees, err := it.deps.FeeProvider.FetchSettleFees()
 	if err != nil {
 		return errors.Wrap(err, "could not fetch fees")
```
zolia:
```suggestion
		return errors.Wrap(err, "could not fetch settlement fees")
```

session/pingpong/invoice_tracker.go
```diff
@@ -213,43 +292,54 @@ func (it *InvoiceTracker) Start() error {
...
+
 	// give the consumer a second to start up his payments before sending the first request
 	firstSend := time.After(time.Second)
 	for {
 		select {
 		case <-firstSend:
-			err := it.sendInvoiceExpectExchangeMessage()
+			err := it.sendInvoice()
 			if err != nil {
 				return err
```
zolia:
```suggestion
				return errors.wrap(err, "sending first invoice failed")
```

session/pingpong/invoice_tracker.go
```diff
@@ -213,43 +292,54 @@ func (it *InvoiceTracker) Start() error {
...
 			return nil
-		case <-time.After(it.chargePeriod):
-			err := it.sendInvoiceExpectExchangeMessage()
+		case <-time.After(it.deps.ChargePeriod):
+			err := it.sendInvoice()
 			if err != nil {
 				return err
 			}
+		case emErr := <-emErrors:
+			if emErr != nil {
```
zolia: is this condition needed?

session/pingpong/invoice_tracker.go
```diff
@@ -154,57 +137,153 @@ type InvoiceTrackerDeps struct {
...
+		case <-it.stop:
+			return nil
+		}
+	}
+}
+
+func (it *InvoiceTracker) handleExchangeMessage(pm crypto.ExchangeMessage) error {
+	invoice, ok := it.getMarkedInvoice(pm.Promise.Hashlock)
+	if !ok {
+		log.Debug().Msgf("consumer sent exchange message with missing expired hashlock %v, skipping", invoice.invoice.Hashlock)
```
anjmao: When using %v it looks like Hashlock is some struct. Better use %s since Hashlock is of type string.

session/pingpong/invoice_tracker.go
```diff
@@ -154,57 +137,153 @@ type InvoiceTrackerDeps struct {
...
 	}
 
-	r := it.generateR()
-	invoice := crypto.CreateInvoice(agreementID, it.paymentInfo.GetPrice().Amount, 0, r)
-	invoice.Provider = it.providerID.Address
-	it.lastInvoice = lastInvoice{
-		invoice: invoice,
-		r:       r,
+	it.lastExchangeMessage = pm
+	it.markInvoicePaid(pm.Promise.Hashlock)
```
anjmao: Do `markInvoicePaid` need mutex if handleExchangeMessage is executed in single goroutine?

session/pingpong/invoice_tracker.go
```diff
@@ -154,57 +137,153 @@ type InvoiceTrackerDeps struct {
...
+	if err != nil {
+		return errors.Wrap(err, "could not store r")
+	}
+
+	promise, err := it.deps.AccountantCaller.RequestPromise(pm)
+	if err != nil {
+		log.Warn().Err(err).Msg("Could not call accountant")
+
+		// TODO: handle this better
+		if strings.Contains(err.Error(), "400 Bad Request") {
```
anjmao: RequestPromise could return err which could be checked like:

```go
if err == accountant_caller_pkg.ErrBadRequest {
}
```

session/pingpong/invoice_tracker_test.go
```diff
@@ -538,43 +531,45 @@ func TestInvoiceTracker_receiveExchangeMessageOrTimeout(t *testing.T) {
...
-				accountantPromiseStorage:   tt.fields.accountantPromiseStorage,
-				accountantID:               tt.fields.accountantID,
-				lastExchangeMessage:        tt.fields.lastExchangeMessage,
-				accountantCaller:           tt.fields.accountantCaller,
-				lastInvoice:                tt.fields.lastInvoice,
-				registryAddress:            tt.fields.registryAddress,
-				publisher:                  &mockPublisher{},
-				invoiceStorage:             NewProviderInvoiceStorage(NewInvoiceStorage(bolt)),
-				channelAddressCalculator:   NewChannelAddressCalculator(tt.fields.accountantID.Address, tt.fields.channelImplementation, tt.fields.registryAddress),
+			deps := InvoiceTrackerDeps{
```
anjmao: Check unit tests coverage report. Would be good to have higher coverage for payments related stuff :)

session/pingpong/invoice_tracker.go
vkuznecovas: yes, because on channel close we'll get a nil value

session/pingpong/invoice_tracker.go
vkuznecovas: there are 3 routines - one is sending the invoices and marking them as sent, while the other is receiving the exchange messages and marking the relevant invoices as sent. There's also a timeout which is running on a separate routine as well.

session/pingpong/invoice_tracker.go
vkuznecovas: this should in general unified with accountant via error codes, because 400 is not descriptive enough.

https://github.com/mysteriumnetwork/accountant/issues/50

session/pingpong/invoice_tracker_test.go
vkuznecovas: it should really be covered in an e2e flow, see:

https://github.com/mysteriumnetwork/node/issues/1330

PR #1489, opened 2020-01-08
Fix leaking goroutines in the connection manager

soffokl:
It now clean channels correctly for statistics and IP checking.

The goroutine profile from the leaking instance:
```
1393 @ 0x4345d0 0x409248 0x40921e 0x408f4b 0xaa9f02 0x462891
#	0xaa9f01	github.com/mysteriumnetwork/node/core/connection.(*connectionManager).consumeStats+0x191	/go/src/github.com/mysteriumnetwork/node/core/connection/manager.go:480
1293 @ 0x4345d0 0x40862b 0x408601 0x4083e5 0xaaabf9 0xaa79c6 0x462891
#	0xaaabf8	github.com/mysteriumnetwork/node/core/connection.(*connectionManager).checkSessionIP.func1+0x38	/go/src/github.com/mysteriumnetwork/node/core/connection/manager.go:204
#	0xaa79c5	github.com/mysteriumnetwork/node/core/connection.(*connectionManager).checkSessionIP+0x255	/go/src/github.com/mysteriumnetwork/node/core/connection/manager.go:218
```

### DIFF
--- a/core/connection/manager.go
+++ b/core/connection/manager.go
@@ -197,6 +197,12 @@ func (manager *connectionManager) Connect(consumerID, accountantID identity.Iden
 	}
 
 	go manager.checkSessionIP(dialog, sessionDTO.ID, originalPublicIP)
+
+	go func() {
+		<-manager.ipCheckParams.Done
+		log.Trace().Msgf("IP check is done for session %v", sessionDTO.ID)
+	}()
+
 	return nil
 }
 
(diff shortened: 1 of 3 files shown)


core/connection/manager.go
```diff
@@ -484,12 +492,17 @@ func (manager *connectionManager) consumeConnectionStates(stateChannel <-chan St
 	logDisconnectError(manager.Disconnect())
 }
 
-func (manager *connectionManager) consumeStats(statisticsChannel <-chan consumer.SessionStatistics) {
-	for stats := range statisticsChannel {
-		manager.eventPublisher.Publish(StatisticsEventTopic, SessionStatsEvent{
-			Stats:       stats,
-			SessionInfo: manager.getCurrentSession(),
-		})
+func (manager *connectionManager) consumeStats(statisticsChannel <-chan consumer.SessionStatistics, stop <-chan struct{}) {
```
vkuznecovas: could we perhaps return an "unsubscribe" method instead of passing in a stop channel?

Passing in another channel only adds to confusion

soffokl: Did it with a returned unsubscribe function.

PR #1490, opened 2020-01-08
generate agreement id randomly

vkuznecovas:


session/pingpong/invoice_tracker.go
```diff
@@ -19,9 +19,9 @@ package pingpong
 
 import (
 	"bytes"
-	"crypto/rand"
 	"encoding/hex"
 	"math"
+	"math/rand"
```
soffokl: Is it ok to use "math/rand" instead of "crypto/rand" here? Do we need to protect this ID somehow?

session/pingpong/invoice_tracker.go
```diff
@@ -184,10 +183,8 @@ func calculateMaxNotReceivedExchangeMessageCount(chargeLeeway, chargePeriod time
 }
 
 func (it *InvoiceTracker) generateInitialInvoice() error {
```
soffokl: There is no error could be returned now, maybe remove returned value?

session/pingpong/invoice_tracker.go
vkuznecovas: I think it's good enough, we don't really need to be super duper secure here.

vkuznecovas: Besides, it's in the providers interest to not to tamper with this, if he does - well, it's his own responsibility :)

PR #1491, opened 2020-01-08
Improve node disconnect time

tadaskay:
Restore connections to broker after connect/disconnect.

Fixes: #1484

### DIFF
--- a/communication/nats/connection_wrap.go
+++ b/communication/nats/connection_wrap.go
@@ -23,7 +23,6 @@ import (
 	"strings"
 	"time"
 
-	"github.com/mysteriumnetwork/node/firewall"
 	nats_lib "github.com/nats-io/go-nats"
 	"github.com/pkg/errors"
 	"github.com/rs/zerolog/log"
@@ -58,6 +57,7 @@ func ParseServerURI(serverURI string) (*url.URL, error) {
 func newConnection(serverURIs ...string) (*ConnectionWrap, error) {
 	connection := &ConnectionWrap{
 		servers: make([]string, len(serverURIs)),
+		onClose: func() {},
 	}
 
 	for i, server := range serverURIs {
@@ -71,27 +71,14 @@ func newConnection(serverURIs ...string) (*ConnectionWrap, error) {
 	return connection, nil
 }
 
-// OpenConnection creates connection instances and connects instantly
-func OpenConnection(serverURIs ...string) (*ConnectionWrap, error) {
-	connection, err := newConnection(serverURIs...)
-	if err != nil {
-		return connection, err
-	}
-
-	log.Debug().Msg("Connecting to NATS servers: " + strings.Join(serverURIs, ","))
-
-	return connection, connection.Open()
-}
-
 // ConnectionWrap defines wrapped connection to NATS server(s)
 type ConnectionWrap struct {
 	*nats_lib.Conn
-	servers     []string
-	removeRules func()
+	servers []string
+	onClose func()
 }
 
-// Open starts the connection
-func (c *ConnectionWrap) Open() (err error) {
+func (c *ConnectionWrap) connectOptions() nats_lib.Options {
 	options := nats_lib.GetDefaultOptions()
 	options.Servers = c.servers
 	options.MaxReconnect = BrokerMaxReconnect
@@ -101,13 +88,13 @@ func (c *ConnectionWrap) Open() (err error) {
 	options.ClosedCB = func(conn *nats_lib.Conn) { log.Warn().Msg("NATS: connection closed") }
 	options.DisconnectedCB = func(nc *nats_lib.Conn) { log.Warn().Msg("NATS: disconnected") }
 	options.ReconnectedCB = func(nc *nats_lib.Conn) { log.Warn().Msg("NATS: reconnected") }
+	return options
+}
 
-	c.removeRules, err = firewall.AllowURLAccess(c.servers...)
-	if err != nil {
-		return errors.Wrapf(err, `failed to allow NATS servers "%v" in firewall`, c.servers)
-	}
-
-	c.Conn, err = options.Connect()
+// Open starts the connection: left for test compatibility.
+// Deprecated: Use nats.BrokerConnector#Connect() instead.
+func (c *ConnectionWrap) Open() (err error) {
+	c.Conn, err = c.connectOptions().Connect()
 	if err != nil {
 		return errors.Wrapf(err, `failed to connect to NATS servers "%v"`, c.servers)
 	}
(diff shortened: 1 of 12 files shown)


communication/nats/connection_wrap.go
```diff
@@ -123,14 +114,17 @@ func (c *ConnectionWrap) Close() {
 		c.removeRules()
 	}
 	c.removeRules = nil
+	if c.onClose != nil {
```
vkuznecovas: is this ever nil though? do we need this check?

tadaskay: Initialized `onClose` with noop

communication/nats/connection_wrap.go
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2019 The "MysteriumNetwork/node" Authors.
+ * Copyright (C) 2020 The "MysteriumNetwork/node" Authors.
```
anjmao: There is no need to update old files headers.

communication/nats/connector.go
```diff
@@ -0,0 +1,91 @@
...
+type BrokerConnector struct {
+	registry map[uuid.UUID]Connection
+	mu       sync.Mutex
+}
+
+// NewBrokerConnector creates a new BrokerConnector.
+func NewBrokerConnector() *BrokerConnector {
+	return &BrokerConnector{
+		registry: make(map[uuid.UUID]Connection),
+		mu:       sync.Mutex{},
```
anjmao: No need to initialise mutex, it can be used by it's default value.

communication/nats/connector.go
```diff
@@ -0,0 +1,91 @@
...
+
+// Connect establishes a new connection to the broker(s).
+func (b *BrokerConnector) Connect(serverURIs ...string) (Connection, error) {
+	log.Debug().Msg("Connecting to NATS servers: " + strings.Join(serverURIs, ","))
+
+	connection, err := newConnection(serverURIs...)
+	if err != nil {
+		return connection, err
+	}
+	conn, err := connection, connection.Open()
```
anjmao: There is no need to assign the same variable again. Could be:

```go
conn, err := newConnection(serverURIs...)
if err != nil {
   return nil, err
}
if err := connection.Open(); err != nil {
   return nil, err
}
```

communication/nats/connector.go
```diff
@@ -0,0 +1,91 @@
...
+	"strings"
+	"sync"
+
+	"github.com/gofrs/uuid"
+	"github.com/pkg/errors"
+	"github.com/rs/zerolog/log"
+)
+
+// BrokerConnector establishes new connections to NATS servers and handles reconnects.
+type BrokerConnector struct {
```
anjmao: Maybe add unit test?

communication/nats/connection_wrap.go
anjmao: As I understand date is when file was created. For example https://github.com/golang/go/blob/master/src/runtime/alg.go#L1 alg.go created on 2014 and modified 29 days ago.

communication/nats/connection_wrap.go
```diff
@@ -49,16 +46,16 @@ func ParseServerURI(serverURI string) (*url.URL, error) {
...
-		serverURL.Host = fmt.Sprintf("%s:%d", serverURL.Host, BrokerPort)
+		serverURL.Host = fmt.Sprintf("%s:%d", serverURL.Host, DefaultBrokerPort)
 	}
 
 	return serverURL, nil
 }
 
-// NewConnection create new ConnectionWrap to given servers`
-func NewConnection(serverURIs ...string) (*ConnectionWrap, error) {
+func newConnection(serverURIs ...string) (*ConnectionWrap, error) {
```
Waldz: It there a point to make this constructor private?

communication/nats/connection_wrap.go
```diff
@@ -72,40 +69,35 @@ func NewConnection(serverURIs ...string) (*ConnectionWrap, error) {
...
-
-	return connection, connection.Open()
-}
-
 // ConnectionWrap defines wrapped connection to NATS server(s)
 type ConnectionWrap struct {
 	*nats_lib.Conn
 	servers     []string
 	removeRules func()
+	onClose     func()
```
Waldz: `onClose()` looks very similar to what `removeRules()` does, lets merge them

communication/nats/connection_wrap.go
```diff
@@ -72,40 +69,35 @@ func NewConnection(serverURIs ...string) (*ConnectionWrap, error) {
...
-	options.DisconnectedCB = func(nc *nats_lib.Conn) { log.Warn().Msg("Disconnected") }
-	options.ReconnectedCB = func(nc *nats_lib.Conn) { log.Warn().Msg("Reconnected") }
+	options.ClosedCB = func(conn *nats_lib.Conn) { log.Warn().Msg("NATS: connection closed") }
+	options.DisconnectedCB = func(nc *nats_lib.Conn) { log.Warn().Msg("NATS: disconnected") }
+	options.ReconnectedCB = func(nc *nats_lib.Conn) { log.Warn().Msg("NATS: reconnected") }
+	return options
+}
 
+// Open starts the connection
+func (c *ConnectionWrap) Open() (err error) {
```
Waldz: After connection factory `BrokerConnector` was introduced - is it a point to have `ConnectionWrap.Open/Close()` or `ConnectionWrap` at all? Because it had purposes - to encapsulate connection open and reconnection logic, which now is done by connector

communication/nats/connection_wrap.go
tadaskay: Yes, it's not used and not intended to be used outside of the package.

communication/nats/connection_wrap.go
tadaskay: Agreed. I left it deprecated for test compatibility. We should rewrite those unit tests using a normal broker instead of mock festival.

communication/nats/connector_test.go
```diff
@@ -0,0 +1,65 @@
...
+import (
+	"testing"
+	"time"
+
+	"github.com/nats-io/go-nats"
+	"github.com/nats-io/nats-server/server"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestBrokerConnector(t *testing.T) {
```
anjmao: 1. Would be better to have isolated unit test which doesn't start real nats server. Calling connector.Connect could assert that mock conn Open was called and is added to registry map. Also it could assert that after closing conn it's removed from registry.
2. ReconnectAll is not tested. Again if mock conn is used it's enough to test if Check method is called.

tadaskay: How is testing with mocks better than a proper unit test?

PR #1493, opened 2020-01-10
Increase proposal ping period

zolia:
Enabling wireguard on RPIs by default will effectively double amount of proposals discovery has to process. Even now we are operating almost on edge.

Reducing proposal ping to every 3 min.


vkuznecovas:
also - this might break dashboard a bit, since the availability there might be assumed to be reported every minute - please double check this

PR #1494, opened 2020-01-10
Do not crash on specifying discovery.type flag

zolia:
still flag does not work as I would expect it. It appends, not overrides flag value. Filling separate bug issue since it is cli package related.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -218,11 +218,12 @@ func (cfg *Config) GetString(key string) string {
 
 // GetStringSlice returns config value as []string.
 func (cfg *Config) GetStringSlice(key string) []string {
-	switch cfg.Get(key).(type) {
+	val := cfg.Get(key)
+	switch val.(type) {
 	case *cli.StringSlice:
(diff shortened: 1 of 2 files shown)


config/config.go
```diff
@@ -218,8 +218,12 @@ func (cfg *Config) GetString(key string) string {
 
 // GetStringSlice returns config value as []string.
 func (cfg *Config) GetStringSlice(key string) []string {
-	value := cfg.Get(key).(*cli.StringSlice)
-	return cast.ToStringSlice([]string(*value))
+	switch cfg.Get(key).(type) {
+	case *cli.StringSlice:
+		return cast.ToStringSlice([]string(*cfg.Get(key).(*cli.StringSlice)))
```
Waldz: Why to `Get()` value twice, while u have it 2 lines before

tadaskay:
db9ff43 seems unrelated to this PR

zolia:
resolved by #1506

PR #1513, opened 2020-01-14
Improve OS info collector for MMN

tadaskay:
- `source` tool is not available out of the box on alpine (crashes) -> rewrite with scanner
- Try alternate os-release location
- Remove usage of `go-ci/shell` (CI only)
- Add debug statements for calls to MMN with the payload

### DIFF
--- a/mmn/collector.go
+++ b/mmn/collector.go
@@ -18,12 +18,15 @@
 package mmn
 
 import (
+	"bufio"
 	"crypto/sha256"
 	"encoding/hex"
+	"fmt"
+	"os"
+	"os/exec"
 	"runtime"
 	"strings"
 
-	"github.com/mysteriumnetwork/go-ci/shell"
 	"github.com/pkg/errors"
 	"github.com/rs/zerolog/log"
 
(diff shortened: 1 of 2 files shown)


mmn/collector.go
```diff
@@ -110,41 +113,64 @@ func hashMACAddress(mac string) string {
...
+	output, err := exec.Command("lsb_release", "-d").Output()
+	if err == nil {
+		return strings.TrimSpace(strings.TrimPrefix(string(output), "Description:")), nil
+	}
 
-	version = strings.TrimSpace(strings.Replace(output, "Caption=", "", 1))
-	version = strings.TrimSpace(strings.Replace(version, "Microsoft", "", 1))
+	const etcOsRelease = "/etc/os-release"
+	const altOsRelease = "/usr/lib/os-release"
+	osReleaseFile, err := os.Open(etcOsRelease)
```
Waldz: Much simpler would be `source /etc/os-release && echo $PRETTY_NAME`.

soffokl: I'm not sure if it will work for all OS because for CentOS this file called `/etc/centos-release`

soffokl: Oh, looks like `/etc/os-release` exists there too.

anjmao: Description mentions that source tools is not available on some environments.

Waldz: I am about complex variable parsing here.
Regarding `centos-release`, CentOS version is not needed, here we're parsing Linux version.

mmn/collector.go
```diff
@@ -110,41 +113,64 @@ func hashMACAddress(mac string) string {
...
+	defer osReleaseFile.Close()
+
+	var prettyName string
+	scanner := bufio.NewScanner(osReleaseFile)
+	for scanner.Scan() {
+		line := scanner.Text()
+		if strings.HasPrefix(line, "PRETTY_NAME") {
+			tokens := strings.SplitN(line, "=", 2)
+			if len(tokens) == 2 {
+				prettyName = strings.Trim(tokens[1], "\"")
```
anjmao: At this point you can stop loop.

```go
prettyName = strings.Trim(tokens[1], "\"")
break
```

mmn/collector.go
tadaskay: Source is unavailable in alpine. It’s not really complex

mmn/collector.go
tadaskay: Yes, but it doesn’t matter as the file is tiny. So I’m intentionally reducing complexity here

anjmao: Why should someone care if file is tiny. Breaking the loop when you done is also much more clear and indicates that after name is one no more cycles are needed.

tadaskay: Because every control instruction, code branch, etc. increases complexity for the reader and since there are only a few cycles to be wasted, that is the rationale for not accounting for them (as already explained). Here, for instance, is the same logic implemented in docker core: https://github.com/moby/moby/blob/master/pkg/parsers/operatingsystem/operatingsystem_linux.go#L64

anjmao: I guess it's because value can be overwritten so in this case the last one will be taken.

PR #1517, opened 2020-01-14
Configurable proposal fetch interval

Waldz:


communication/nats/logger.go
```diff
@@ -18,16 +18,18 @@
 package nats
 
 import (
-	"github.com/nats-io/go-nats"
 	"github.com/rs/zerolog"
 )
 
 // Change this to slice as needed
```
soffokl: Is this comment still needed?

PR #1518, opened 2020-01-14
Possibility to subscribe for proposal changes

Waldz:
Possibility for mobile to update proposal selector wo explicit fetch

### DIFF
--- a/core/discovery/event.go
+++ b/core/discovery/event.go
@@ -19,10 +19,12 @@ package discovery
 
 // Topic represents the different topics a consumer can subscribe to
 const (
-	// ProposalAddedEventTopic represents newly announced proposal
-	ProposalAddedEventTopic = "ProposalAdded"
-	// ProposalUpdatedEventTopic represents re-announced proposal
-	ProposalUpdatedEventTopic = "ProposalUpdated"
-	// ProposalRemovedEventTopic represents newly de-announced proposal
-	ProposalRemovedEventTopic = "ProposalRemoved"
+	// EventTopicProposalAdded represents newly announced proposal
+	EventTopicProposalAdded = "ProposalAdded"
+	// EventTopicProposalUpdated represents re-announced proposal
+	EventTopicProposalUpdated = "ProposalUpdated"
+	// EventTopicProposalRemoved represents newly de-announced proposal
+	EventTopicProposalRemoved = "ProposalRemoved"
+	// EventTopicProposalAnnounce represent proposal events topic.
+	EventTopicProposalAnnounce = "proposalEvent"
 )
--- a/core/discovery/event.go
+++ b/core/discovery/event.go
@@ -1,0 +1,28 @@
+/*
+ * Copyright (C) 2020 The "MysteriumNetwork/node" Authors.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package discovery
+
+// Topic represents the different topics a consumer can subscribe to
+const (
+	// ProposalAddedEventTopic represents newly announced proposal
+	ProposalAddedEventTopic = "ProposalAdded"
+	// ProposalUpdatedEventTopic represents re-announced proposal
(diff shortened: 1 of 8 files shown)


core/discovery/event.go
```diff
@@ -0,0 +1,28 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package discovery
+
+// Topic represents the different topics a consumer can subscribe to
+const (
+	// ProposalAddedEventTopic represents newly announced proposal
+	ProposalAddedEventTopic = "ProposalAdded"
```
anjmao: Go's naming convention for multiple constants is to to prefix. See example https://github.com/golang/go/blob/c04e47f82159f1010d7403276b3dff5ab836fd00/src/net/http/status.go#L9

```go
const (
    EventTopicProposalAdded = "ProposalAdded"
    EventTopicProposalUpdated = "ProposalUpdated"
    EventTopicProposalRemovedE = "ProposalRemoved"
)
```

When it is more easy to find all topics when writing in IDE `discovery.EventTopic*`

core/discovery/storage.go
```diff
@@ -41,14 +41,17 @@ type ProposalFilter interface {
 }
 
 // NewStorage creates new instance of ProposalStorage
-func NewStorage() *ProposalStorage {
+func NewStorage(eventPublisher Publisher) *ProposalStorage {
```
anjmao: I saw Publisher interface in 10 places. There is already pacakge eventbus which contains this interface

```go
package eventbus

// Publisher is responsible for publishing given events.
type Publisher interface {
	Publish(topic string, data interface{})
}
```

I suggest to remove all other interface copies and use this one.

mobile/mysterium/entrypoint.go
```diff
@@ -212,6 +207,35 @@ func (mb *MobileNode) GetProposal(req *GetProposalRequest) ([]byte, error) {
...
 
+// ProposalChangeCallback represents proposal callback.
+type ProposalChangeCallback interface {
+	OnChange(proposal []byte)
+}
+
+// RegisterProposalAddedCallback registers callback which is called on newly announced proposals
+func (mb *MobileNode) RegisterProposalAddedCallback(cb ProposalChangeCallback) {
+	_ = mb.eventBus.SubscribeAsync(discovery.ProposalAddedEventTopic, func(proposal market.ServiceProposal) {
+		proposalPayload, _ := mb.proposalsManager.mapToProposalResponse(&proposal)
```
anjmao: Would be good to log possible error.

mobile/mysterium/entrypoint.go
```diff
@@ -212,6 +207,35 @@ func (mb *MobileNode) GetProposal(req *GetProposalRequest) ([]byte, error) {
...
+		cb.OnChange(proposalPayload)
```
anjmao: Sad news are that mapToProposalResponse is not adding quality level data. Before it was used only for single GetProposal which doesn't need quality data. But now these proposals need to have such data as they will update actual UI. See `func (m *proposalsManager) addQualityData(proposals []*proposal)` for usage how it was done. Probably quality metrics need to be cached on initial GetProposals etc. to be accessible for each event.

mobile/mysterium/entrypoint.go
```diff
@@ -212,6 +207,35 @@ func (mb *MobileNode) GetProposal(req *GetProposalRequest) ([]byte, error) {
...
+
+// RegisterProposalAddedCallback registers callback which is called on re-announced proposals
+func (mb *MobileNode) RegisterProposalUpdatedCallback(cb ProposalChangeCallback) {
+	_ = mb.eventBus.SubscribeAsync(discovery.ProposalUpdatedEventTopic, func(proposal market.ServiceProposal) {
+		proposalPayload, _ := mb.proposalsManager.mapToProposalResponse(&proposal)
+		cb.OnChange(proposalPayload)
+	})
+}
+
+// RegisterProposalAddedCallback registers callback which is called on re-announced proposals
```
zolia:
```suggestion
// RegisterProposalRemovedCallback registers callback which is called on unregistered proposals
```

core/discovery/event.go
Waldz: Named by existing event convention. Renamed

core/discovery/storage.go
Waldz: Improved, just not sure what it improves. Usually Go encourages to have interfaces on caller side.

core/discovery/storage.go
```diff
@@ -85,6 +88,19 @@ func (s *ProposalStorage) Set(proposals []market.ServiceProposal) {
 	s.mutex.Lock()
 	defer s.mutex.Unlock()
 
+	proposalsOld := s.proposals
```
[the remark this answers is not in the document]
Waldz: Simplified parallelism

mobile/mysterium/entrypoint.go
Waldz: Looks like need to addressed in another PR

anjmao: Could be possible addressed in https://github.com/mysteriumnetwork/node/issues/1524

PR #1522, opened 2020-01-15
WireGuard NAT hole punching

anjmao:
Implement NAT hole punching the same way as it is done for OpenVPN service by reusing NAT pinger and UDP Proxy. One difference is that for mobile consumser side separate UDP proxy is now not needed as because WireGuard does not change source port for outgoing packets.

TODO:
- [x] Add more tests including compatibility with old node versions.
- [x] Do more manual testing from desktop and mobile.
- [x] Fix e2e tests. Check way it's failing. Can be due to stability issues.


services/wireguard/handshake.go
```diff
@@ -0,0 +1,52 @@
...
+
+import (
+	"context"
+	"errors"
+	"net"
+	"time"
+)
+
+// WaitHandshake waits until WireGuard does initial handshake.
+func WaitHandshake(statsFetch func() (*Stats, error), timeout time.Duration, stop chan struct{}) error {
```
vkuznecovas: return a cancel func instead of passing in a stop.

anjmao: I can't return cancel func since there is a for loop which checks for latests handshake and returns err if any. It make sense to return cancel func for similar cases like context.WithTimeout. In this case stop channel can be use to get similar cleanup logic as with defer cancel.

```go
done := make(chan struct{})
defer close(done)
doSomething(done)
```

bin/docker/dev/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM golang:1.13-alpine
+
+# Install packages
+RUN apk add --update --no-cache iptables ca-certificates openvpn wireguard-tools bash sudo openresolv gcc musl-dev make linux-headers vim curl tcpdump && rm -rf /var/cache/apk/*
```
soffokl: No need to clean manually with `--no-cache`
```suggestion
RUN apk add --update --no-cache iptables ca-certificates openvpn wireguard-tools bash sudo openresolv gcc musl-dev make linux-headers vim curl tcpdump
```

bin/docker/dev/README.md
```diff
@@ -0,0 +1,41 @@
...
+docker exec -it $(docker ps -aqf "name=node_e2e_nat_test_myst-consumer_1") /bin/bash
+```
+
+5. Use helper scripts to build source code and run provider or consumer.
+```
+# Build and run provider
+./bin/build && ./bin/docker/dev/provider.sh
+
+# Build and run consumser daemon
+./bin/build && ./bin/docker/dev/consumser.sh
```
soffokl:
```suggestion
./bin/build && ./bin/docker/dev/consumer.sh
```

bin/docker/dev/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+# Run consumser CLI
```
soffokl:
```suggestion
# Run consumer CLI
```

bin/docker/dev/README.md
```diff
@@ -0,0 +1,41 @@
...
+```
+docker exec -it $(docker ps -aqf "name=node_e2e_nat_test_myst-consumer_1") /bin/bash
+```
+
+5. Use helper scripts to build source code and run provider or consumer.
+```
+# Build and run provider
+./bin/build && ./bin/docker/dev/provider.sh
+
+# Build and run consumser daemon
```
soffokl:
```suggestion
# Build and run consumer daemon
```

bin/docker/dev/README.md
```diff
@@ -0,0 +1,41 @@
...
+2. Run docker-compose stack
+
+```
+go run mage.go -v NATDevUp
+```
+
+3. Connect to provider container
+
+```
+docker exec -it $(docker ps -aqf "name=node_e2e_nat_test_myst-provider_1") /bin/bash
```
soffokl: Can this be just a container name?
```suggestion
docker exec -it node_e2e_nat_test_myst-provider_1 bash
```

bin/docker/dev/README.md
```diff
@@ -0,0 +1,41 @@
...
+```
+
+4. Connect to consumer container
+```
+docker exec -it $(docker ps -aqf "name=node_e2e_nat_test_myst-consumer_1") /bin/bash
```
soffokl:
```suggestion
docker exec -it node_e2e_nat_test_myst-consumer_1 bash
```

ci/test/e2e.go
```diff
@@ -57,8 +57,7 @@ func TestE2ENAT() error {
 	composeFiles := []string{
 		"e2e/traversal/docker-compose.yml",
 	}
-	runner := newRunner(composeFiles, "node_e2e_nat_test", "openvpn")
-	defer runner.cleanup()
```
soffokl: Why we should not cleanup the runner?

mobile/mysterium/openvpn3/openvpn3_mock.go
```diff
@@ -0,0 +1,132 @@
...
+type Session struct {
+}
+
+// Start mock
+func (s Session) Start() {
+}
+
+// Wait mock
+func (s Session) Wait() error {
+	return errors.New("using mock openvp3")
```
soffokl:
```suggestion
	return errors.New("using mock openvpn3")
```

mobile/mysterium/openvpn_connection_setup.go
```diff
@@ -94,20 +94,20 @@ func (wrapper *openvpnConnection) Wait() error {
...
-
 	switch wrapper.natPinger.(type) {
 	case *traversal.NoopPinger:
 		log.Info().Msg("Noop pinger detected, returning nil client config.")
 		return nil, nil
 	}
 
+	publicIP, err := wrapper.ipResolver.GetPublicIP()
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to get consumer config")
```
soffokl:
```suggestion
		return nil, errors.Wrap(err, "failed to get public IP")
```

mobile/mysterium/wireguard_connection_setup.go
```diff
@@ -162,91 +131,161 @@ func newTunnDevice(wgTunnSetup WireguardTunnelSetup, config *wireguard.ServiceCo
...
+
 	return wireguard.ConsumerConfig{
 		PublicKey: publicKey,
+		IP:        publicIP,
 	}, nil
 }
 
-var _ connection.Connection = &wireguardConnection{}
+func (c *wireguardConnection) isNoopPinger() bool {
+	switch c.natPinger.(type) {
```
soffokl: This can be done simpler
```golang
_, ok := c.natPinger.(*traversal.NoopPinger)
return ok
```

nat/traversal/pinger.go
```diff
@@ -42,6 +40,18 @@ var (
 	errNATPunchAttemptTimedOut = errors.New("NAT punch attempt timed out")
 )
 
+// NatPinger is responsible for pinging nat holes
+type NatPinger interface {
```
soffokl:
```suggestion
type NATPinger interface {
```

nat/traversal/pinger.go
```diff
@@ -50,8 +60,7 @@ type Pinger struct {
 	stopNATProxy   chan struct{}
 	once           sync.Once
 	natEventWaiter NatEventWaiter
```
soffokl: Can we call it with a constant case everywhere?
```suggestion
	natEventWaiter NATEventWaiter
```

services/wireguard/connection/connection.go
```diff
@@ -127,38 +137,99 @@ func (c *Connection) Start(options connection.ConnectOptions) (err error) {
...
+		}
 	}
 	return wg.ConsumerConfig{
 		PublicKey: publicKey,
+		IP:        publicIP,
 	}, nil
 }
 
+func (c *Connection) isNoopPinger() bool {
+	switch c.natPinger.(type) {
```
soffokl: The same here:
```suggestion
	_, ok := c.natPinger.(*traversal.NoopPinger)
	return ok
```

services/wireguard/connection/connection.go
```diff
@@ -127,38 +137,99 @@ func (c *Connection) Start(options connection.ConnectOptions) (err error) {
...
+		return nil, errors.Wrap(err, "failed to start connection endpoint")
+	}
+
+	return conn, nil
+}
+
+func (c *Connection) addProviderPeer(conn wg.ConnectionEndpoint, endpoint net.UDPAddr, publicKey string, natPunchingEnabled bool) error {
+	var keepAlivePeriodSeconds int
+	if natPunchingEnabled {
+		keepAlivePeriodSeconds = 25
```
soffokl: Why can't we have the same keep alive period for both with NAT punching and without it?

services/wireguard/endpoint/kernelspace/client.go
```diff
@@ -65,18 +62,36 @@ func (c *client) ConfigureDevice(iface string, config wg.DeviceConfig, ipAddr ne
...
 	publicKey, err := stringToKey(peer.PublicKey)
 	if err != nil {
-		return err
+		return errors.Wrap(err, "could not convert string key to wgtypes.Key")
+	}
+
+	// Apply keep alive interval
+	var keepAliveInterval *time.Duration
+	if peer.KeepAlivePeriodSeconds > 0 {
+		interval := time.Duration(peer.KeepAlivePeriodSeconds) * time.Second
```
soffokl: We do we storing `peer.KeepAlivePeriodSeconds` as int and converting to the `time.Duration` here? It can be `time.Duration` from the very beginning.

services/wireguard/handshake.go
```diff
@@ -0,0 +1,52 @@
...
+		select {
+		case <-time.After(100 * time.Millisecond):
+			stats, err := statsFetch()
+			if err != nil {
+				return err
+			}
+			if !stats.LastHandshake.IsZero() {
+				return nil
+			}
+		case <-ctx.Done():
```
soffokl: You don't need context here:
```suggestion
		case <-time.After(timeout):
```

mobile/mysterium/wireguard_connection_setup.go
```diff
@@ -140,14 +105,18 @@ func setupWireguardDevice(devApi *device.Device, config *wireguard.ServiceConfig
...
 }
 
-func newTunnDevice(wgTunnSetup WireguardTunnelSetup, config *wireguard.ServiceConfig) (tun.Device, error) {
+func newTunnDevice(wgTunnSetup WireguardTunnelSetup, config wireguard.ServiceConfig) (tun.Device, error) {
 	consumerIP := config.Consumer.IPAddress
 	prefixLen, _ := consumerIP.Mask.Size()
+	wgTunnSetup.NewTunnel()
+	wgTunnSetup.SetSessionName("wg-tun-session")
+	//TODO fetch from user connection options
+	wgTunnSetup.AddDNS("8.8.8.8")
```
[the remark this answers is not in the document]
tadaskay: This should be autonegotiated, similar to openvpn's tunnel setup using `connection.DNSOptionAuto` (provider for some reason might not have DNS proxy started):

```go
	dnsIPs, err := dnsOption.ResolveIPs(vpnConfig.DNSIPs)
	if err != nil {
		return nil, err
	}
	for _, ip := range dnsIPs {
		clientFileConfig.SetParam("dhcp-option", "DNS", ip)
	}
```

mobile/mysterium/wireguard_connection_setup.go
```diff
@@ -36,8 +36,10 @@ import (
 )
 
 const (
-	//taken from android-wireguard project
+	// Taken from android-wireguard project
 	androidTunMtu = 1280
```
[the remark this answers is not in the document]
anjmao: On wireguard-go default MTU is 1420, but on wireguard-android it is 1280. Need to check why it is lower, do you have any ideas? Just want to be sure it will not impact performance.

mobile/mysterium/wireguard_connection_setup.go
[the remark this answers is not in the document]
anjmao: Created separate issue for this https://github.com/mysteriumnetwork/node/issues/1541 will add it in separate PR as it will need more testing.

mobile/mysterium/wireguard_connection_setup.go
```diff
@@ -157,143 +133,208 @@ func newTunnDevice(wgTunnSetup WireguardTunnelSetup, config *wireguard.ServiceCo
...
+		name, nameErr := tunDevice.Name()
 		log.Info().Err(nameErr).Msg("Name value: " + name)
 	}
 
-	return tun, err
+	return tunDevice, err
+}
+
+func waitUDPPortReadyFor(port int, timeout time.Duration) error {
+	timeoutChan := time.After(timeout)
```
soffokl: This can be smaller, not sure if it's better, but smaller. 😂
```golang
done := time.Now().Add(timeout)
for t := time.NewTicker(500 * time.Millisecond); time.Now().Before(done); <-t.C {
	p, err := net.ListenPacket("udp", fmt.Sprintf(":%d", port))
	if err == nil {
		p.Close()
		return nil
	} 
	log.Err(err).Msgf("Port %d is in use. Trying to check again...", port)
}
return fmt.Errorf("timeout waiting for UDP port %d", port)
```

mobile/mysterium/wireguard_connection_setup.go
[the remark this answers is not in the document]
zolia: Most likely it due to IPv6 compatibility. Minimum IPv6 packet MTU should be 1280. Why it is that low? Not sure, most likely due to bet possible traversal without actual fragmentation.

PR #1526, opened 2020-01-16
Cli settle

### LINKED ISSUE #1519: CLI call to force settlement on provider side
At the moment settlements are automatic, we need to add a tequila endpoint and cli call to force the settlement.

vkuznecovas:
Adds for the ability to force a settlement without an ongoing session.

Closes #1519

### DIFF
--- a/session/pingpong/accountant_promise_settler.go
+++ b/session/pingpong/accountant_promise_settler.go
@@ -40,6 +40,10 @@ type transactor interface {
 	SettleAndRebalance(id string, promise crypto.Promise) error
 }
 
+type promiseStorage interface {
+	Get(providerID, accountantID identity.Identity) (AccountantPromise, error)
+}
+
 type receivedPromise struct {
 	provider identity.Identity
 	promise  crypto.Promise
@@ -54,6 +58,7 @@ type AccountantPromiseSettler struct {
 	registrationStatusProvider registrationStatusProvider
 	ks                         ks
 	transactor                 transactor
+	promiseStorage             promiseStorage
 
 	currentState map[identity.Identity]state
 	settleQueue  chan receivedPromise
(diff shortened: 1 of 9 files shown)


session/pingpong/accountant_promise_settler.go
```diff
@@ -267,13 +319,23 @@ func (aps *AccountantPromiseSettler) settle(p receivedPromise) {
 		defer aps.setSettling(p.provider, false)
 		select {
 		case <-aps.stop:
+			// send a signal to waiter that the settlement was cancelled
+			if waiter != nil {
```
Waldz: Strange conditions here, can waiter be nilable? Maybe different function or nop waiter

session/pingpong/accountant_promise_settler.go
```diff
@@ -248,12 +255,57 @@ func (aps *AccountantPromiseSettler) listenForSettlementRequests() {
...
 	}
 }
 
-func (aps *AccountantPromiseSettler) settle(p receivedPromise) {
+// ErrNothingToSettle indicates that there is nothing to settle.
+var ErrNothingToSettle = errors.New("nothing to settle for the given provider")
+
+// ForceSettleAsync forces the settling of promises for the given accountant and provider combo without waiting for the result.
+func (aps *AccountantPromiseSettler) ForceSettleAsync(providerID, accountantID identity.Identity) error {
+	return aps.forceSettle(providerID, accountantID, nil)
```
anjmao: This is not very Go way. If I want to make it async I would like to decide it from calling side.

```go
go ForceSettle(...)
```

It will also simplify internal settle logic.

tequilapi/endpoints/transactor.go
```diff
@@ -87,7 +95,82 @@ func (te *transactorEndpoint) TransactorFees(resp http.ResponseWriter, _ *http.R
...
+// responses:
+//   202:
+//     description: settle request accepted
+//   500:
+//     description: Internal server error
+//     schema:
+//       "$ref": "#/definitions/ErrorMessageDTO"
+func (te *transactorEndpoint) SettleAsync(resp http.ResponseWriter, request *http.Request, _ httprouter.Params) {
+	err := te.settle(request, func(provider, accountant identity.Identity) error {
+		go func() { te.promiseSettler.ForceSettle(provider, accountant) }()
```
anjmao: ForceSettle returns error which is not handled. Maybe log it in such case?

session/pingpong/accountant_promise_settler.go
```diff
@@ -285,6 +324,9 @@ func (aps *AccountantPromiseSettler) settle(p receivedPromise) {
 			return
 		case <-time.After(aps.config.MaxWaitForSettlement):
 			log.Info().Msgf("Settle timeout for %v", p.provider)
+
+			// send a signal to waiter that the settlement has timedout
```
anjmao: timedout -> timeout

session/pingpong/accountant_promise_settler.go
```diff
@@ -248,23 +255,54 @@ func (aps *AccountantPromiseSettler) listenForSettlementRequests() {
...
 	log.Info().Msgf("Marked provider %v as requesting setlement", p.provider)
 	sink, cancel, err := aps.bc.SubscribeToPromiseSettledEvent(p.provider.ToCommonAddress(), aps.config.AccountantAddress)
 	if err != nil {
 		aps.setSettling(p.provider, false)
 		log.Error().Err(err).Msg("Could not subscribe to promise settlement")
-		return
+		return err
 	}
+
+	ch := make(chan error)
```
anjmao: Could be a bit more clear name for ch. Maybe name it errc which means error channel.

PR #1530, opened 2020-01-17
put more trust to accountant when calculating balance

vkuznecovas:
We'll now check balance with the accountant and trust it just a little bit more.

This will hopefully simplify the balance calculations just by a little.

### DIFF
--- a/session/pingpong/consumer_balance_tracker.go
+++ b/session/pingpong/consumer_balance_tracker.go
@@ -18,7 +18,6 @@
 package pingpong
 
 import (
-	"math/big"
 	"sync"
 
 	"github.com/ethereum/go-ethereum/common"
@@ -62,7 +61,6 @@ func NewConsumerBalanceTracker(publisher eventbus.Publisher, mystSCAddress commo
 }
 
 type consumerBalanceChecker interface {
-	GetConsumerBalance(channel, mystSCAddress common.Address) (*big.Int, error)
 	SubscribeToConsumerBalanceEvent(channel, mystSCAddress common.Address) (chan *bindings.MystTokenTransfer, func(), error)
 }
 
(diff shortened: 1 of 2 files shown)


session/pingpong/consumer_balance_tracker.go
```diff
@@ -211,43 +190,48 @@ func (cbt *ConsumerBalanceTracker) decreaseBalance(id identity.Identity, b uint6
...
 		}
-		go cbt.publishChangeEvent(id, 0, bcBalance-accountantPromised)
 	}
 }
 
-func (cbt *ConsumerBalanceTracker) getBCBalance(id identity.Identity) (uint64, error) {
-	addr, err := cbt.channelAddressCalculator.GetChannelAddress(id)
-	if err != nil {
-		return 0, err
+func safeSub(a, b uint64) uint64 {
```
anjmao: Is is safe to ignore possible underflows with money? What if there is a bug and balance was decreased by more than it supposed to?

session/pingpong/consumer_balance_tracker.go
```diff
@@ -211,43 +190,48 @@ func (cbt *ConsumerBalanceTracker) decreaseBalance(id identity.Identity, b uint6
 	}
 }
 
-func (cbt *ConsumerBalanceTracker) updateBCBalance(id identity.Identity, bcBalance, accountantPromised uint64) {
+func (cbt *ConsumerBalanceTracker) updateBalanceFromAccountant(id identity.Identity) {
+	cd, err := cbt.accountantBalanceFetcher(id.Address)
```
soffokl: Is this `cd` for consumer balance? Then it should be `cb`.

session/pingpong/consumer_balance_tracker.go
vkuznecovas: The truth is always in the blockchain and there no underflows or overflows will happen. This is just a counter that doesn't do much really, just show balance estimates.

PR #1537, opened 2020-01-18
Cleanup e2e infrastructre

anjmao:
1. Remove old unused scripts and files related to e2e.
2. Move compose files to root for easer paths mapping.
3. Create new localnet environment for local development.


docker-compose.e2e-basic.yml
```diff
@@ -0,0 +1,201 @@
+version: '3.0'
+services:
+  morqa:
+    image: mysteriumnetwork/quality-oracle:latest
```
Waldz: Not sure if about this

anjmao: When we use latest images of our services e2e tests will be able to catch possible issues faster regarding backwards compatibility.

Issue #1497, opened 2020-01-12
Install script broken on Raspberry


anjmao:
Running [install script](https://github.com/mysteriumnetwork/node/blob/master/install.sh#L147) on Raspberry node fails with error:

```
Reading state information... Done
E: Unable to locate package linux-headers-4.19.75-v7
E: Couldn't find any package by glob 'linux-headers-4.19.75-v7'
E: Couldn't find any package by regex 'linux-headers-4.19.75-v7'
```

Before trying to install linux-header we could check if they are not already installed with
```
ls -l /usr/src/linux-headers-$(uname -r)
```

PR #1544, opened 2020-01-20
Use raspberrypi-kernel-headers instead of linux-headers on raspbian

tadaskay:
Fixes: #1497

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -6,6 +6,8 @@ set -o errtrace
 set -o errexit
 set -o pipefail
 
+export DEBIAN_FRONTEND="noninteractive"
+
 # Install latest release of myst for debian/ubuntu/raspbian
 #
 # Variables:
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -142,7 +143,12 @@ install_dependencies() {
     apt-key adv --keyserver keyserver.ubuntu.com --recv-keys AE33835F504A1A25
     apt-get update
     apt-get install -y wireguard
-    apt-get install -y libmnl-dev libelf-dev linux-headers-$(uname -r) build-essential pkg-config
+    apt-get install -y libmnl-dev libelf-dev build-essential pkg-config
+    if [[ "$distro" == "raspbian" ]]; then
+      apt-get install -y raspberrypi-kernel-headers
```
soffokl: There are 4 spaces in other places, let's keep it consistent. 
```suggestion
        apt-get install -y raspberrypi-kernel-headers
```

install.sh
```diff
@@ -142,7 +143,12 @@ install_dependencies() {
...
+    else
+      apt-get install -y linux-headers-$(uname -r)
```
soffokl:
```suggestion
        apt-get install -y linux-headers-$(uname -r)
```

Issue #1520, opened 2020-01-15
Pass in payment information from actual proposal


vkuznecovas:
Currently, the payment info is hardcoded on both the consumer and provider side. Need to take in the actual proposal info on both ends.

PR #1546, opened 2020-01-21
1mystt = 1$

vkuznecovas:
One test myst is now equal to one dollar to ease computations in users heads.

updates #1547

### DIFF
--- a/services/openvpn/discovery/dto/payment_method_per_time.go
+++ b/services/openvpn/discovery/dto/payment_method_per_time.go
@@ -36,7 +36,7 @@ type PaymentRate struct {
 
 // DefaultPaymentInfo represents the default payment rate
 var DefaultPaymentInfo = PaymentRate{
-	Price:    money.NewMoney(1000000, money.CurrencyMyst),
+	Price:    money.NewMoney(50000, money.CurrencyMyst),
 	Duration: time.Minute,
 }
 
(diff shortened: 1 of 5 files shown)


soffokl:
This will break compatibility with old versions, right?

vkuznecovas:
> This will break compatibility with old versions, right?

old consumers will work with new providers.

old providers will not work with new consumers.

vkuznecovas:
In theory, implementing #1520 and releasing it before releasing this would at least lower the impact this will have.

vkuznecovas:
However, implementing #1520 would probably create a misbalance with the point calculation between providers. Providers running old versions would get paid a ton more than the current providers, meaning that payments based on points would have huge discrepancies between them.

zolia:
Worst for this change is that even if we solve all compatibility issues, still the same identity might come across different versions of nodes and we would not be able to calculate the real earned amount.

In this issue we should stress that MYSTT == 1$ is just an assumption. Real payout model might differ from this model.

My suggestion is than when we will start our payouts based on MYSTT tokens earned we should do it only for nodes with appropriate node version ( >= 0.20.0 ?). Eligibility rule should include node version. @ignasbernotas ?

That also means we should release such version at least prior 3 days before beginning of next month.

vkuznecovas:
will reopen once #1520 moves ahead

PR #1549, opened 2020-01-21
solve race between node and accountant with a retry

### LINKED ISSUE #1548: node/accountant race
After registration, the lookup of balance on accountant fails due to a race where accountant has not yet updated the balance, but the consumer is already asking for it.

2020-01-21T10:35:28.745 ERR session/pingpong/consumer_balance_tracker.go:196 > could not get accountant balance error="could not request consumer data accountant: server response invalid: 400 Bad Request (https://testnet-accountant.mysterium.network/api/v1/data/consumer/0xe3f4baaae014dbaa49e390118ff7b77ae9322526)"

vkuznecovas:
added a retry for accountant balance calls after registration. This will now solve the race of who gets notified first of the registration.

fixes #1548

### DIFF
--- a/session/pingpong/consumer_balance_tracker.go
+++ b/session/pingpong/consumer_balance_tracker.go
@@ -18,8 +18,11 @@
 package pingpong
 
 import (
+	"context"
 	"sync"
+	"time"
 
+	"github.com/cenkalti/backoff/v4"
 	"github.com/ethereum/go-ethereum/common"
 	nodevent "github.com/mysteriumnetwork/node/core/node/event"
 	"github.com/mysteriumnetwork/node/eventbus"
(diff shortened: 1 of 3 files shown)


session/pingpong/consumer_balance_tracker.go
```diff
@@ -144,7 +147,32 @@ func (cbt *ConsumerBalanceTracker) handleTopUpEvent(id string) {
...
+		eback := backoff.NewExponentialBackOff()
+		eback.MaxElapsedTime = time.Minute
+		eback.InitialInterval = time.Second * 2
+
+		boff = backoff.WithMaxRetries(eback, 10)
+		ctx, cancel := context.WithCancel(context.Background())
+		defer cancel()
+
+		go func() {
+			defer cancel()
```
anjmao: Why is this second defer cancel needed and in separate goroutine? Maybe add some comment as it is a bit hard to understand.

session/pingpong/consumer_balance_tracker.go
```diff
@@ -144,7 +147,32 @@ func (cbt *ConsumerBalanceTracker) handleTopUpEvent(id string) {
...
+			select {
+			case <-cbt.stop:
+			case <-ctx.Done():
+			}
+		}()
+
+		boff = backoff.WithContext(boff, ctx)
+		toRetry := func() error {
+			return cbt.updateBalanceFromAccountant(event.ID)
```
anjmao: What if registration success event is send to node only after balance is ready?

vkuznecovas: In that case, we'd not need the retry here.

PR #1556, opened 2020-01-21
Add compatibility e2e tests

anjmao:
1. Add compatibility tests to test with agains older released version.
2. Use local location and quality oracle in e2e tests.


docker-compose.e2e-compatibility.yml
```diff
@@ -0,0 +1,278 @@
...
+      -operator 0x354bd098b4ef8c9e70b7f21be2d455df559705d7
+      -chImplementation 0x599d43715DF3070f83355D9D90AE62c159E62A75
+      -keystoreDir /keystore
+      -accountantAddress 0x241F6e1d0bB17f45767DC60A6Bd3D21Cdb543a0c
+      -boltPath /bolt.db
+    volumes:
+      - ./e2e/blockchain/keystore:/keystore
+
+  myst-provider:
+    image: mysteriumnetwork/myst:0.19.1-alpine
```
soffokl: Are we testing compatibility with `0.19.1` only? 
What is the flow to change the version that should be tested for compatibility?

anjmao: Actually there is no need to test with different version if all new versions are compatible with this fixed one we are fine. Aren't we?

soffokl: Ok, I see the plan. Let's see how it will work.

PR #1561, opened 2020-01-22
Synchronise access policies with TrustOracle

Waldz:
Validates policy name by fetching it's rules from TrustOracle before starting service and subscribes for improvements of rules


core/policy/policy_repository.go
```diff
@@ -0,0 +1,168 @@
...
+	for i, policyID := range policyIDs {
+		policies[i] = pr.Policy(policyID)
+	}
+	return policies
+}
+
+// AddPolicies adds given policy to repository. Also syncs policy rules from TrustOracle
+func (pr *PolicyRepository) AddPolicies(policies []market.AccessPolicy) error {
+	pr.policyLock.Lock()
+	policyRulesNew := pr.policyRules
```
soffokl: `policyRules` is a map. Assigning it to another variable not copy data, just assign the reference. It does not protect the map from a race condition.

core/policy/policy_repository.go
```diff
@@ -0,0 +1,168 @@
...
+		client:        client,
+		policyURL:     policyURL,
+		policyRules:   make(map[market.AccessPolicy]market.AccessPolicyRuleSet),
+		fetchInterval: interval,
+	}
+}
+
+// Start begins fetching proposals to repository
+func (pr *PolicyRepository) Start() {
+	pr.fetchShutdown = make(chan bool, 1)
```
anjmao: 1. Why do you need buffered channel here?
2. Stop is usually done simpler with empty struct channel like `chan struct`, you will find it in 100 places in node repo. When just send `fetchShutdown <- struct{}` to stop. If you use bool it is confusing. For example can I send only true, or false is also acceptable.

cmd/di_desktop.go
```diff
@@ -232,33 +235,25 @@ func (di *Dependencies) bootstrapServiceComponents(nodeOptions node.Options) err
...
 	storage := session.NewEventBasedStorage(di.EventBus, session.NewStorageMemory())
+	if err := storage.Subscribe(); err != nil {
+		return errors.Wrap(err, "could not subscribe session to node events")
+	}
 	di.ServiceSessionStorage = storage
 
-	err := storage.Subscribe()
-	if err != nil {
-		return errors.Wrap(err, "could not bootstrap service components")
+	policyRepo := policy.NewPolicyRepository(di.HTTPClient, nodeOptions.AccessPolicyEndpointAddress, 10*time.Minute)
```
anjmao: Can we make this interval configurable? Could be used for e2e tests etc.

core/policy/policy_repository.go
Waldz: 1. Why’d  not? This channel wait just for single stop message + don't want to block sending received isn't ready
2. Don't see a difference, sending as minimal structure as possible. Changed to s`truct{}`

core/policy/policy_repository.go
Waldz: What do do then? I don't want to lock reads of this variable for time spent on fetching data to this variables

tadaskay:
Do we really need to poll and cache access policies in node? This could be simply an on-demand call, could it not? @Waldz

core/policy/policy_repository.go
anjmao: Could initialise fetchShutdown inside NewPolicyRepository to be consistent.

cmd/di_desktop.go
```diff
@@ -232,33 +235,25 @@ func (di *Dependencies) bootstrapServiceComponents(nodeOptions node.Options) err
...
+	policyRepo.Start()
+	if err := di.EventBus.SubscribeAsync(string(event.StatusStopped), policyRepo.Stop); err != nil {
```
anjmao: Why is is stopped differently from other services? There is method Shutdown which stops stuff.

core/policy/policy_repository.go
anjmao: 1. This stop is only called if node is shutting down. In such case there is no point to Stop in async fashion.
2. As minimal as possible is empty struct.

core/policy/policy_repository.go
```diff
@@ -0,0 +1,168 @@
...
+
+func (pr *PolicyRepository) fetchPolicyRules(policy market.AccessPolicy) (market.AccessPolicyRuleSet, error) {
+	var policyRules market.AccessPolicyRuleSet
+
+	req, err := requests.NewGetRequest(policy.Source, "", nil)
+	if err != nil {
+		return policyRules, errors.Wrap(err, "failed to create policy request")
+	}
+
+	err = pr.client.DoRequestAndParseResponse(req, &policy)
```
anjmao: Shouldn't it be `pr.client.DoRequestAndParseResponse(req, &policyRules)` ? I mean response is parsed into policy not policyRules

core/policy/policy_repository.go
anjmao: 1. Lock pr.policyRules to read  source ids into string array etc. since fetchPolicyRules could accept source only.
2. Now iterate all sources and fetch into new map which don't need any lock.
3. Now lock again and remap from new map into your destination.

cmd/di_desktop.go
Waldz: Introduced

cmd/di_desktop.go
Waldz: Looks more neat approach to configure feature single place, not sharing variable `policyRepo` to DI if no other module is using it

core/policy/policy_repository.go
Waldz: Good point, thanks

cmd/di_desktop.go
anjmao: But when this is not waiting for stop. We already had problems what even if node is killed by sending OS signals (ctrl + c) some components was not stopped correctly as they was doing async stop.

core/policy/policy_repository.go
anjmao: 1. fetchPolicyRules could accept source instead of whole policy.
2. Now you can lock and read into source array and unlock.
3. Now iterate all sources and fetch into new map which don't need to be locked.
4. Now lock again and merge new map values into policyRules. Unlock. Done.

Waldz: Anyways, I've just remade to array instead of map

cmd/di_desktop.go
vkuznecovas: I'd listen to node shutdown for the reasons @anjmao lists.

core/policy/policy_repository.go
```diff
@@ -0,0 +1,196 @@
...
+
+// Start begins fetching proposals to repository
+func (pr *PolicyRepository) Start() {
+	pr.fetchShutdown = make(chan struct{})
+	go pr.fetchLoop()
+}
+
+// Stop ends fetching proposals to repository
+func (pr *PolicyRepository) Stop() {
+	pr.fetchShutdown <- struct{}{}
```
vkuznecovas: close it, since sending a single struct will fail to signal multiple listeners(if you ever have them)

core/policy/policy_repository.go
```diff
@@ -0,0 +1,196 @@
...
+		return policyRules, errors.Wrapf(err, "failed fetch policy rule %s", policy)
+	}
+	return policyRules, nil
+}
+
+func (pr *PolicyRepository) fetchLoop() {
+	for {
+		select {
+		case <-pr.fetchShutdown:
+			break
```
vkuznecovas: you're breaking the select here, you should return

Waldz:
All improvements addressed

core/policy/policy_repository.go
```diff
@@ -0,0 +1,212 @@
...
+		case <-pr.fetchShutdown:
+			return
+		case <-time.After(pr.fetchInterval):
+			pr.policyLock.Lock()
+			policyListActive := pr.policyList
+			pr.policyLock.Unlock()
+
+			for index := range policyListActive {
+				var err error
+				pr.fetchPolicyRules(&policyListActive[index])
```
vkuznecovas: you're not assigning this to the error

core/policy/policy_repository.go
tadaskay: Your latest changes didn't address @soffokl 's comment. You are copying reference, but releasing the lock, allowing race conditions with other `AddPolicies` calls.

You don't need to lock reads at all, only in `AddPolicies`:
```go
pr.policyLock.Lock()
defer pr.policyLock.Unlock()
policyListNew := deep copy pr.policyList
update policyListNew
pr.policyList = policyListNew
return nil
```
This will now allow concurrent `AddPolicies`, but will allow concurrent reads at any point: before new reference is updated on `pr`, all reads will return current access policies.

core/policy/policy_repository.go
```diff
@@ -0,0 +1,212 @@
...
+	for i, policyID := range policyIDs {
+		policies[i] = pr.Policy(policyID)
+	}
+	return policies
+}
+
+// AddPolicies adds given policy to repository. Also syncs policy rules from TrustOracle
+func (pr *PolicyRepository) AddPolicies(policies []market.AccessPolicy) error {
+	pr.policyLock.Lock()
+	policyListNew := pr.policyList
```
anjmao: Add unit test which in one goroutine calls this method and in another calls read RulesForPolicies. I should fail with race flag enabled.

anjmao: This algorithm looks a bit unclear. If you just want to merge new policies into current list I would do something like this:

```go
pr.policyLock.Lock()
policiesLookupTable := map[key]struct{}{}
// Add policies keys from pr.policyList to map for fast lookup check.
pr.policyLock.Unlock()

var newPolicies []market.AccessPolicy
for _, policy := range policies {
    // now check if policy exist using lookup map, no need for getPolicyIndex method.
   // do your stuff with fetch and add to newPolicies
}

pr.policyLock.Lock()
// now can update original pr.policyList with new added policies
pr.policyLock.Unlock()
```

Waldz: Talked over algorithm with @anjmao, decided that suggestion irrelevant

core/policy/policy_repository.go
Waldz: Thanks on ideas. Improved locking of list

core/policy/policy_repository.go
```diff
@@ -0,0 +1,214 @@
...
+
+// Stop ends fetching policies to repository
+func (pr *PolicyRepository) Stop() {
+	pr.fetchShutdown <- struct{}{}
+}
+
+// Policy converts given value to valid policy rule
+func (pr *PolicyRepository) Policy(policyID string) market.AccessPolicy {
+	policyURL := pr.policyURL
+	if !strings.HasSuffix(policyURL, "/") {
```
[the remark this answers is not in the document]
Waldz: It can be 3rd party service, just follows this contract:

1. While starting service, provide Policy Oracle via:
  `--access-policy.address=https://testnet-trust.mysterium.network/api/v1/access-policies/` 
  `--access-policy.address:https://testnet-trust.mysterium.network/api/v1/access-policies`
2. Provider policy IDs via:
  `--access-policy.list= ipinfo,wikipedia`

So yes, provided custom URL still have to match certain contract

core/policy/policy_repository.go
```diff
@@ -0,0 +1,214 @@
...
+
+	pr.policyLock.Lock()
+	pr.policyList = policyListNew
+	pr.policyLock.Unlock()
+
+	return nil
+}
+
+// RulesForPolicy gives rules of given policy
+func (pr *PolicyRepository) RulesForPolicy(policy market.AccessPolicy) (market.AccessPolicyRuleSet, error) {
```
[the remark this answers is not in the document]
Waldz: Actually there is difference in return type, multiple retrieve returns list of asked rulesets

core/policy/policy_repository.go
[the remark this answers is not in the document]
soffokl: But you don't need to add `/` at the end of the user URL. 
If you need it for URL `https://testnet-trust.mysterium.network/api/v1/access-policies/` you passing it through the command argument. 

If I need to pass the URL `https://myown-trust.com/policies.json`, it should not become `https://myown-trust.com/policies.json/`.

core/policy/repository.go
```diff
@@ -0,0 +1,212 @@
...
+			pr.policyLock.Unlock()
+
+			for index := range policyListActive {
+				var err error
+				if err = pr.fetchPolicyRules(&policyListActive[index]); err != nil {
+					log.Warn().Err(err).Msg("synchronise fetch failed")
+				}
+
+				pr.policyLock.Lock()
+				pr.policyList[index] = policyListActive[index]
```
anjmao: This could create some nasty bugs. You updating here original policyList item by index of copy of policy, but in theory at this time original pr.policyList may change and it's indexes will not mach any more so you will end up updating wrong item.

I suggest modify `AddPolicies` so it only adds request to channel and then you can handle it here if fetchLoop so you have only once place which modifies pr.policyList.

```go
func (pr *Repository) fetchLoop() {
	for {
		select {
		case <-pr.fetchShutdown:
                        return;
                case policies := <-pr.newPolicies:
                       // Handle new policies
                case <-time.After(pr.fetchInterval):
                      // ...
 
```

core/policy/repository.go
```diff
@@ -0,0 +1,212 @@
...
+// AddPolicies adds given policy to repository. Also syncs policy rules from TrustOracle
+func (pr *Repository) AddPolicies(policies []market.AccessPolicy) error {
+	pr.policyLock.Lock()
+	defer pr.policyLock.Unlock()
+
+	policyListNew := make([]policyMetadata, len(pr.policyList))
+	copy(policyListNew, pr.policyList)
+
+	for _, policy := range policies {
+		index, exist := pr.getPolicyIndex(policyListNew, policy)
```
anjmao: Maybe it make sense to fetch all active policies at once by providerID? As there is no a lot of policies it would simplify this whole logic a lot. And fetch loop will be fetching all active policies once by one anyway so could fetch in one request too.

PR #1563, opened 2020-01-22
Test install script

tadaskay:
Part of #1545 
Need to integrate it into CI

For now, you can run manually: `bash ./testinstall/test.sh`

Also fixed ant tested on:

- raspbian buster
- debian buster
- debian stretch
- ubuntu bionic
- ubuntu xenial


test-install/Vagrantfile
```diff
@@ -0,0 +1,24 @@
+# -*- mode: ruby -*-
```
anjmao: Maybe rename this folder to test_install to match other folders naming. If we add here go files someday they would want to be in test_install folder to feel a bit better 😄

test-install/test.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
+
+shopt -s extglob
+set -e
+set -o errtrace
+set -o errexit
+set -o pipefail
+
+script_url=https://raw.githubusercontent.com/mysteriumnetwork/node/master/install.sh
```
soffokl: Should this be a local version of `install.sh`? You are pulling always from the master.
```suggestion
script_url=https://raw.githubusercontent.com/mysteriumnetwork/node/master/install.sh
```

tadaskay: I chose this so there's no need to mount anything. Users always install it from the master branch anyway.

tadaskay: It may change as I am now trying to implement it using containers instead of VM (so we can run it on CI).

PR #1573, opened 2020-01-27
Introduce service for SOCKS5 

Waldz:
Providers can start additional service type now - "socks5" and shared their IP for cases like crawling, http request proxying, etc.
Crawling case enduser needed to establish proxy frontend on top of VPN connection, which is heavy, endusers can connection in P2P style.

Not implemented yet:
- connection negotiation part to this service
- authentication layer to allow only negotiated connections
- calculation of established session statistics

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,6 +6,7 @@ require (
 	github.com/BurntSushi/toml v0.3.1
 	github.com/DataDog/zstd v1.4.1 // indirect
 	github.com/Sereal/Sereal v0.0.0-20190618215532-0b8ac451a863 // indirect
+	github.com/armon/go-socks5 v0.0.0-20160902184237-e75332964ef5
 	github.com/arthurkiller/rollingwriter v1.1.2
 	github.com/asaskevich/EventBus v0.0.0-20180315140547-d46933a94f05
 	github.com/asdine/storm v2.1.2+incompatible
--- a/services/socks5/service/service.go
+++ b/services/socks5/service/service.go
@@ -1,0 +1,202 @@
+/*
+ * Copyright (C) 2020 The "MysteriumNetwork/node" Authors.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package service
+
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/mysteriumnetwork/node/core/location"
+	"github.com/mysteriumnetwork/node/core/port"
+	"github.com/mysteriumnetwork/node/identity"
+	"github.com/mysteriumnetwork/node/market"
+	"github.com/mysteriumnetwork/node/nat/event"
+	"github.com/mysteriumnetwork/node/nat/mapping"
+	"github.com/mysteriumnetwork/node/nat/traversal"
+	"github.com/mysteriumnetwork/node/services/openvpn/discovery/dto"
+	"github.com/mysteriumnetwork/node/services/socks5"
+	sock5_session "github.com/mysteriumnetwork/node/services/socks5/session"
+	"github.com/mysteriumnetwork/node/session"
+	"github.com/pkg/errors"
+	"github.com/rs/zerolog/log"
+)
+
+// GetProposal returns the proposal for SOCKS5 service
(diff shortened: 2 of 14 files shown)


services/socks5/service/service.go
```diff
@@ -0,0 +1,252 @@
...
+
+	conf := &socks5_lib.Config{}
+	server, err := socks5_lib.New(conf)
+	if err != nil {
+		return errors.Wrap(err, "could not initiate SOCKS5 proxy server")
+	}
+
+	log.Info().Msg("SOCKS5 service started successfully now")
+	err = m.runServer(server, fmt.Sprintf(":%d", m.servicePort))
+	if err != nil && err != errServerStopped {
```
anjmao: Not sure if errServerStopped is needed here. If server is stoped gracefully could return nil and this extra check will not be needed too.

go.mod
```diff
@@ -6,6 +6,7 @@ require (
 	github.com/BurntSushi/toml v0.3.1
 	github.com/DataDog/zstd v1.4.1 // indirect
 	github.com/Sereal/Sereal v0.0.0-20190618215532-0b8ac451a863 // indirect
+	github.com/armon/go-socks5 v0.0.0-20160902184237-e75332964ef5
```
anjmao: This library was last time updated 3 years ago. It may still be stable and good quality but maybe you could take a look at golang/x/net is suitable https://github.com/golang/net/blob/master/proxy/socks5.go?

services/socks5/service/service.go
```diff
@@ -0,0 +1,252 @@
...
+var errServerStopped = errors.New("server stopped")
+
+// runServer listens on the TCP network address and then handles incoming connections.
+// runServer always returns a non-nil error.
+// After Stop(), the returned error is errServerStopped.
+func (m *Manager) runServer(server *socks5_lib.Server, address string) error {
+	var err error
+
+	m.listenerLock.Lock()
+	m.listenerStop = make(chan struct{})
```
vkuznecovas: move this to the constructor and initialize the channel there.

services/socks5/service/service.go
```diff
@@ -0,0 +1,252 @@
...
+// Stop stops service.
+func (m *Manager) Stop() error {
+	log.Info().Msg("SOCKS5 service stopping..")
+
+	m.listenerLock.Lock()
+	close(m.listenerStop)
+	m.listenerLock.Unlock()
+
+	if m.listener != nil {
+		return m.listener.Close()
```
vkuznecovas: remove this, and just close the listener if stop is called from the for loop in start method.

services/socks5/service/service_test.go
```diff
@@ -0,0 +1,93 @@
...
+			},
+			PaymentMethodType: "PER_TIME",
+			PaymentMethod:     dto.DefaultPaymentInfo,
+		},
+		GetProposal(location.Location{Country: country}),
+	)
+}
+
+func Test_Manager_Stop(t *testing.T) {
+	manager := newManagerStub(servicePort, publicIP, outboundIP)
```
vkuznecovas: Don't just launch a go routine without acknowledging that it's actually done doing what it's supposed to do.

```suggestion
	manager := newManagerStub(servicePort, publicIP, outboundIP)
        errChan := make(chan error)
	go func() {
		err := manager.Serve(providerID)
		errChan <- err
	}()

	waitABit()
	err := manager.Stop()
	assert.NoError(t, err)
	
	assert.NoError(t, <-errChan)
```

services/socks5/service/service.go
```diff
@@ -0,0 +1,252 @@
...
+		}
+		go server.ServeConn(conn)
+	}
+}
+
+// Stop stops service.
+func (m *Manager) Stop() error {
+	log.Info().Msg("SOCKS5 service stopping..")
+
+	m.listenerLock.Lock()
```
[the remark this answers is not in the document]
vkuznecovas: as @anjmao states, the lock is absolutely redundant. I'd rather wrap the close in a sync.Once though, in case stop is called more than once.

services/socks5/service/service.go
```diff
@@ -0,0 +1,252 @@
...
+	}
+
+	natPingerEnabled := m.natPinger.Valid() && m.isBehindNAT() && m.portMappingFailed()
+	traversalParams, err := m.newTraversalParams(natPingerEnabled, sessionRequest)
+	if err != nil {
+		return nil, errors.Wrap(err, "could not create traversal params")
+	}
+
+	if natPingerEnabled {
+		log.Info().Msgf("NAT Pinger enabled, binding service port for proxy, key: %s", traversalParams.ProxyPortMappingKey)
```
anjmao: Will it work with NAT hole punching? SOCKS5 is running on TCP and hole punching is using UDP.

zolia: Yes, as @anjmao noted, TCP hole punching mechanism is not supported in our implementation atm. For this we will need completely new TCP based pinger. I suspect that success rate for TCP hole punching will much lower. Idea is very similar at to one with UDP NAT, but now we need to control how we send TCP SYN packet to remote peer. 

Created related ticket #1680 

For now, I would would not attempt NAT hole punching in case upnp fails. We could even stop socks service if it is behind the nat and upnp failed.

tadaskay:
@Waldz can you update this or close?

PR #1574, opened 2020-01-27
Contributors updated

zolia:
I think, most of our core team is experienced enough and familiar with our codebase to be code owners. Propose these changes. 
Please let me know if I was not accurate enough and will update accordingly.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,1 @@
-* @tadovas @zolia @vkuznecovas @soffokl
+* @zolia @vkuznecovas @soffokl @waldz @tadaskay @anjmao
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -88,14 +90,19 @@ For possible issues while running a node refer to our [FAQ](https://github.com/m
 
 Please read [CONTRIBUTING.md](./CONTRIBUTING.md) for details on our code of conduct, and the process for submitting pull requests to us.
 
-## Authors
-* **Valdas Petrulis** - *Lead developer, go evangelist* - [Waldz](https://github.com/Waldz)
+## Contributors
```
soffokl: Maybe this should be just a link to the https://github.com/mysteriumnetwork/node/graphs/contributors page?

.github/CODEOWNERS
```diff
@@ -1 +1 @@
-* @tadovas @zolia @vkuznecovas @soffokl
+* @zolia @vkuznecovas @soffokl @waldz @tadaskay @anjmao
```
soffokl: Do we really need codeowners if everyone is code owner? Maybe just write access to the repo is enough?

zolia: Not everyone, but at this stage, majority of our team should be owners. There is too little of us and we are all experienced enough. Future contributors should not be owners automatically.

README.md
zolia: We have that link. But IMHO its good to have a list of ppl that contributed the most. For example added @chompomonim too since he contributed great deal to other repos, that are not properly reflected by this repo.

PR #1575, opened 2020-01-27
Wireguard port mapping fixes

anjmao:
- Port mapper now could be used in blocking way when port mapping result is needed. This fixes case when only WireGuard service is started and port mapping result may not yet be available since it implemented by checking from global event tracker store which uses even bus (async).
- Port mapper will now not update port mapping in intervals if it was created with permanent lease.
- Refactor pinger to support disabling NAT proxy for WireGuard consumer.


nat/mapping/port_mapping.go
```diff
@@ -21,76 +21,110 @@ import (
...
-		mapperQuit,
-		protocol,
-		port,
-		port,
-		name,
-		publisher)
-
-	return func() { close(mapperQuit) }
+type portMapper struct {
+	publicIP, outboundIP string
```
soffokl: > field `outboundIP` is unused (unused)

nat/mapping/port_mapping.go
```diff
@@ -21,76 +21,110 @@ import (
...
+	if err != nil {
+		p.publisher.Publish(event.Topic, event.BuildFailureEvent(StageName, err))
+	} else {
+		p.publisher.Publish(event.Topic, event.BuildSuccessfulEvent(StageName))
 	}
 }
 
-func addMapping(m portmap.Interface, protocol string, extPort, intPort int, name string, publisher Publisher) {
-	if err := m.AddMapping(protocol, extPort, intPort, name, mapTimeout); err != nil {
+func (p *portMapper) addMapping(protocol string, extPort, intPort int, name string) (err error, permanent bool) {
```
soffokl: > error should be the last type when returning multiple items (golint)

nat/traversal/pinger.go
```diff
@@ -32,68 +33,66 @@ import (
...
-const pingTimeout = 10000
 
 var (
 	errNATPunchAttemptStopped  = errors.New("NAT punch attempt stopped")
 	errNATPunchAttemptTimedOut = errors.New("NAT punch attempt timed out")
 )
 
+// NATProviderPinger pings provider and optionally hands off connection to consumer proxy.
+type NATProviderPinger interface {
+	PingProvider(ip string, port, consumerPort, proxyPort int, stop <-chan struct{}) error
```
soffokl: These ports starting to be unreadable. Should the`port` be a `providerPort`?

mobile/mysterium/wireguard_connection_setup.go
```diff
@@ -157,20 +157,27 @@ func newTunnDevice(wgTunnSetup WireguardTunnelSetup, config wireguard.ServiceCon
 	return tunDevice, err
 }
 
-func waitUDPPortReadyFor(port int, timeout time.Duration) error {
+func waitUDPPortReadyFor(port int, timeout, checkInterval time.Duration) error {
 	timeoutChan := time.After(timeout)
+	check := func() error {
```
vkuznecovas: why not extract this to a separate method? Looks cleaner imho than inlining

mobile/mysterium/wireguard_connection_setup.go
```diff
@@ -157,20 +157,27 @@ func newTunnDevice(wgTunnSetup WireguardTunnelSetup, config wireguard.ServiceCon
 	return tunDevice, err
 }
 
-func waitUDPPortReadyFor(port int, timeout time.Duration) error {
+func waitUDPPortReadyFor(port int, timeout, checkInterval time.Duration) error {
```
vkuznecovas: I'd return a cancel func here as well, to allow for early exit to the caller. as currently this might be stuck for quite a bit if I want to exit node.

mobile/mysterium/wireguard_connection_setup_test.go
```diff
@@ -0,0 +1,47 @@
...
+	"time"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestUDPPortChecker(t *testing.T) {
+	port := 51000
+
+	udpServerStarted := make(chan struct{})
+	go func() {
```
vkuznecovas: this whole test is super long running, 0.2s for a single test. Can we optimize/redesign?

nat/mapping/port_mapping_test.go
```diff
@@ -0,0 +1,148 @@
...
+}
+
+type mapping struct {
+	protocol         string
+	extport, intport int
+	name             string
+	lifetime         time.Duration
+}
+
+type mockRouter struct {
```
zolia: nice notion for testing. We could even start having templates for different router models, which would cover e2e like testing based on their internal functionality.

nat/traversal/pinger.go
```diff
@@ -164,27 +168,33 @@ func (p *Pinger) PingProvider(ip string, port int, consumerPort int, stop <-chan
...
-	p.pingCancelled <- struct{}{}
-
-	if consumerPort > 0 {
-		consumerAddr := fmt.Sprintf("127.0.0.1:%d", consumerPort+1)
+	if proxyPort > 0 {
+		consumerAddr := fmt.Sprintf("127.0.0.1:%d", proxyPort)
 		log.Info().Msg("Handing connection to consumer NATProxy: " + consumerAddr)
+
+		// Set higher read deadline when NAT proxy is used.
+		conn.SetReadDeadline(time.Now().Add(12 * time.Hour))
```
zolia: We can relate this with payment promise timeout or not set timeout at all?

PR #1577, opened 2020-01-27
Sensible limits payments

### LINKED ISSUE #1570: Sensible consumer price range
#1569 introduces payment information parsing from proposal, but this introduces a conundrum - as a consumer, I'd like to only see the proposals that fit my accepted price range. 

It would be good to introduce a flag that sets a sensible limit for the consumer and filter away all proposals that do not fall into that range.

It's also a prerequisite for a less cheating prone bounty system.

The sensible limit should be 1000000 mystt for starters as we want backwards compatibility. After a month or so, once the providers/consumers have been updated the sensible limit should probably be related to the amount set in #1547.

vkuznecovas:
Added a sensible limit for consumer in terms of price. Consumers can still force-connect if they have the required id's, but otherwise will not see proposals with exceedingly large prices.

Closes #1570

### DIFF
--- a/config/flags_payments.go
+++ b/config/flags_payments.go
@@ -65,6 +65,18 @@ var (
 		Name:  "payments.disable",
 		Usage: "Disables payments and moves to a backwards compatible legacy mode",
 	}
+	// FlagPaymentsConsumerUpperPriceBound sets the upper price bound to a set value.
+	FlagPaymentsConsumerUpperPriceBound = cli.Uint64Flag{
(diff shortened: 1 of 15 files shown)


config/flags_payments.go
```diff
@@ -65,6 +65,18 @@ var (
 		Name:  "payments.disable",
```
soffokl: Do we need to add this empty file `config/config.toml`?

vkuznecovas: No we don't, thx :)

Issue #1534, opened 2020-01-17
"Limit bandwidth to 5Mbps" option does not appear to work


ninjit:
The checkbox under settings does not remain enabled.
And monitoring system network usage indicates it's not functioning either – sessions regularly use more than that.
<img width="506" alt="Screen Shot 2020-01-17 at 8 57 04 AM" src="https://user-images.githubusercontent.com/753263/72630538-67f6c000-3907-11ea-9470-ca2c7f856092.png">

PR #1578, opened 2020-01-28
Bump go-wondershaper to fix bandwidth limiting.

soffokl:
Closes: https://github.com/mysteriumnetwork/node/issues/1534


bin/package/sudoers/mysterium-node
```diff
@@ -5,3 +5,4 @@ mysterium-node    ALL=NOPASSWD: /sbin/iptables
 mysterium-node    ALL=NOPASSWD: /sbin/ip
 mysterium-node    ALL=NOPASSWD: /etc/mysterium-node/prepare-env.sh
 mysterium-node    ALL=NOPASSWD: /sbin/tc
+mysterium-node    ALL=NOPASSWD: /sbin/modeprobe
```
tadaskay: shouldn't this be `modprobe` @soffokl ?

soffokl: Thanks, it should be `modprobe`, I'll fix it.

PR #1603, opened 2020-01-31
Port testinstall to docker+go

tadaskay:
Fixes: #1545

Rewrote bash+vagrant to go+docker


ci/test/testinstall/install_integration_test.go
```diff
@@ -0,0 +1,128 @@
...
+	installFile = "install.sh"
+)
+
+// TestInstall tests installation on various Linux distributions.
+// SystemD docker images based on https://github.com/j8r/dockerfiles
+// Installs myst using install.sh, tests for healthcheck output.
+func TestInstall(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
```
vkuznecovas: why not []string?

ci/test/testinstall/install_integration_test.go
```diff
@@ -0,0 +1,128 @@
...
+	failIf(err)
+
+	var containerId string
+	scanner = bufio.NewScanner(strings.NewReader(runOutput))
+	for scanner.Scan() {
+		line := scanner.Text()
+		fmt.Println(line)
+		containerId = strings.TrimSpace(line)[0:12]
+	}
+	defer func() {
```
vkuznecovas: might want to move it up a couple of lines just after the docker run

ci/test/testinstall/install_integration_test.go
```diff
@@ -0,0 +1,128 @@
...
+	}()
+	failIf(scanner.Err())
+	assert.NotEmpty(containerId)
+
+	err = sh.RunV("docker", "exec", containerId, "curl", "-o", installFile, installURL)
+	failIf(err)
+	err = sh.RunV("docker", "exec", containerId, "bash", installFile)
+	failIf(err)
+
+	time.Sleep(5 * time.Second)
```
vkuznecovas: you could make this more resilient with like this:

```go
timeout := time.After(5 *time.Second)

for {
select {
  case <-time.After(time.Millisecond*500):
            healthcheckJSON, err := sh.Output("docker", "exec", containerId, "curl", "-s", "localhost:4050/healthcheck")
            if err != nil {
                 log...
            }
            err = json.Unmarshal([]byte(healthcheckJSON), &healthcheckOutput)
	    failIf(err)
	    assert.NotEmpty(healthcheckOutput.Uptime)


      case <-timeout:
                 assert.Fail(t, "failed)
    }
}
```

ci/test/testinstall/install.go
```diff
@@ -0,0 +1,27 @@
...
+
+package testinstall
+
+import (
+	"github.com/magefile/mage/sh"
+)
+
+// TestInstallScript tests install.sh
+func TestInstallScript() error {
+	return sh.RunV("go", "test", "-v", "-tags=integration", "github.com/mysteriumnetwork/node/ci/test/testinstall")
```
Waldz: would be nice to have all e2e tests in `/e2e`

ci/test/testinstall/install_integration_test.go
tadaskay: I'm killing a particular `containerId`. The defer is right after we are able to parse `containerId` from `docker run` output.

ci/test/testinstall/install_integration_test.go
tadaskay: Auto-generated. But right, it could be simplified.

ci/test/testinstall/install_integration_test.go
tadaskay: Rewrote using `assert.Eventually`.

tadaskay: I had to upgrade to the latest unreleased `testify` because https://github.com/stretchr/testify/pull/808 hasn't been released yet.

ci/test/testinstall/install.go
tadaskay: I'll keep it separate because as they are now, e2e tests check node from functional perspective, and this merely tests the installation script.

Issue #1499, opened 2020-01-13
Wireguard does not cleanup when connection fails


anjmao:
At this point handshake https://github.com/mysteriumnetwork/node/blob/master/services/wireguard/connection/connection.go#L105 never finishes and if consumer is closed already applied NAT/Firewall rules are not applied which break whole network.

PR #1606, opened 2020-01-31
Fix WireGuard connection cleanup if start fails

anjmao:
Sometimes connection Start could finish setup tunnel but if handshake wait timeout (on any other error) occurs it will not cleanup. Now Start always calls Stop on error which does needed cleanup logic.

Fixes https://github.com/mysteriumnetwork/node/issues/1499

### DIFF
--- a/cmd/di_desktop.go
+++ b/cmd/di_desktop.go
@@ -21,6 +21,7 @@ package cmd
 
 import (
 	"fmt"
+	"time"
 
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/mysteriumnetwork/node/communication"
@@ -44,6 +45,8 @@ import (
 	openvpn_service "github.com/mysteriumnetwork/node/services/openvpn/service"
 	"github.com/mysteriumnetwork/node/services/wireguard"
 	wireguard_connection "github.com/mysteriumnetwork/node/services/wireguard/connection"
+	"github.com/mysteriumnetwork/node/services/wireguard/endpoint"
+	"github.com/mysteriumnetwork/node/services/wireguard/resources"
 	wireguard_service "github.com/mysteriumnetwork/node/services/wireguard/service"
 	"github.com/mysteriumnetwork/node/session"
 	"github.com/mysteriumnetwork/node/session/connectivity"
@@ -290,8 +293,19 @@ func (di *Dependencies) registerConnections(nodeOptions node.Options) {
 
 func (di *Dependencies) registerWireguardConnection(nodeOptions node.Options) {
 	wireguard.Bootstrap()
+	dnsManager := wireguard_connection.NewDNSManager()
+	handshakeWaiter := wireguard_connection.NewHandshakeWaiter()
+	endpointFactory := func() (wireguard.ConnectionEndpoint, error) {
+		resourceAllocator := resources.NewAllocator(nil, wireguard_service.DefaultOptions.Subnet)
+		return endpoint.NewConnectionEndpoint(nil, resourceAllocator, 0)
+	}
 	connFactory := func() (connection.Connection, error) {
-		return wireguard_connection.NewConnection(nodeOptions.Directories.Config, di.IPResolver, di.NATPinger)
+		opts := wireguard_connection.Options{
+			DNSConfigDir:        nodeOptions.Directories.Config,
+			StatsUpdateInterval: 1 * time.Second,
+			HandshakeTimeout:    1 * time.Minute,
+		}
+		return wireguard_connection.NewConnection(opts, di.IPResolver, di.NATPinger, endpointFactory, dnsManager, handshakeWaiter)
 	}
 	di.ConnectionRegistry.Register(wireguard.ServiceType, connFactory)
 }
--- a/mobile/mysterium/wireguard_connection_setup.go
+++ b/mobile/mysterium/wireguard_connection_setup.go
@@ -21,6 +21,7 @@ import (
 	"bufio"
 	"encoding/json"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/mysteriumnetwork/node/consumer"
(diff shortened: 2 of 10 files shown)


cmd/di_desktop.go
```diff
@@ -290,8 +293,19 @@ func (di *Dependencies) registerConnections(nodeOptions node.Options) {
...
+	handshakeWaiter := wireguard_connection.NewHandshakeWaiter()
+	endpointFactory := func() (wireguard.ConnectionEndpoint, error) {
+		resourceAllocator := resources.NewAllocator(nil, wireguard_service.DefaultOptions.Subnet)
+		return endpoint.NewConnectionEndpoint(nil, resourceAllocator, 0)
+	}
 	connFactory := func() (connection.Connection, error) {
-		return wireguard_connection.NewConnection(nodeOptions.Directories.Config, di.IPResolver, di.NATPinger)
+		opts := wireguard_connection.Options{
+			DNSConfigDir:        nodeOptions.Directories.Config,
+			StatsUpdateInterval: 1 * time.Second,
```
vkuznecovas: can we expose these through flags?

mobile/mysterium/wireguard_connection_setup.go
```diff
@@ -164,19 +154,18 @@ func (c *wireguardConnection) Wait() error {
...
-	c.updateStatistics()
-	if c.device != nil {
-		c.device.Close()
-		c.device.Wait()
-	}
-	c.stateCh <- connection.NotConnected
+	c.closeOnce.Do(func() {
+		c.stateCh <- connection.Disconnecting
+		if c.device != nil {
+			c.device.Stop()
```
vkuznecovas: in reality, can this even be nil?

services/wireguard/connection/connection.go
```diff
@@ -223,29 +228,32 @@ func (c *Connection) isNoopPinger() bool {
...
-
-	if c.connectionEndpoint != nil {
-		if err := cleanDNS(c.configDir, c.connectionEndpoint.InterfaceName()); err != nil {
-			log.Error().Err(err).Msg("Failed to clear DNS")
+	c.stopOnce.Do(func() {
+		log.Info().Msg("Stopping WireGuard connection")
+		c.stateCh <- connection.Disconnecting
+
+		if c.connectionEndpoint != nil {
+			// c.sendStats()
```
vkuznecovas: remove the redundant comment

services/wireguard/connection/connection.go
```diff
@@ -223,29 +228,32 @@ func (c *Connection) isNoopPinger() bool {
...
+				log.Error().Err(err).Msg("Failed to clear DNS")
+			}
+			if err := c.connectionEndpoint.Stop(); err != nil {
+				log.Error().Err(err).Msg("Failed to close wireguard connection")
+			}
 		}
-		if err := c.connectionEndpoint.Stop(); err != nil {
-			log.Error().Err(err).Msg("Failed to close wireguard connection")
+
+		if c.removeAllowedIPRule != nil {
```
vkuznecovas: in constructor, you could set this to func() {}

Then you would not need this check. You would also be able to do the following:
```golang
defer  c.removeAllowedIPRule()
if c.connectionEndpoint == nil {
return
}

if err := c.dnsManager.Clean(c.opts.DNSConfigDir, c.connectionEndpoint.InterfaceName()); err != nil {
				log.Error().Err(err).Msg("Failed to clear DNS")
			}
			if err := c.connectionEndpoint.Stop(); err != nil {
				log.Error().Err(err).Msg("Failed to close wireguard connection")
			}

File: services/wireguard/connection/dns.go [hunk 5]
```diff
@@ -25,14 +25,29 @@ import (
 	"path"
 )
 
-func setDNS(configDir, dev, dns string) error {
+// DNSManager is connection DNS configuration manager.
+type DNSManager interface {
```
vkuznecovas: this definition should be near where it's being used, not where it's implemented

anjmao: If interface can be reused there is no point to duplicate it in different places. It just happens that there is only one implementation which is also placed in this file. There are many interfaces which you can reuse like io.Reader, io.Writer etc.

The same applies to `type HandshakeWaiter interface` which I'm using from different places. If I copy it to all used places when I also need to keep documentation in sync which is strange.

services/wireguard/connection/connection.go
anjmao: This looks more clear as it shows that `removeAllowedIPRule` was actually applied on Start().
Also I can't return if connectionEndpoint is nil as I need to close all channels at the end.

mobile/mysterium/wireguard_connection_setup.go
```diff
@@ -164,19 +154,16 @@ func (c *wireguardConnection) Wait() error {
 }
 
 func (c *wireguardConnection) Stop() {
-	c.stateCh <- connection.Disconnecting
-	c.updateStatistics()
```
soffokl: This line was updating stats on disconnect. With this, you don't need to wait 5 seconds on e2e test to make sure we updated stats. And we are not losing the last part of stats.

e2e/connection_test.go
```diff
@@ -157,6 +157,9 @@ func consumerConnectFlow(t *testing.T, tequilapi *tequilapi_client.Client, consu
 	assert.Equal(t, connectionStatus.SessionID, se.SessionID)
 	assert.Equal(t, "New", se.Status)
 
+	// Keep active connection for some time to check for statistics change.
+	time.Sleep(5 * time.Second)
```
soffokl: It should be enough to update stats on disconnect, no need to wait 5 seconds.

anjmao: e2e was not testing statistics update loop at all since it was stopping connection immediately. I added it to be at least a bit realistic.

PR #1618, opened 2020-02-03
Handle accountant errors

vkuznecovas:
Until now, all errors from accountant were treated equally. They are now classified and acted upon accordingly.

### DIFF
--- a/session/pingpong/invoice_tracker.go
+++ b/session/pingpong/invoice_tracker.go
@@ -86,6 +86,8 @@ type accountantCaller interface {
 	RevealR(r string, provider string, agreementID uint64) error
 }
 
+type settler func(providerID, accountantID identity.Identity) error
+
 const chargePeriodLeeway = time.Hour * 2
 
 type sentInvoice struct {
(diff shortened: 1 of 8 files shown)


session/pingpong/invoice_tracker.go
```diff
@@ -288,18 +284,12 @@ func (it *InvoiceTracker) revealPromise() error {
...
-	err = it.deps.AccountantCaller.RevealR(accountantPromise.R, it.deps.ProviderID.Address, accountantPromise.AgreementID)
-	if err != nil {
-		log.Error().Err(err).Msg("Could not reveal R")
-		it.incrementAccountantFailureCount()
-		if it.getAccountantFailureCount() > it.deps.MaxAccountantFailureCount {
-			return errors.Wrap(err, "max failure count calling accountant reached")
-		}
-		log.Warn().Msg("Ignoring accountant error, we haven't reached the error threshold yet")
-		return errHandled
+	err = it.deps.AccountantCaller.RevealR("wazap", it.deps.ProviderID.Address, accountantPromise.AgreementID)
```
soffokl: Why it started to be `wazap` instead of `accountantPromise.R`? Is it ok to have here a const string?

vkuznecovas: leftovers from testing, fixed

Issue #1613, opened 2020-02-03
SystemD service failed to stop mysterium node gracefully within 10 seconds. 


soffokl:
**Describe the bug**
SystemD service failed to stop mysterim node gracefully withing 10 seconds. 
This prevents correct system cleaning.

**To Reproduce**
`systemctl stop mysterium-node`

**Expected behavior**
Service should have enough time to clean up everything. Make sure we are not blocking anywhere. 
We can increase timeout, but we need to check other options first.

**Additional context**
```
Feb 03 00:28:56 ubuntu systemd[1]: Stopping Server for Mysterium - decentralised VPN Network...
Feb 03 00:28:56 ubuntu myst[11904]: 2020-02-03T00:28:56.907 INF cmd/commands/service/command.go:91       > Stopping application
Feb 03 00:28:56 ubuntu myst[11904]: 2020-02-03T00:28:56.908 INF logconfig/openvpnlog.go:54               > [openvpn-mgmt] Shutdown
Feb 03 00:28:56 ubuntu myst[11904]: 2020-02-03T00:28:56.909 INF core/discovery/discovery.go:151          > Stopping discovery loop..
Feb 03 00:28:56 ubuntu myst[11904]: 2020-02-03T00:28:56.909 INF logconfig/openvpnlog.go:54               > [openvpn-mgmt] Shutdown finished
Feb 03 00:28:56 ubuntu myst[11904]: 2020-02-03T00:28:56.909 DBG logconfig/openvpnlog.go:63               > [openvpn-proc] Stdout: Mon Feb  3 00:28:56 2020 event_wait : Interrupted system call (code=4)
Feb 03 00:28:56 ubuntu myst[11904]: 2020-02-03T00:28:56.910 DBG logconfig/openvpnlog.go:63               > [openvpn-proc] Stdout: Mon Feb  3 00:28:56 2020 SENT CONTROL [UNDEF]: 'RESTART' (status=1)
Feb 03 00:28:56 ubuntu myst[11904]: 2020-02-03T00:28:56.910 DBG logconfig/openvpnlog.go:63               > [openvpn-proc] Stdout: Mon Feb  3 00:28:56 2020 SENT CONTROL [UNDEF]: 'RESTART' (status=1)
Feb 03 00:28:56 ubuntu sudo[12143]: mysterium-node : TTY=unknown ; PWD=/ ; USER=root ; COMMAND=/sbin/ip addr flush dev tun0
[...]

PR #1619, opened 2020-02-03
Improvements for on-demand proposal fetching 

Waldz:
Fixes: #1613

### DIFF
--- a/core/discovery/brokerdiscovery/repository.go
+++ b/core/discovery/brokerdiscovery/repository.go
@@ -34,10 +34,12 @@ type Repository struct {
 	receiver        communication.Receiver
 	timeoutInterval time.Duration
 
-	watchdogStop chan struct{}
-	watchdogStep time.Duration
-	watchdogLock sync.Mutex
-	watchdogSeen map[market.ProposalID]time.Time
+	stopOnce sync.Once
+	stopChan chan struct{}
+
+	timeoutCheckStep  time.Duration
+	watchdogLock      sync.Mutex
+	timeoutCheckSeens map[market.ProposalID]time.Time
 }
 
 // NewRepository constructs a new proposal repository (backed by the broker).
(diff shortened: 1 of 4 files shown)


core/discovery/brokerdiscovery/repository.go
```diff
@@ -86,18 +85,14 @@ func (s *Repository) Start() error {
...
 
 	return nil
 }
 
 // Stop ends proposals synchronization to storage
 func (s *Repository) Stop() {
-	s.stopOnce.Do(func() {
-		close(s.stopChan)
-	})
+	s.stopChan <- struct{}{}
```
vkuznecovas: why? this means only a single listener will receive the signal, do close it.

Waldz: There is just one listener. I dont get what will it give, because then you have to do:
```
pr.once.Do(func() {
		log.Info().Msg("Stopping provider registrar")
		close(pr.stopChan)
	})
```

It was added to handle multiple stops.

core/discovery/brokerdiscovery/repository.go
```diff
@@ -137,7 +132,7 @@ func (s *Repository) proposalPingMessage(message pingMessage) error {
 func (s *Repository) proposalWatchdog() {
 	for {
 		select {
-		case <-s.stopChan:
+		case <-s.whatchdogStop:
```
tadaskay: First, a mistype. I believe that you meant `watchdog`.
Secondly, I'm not familiar with the terminology here. 

> A watchdog timer (sometimes called a computer operating properly or COP timer, or simply a watchdog) is an electronic timer that is used to detect and recover from computer malfunctions. During normal operation, the computer regularly resets the watchdog timer to prevent it from elapsing, or "timing out". If, due to a hardware fault or program error, the computer fails to reset the watchdog, the timer will elapse and generate a timeout signal. The timeout signal is used to initiate corrective action or actions. The corrective actions typically include placing the computer system in a safe state and restoring normal system operation.

https://en.wikipedia.org/wiki/Watchdog_timer

Is this the case? It appears to me that it just expires proposals and has nothing to do with system restart/recovery.

core/discovery/brokerdiscovery/repository.go
vkuznecovas: I don't understand what you've written.

core/discovery/brokerdiscovery/repository.go
Waldz: Improved naming

core/discovery/brokerdiscovery/repository.go
Waldz: Rephrased

Waldz: Added clearing opened channels

Issue #1597, opened 2020-01-30
Provider traffic policies by DNS names


Waldz:
As as a provider I want to share my network
While blocking everything and whitelisting domains/services which I allow 
While allowing everything and blacklisting domains/services which I block

Tasks:
- [x] Install `ipset` utility
- [x] Find commands needed for blocking block
- [x] Enable/disable DNS firewall chain on provider start/stop #1602, #1632
- [x] Whitelist DNS query results to DNS firewall #1733, #1741
- [x] Match DNS query results with TrustOracle rules #1733
- [x] Add WireGuard firewall #1754

Resources:
- https://blog.ls20.com/securing-your-server-using-ipset-and-dynamic-blocklists/
- https://forums.gentoo.org/viewtopic-t-863121.html
- http://ipset.netfilter.org/iptables-extensions.man.html

1. Create custom chain, which can be turned on/off:
```
nslookup ipinfo.io

iptables -S PROVIDER_DNS_FIREWALL

iptables -N PROVIDER_DNS_FIREWALL
iptables -A PROVIDER_DNS_FIREWALL -i tun0 -d 216.239.32.21 -j ACCEPT
iptables -A PROVIDER_DNS_FIREWALL -i tun0 -d 216.239.34.21 -j ACCEPT
iptables -A PROVIDER_DNS_FIREWALL -i tun0 -d 216.239.36.21 -j ACCEPT
iptables -A PROVIDER_DNS_FIREWALL -i tun0 -d 216.239.38.21 -j ACCEPT
iptables -A PROVIDER_DNS_FIREWALL -i tun0 -j DROP

iptables -I INPUT -i tun0 -j PROVIDER_DNS_FIREWALL
```

2. Create set of IPs which are allowed:
```
nslookup ipinfo.io

ipset create dns-firewall hash:ip --netmask 24 --hashsize 64
ipset add dns-firewall 216.239.32.21
ipset add dns-firewall 216.239.34.21
ipset add dns-firewall 216.239.36.21
ipset add dns-firewall 216.239.38.21

ipset list dns-firewall
```

3. Block by this dynamic set of IPs:
```
iptables -A PROVIDER_DNS_FIREWALL -m set --match-set dns-firewall dst -j ACCEPT
iptables -A PROVIDER_DNS_FIREWALL -i tun0 -j DROP

iptables -I INPUT -i tun0 -j PROVIDER_DNS_FIREWALL
```

4. Allow all Mysterium services to custom chain

[...]

PR #1632, opened 2020-02-04
Enable traffic blocking for Openvpn service

Waldz:
Updates: #1597

- When provider now start Openvpn service we setup firewall to inspect incoming consumer packets (based on iptables so just for Linux)
- This enables provider possibilities to share his internet by filtering consumer packets by port, destination or any other filters. Make these cases possible:  
-- block all & allow by whitelist
-- allow all & block by blacklist
- Currently no rules are enabled just prepares custom iptables chain wo activating it for a device

### DIFF
--- a/bin/docker/alpine/Dockerfile
+++ b/bin/docker/alpine/Dockerfile
@@ -20,7 +20,7 @@ RUN bin/build
 FROM alpine:3.6
 
 # Install packages
-RUN apk add --update --no-cache iptables ca-certificates openvpn bash sudo openresolv
+RUN apk add --update --no-cache iptables ipset ca-certificates openvpn bash sudo openresolv
 
 COPY bin/helpers/prepare-run-env.sh /usr/local/bin/prepare-run-env.sh
 COPY bin/docker/docker-entrypoint.sh /usr/local/bin/docker-entrypoint.sh
--- a/core/service/manager.go
+++ b/core/service/manager.go
@@ -112,6 +112,7 @@ func (manager *Manager) Start(providerID identity.Identity, serviceType string, 
 	if len(policyIDs) > 0 {
 		policies := manager.policyOracle.Policies(policyIDs)
 		if err = manager.policyOracle.SubscribePolicies(policies, policyRules); err != nil {
+			log.Warn().Err(err).Msg("Can't find given access policies")
 			return id, ErrUnsupportedAccessPolicy
 		}
 		proposal.SetAccessPolicies(&policies)
--- a/firewall/incoming_blocker_iptables.go
+++ b/firewall/incoming_blocker_iptables.go
@@ -27,7 +27,7 @@ import (
 )
 
 const (
-	dnsFirewallChain = "PROVIDER_DNS_FIREWALL"
+	dnsFirewallChain = "MYST_PROVIDER_FIREWALL"
 	dnsFirewallIpset = "myst-provider-dst-whitelist"
 )
 
@@ -126,7 +126,7 @@ func (ibi *incomingBlockerIptables) cleanupStaleRules() error {
 		return err
 	}
 	for _, rule := range rules {
-		// detect if any references exist in FORWARD chain like -j PROVIDER_DNS_FIREWALL
+		// detect if any references exist in FORWARD chain like -j MYST_PROVIDER_FIREWALL
 		if strings.HasSuffix(rule, dnsFirewallChain) {
 			deleteRule := strings.Replace(rule, "-A", "-D", 1)
 			deleteRuleArgs := strings.Split(deleteRule, " ")
--- a/firewall/incoming_blocker_iptables.go
+++ b/firewall/incoming_blocker_iptables.go
@@ -26,9 +26,9 @@ import (
 	"github.com/rs/zerolog/log"
 )
 
-var (
+const (
 	dnsFirewallChain = "PROVIDER_DNS_FIREWALL"
-	dnsFirewallIpset = "dns-firewall-whitelist"
+	dnsFirewallIpset = "myst-provider-dst-whitelist"
 )
 
 // NewIncomingTrafficBlockerIptables creates instance iptables based traffic blocker
@@ -48,9 +48,9 @@ func (ibi *incomingBlockerIptables) Setup() error {
 	if err := ibi.cleanupStaleRules(); err != nil {
 		return err
 	}
-	ipset.Exec(ipset.OpSetDelete(dnsFirewallIpset))
+	ipset.Exec(ipset.OpDelete(dnsFirewallIpset))
 
-	op := ipset.OpSetCreate(dnsFirewallIpset, ipset.SetTypeHashIP, net.CIDRMask(24, 32), 64)
+	op := ipset.OpCreate(dnsFirewallIpset, ipset.SetTypeHashIP, nil, 0)
 	if _, err := ipset.Exec(op); err != nil {
 		return err
 	}
@@ -61,7 +61,7 @@ func (ibi *incomingBlockerIptables) Teardown() {
 	if err := ibi.cleanupStaleRules(); err != nil {
 		log.Warn().Err(err).Msg("Error cleaning up iptables rules, you might want to do it yourself")
 	}
-	if errOutput, err := ipset.Exec(ipset.OpSetDelete(dnsFirewallIpset)); err != nil {
+	if errOutput, err := ipset.Exec(ipset.OpDelete(dnsFirewallIpset)); err != nil {
 		log.Warn().Err(err).Msgf("Error deleting ipset table. %s", strings.Join(errOutput, ""))
 	}
 }
@@ -80,11 +80,11 @@ func (ibi *incomingBlockerIptables) BlockIncomingTraffic(network net.IPNet) (Inc
 }
 
 func (ibi *incomingBlockerIptables) AllowIPAccess(ip net.IP) (IncomingRuleRemove, error) {
-	if _, err := ipset.Exec(ipset.OpSetIPAdd(dnsFirewallIpset, ip)); err != nil {
+	if _, err := ipset.Exec(ipset.OpIPAdd(dnsFirewallIpset, ip)); err != nil {
 		return nil, err
 	}
 	return func() error {
-		_, err := ipset.Exec(ipset.OpSetIPRemove(dnsFirewallIpset, ip))
+		_, err := ipset.Exec(ipset.OpIPRemove(dnsFirewallIpset, ip))
 		return err
 	}, nil
 }
@@ -106,12 +106,12 @@ func (ibi *incomingBlockerIptables) setupDNSFirewallChain() error {
 		return err
 	}
 
-	// Append rule - by default all packets going to dns firewall chain are rejected
+	// Append rule - packets going to DNS firewall with these destination IPs are whitelisted
 	if _, err := iptables.Exec("-A", dnsFirewallChain, "-m", "set", "--match-set", dnsFirewallIpset, "dst", "-j", "ACCEPT"); err != nil {
 		return err
 	}
 
-	// Append rule - by default all packets going to dns firewall chain are rejected
+	// Append rule - by default all packets going to DNS firewall chain are rejected
 	if _, err := iptables.Exec("-A", dnsFirewallChain, "-j", "REJECT"); err != nil {
 		return err
 	}
--- a/firewall/incoming_blocker_iptables.go
+++ b/firewall/incoming_blocker_iptables.go
@@ -31,12 +31,12 @@ const (
 	dnsFirewallIpset = "myst-provider-dst-whitelist"
 )
 
-// NewIncomingTrafficBlockerIptables creates instance iptables based traffic blocker
+// NewIncomingTrafficBlockerIptables creates instance iptables based traffic blocker.
 func NewIncomingTrafficBlockerIptables() IncomingTrafficBlocker {
 	return &incomingBlockerIptables{}
 }
 
-// incomingBlockerIptables allows incoming traffic blocking in IP granularity
+// incomingBlockerIptables allows incoming traffic blocking in IP granularity.
 type incomingBlockerIptables struct{}
 
 func (ibi *incomingBlockerIptables) Setup() error {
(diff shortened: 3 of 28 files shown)


firewall/ipset/operations.go
```diff
@@ -0,0 +1,64 @@
...
+	SetTypeHashIP = SetType("hash:ip")
+)
+
+// OpVersion is an operation which prints version information.
+func OpVersion() []string {
+	return []string{"version"}
+}
+
+// OpSetCreate is an operation which creates a new set.
+func OpSetCreate(setName string, setType SetType, netMask net.IPMask, hashSize int) []string {
```
soffokl: Why do you need this `OpSet` prefix, this just `ipset.Create`
```suggestion
func Create(setName string, setType SetType, netMask net.IPMask, hashSize int) []string {
```

soffokl: The same for all other methods.

firewall/ipset/operations.go
```diff
@@ -0,0 +1,64 @@
...
+	args := []string{"create", setName, string(setType)}
+	if netMask != nil {
+		ones, _ := netMask.Size()
+		args = append(args, "--netmask", strconv.Itoa(ones))
```
soffokl: You are setting netmask `/24` for all IPs that added. This will allow exposing more IPs then expected. Just remove this argument.

firewall/ipset/operations.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+	if hashSize != 0 {
+		args = append(args, "--hashsize", strconv.Itoa(hashSize))
```
soffokl: What the point to limit this initial hash table size. The default value is 1024, but you set it for 64. I think we can keep the default value for sure.

firewall/incoming_blocker_iptables.go
```diff
@@ -0,0 +1,156 @@
...
+	"net"
+	"strings"
+
+	"github.com/mysteriumnetwork/node/firewall/ipset"
+	"github.com/mysteriumnetwork/node/firewall/iptables"
+	"github.com/rs/zerolog/log"
+)
+
+var (
+	dnsFirewallChain = "PROVIDER_DNS_FIREWALL"
```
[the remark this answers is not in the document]
Waldz: We already have custom chain `CONSUMER_KILL_SWITCH` and it's wo prefix

firewall/incoming_blocker_iptables.go
```diff
@@ -0,0 +1,156 @@
...
+func (ibi *incomingBlockerIptables) Setup() error {
+	if err := ibi.checkIpsetVersion(); err != nil {
+		return err
+	}
+
+	// Clean up setups from previous runs, just in case
+	if err := ibi.cleanupStaleRules(); err != nil {
+		return err
+	}
+	ipset.Exec(ipset.OpSetDelete(dnsFirewallIpset))
```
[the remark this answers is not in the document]
Waldz: there is logging inside of command

firewall/incoming_blocker_iptables.go
```diff
@@ -0,0 +1,156 @@
...
+func (ibi *incomingBlockerIptables) Teardown() {
+	if err := ibi.cleanupStaleRules(); err != nil {
+		log.Warn().Err(err).Msg("Error cleaning up iptables rules, you might want to do it yourself")
+	}
+	if errOutput, err := ipset.Exec(ipset.OpSetDelete(dnsFirewallIpset)); err != nil {
+		log.Warn().Err(err).Msgf("Error deleting ipset table. %s", strings.Join(errOutput, ""))
+	}
+}
+
+func (ibi *incomingBlockerIptables) BlockIncomingTraffic(network net.IPNet) (IncomingRuleRemove, error) {
```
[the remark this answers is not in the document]
Waldz: It's common interface for provider firewall - to block traffic incoming from consumers, so that's why not binding to naming of iptables here.

firewall/ipset/ipset.go
```diff
@@ -0,0 +1,46 @@
...
+	"github.com/pkg/errors"
+	"github.com/rs/zerolog/log"
+)
+
+// Exec activates given args
+var Exec = func(args []string) ([]string, error) {
+	args = append([]string{"sudo", "/usr/sbin/ipset"}, args...)
+
+	log.Debug().Msgf("[cmd] %v", args)
+	output, err := exec.Command("sudo", args...).CombinedOutput()
```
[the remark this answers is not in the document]
Waldz: Fixed sudo.
Don't see a point to reuse `cmdutil` util for every case, I need raw command output here

firewall/ipset/operations.go
Waldz: Improved

firewall/ipset/operations.go
Waldz: Do you know what for is hashsize parameter? http://ipset.netfilter.org/ipset.man.html#lbAN
It’s just initial rows in table, correct?

firewall/ipset/operations.go
Waldz: Not passing anymore

firewall/incoming_blocker_iptables.go
```diff
@@ -0,0 +1,156 @@
...
+	"strings"
+
+	"github.com/mysteriumnetwork/node/firewall/ipset"
+	"github.com/mysteriumnetwork/node/firewall/iptables"
+	"github.com/rs/zerolog/log"
+)
+
+var (
+	dnsFirewallChain = "PROVIDER_DNS_FIREWALL"
+	dnsFirewallIpset = "dns-firewall-whitelist"
```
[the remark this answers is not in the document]
Waldz: improved name

firewall/ipset/ipset.go
```diff
@@ -0,0 +1,46 @@
...
+	"bufio"
+	"bytes"
+	"os/exec"
+
+	"github.com/pkg/errors"
+	"github.com/rs/zerolog/log"
+)
+
+// Exec activates given args
+var Exec = func(args []string) ([]string, error) {
```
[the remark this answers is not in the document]
Waldz: It's a way of mocking, with some tradeoffs, don't see a problem here

firewall/incoming_blocker_iptables.go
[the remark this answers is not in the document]
tadaskay: I think prefix is a good idea since its scope is the whole system.

firewall/ipset/ipset.go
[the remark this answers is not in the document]
tadaskay: Is there a reason why you are not using cmdutil here? It does the same, except that the error logging is not unified.

firewall/ipset/operations.go
```diff
@@ -0,0 +1,64 @@
...
+// SetType defines type of IP set.
+type SetType string
+
+var (
+	// SetTypeHashIP set type uses a hash to store IP addresses where clashing is resolved by storing the clashing elements in an array and, as a last resort, by dynamically growing the hash.
+	SetTypeHashIP = SetType("hash:ip")
+)
+
+// OpVersion is an operation which prints version information.
+func OpVersion() []string {
```
tadaskay: Why need to prefix all functions with `Op`? Isn't that just noise?

firewall/incoming_blocker_iptables.go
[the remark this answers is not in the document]
Waldz: Should I migrate existing chain `CONSUMER_KILL_SWITCH` also?

firewall/ipset/ipset.go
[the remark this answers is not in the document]
Waldz: It doesn't do the same, I this place needs scanning command output line by line

firewall/ipset/operations.go
Waldz: Because it's not a functions, but rather factories for different operations

firewall/ipset/ipset.go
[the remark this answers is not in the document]
tadaskay: It's not related to scanning. You can perform the same scanning as below on a string

firewall/incoming_blocker_iptables.go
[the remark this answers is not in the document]
soffokl: Yes, it will be good to make everything in a common way.

firewall/ipset/operations.go
soffokl: It's still there. 
It looks like you didn't commit changes but wrote a comment that it's done.

firewall/ipset/operations.go
soffokl: Yes, I know. This is the initial size. 
What you are trying to achieve by setting it to non-default value?

firewall/ipset/ipset.go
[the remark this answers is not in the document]
soffokl: It just makes it more complicated to read.

firewall/ipset/ipset.go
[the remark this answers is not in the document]
soffokl: > Fixed sudo.

This is still `sudo sudo /usr/sbin/ipset`.

firewall/incoming_blocker_iptables.go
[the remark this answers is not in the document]
Waldz: done

firewall/ipset/ipset.go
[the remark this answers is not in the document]
Waldz: Improved readability

firewall/ipset/ipset.go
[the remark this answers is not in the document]
Waldz: Reused

firewall/ipset/operations.go
Waldz: I dont pass this optional option, but yes operation still supports it

utils/cmdutil/command.go
```diff
@@ -52,8 +52,5 @@ func ExecOutput(args ...string) (output string, err error) {
 	out, err := exec.Command(args[0], args[1:]...).CombinedOutput()
 	logSkipFrame := log.With().CallerWithSkipFrameCount(3).Logger()
 	(&logSkipFrame).Debug().Msgf("%q output:\n%s", strings.Join(args, " "), out)
-	if err != nil {
-		return string(out), errors.Errorf("%q: %v output: %s", strings.Join(args, " "), err, out)
-	}
 	return string(out), nil
```
soffokl:
```suggestion
	return string(out), err
```

Waldz: Ups, restored previous code

firewall/outgoing_blocker_default.go
```diff
@@ -27,35 +27,35 @@ const (
 )
 
 // DefaultTrackingBlocker traffic blocker bootstrapped for global calls
-var DefaultTrackingBlocker TrafficBlocker = &noopTrafficBlocker{}
+var DefaultTrackingBlocker OutgoingTrafficBlocker = &outgoingBlockerNoop{}
 
-// TrafficBlocker interface neededs to be satisfied by any implementations which provide firewall capabilities, like iptables
-type TrafficBlocker interface {
+// OutgoingTrafficBlocker interface neededs to be satisfied by any implementations which provide firewall capabilities, like iptables
```
tadaskay: `OutgoingTrafficBlocker` and `IncomingTrafficBlocker` have the same godoc, but obviously they are 2 different interfaces providing different functionality. I think you could explain what is the purpose of each, and also skip the `interface neededs to be satisfied by any implementations which provide` - because this is true for any interface.

firewall/iptables_exec_mock.go
```diff
@@ -0,0 +1,49 @@
...
+	called bool
+	output []string
+	err    error
+}
+
+type iptablesExecMock struct {
+	mocks map[string]iptablesExecResult
+}
+
+func (mce *iptablesExecMock) Exec(args ...string) ([]string, error) {
```
tadaskay: What does `mce` stand for?

firewall/outgoing_blocker_default.go
tadaskay: For future reference, godoc should be a complete sentence, notably ending with a dot.

bin/docker/alpine/Dockerfile
```diff
@@ -20,7 +20,7 @@ RUN bin/build
 FROM alpine:3.6
 
 # Install packages
-RUN apk add --update --no-cache iptables ca-certificates openvpn bash sudo openresolv
+RUN apk add --update --no-cache iptables ipset ca-certificates openvpn bash sudo openresolv
```
tadaskay: You are adding new dependencies thus it is necessary to do `[ci full]` build on this PR.

utils/cmdutil/command.go
soffokl: It's still not restored.

core/service/manager.go
```diff
@@ -112,6 +112,7 @@ func (manager *Manager) Start(providerID identity.Identity, serviceType string,
 	if len(policyIDs) > 0 {
 		policies := manager.policyOracle.Policies(policyIDs)
 		if err = manager.policyOracle.SubscribePolicies(policies, policyRules); err != nil {
+			log.Warn().Err(err).Msg("Cant find given access policies")
```
soffokl:
```suggestion
			log.Warn().Err(err).Msg("Can't find given access policies")
```

firewall/incoming_blocker_iptables.go
```diff
@@ -0,0 +1,156 @@
...
+func (ibi *incomingBlockerIptables) Setup() error {
+	if err := ibi.checkIpsetVersion(); err != nil {
+		return err
+	}
+
+	// Clean up setups from previous runs, just in case
+	if err := ibi.cleanupStaleRules(); err != nil {
+		return err
+	}
+	ipset.Exec(ipset.OpDelete(dnsFirewallIpset))
```
soffokl: Error handling is missing here.

core/service/manager.go
Waldz: fixed

firewall/incoming_blocker_iptables.go
Waldz: It's just not needed here

PR #1636, opened 2020-02-05
Run all e2e on high perf runners

tadaskay:
Trying to find an optimal setting. Increased high perf runner count.


soffokl:
It's pending a long now waiting for these runners :\
image

PR #1641, opened 2020-02-05
Check valid until & send transactor fee

### LINKED ISSUE #1435: Transactor fee endpoint response should have `valid_until` field
The problem is that ethereum and gas price is changing all the time, so transactor can't have fixed fee in tokens. This fee have to be dynamic. Node should know when fee will be updated and call for new fee when needed.

* [x] Transactor: Add dynamic pricing and `valid_until` field in fee endpoint response.
* [x] Node: call transactor for new fee during invoice issuing if known fee is already not valid (`valid_until` time have passed)

### LINKED ISSUE #1628: Send transactor fee when requesting accountant promise
After https://github.com/mysteriumnetwork/accountant/issues/70 is implemented make sure to set appropriate transactor fee when requesting a promise.

vkuznecovas:
Up till now the provider was not sending the settlement fee to accountant, which meant that the transactor would get no cut from the promises.

This means that consumer does not have to set a transactor fee in its promise, while we need to ask for one from accountant.

Fixes #1435
Fixes #1628

### DIFF
--- a/localnet/consumer.sh
+++ b/localnet/consumer.sh
@@ -17,7 +17,7 @@ exec /node/build/myst/myst \
   --transactor.channel-implementation=0x599d43715DF3070f83355D9D90AE62c159E62A75 \
   --transactor.registry-address=0xbe180c8CA53F280C7BE8669596fF7939d933AA10 \
   --accountant.accountant-id=0xf2e2c77D2e7207d8341106E6EfA469d1940FD0d8 \
-  --accountant.address=http://accountant:8889/api/v1 \
+  --accountant.address=http://accountant:8889/api/v2 \
   --transactor.address=http://transactor:8888/api/v1 \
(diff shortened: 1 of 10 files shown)


localnet/consumer.sh
```diff
@@ -17,7 +17,7 @@ exec /node/build/myst/myst \
   --transactor.channel-implementation=0x599d43715DF3070f83355D9D90AE62c159E62A75 \
   --transactor.registry-address=0xbe180c8CA53F280C7BE8669596fF7939d933AA10 \
   --accountant.accountant-id=0xf2e2c77D2e7207d8341106E6EfA469d1940FD0d8 \
-  --accountant.address=http://accountant:8889/api/v1 \
+  --accountant.address=http://accountant:8889/api/21 \
```
soffokl:
```suggestion
  --accountant.address=http://accountant:8889/api/2 \
```

PR #1651, opened 2020-02-06
Enable wireguard by default

soffokl:
Closes: https://github.com/mysteriumnetwork/node/issues/1502


soffokl:
@tadaskay Wireguard performance is much better than OpenVPN. It should be ok.

PR #1700, opened 2020-02-12
Apply DNS policy rules for incoming consumer traffic

Waldz:
Updates: #1597

While PR #1632 blocked all traffic on particular device if DNS policies exist, this puts in action whitelisting part:
- Implemented rule: block all & allow certain DNS hostnames
- Implemented rule: block all & allow certain DNS wildcard hostnames
- To whitelist traffic consumer must announce hostnames through it MUST use provider's DNS proxy. Otherwise all traffic is blocked by default


dns/proxy_test.go
```diff
@@ -0,0 +1,189 @@
...
+	}
+)
+
+func Test_Proxy_WhitelistsAllowedHostnames(t *testing.T) {
+	mockedBlocker := &trafficBlockerMock{
+		allowIPCalls: map[string]int{},
+	}
+	proxy := NewProxy("localhost", 10053, mockedBlocker, createPolicies())
+
+	err := proxy.whitelistByAnswer(&dns.Msg{
```
tadaskay: It can be simplified using a test table.

vkuznecovas:
Please re-create this on origin, forks do not get CI checks.

core/policy/repository.go
```diff
@@ -114,6 +115,54 @@ func (r *Repository) Rules() []market.AccessPolicyRuleSet {
...
+			if rule.Type == market.AccessPolicyTypeDNSHostname {
+				return true
+			}
+		}
+	}
+
+	return false
+}
+
+// IsHostAllowed return flag is given FQDN host should be allowed by rules
```
anjmao: `return flag is given` -> `return flag if given`

firewall/incoming_blocker_iptables_test.go
```diff
@@ -0,0 +1,152 @@
+/*
+ * Copyright (C) 2019 The "MysteriumNetwork/node" Authors.
```
anjmao: 2019 -> 2020

firewall/incoming_blocker_noop.go
```diff
@@ -0,0 +1,64 @@
+/*
+ * Copyright (C) 2019 The "MysteriumNetwork/node" Authors.
```
anjmao: 2019 -> 2020

dns/proxy.go
```diff
@@ -86,16 +100,48 @@ func (p *Proxy) proxyHandler() dns.Handler {
...
+	}
+	return nil
+}
+
+func (p *Proxy) whitelistByARecord(record *dns.A) error {
+	host := strings.TrimRight(record.Hdr.Name, ".")
+	ip := record.A
+
+	if p.policies.IsHostAllowed(host) {
+		_, err := p.trafficBlocker.AllowIPAccess(ip)
```
anjmao: Would be good to check how much this impacts performances and maybe some caching is needed. AllowIPAccess will shell out to ipset every time even if ip access is already added.

firewall/incoming_blocker_interface.go
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+package firewall
+
+import (
+	"net"
+)
+
+// IncomingTrafficBlocker interface neededs to be satisfied by any implementations which provide firewall capabilities, like iptables
+type IncomingTrafficBlocker interface {
```
anjmao: 1. Interface name is IncomingTrafficBlocker but it also contains AllowIPAccess which doesn't block but allow access. Seems that it could be split into separate interfaces.
2. It contains Setup and Teardown which does unrelated things like DNS firewall setup which could itself be separate structure.

Could be separate interfaces.

```go
type IncomingTrafficBlocker interface {
    Block(network net.IPNet) (IncomingRuleRemove, error)
}

type IncomingTrafficAllower interface {
    Allow(network net.IPNet) (IncomingRuleRemove, error)
}
```

When Setup and Teardown could be moved to DNS related logic and firewall package will be decoupled from DNS rules setup.

bin/docker/alpine/Dockerfile
```diff
@@ -20,7 +20,7 @@ RUN bin/build
 FROM alpine:3.6
 
 # Install packages
-RUN apk add --update --no-cache iptables ca-certificates openvpn bash sudo openresolv
+RUN apk add --update --no-cache iptables ipset ca-certificates openvpn bash sudo openresolv
```
anjmao: Does our ubuntu image contains ipset already?

dns/proxy_test.go
Waldz: added

dns/proxy.go
```diff
@@ -86,16 +100,48 @@ func (p *Proxy) proxyHandler() dns.Handler {
...
+			resp, _, err := client.Exchange(req, addr)
+			if err != nil {
 				log.Error().Err(err).Msg("Error proxying DNS query to " + addr)
-			} else {
-				writer.WriteMsg(resp)
-				return
+				continue
 			}
+
+			writer.WriteMsg(resp)
```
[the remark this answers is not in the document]
Waldz: added error message: `RcodeNameError      = 3  // NXDomain  - Non-Existent Domain               [DNS]`

core/policy/repository.go
```diff
@@ -133,6 +134,35 @@ func (r *Repository) HasDNSRules() bool {
 	return false
 }
 
+// IsHostAllowed return flag if	given FQDN host should be allowed by rules
```
soffokl:
```suggestion
// IsHostAllowed return flag if given FQDN host should be allowed by rules
```

core/policy/repository.go
```diff
@@ -133,6 +134,35 @@ func (r *Repository) HasDNSRules() bool {
...
+			if rule.Type == market.AccessPolicyTypeDNSHostname {
+				hasDNSRules = true
+				if host == rule.Value {
+					return true
+				}
+			}
+		}
+	}
+
+	if !hasDNSRules {
```
soffokl: > S1008: should use 'return <expr>' instead of 'if <expr> { return <bool> }; return <bool>' (gosimple)
```suggestion
	return !hasDNSRules
```

dns/handler_proxy.go
```diff
@@ -0,0 +1,71 @@
...
+
+func (ph *proxyHandler) ServeDNS(writer dns.ResponseWriter, req *dns.Msg) {
+	for _, addr := range ph.proxyAddrs {
+		resp, _, err := ph.client.Exchange(req, addr)
+		if err != nil {
+			log.Error().Err(err).Msg("Error proxying DNS query to " + addr)
+			continue
+		}
+
+		writer.WriteMsg(resp)
```
soffokl:
```suggestion
		writer.WriteMsg(resp)
		return
```

dns/handler_whitelist.go
```diff
@@ -0,0 +1,90 @@
...
+		return
+	}
+
+	writer.WriteMsg(resp)
+}
+
+func (wh *whitelistHandler) whitelistByAnswer(response *dns.Msg) error {
+	for _, record := range response.Answer {
+		switch recordValue := record.(type) {
+		case *dns.A:
```
soffokl: Is it enough to handle only A record? What about MX, CNAME and other?

Waldz: I don't known yet.. @zolia, @soffokl this is exactly feedback i'm expecting from you. Currently only A resolves are whitelisted for later packets.

bin/docker/alpine/Dockerfile
Waldz: Comment addressed in #1632. Added .deb dependency.

core/policy/repository.go
Waldz: simplified

firewall/incoming_blocker_interface.go
Waldz: traffic blocker, neither blocks neither allows, it's interface to control how blocking rules are applied. If blocker suffix triggers you, i can rename it to IncomingTrafficFirewall :)

> When Setup and Teardown could be moved to DNS related logic and firewall package will be decoupled from DNS rules setup.

i can't add there. Setup needs to be initiated at very beginning of all system bootstrap

dns/handler_whitelist.go
```diff
@@ -0,0 +1,90 @@
...
+type whitelistHandler struct {
+	resolver       dns.Handler
+	trafficBlocker firewall.IncomingTrafficBlocker
+	policies       *policy.Repository
+}
+
+func (wh *whitelistHandler) ServeDNS(writer dns.ResponseWriter, req *dns.Msg) {
+	resolverWriter := &recordingWriter{writer: writer}
+	wh.resolver.ServeDNS(resolverWriter, req)
+	resp := resolverWriter.responseMsg
```
tadaskay: You should check the result of DNS resolution. Is there a point in whitelisting if DNS lookup fails, i.e. `resp.Rcode != dns.RcodeSuccess`?

dns/handler_whitelist_test.go
```diff
@@ -0,0 +1,187 @@
...
+			},
+			map[string]int{
+				"0.0.0.6": 1,
+				"0.0.0.7": 1,
+			},
+			"should allow zone of whitelisted wildcard hostname",
+		},
+	}
+
+	for _, test := range tests {
```
tadaskay: Do parameterized test cases appear properly named in test runners, e.g. IDE?
Usually, when doing data-driven tests, they come in the form of:

```
	for _, tt := range tests {
		t.Run(tt.name, func(t *testing.T) {
```

dns/handler_whitelist_test.go
```diff
@@ -0,0 +1,187 @@
...
+						Hdr: dns.RR_Header{Name: "wildcard.com.", Rrtype: dns.TypeA, Class: dns.ClassINET, Ttl: 0},
+						A:   net.ParseIP("0.0.0.9"),
+					},
+				},
+			},
+			map[string]int{
+				"0.0.0.8": 1,
+				"0.0.0.9": 1,
+			},
+			"should allow whitelisted wildcard hostname",
```
tadaskay: It would be clearer if the name were the first field, naming the test case. It's not simply a message.

firewall/incoming_blocker_interface.go
tadaskay: _Firewall_ is a good name.

dns/handler_whitelist.go
Waldz: DNS resolution answers are traversed here, so errors don't have them.

firewall/incoming_blocker_interface.go
Waldz: Renamed

dns/handler_whitelist_test.go
Waldz: Improved

dns/handler_whitelist_test.go
Waldz: Added

services/wireguard/service/service_unix.go
```diff
@@ -309,12 +312,17 @@ func (m *Manager) Serve(instance *service.Instance) error {
...
-	m.dnsProxy = dns.NewProxy("", m.dnsPort)
-	if err := m.dnsProxy.Run(); err != nil {
-		log.Warn().Err(err).Msg("Provider DNS will not be available")
+	dnsHandler, err := dns.ResolveViaSystem()
+	if err == nil {
+		m.dnsProxy = dns.NewProxy("", m.dnsPort, dnsHandler)
+		if err := m.dnsProxy.Run(); err != nil {
+			log.Warn().Err(err).Msg("Provider DNS will not be available")
+		} else {
+			// m.dnsProxy = dnsProxy
```
soffokl: Leftover comment

Waldz:
Repeating PR in #1733

PR #1708, opened 2020-02-13
Payments based on bytes + duration

### LINKED ISSUE #1668: Don't allow consumers who are not supporting payments
At the moment our provider node is accepting both paying and free (with version until 0.17.0) consumer apps. After 3 months of testing of the payment system, we should disable support for old (not paying consumer) apps.

### LINKED ISSUE #1687: Payments based on traffic calculations
Pingpong should be able to work with time + bytes.

### LINKED ISSUE #1689: Consumer side byte count events in intervals
We need configurable byte transferred reporting frequency. Initially we might assume that reporting every 3 sec. should be enough. Given that payment invoicing frequency is 60 sec. atm. this should yield around 5% discrepancy.

relates: #1686

vkuznecovas:
Unified payment method in proposals. This is backwards incompatible - new consumers will not see old proposals. New providers will not emit old proposals. Old consumers will only see old providers' proposals.

Ensured that consumers will not see proposals above 0.07myst/GB or 0.0005myst/min.

Added configuration flags for wireguard and openvpn consumers to configure the frequency of byte statistics reporting. Set them to a default of 2 seconds.

Extracted payment calculations to a separate function and used that in both invoice tracking and exchange message tracking.

Disabled compatibility tests due to the backwards incompatible change introduced in this PR.

Removed a ton of redundant code.

Removed old packages:

`core/promise`
`session/balance`
`session/payment`
`session/promise`

Closes #1689 
Closes #1687
Closes #1668


core/discovery/reducer/proposal.go
```diff
@@ -63,14 +66,49 @@ func LocationType(proposal market.ServiceProposal) interface{} {
...
-// Price checks if the price is below the given value
-func Price(lowerBound, upperBound uint64) func(market.ServiceProposal) bool {
+// PriceMinute checks if the price per minute is below the given value
+func PriceMinute(lowerBound, upperBound uint64) func(market.ServiceProposal) bool {
+	return pricePerTime(lowerBound, upperBound, time.Minute)
+}
+
+const bytesInGigabyte = 1000000000
+
+// PriceGB checks if the price per GV is below the given value
```
Waldz: GV -> GB

mobile/mysterium/proposals_manager.go
```diff
@@ -83,24 +83,30 @@ func newProposalsManager(
...
-	cache           []market.ServiceProposal
-	mysteriumAPI    mysteriumAPI
-	qualityFinder   qualityFinder
-	upperPriceBound uint64
-	lowerPriceBound uint64
+	repository          proposal.Repository
+	cache               []market.ServiceProposal
+	mysteriumAPI        mysteriumAPI
+	qualityFinder       qualityFinder
+	upperTimePriceBound uint64
```
Waldz: I suggest to inject `proposal.Filter` instead of proxying every filter variable

services/openvpn/bootstrap.go
```diff
@@ -40,19 +41,9 @@ func Bootstrap() {
 	)
 
 	market.RegisterPaymentMethodUnserializer(
-		dto_openvpn.PaymentMethodPerTime,
+		pingpong.DefaultPaymentMethod.Type,
```
Waldz: Seems like you're registering same payment method unserialised several times, while only one is needed

session/pingpong/exchange_message_tracker.go
```diff
@@ -258,6 +257,46 @@ func (emt *ExchangeMessageTracker) issueExchangeMessage(invoice crypto.Invoice)
...
+	// This way, the session will show that it downloaded the bytes in a manner that is easier to comprehend.
+	emt.updateDataTransfer(e.Stats.BytesSent, e.Stats.BytesReceived)
+}
+
+func (emt *ExchangeMessageTracker) updateDataTransfer(up, down uint64) {
+	emt.dataTransferedLock.Lock()
+	defer emt.dataTransferedLock.Unlock()
+
+	newUp := emt.dataTransfered.up
+	if up > emt.dataTransfered.up {
```
Waldz: Why can't you trust data coming in? If there is bugs, should they be fixes

session/pingpong/factory.go
```diff
@@ -24,27 +24,54 @@ import (
...
-// DefaultPaymentInfo represents the default payment info for the alpha release
-var DefaultPaymentInfo = dto.PaymentRate{
+// DefaultPaymentMethod represents the the default payment method of time + bytes.
+// The rate is frozen at 0.07MYSTT per GB of data transfered and 0.0005MYSTT/minute.
+var DefaultPaymentMethod = PaymentMethod{
 	Price:    money.NewMoney(50000, money.CurrencyMyst),
-	Duration: 1 * time.Minute,
+	Duration: time.Minute,
+	Type:     "BYTES_TRANSFERED_WITH_TIME",
+	Bytes:    7142857,
```
Waldz: Not clear what kind of number is here. And I don't see expression of 0.07MYSTT per GB

vkuznecovas: added explanation

session/pingpong/exchange_message_tracker.go
vkuznecovas: never trust anyone, if data starts decreasing we'll end up either way be notified by the desync of the provider.

services/openvpn/bootstrap.go
vkuznecovas: removed redundant ones

mobile/mysterium/proposals_manager.go
vkuznecovas: done

config/flags_payments.go
```diff
@@ -65,16 +65,28 @@ var (
...
 		Usage: "Disables payments and moves to a backwards compatible legacy mode",
 	}
-	// FlagPaymentsConsumerUpperPriceBound sets the upper price bound to a set value.
-	FlagPaymentsConsumerUpperPriceBound = cli.Uint64Flag{
-		Name:  "payments.consumer.price.upper.bound",
-		Usage: "Sets the maximum price of the service. All proposals with a price above this bound will be filtered out and not visible.",
-		Value: 1000000,
+	// FlagPaymentsConsumerPricePerMinuteUpperBound sets the upper price bound per minute to a set value.
+	FlagPaymentsConsumerPricePerMinuteUpperBound = cli.Uint64Flag{
+		Name:  "payments.consumer.price.perminute.upper.bound",
```
tadaskay: No need to separate every word with a dot. It only makes sense if you are grouping something. Instead of 6 level hierarchy, it could be smth like:

```suggestion
		Name:  "payments.consumer.price-perminute-max,
```
(also replaced upper bound by _max_)

config/flags_payments.go
```diff
@@ -65,16 +65,28 @@ var (
...
+		Usage: "Sets the maximum price of the service per minute. All proposals with a price above this bound will be filtered out and not visible.",
+		Value: 50000,
 	}
-	// FlagPaymentsConsumerLowerPriceBound sets the lower price bound to a set value.
-	FlagPaymentsConsumerLowerPriceBound = cli.Uint64Flag{
-		Name:  "payments.consumer.price.lower.bound",
-		Usage: "Sets the minimum price of the service. All proposals with a below above this bound will be filtered out and not visible.",
+	// FlagPaymentsConsumerPricePerMinuteLowerBound sets the lower price bound per minute to a set value.
+	FlagPaymentsConsumerPricePerMinuteLowerBound = cli.Uint64Flag{
+		Name:  "payments.consumer.price.perminute.lower.bound",
```
tadaskay: Similarly, 
```suggestion
		Name:  "payments.consumer.price-perminute-min",
```
And so on.

core/discovery/proposal/filter_test.go
```diff
@@ -19,12 +19,15 @@ package proposal
...
 import (
 	"testing"
+	"time"
 
 	"github.com/mysteriumnetwork/node/market"
 	"github.com/mysteriumnetwork/node/money"
 	"github.com/stretchr/testify/assert"
 )
 
+const bytesInGigabyte = 1000000000
```
tadaskay: I also love zeros, but:
```suggestion
const bytesInGigabyte = 1_000_000_000
```

Also, I don't get the discrepancy between this and `datasize.BitSize`?

vkuznecovas: bitsize is exactly this - bits, we're not counting bits here.

tadaskay: I was referring to the logic in the given package, where binary conversion is used instead of decimal SI:
`Kilobyte = 1024 * Byte`

tadaskay: What I meant is you should either use it or change it, but don't create 2 different versions of data counting - it will create discrepancies if data is interpreted in different ways throughout the app.

tadaskay: We can change this in a separate PR.

zolia: Calculation should be more accurate. There is some notion where you specify kilo as 1000, but you really imply 1024. That is, in reality you do all calculations with 1024 (cisco style). But i suggest to always use 1024 as  a kilo.. Ppl these days are smart enough to understand power of 2 calculations.

vkuznecovas: moved to GiB instead of GB

Issue #1665, opened 2020-02-07
Flaky tests: wireguard connection test


vkuznecovas:
This will occasionally fail on CI

```
{"level":"info","time":"2020-02-07T12:00:54+01:00","message":"Allow IP 127.0.0.1 access"}
 {"level":"info","time":"2020-02-07T12:00:54+01:00","message":"Starting new connection"}
 {"level":"info","time":"2020-02-07T12:00:54+01:00","message":"Starting connection endpoint"}
 {"level":"info","time":"2020-02-07T12:00:54+01:00","message":"Adding connection peer"}
 {"level":"info","time":"2020-02-07T12:00:54+01:00","message":"Configuring routes"}
 {"level":"info","time":"2020-02-07T12:00:54+01:00","message":"Waiting for initial handshake"}
 {"level":"debug","time":"2020-02-07T12:00:54+01:00","message":"Selecting DNS servers using strategy: 1.2.3.4"}
 {"level":"info","time":"2020-02-07T12:00:54+01:00","message":"Stopping WireGuard connection"}
 coverage: 46.2% of statements
 panic: test timed out after 3m0s
 goroutine 15 [running]:
 testing.(*M).startAlarm.func1()
 	/usr/lib/go-1.13/src/testing/testing.go:1377 +0x11c
 created by time.goFunc
 	/usr/lib/go-1.13/src/time/sleep.go:168 +0x52
 goroutine 1 [chan receive]:
 testing.(*T).Run(0xc00018e500, 0x12298b5, 0x17, 0x127e3d8, 0x0)
 	/usr/lib/go-1.13/src/testing/testing.go:961 +0x68a
 testing.runTests.func1(0xc00018e500)
 	/usr/lib/go-1.13/src/testing/testing.go:1202 +0xa7
 testing.tRunner(0xc00018e500, 0xc000283cd8)
 	/usr/lib/go-1.13/src/testing/testing.go:909 +0x19a
 testing.runTests(0xc00039c220, 0x1c5b2a0, 0x3, 0x3, 0x0)
 	/usr/lib/go-1.13/src/testing/testing.go:1200 +0x522
 testing.(*M).Run(0xc00018d080, 0x0)
 	/usr/lib/go-1.13/src/testing/testing.go:1117 +0x300
 main.main()
 	_testmain.go:102 +0x338
 goroutine 20 [chan receive]:
[...]

PR #1709, opened 2020-02-13
Increase stats send interval in unit tests 

anjmao:
In unit tests stats check period is very small (1 micro second). Sometimes stats channel can become full too fast and Stop will block forever since it tried to send last stats.

Fixes https://github.com/mysteriumnetwork/node/issues/1665

### DIFF
--- a/e2e/connection_test.go
+++ b/e2e/connection_test.go
@@ -157,6 +157,9 @@ func consumerConnectFlow(t *testing.T, tequilapi *tequilapi_client.Client, consu
 	assert.Equal(t, connectionStatus.SessionID, se.SessionID)
 	assert.Equal(t, "New", se.Status)
 
+	// Wait some time for session to collect stats.
+	time.Sleep(2 * time.Second)
+
 	err = tequilapi.ConnectionDestroy()
 	assert.NoError(t, err)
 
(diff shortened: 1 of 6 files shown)


e2e/connection_test.go
```diff
@@ -157,6 +157,9 @@ func consumerConnectFlow(t *testing.T, tequilapi *tequilapi_client.Client, consu
 	assert.Equal(t, connectionStatus.SessionID, se.SessionID)
 	assert.Equal(t, "New", se.Status)
 
+	// Wait some time for session to collect stats.
+	time.Sleep(2 * time.Second)
```
anjmao: @vkuznecovas After this change NAT e2e test started to fail with error

```
core/connection/manager.go:468           > Payment error error="could not calculate amount to promise: could not request consumer data accountant: could not execute request: Get http://accountant:8889/api/v2/data/consumer/0x7ab7f1f3838b94ca5dd539c3c0836c76c6f9ee9d: net/http: request canceled (Client.Timeout exceeded while awaiting headers)"
```

I see that you are using new http client instance https://github.com/mysteriumnetwork/node/blob/master/cmd/di.go#L497 instead of di.HttpClient. In such case this http client is not reconnected as implemented in https://github.com/mysteriumnetwork/node/blob/master/cmd/di.go#L873

Do you need such low timeout or can I change it to use global HTTPClient ?( which will also work more stable)

vkuznecovas: I think we can change the client.

Issue #1686, opened 2020-02-12
Add byte count on the provider side


vkuznecovas:
Make sure that byte counts are emitted for ongoing sessions.

zolia:
We will need to update consumer side too since we will need configurable reporting frequency. Initially we might assume that reporting every 3 sec. should be enough. Given that payment invoicing frequency is 60 sec. atm. this should yield around 5% discrepancy.

zolia:
Most likely we will need to address this bug too: #1674

tadaskay:
> We will need to update consumer side too since we will need configurable reporting frequency
I suggest opening a separate issue for that

PR #1710, opened 2020-02-13
Emit wireguard session statistics (provider side)

tadaskay:
Fixes: #1686

Gratis: some extra fixes, most notably - wireguard session cleanup if provider stops the service


mocks/mock_publisher.go
```diff
@@ -0,0 +1,59 @@
...
+	return nil
+}
+
+// Subscribe fakes subscribe.
+func (mp *EventBus) Subscribe(topic string, fn interface{}) error {
+	return nil
+}
+
+// Pop pops the last event for assertions.
+func (mp *EventBus) Pop() interface{} {
```
anjmao: If it called Pop maybe Publish should add to slice and then Pop will remove only last item. Just thinking about case there code published multiple times and you want to assert all published events.

services/wireguard/service/stats_publisher.go
```diff
@@ -0,0 +1,71 @@
...
+	return statsPublisher{
+		done:      make(chan struct{}),
+		bus:       bus,
+		frequency: frequency,
+	}
+}
+
+func (s statsPublisher) start(sessionID string, supplier statsSupplier) {
+	for {
+		tick := time.After(s.frequency)
```
anjmao: Could be more clear if `time.After(s.frequency)` is used inline in case block.

services/wireguard/service/service_unix.go
```diff
@@ -299,23 +317,36 @@ func (m *Manager) Serve(instance *service.Instance) error {
...
-	m.dnsProxyMu.Lock()
-	defer m.dnsProxyMu.Unlock()
 	if m.dnsProxy != nil {
 		if err := m.dnsProxy.Stop(); err != nil {
 			log.Error().Err(err).Msg("Failed to stop DNS server")
 		}
 	}
 
-	log.Info().Msg("Wireguard service stopped")
+	m.startStopMu.Unlock()
```
anjmao: If possible use `defer m.startStopMu.Unlock()` for bigger code blocks.

services/wireguard/service/service_unix.go
```diff
@@ -175,7 +180,17 @@ func (m *Manager) ProvideConfig(sessionConfig json.RawMessage) (*session.ConfigP
 		return nil, errors.Wrap(err, "failed to setup NAT/firewall rules")
 	}
 
+	statsPublisher := newStatsPublisher(m.publisher, 3*time.Second)
```
anjmao: In https://github.com/mysteriumnetwork/node/pull/1708 stats update interval is 2 seconds. Not sure if this is related.

vkuznecovas: We can have more frequent emitting on consumer side, as that is only tracking a single session. The more frequent the interval, the more accurate the payments.

PR #1722, opened 2020-02-17
Polish

tadaskay:


datasize/bitsize.go
```diff
@@ -34,35 +34,35 @@ const (
...
-	Petabyte = 1024 * Terabyte
-	// PB is short for Petabyte
-	PB = Petabyte
-
-	// Exabyte represents 1024 petabytes
-	Exabyte = 1024 * Petabyte
-	// EB is short for Exabyte
-	EB = Exabyte
+	// Kibibyte represents 1024 bytes
+	Kibibyte = 1024 * Byte
```
zolia: I think this will be yet another confusion from user side. From pure scientific notation maybe you are right, from majority of users - confusion. I would leave Kilobyte (KB) == 1024.

vkuznecovas: I'd rather leave it as kibibyte, the consumer will not be digging in code.

tadaskay: Example from docker: https://github.com/docker/go-units/blob/master/size.go

tadaskay: In user-facing UIs we may use approximations if we see fit, but for the core code, we should call things as they are.

Issue #1674, opened 2020-02-10
Insane speed reports


zolia:
Right after connect: 

```
2020-02-10T11:59:32.097 INF communication/nats/connector.go:93       > Re-establishing broker connection 8ebde9ce-0d84-4b1f-a470-fcfa4531f74d DONE (check result=nats: timeout)
2020-02-10T11:59:32.097 INF communication/nats/connector.go:93       > Re-establishing broker connection cec06223-7de5-46bb-b938-8d05571bf9e2 DONE (check result=nats: timeout)
2020-02-10T11:59:32.097 INF communication/nats/connector.go:97       > All broker connections re-established
2020-02-10T11:59:32.097 INF core/connection/manager.go:435           > Connection state: Connecting → Connected
2020-02-10T11:59:32.097 DBG consumer/bandwidth/tracker.go:87         > Download speed: 33.3 kbps
2020-02-10T11:59:32.097 DBG consumer/bandwidth/tracker.go:88         > Upload speed: 40.5 kbps
2020-02-10T11:59:32.098 DBG consumer/bandwidth/tracker.go:87         > Download speed: 834.9 Mbps
2020-02-10T11:59:32.098 DBG consumer/bandwidth/tracker.go:88         > Upload speed: 842.2 Mbps
2020-02-10T11:59:32.098 DBG consumer/bandwidth/tracker.go:87         > Download speed: -9223372036854775808 bps
2020-02-10T11:59:32.098 DBG consumer/bandwidth/tracker.go:88         > Upload speed: -9223372036854775808 bps
2020-02-10T11:59:33.028 DBG consumer/bandwidth/tracker.go:87         > Download speed: 414.0 kbps
2020-02-10T11:59:33.029 DBG consumer/bandwidth/tracker.go:88         > Upload speed: 300.2 kbps
```

But data transferred statistics looks sane: 

```
[INFO] Location: Occitanie, FR (residential - Free SAS)
[INFO] Proposal: Id: 1 , Provider: 0xf0160d53dfbb7ce040791192b5c838b4762bbd29, Country: FR
[INFO] Balance: 12397903822
[INFO] Connection duration: 261s
[INFO] Bytes sent: 1193304
[INFO] Bytes received: 3825150
```

soffokl:
```
2020-02-10T11:59:32.098 DBG consumer/bandwidth/tracker.go:87         > Download speed: -9223372036854775808 bps
[...]

PR #1726, opened 2020-02-17
Fix insane speed reports

tadaskay:
Fixes: #1674 

I've got some more fixes in mind, but since this is going to be backported to 0.21, keeping it to the minimum on this PR. Will do the rest of the fixes separately.


tadaskay:
Applied to `release/0.21` as f1a14132c1c496eb1e6fb861a36c6102fc8e0f07

PR #1727, opened 2020-02-18
TTL based NAT hole punching with parallel pings

soffokl:
This TTL based NAT hole punching tries to open 10 connections with different ports at the same time on a provider side. It sends ping with different TTL to every connection(TTL from 2 to 11). 

The consumer at that moment trying to send a ping to every connection. First delivered packed mean that NAT hole punched and we continue using this connection.

What is missing:
- [ ] NAT hole punching from the consumer if provider side NAT hole punching failed (or parallel)
- [ ] Test mobile implementation
- [ ] Windows provider support?

Closes: #1718

### DIFF
--- a/mobile/mysterium/openvpn_connection_setup.go
+++ b/mobile/mysterium/openvpn_connection_setup.go
@@ -19,6 +19,8 @@ package mysterium
 
 import (
 	"encoding/json"
+	"net"
+	"strconv"
 	"sync"
 	"time"
 
@@ -38,7 +40,7 @@ type natPinger interface {
 	SetProtectSocketCallback(SocketProtect func(socket int) bool)
 }
 
-type openvpn3SessionFactory func(connection.ConnectOptions) (*openvpn3.Session, *openvpn.ClientConfig, error)
+type openvpn3SessionFactory func(connection.ConnectOptions, *openvpn.VPNConfig) (*openvpn3.Session, *openvpn.ClientConfig, error)
 
 var errSessionWrapperNotStarted = errors.New("session wrapper not started")
 
@@ -50,21 +52,12 @@ func NewOpenVPNConnection(sessionTracker *sessionTracker, signerFactory identity
 		ipResolver: ipResolver,
 		pingerStop: make(chan struct{}),
 	}
-	sessionFactory := func(options connection.ConnectOptions) (*openvpn3.Session, *openvpn.ClientConfig, error) {
-		sessionConfig := &openvpn.VPNConfig{}
-		err := json.Unmarshal(options.SessionConfig, sessionConfig)
-		if err != nil {
-			return nil, nil, err
-		}
-
+	sessionFactory := func(options connection.ConnectOptions, sessionConfig *openvpn.VPNConfig) (*openvpn3.Session, *openvpn.ClientConfig, error) {
 		// override vpnClientConfig params with proxy local IP and pinger port
 		// do this only if connecting to natted provider
 		if sessionConfig.LocalPort > 0 {
 			sessionConfig.OriginalRemoteIP = sessionConfig.RemoteIP
 			sessionConfig.OriginalRemotePort = sessionConfig.RemotePort
-			sessionConfig.RemoteIP = "127.0.0.1"
-			// TODO: randomize this too?
-			sessionConfig.RemotePort = sessionConfig.LocalPort + 1
 		}
 
 		vpnClientConfig, err := openvpn.NewClientConfigFromSession(sessionConfig, "", "", connection.DNSOptionAuto)
@@ -106,7 +99,7 @@ func NewOpenVPNConnection(sessionTracker *sessionTracker, signerFactory identity
 }
 
 type openvpnConnection struct {
-	pingerStop    chan struct{}
+	ports         []int
 	stateCh       chan connection.State
 	stats         connection.Statistics
 	statsMu       sync.RWMutex
@@ -159,24 +152,48 @@ func (c *openvpnConnection) Statistics() (connection.Statistics, error) {
 }
 
 func (c *openvpnConnection) Start(options connection.ConnectOptions) error {
-	newSession, clientConfig, err := c.createSession(options)
+	sessionConfig := &openvpn.VPNConfig{}
+	err := json.Unmarshal(options.SessionConfig, sessionConfig)
 	if err != nil {
 		return err
 	}
 
-	log.Info().Msgf("Client config after session create: %v", clientConfig)
-	if clientConfig.LocalPort > 0 {
-		err := c.natPinger.PingProvider(
-			clientConfig.VpnConfig.OriginalRemoteIP,
-			clientConfig.VpnConfig.OriginalRemotePort,
-			clientConfig.LocalPort,
-			clientConfig.LocalPort+1,
-			c.pingerStop,
+	if sessionConfig.LocalPort == 0 && len(sessionConfig.Ports) > 0 {
+		if len(sessionConfig.Ports) == 0 || len(c.ports) == 0 {
+			c.ports = []int{sessionConfig.LocalPort}
+			sessionConfig.Ports = []int{sessionConfig.RemotePort}
+		}
+
+		conn, err := c.natPinger.PingProvider(
+			sessionConfig.RemoteIP,
+			c.ports,
+			sessionConfig.Ports,
+			sessionConfig.LocalPort,
 		)
 		if err != nil {
 			return err
 		}
+
+		_, lPort, err := net.SplitHostPort(conn.LocalAddr().String())
+		if err != nil {
+			return err
+		}
+
+		_, rPort, err := net.SplitHostPort(conn.RemoteAddr().String())
+		if err != nil {
+			return err
+		}
+
+		sessionConfig.LocalPort, _ = strconv.Atoi(lPort)
+		sessionConfig.RemotePort, _ = strconv.Atoi(rPort)
 	}
+
+	newSession, clientConfig, err := c.createSession(options, sessionConfig)
+	if err != nil {
+		log.Info().Err(err).Msg("Client config factory error")
+		return err
+	}
+	log.Info().Interface("data", clientConfig).Msgf("Openvpn client configuration")
 
 	c.session = newSession
 	c.session.Start()
@@ -188,8 +205,6 @@ func (c *openvpnConnection) Stop() {
 		if c.session != nil {
 			c.session.Stop()
 		}
-		log.Info().Msg("Stopping NATProxy")
-		close(c.pingerStop)
 	})
 }
 
--- a/mobile/mysterium/openvpn_connection_setup.go
+++ b/mobile/mysterium/openvpn_connection_setup.go
@@ -20,7 +20,6 @@ package mysterium
 import (
 	"encoding/json"
 	"net"
-	"strconv"
 	"sync"
 	"time"
 
@@ -158,34 +157,31 @@ func (c *openvpnConnection) Start(options connection.ConnectOptions) error {
 		return err
 	}
 
-	if sessionConfig.LocalPort == 0 && len(sessionConfig.Ports) > 0 {
-		if len(sessionConfig.Ports) == 0 || len(c.ports) == 0 {
+	// TODO this backward compatibility check needs to be removed once we will start using port ranges for all peers.
+	if sessionConfig.LocalPort > 0 || len(sessionConfig.Ports) > 0 {
+		if len(c.ports) == 0 {
 			c.ports = []int{sessionConfig.LocalPort}
 			sessionConfig.Ports = []int{sessionConfig.RemotePort}
 		}
 
-		conn, err := c.natPinger.PingProvider(
-			sessionConfig.RemoteIP,
-			c.ports,
-			sessionConfig.Ports,
-			sessionConfig.LocalPort,
-		)
+		params := traversal.Params{
+			IP:          sessionConfig.RemoteIP,
+			LocalPorts:  c.ports,
+			RemotePorts: sessionConfig.Ports,
+		}
+
+		conn, err := c.natPinger.PingProvider(params, sessionConfig.LocalPort)
 		if err != nil {
 			return err
 		}
 
-		_, lPort, err := net.SplitHostPort(conn.LocalAddr().String())
-		if err != nil {
-			return err
+		if addr, ok := conn.LocalAddr().(*net.UDPAddr); ok {
+			sessionConfig.LocalPort = addr.Port
 		}
 
-		_, rPort, err := net.SplitHostPort(conn.RemoteAddr().String())
-		if err != nil {
-			return err
+		if addr, ok := conn.RemoteAddr().(*net.UDPAddr); ok {
+			sessionConfig.RemotePort = addr.Port
 		}
-
-		sessionConfig.LocalPort, _ = strconv.Atoi(lPort)
-		sessionConfig.RemotePort, _ = strconv.Atoi(rPort)
 	}
 
 	newSession, clientConfig, err := c.createSession(options, sessionConfig)
--- a/mobile/mysterium/openvpn_connection_setup.go
+++ b/mobile/mysterium/openvpn_connection_setup.go
@@ -39,7 +39,7 @@ type natPinger interface {
 	SetProtectSocketCallback(SocketProtect func(socket int) bool)
 }
 
-type openvpn3SessionFactory func(connection.ConnectOptions, *openvpn.VPNConfig) (*openvpn3.Session, *openvpn.ClientConfig, error)
+type openvpn3SessionFactory func(connection.ConnectOptions, openvpn.VPNConfig) (*openvpn3.Session, *openvpn.ClientConfig, error)
 
 var errSessionWrapperNotStarted = errors.New("session wrapper not started")
 
@@ -51,14 +51,8 @@ func NewOpenVPNConnection(sessionTracker *sessionTracker, signerFactory identity
 		ipResolver: ipResolver,
 		pingerStop: make(chan struct{}),
 	}
-	sessionFactory := func(options connection.ConnectOptions, sessionConfig *openvpn.VPNConfig) (*openvpn3.Session, *openvpn.ClientConfig, error) {
-		// override vpnClientConfig params with proxy local IP and pinger port
-		// do this only if connecting to natted provider
-		if sessionConfig.LocalPort > 0 {
-			sessionConfig.OriginalRemoteIP = sessionConfig.RemoteIP
-			sessionConfig.OriginalRemotePort = sessionConfig.RemotePort
-		}
 
+	sessionFactory := func(options connection.ConnectOptions, sessionConfig openvpn.VPNConfig) (*openvpn3.Session, *openvpn.ClientConfig, error) {
 		vpnClientConfig, err := openvpn.NewClientConfigFromSession(sessionConfig, "", "", connection.DNSOptionAuto)
 		if err != nil {
 			return nil, nil, err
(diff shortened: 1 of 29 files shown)


mobile/mysterium/openvpn_connection_setup.go
```diff
@@ -128,24 +118,48 @@ func (c *openvpnConnection) Statistics() <-chan consumer.SessionStatistics {
...
-	log.Info().Msgf("Client config after session create: %v", clientConfig)
-	if clientConfig.LocalPort > 0 {
-		err := c.natPinger.PingProvider(
-			clientConfig.VpnConfig.OriginalRemoteIP,
-			clientConfig.VpnConfig.OriginalRemotePort,
-			clientConfig.LocalPort,
-			clientConfig.LocalPort+1,
-			c.pingerStop,
+	if sessionConfig.LocalPort == 0 && len(sessionConfig.Ports) > 0 {
+		if len(sessionConfig.Ports) == 0 || len(c.ports) == 0 {
```
anjmao: It is not possible that `sessionConfig.Ports` could be 0  here since this branch is executed only if `len(sessionConfig.Ports) > 0`

nat/traversal/pinger.go
```diff
@@ -312,36 +269,66 @@ func (p *Pinger) pingTargetConsumer(pingParams *Params) {
...
+	ch := make(chan res, len(localPorts))
+
+	for i := range localPorts {
+		go func(i int) {
+			conn, err := p.singlePing(remoteIP, localPorts[i], remotePorts[i], initialTTL+i, stop)
+			ch <- res{conn, err}
+		}(i)
+	}
+
+	// First responce wins. Other are not important.
```
zolia: First might be connection establishment failure, then others might be important.

mobile/mysterium/openvpn_connection_setup.go
```diff
@@ -128,24 +118,48 @@ func (c *openvpnConnection) Statistics() <-chan consumer.SessionStatistics {
...
+			sessionConfig.RemoteIP,
+			c.ports,
+			sessionConfig.Ports,
+			sessionConfig.LocalPort,
 		)
 		if err != nil {
 			return err
 		}
+
+		_, lPort, err := net.SplitHostPort(conn.LocalAddr().String())
```
anjmao: I believe you can get port of net.Addr without parsing string. And when no need for conversion from string to int.
```go
if addr, ok := conn.LocalAddr().(*net.UDPAddr); ok {
     addr.Port
}
```

mobile/mysterium/wireguard_connection_setup.go
```diff
@@ -120,17 +121,33 @@ func (c *wireguardConnection) Start(options connection.ConnectOptions) (err erro
...
+			config.Ports,
 			0,
-			c.pingerStop,
 		)
 		if err != nil {
 			return errors.Wrap(err, "could not ping provider")
 		}
+
+		if conn != nil {
+			_, lPort, err := net.SplitHostPort(conn.LocalAddr().String())
```
anjmao: Same here for port parsing.

mobile/mysterium/wireguard_connection_setup.go
```diff
@@ -120,17 +121,33 @@ func (c *wireguardConnection) Start(options connection.ConnectOptions) (err erro
...
+			c.ports,
+			config.Ports,
 			0,
-			c.pingerStop,
 		)
 		if err != nil {
 			return errors.Wrap(err, "could not ping provider")
 		}
+
+		if conn != nil {
```
anjmao: Is this check for conn != nil is needed? I would expect conn to be defined if err is nil.

nat/traversal/pinger.go
```diff
@@ -41,7 +41,7 @@ var (
 
 // NATProviderPinger pings provider and optionally hands off connection to consumer proxy.
 type NATProviderPinger interface {
-	PingProvider(ip string, providerPort, consumerPort, proxyPort int, stop <-chan struct{}) error
+	PingProvider(ip string, cPorts, pPorts []int, proxyPort int) (*net.UDPConn, error)
```
anjmao: Not sure if it's a good idea to remove stop channel. If user cancels connection how will you stop pinger?

nat/traversal/pinger.go
```diff
@@ -280,16 +235,18 @@ func (p *Pinger) pingReceiver(conn *net.UDPConn, stop <-chan struct{}) error {
 		case <-stop:
 			return errNATPunchAttemptStopped
 		default:
+			// Add read deadline to prevent possible conn.Read hang when remote peer doesn't send ping ack.
+			conn.SetReadDeadline(time.Now().Add(p.pingConfig.Timeout * 2))
 			n, err := conn.Read(buf)
-			if err != nil {
+			// Set higher read deadline when NAT proxy is used.
+			conn.SetReadDeadline(time.Now().Add(12 * time.Hour))
```
anjmao: Not sure if this is correct to set read deadline of 12 hours here. My idea was to set it from consumer side only after whole ping is done and consumer still need to use proxy connection but now it will be executed for each ping.

services/openvpn/client.go
```diff
@@ -194,8 +214,20 @@ func (c *Client) GetConfig() (connection.ConsumerConfig, error) {
...
+			return nil, err
+		}
+
+		c.ports = append(c.ports, cp.Num())
+	}
+
 	return &ConsumerConfig{
-		IP: publicIP,
+		IP:    publicIP,
+		Ports: c.ports,
```
zolia: I am almost confident its not a consumer side which has to pick ports. Single provider might have hundreds of consumers therefore change of collision  of ports will be much lower if provider would select ports to ping from.

nat/traversal/pinger.go
```diff
@@ -312,36 +269,66 @@ func (p *Pinger) pingTargetConsumer(pingParams *Params) {
...
+
+	for i := range localPorts {
+		go func(i int) {
+			conn, err := p.singlePing(remoteIP, localPorts[i], remotePorts[i], initialTTL+i, stop)
+			ch <- res{conn, err}
+		}(i)
+	}
+
+	// First responce wins. Other are not important.
+	r := <-ch
```
anjmao: Maybe it also make sense to cancel other pings when you got response?
Also is it ok to get first response even if it returns error. I would think that you want first non error response.

services/openvpn/client.go
```diff
@@ -129,7 +126,44 @@ func (c *Client) Statistics() <-chan consumer.SessionStatistics {
...
+			sessionConfig.RemoteIP,
+			c.ports,
+			sessionConfig.Ports,
+			sessionConfig.LocalPort,
+		)
+		if err != nil {
+			return err
+		}
+
+		_, lPort, err := net.SplitHostPort(conn.LocalAddr().String())
```
anjmao: Same here for port parsing.

services/openvpn/service/manager.go
```diff
@@ -236,27 +234,36 @@ func (m *Manager) ProvideConfig(_ string, sessionConfig json.RawMessage) (*sessi
...
+
+		if m.location.BehindNAT() && m.portMappingFailed() {
+			for range consumerConfig.Ports {
+				pp, err := m.natPingerPorts.Acquire()
+				if err != nil {
+					return nil, err
+				}
+
+				vpnConfig.Ports = append(vpnConfig.Ports, pp.Num())
+				vpnConfig.RemotePort = pp.Num()
```
anjmao: Maybe `vpnConfig.RemotePort` can be assign once outside of this loop if this is only a single port?

services/wireguard/wireguard.go
```diff
@@ -79,14 +79,16 @@ type ProviderModeConfig struct {
...
 }
 
 // ServiceConfig represent a Wireguard service provider configuration that will be passed to the consumer for establishing a connection.
 type ServiceConfig struct {
 	// LocalPort and RemotePort are needed for NAT hole punching only.
-	LocalPort  int `json:"-"`
-	RemotePort int `json:"-"`
+	LocalPort  int   `json:"-"`
+	RemotePort int   `json:"-"`
+	Ports      []int `json:"Ports"`
```
anjmao: Should be from lowercase `json:"ports"` the same is defined in MarshalJSON

mobile/mysterium/openvpn_connection_setup.go
```diff
@@ -128,24 +118,48 @@ func (c *openvpnConnection) Statistics() <-chan consumer.SessionStatistics {
...
 
-	log.Info().Msgf("Client config after session create: %v", clientConfig)
-	if clientConfig.LocalPort > 0 {
-		err := c.natPinger.PingProvider(
-			clientConfig.VpnConfig.OriginalRemoteIP,
-			clientConfig.VpnConfig.OriginalRemotePort,
-			clientConfig.LocalPort,
-			clientConfig.LocalPort+1,
-			c.pingerStop,
+	if sessionConfig.LocalPort == 0 && len(sessionConfig.Ports) > 0 {
```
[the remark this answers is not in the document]
soffokl: Correct, it should be `LocalPort > 0`. Changed it and added comment that this is just a compatibility check and it will be removed in the future.

nat/traversal/pinger.go
```diff
@@ -124,7 +123,7 @@ func (p *Pinger) Start() {
 }
 
 func isPunchingRequired(params *Params) bool {
-	return params.ConsumerPort > 0
+	return true
```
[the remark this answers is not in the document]
soffokl: Remoted this check. It should be decided from outside if traversal required.
Even if this is the direct connection this implementation will establish a connection correctly.

nat/traversal/pinger.go
soffokl: The real error could happen only if we failed to create connection which is significant.
In other cases, this will be just a timeout error, which means that whole pinger timed out.

services/openvpn/client.go
soffokl: The consumer picks his ports and the provider picks his ports. 
There is no collision, both sides provides available ports.

nat/traversal/pinger.go
```diff
@@ -86,24 +86,23 @@ type PortSupplier interface {
...
 }
 
 // Params contains session parameters needed to NAT ping remote peer
 type Params struct {
-	ProviderPort        int
-	ConsumerPort        int
-	ConsumerPublicIP    string
+	ProviderPorts       []int
+	ConsumerPorts       []int
+	IP                  string
```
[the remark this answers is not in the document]
soffokl: This traversal params is used for both consumer and provider now. 
So this is just IP, LocalPorts and RemotePorts.

mobile/mysterium/wireguard_connection_setup.go
soffokl: Added this because of the test which uses mocks which return `nil` connection and `nil` error.

nat/traversal/pinger.go
soffokl: There is a timeout for pinger, even if it is not stopped it will work for couple of seconds in any case and its conflict with nothing.

nat/traversal/pinger.go
soffokl: I was planning to change it later to reset this timeout in another PR to keep this as before.
But anyway resetting this read deadline now. No more 12 hours deadline.

nat/traversal/pinger.go
soffokl: It will be canceled from outside once the first value returned.

services/openvpn/service/manager.go
soffokl: This is part of the backward compatibility and it will be removed soon. 
Now RemotePort should be assigned to the latest version. 
I'd prefer to assign it 10 times with a single line, instead of adding extra block with checking slices length and assigning it outside.

mobile/mysterium/wireguard_connection_setup.go
anjmao: Looks like mock is not correct.

nat/traversal/pinger.go
anjmao: Default timeout is 10 seconds. If user cancel after 2 seconds there will still be 80 (8 seconds * 10 ports) not needed pings.

services/wireguard/service/service_unix.go
```diff
@@ -267,36 +266,45 @@ func (m *Manager) addTraversalParams(config wg.ServiceConfig, traversalParams tr
 	// There is no need to add any connect delay when port mapping failed.
 	config.Consumer.ConnectDelay = 0
 
+	// TODO this backward compatibility block needs to be removed once we will start using port ranges for all peers.
+	config.LocalPort = traversalParams.RemotePorts[len(traversalParams.RemotePorts)-1]
```
Waldz: Are you sure you wont index out of range exception here?

nat/traversal/pinger.go
```diff
@@ -304,44 +256,74 @@ func (p *Pinger) Valid() bool {
...
-		return
-	}
-
-	p.eventPublisher.Publish(event.AppTopicTraversal, event.BuildSuccessfulEvent(StageName))
-
-	log.Info().Msg("Ping received, waiting for a new connection")
-
-	go p.natProxy.handOff(pingParams.ProxyPortMappingKey, conn)
+	err = p.pingReceiver(conn, stop)
+	return conn, errors.Wrap(err, "ping receiver error")
```
Waldz: what if err is nil?

nat/traversal/pinger.go
```diff
@@ -304,44 +256,74 @@ func (p *Pinger) Valid() bool {
...
+	type res struct {
+		conn *net.UDPConn
+		err  error
+	}
+
+	ch := make(chan res, len(params.LocalPorts))
+
+	for i := range params.LocalPorts {
+		go func(i int) {
+			conn, err := p.singlePing(params.IP, params.LocalPorts[i], params.RemotePorts[i], initialTTL+i, stop)
```
Waldz: Do I understand correctly what you assume same order between `RemotePorts` and `LocalPorts`. And each side has pre configured TTL. 
I suggest to make new structure configurable thru communication, TTLs pressented in tuples:
```
natPorts: [
  {local: 10001, remote: 20003, ttl: 124},
  {local: 10003, remote: 20001, ttl: 125},
]
```

nat/traversal/pinger.go
zolia: nothing, Wrap will return nil.

nat/traversal/pinger.go
soffokl: There is a `func (p *Pinger) Stop()` method which can be called to stop pings.

nat/traversal/pinger.go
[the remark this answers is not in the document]
soffokl: This is not a DTO.

mobile/mysterium/wireguard_connection_setup.go
soffokl: Fixed it.

mobile/mysterium/openvpn_connection_setup.go
```diff
@@ -47,21 +48,10 @@ func NewOpenVPNConnection(sessionTracker *sessionTracker, signerFactory identity
...
-		err := json.Unmarshal(options.SessionConfig, sessionConfig)
-		if err != nil {
-			return nil, nil, err
-		}
-
-		// override vpnClientConfig params with proxy local IP and pinger port
-		// do this only if connecting to natted provider
+	sessionFactory := func(options connection.ConnectOptions, sessionConfig *openvpn.VPNConfig) (*openvpn3.Session, *openvpn.ClientConfig, error) {
 		if sessionConfig.LocalPort > 0 {
 			sessionConfig.OriginalRemoteIP = sessionConfig.RemoteIP
```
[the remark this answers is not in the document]
soffokl: Changed it to avoid mutation.

services/openvpn/client.go
```diff
@@ -209,6 +238,7 @@ type VPNConfig struct {
 	RemoteIP        string `json:"remote"`
 	RemotePort      int    `json:"port"`
 	LocalPort       int    `json:"lport"`
+	Ports           []int  `json:"ports"`
```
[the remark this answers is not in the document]
soffokl: No, this is local for one side, but remote for another.

services/wireguard/wireguard.go
```diff
@@ -79,14 +79,16 @@ type ProviderModeConfig struct {
 type ConsumerConfig struct {
 	PublicKey string `json:"PublicKey"`
 	// IP is needed when provider is behind NAT. In such case provider parses this IP and tries to ping consumer.
-	IP string `json:"IP,omitempty"`
+	IP    string `json:"IP,omitempty"`
+	Ports []int  `json:"Ports"`
```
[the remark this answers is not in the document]
soffokl: No, this is local for one side, but remote for another.

services/wireguard/wireguard.go
```diff
@@ -79,14 +79,16 @@ type ProviderModeConfig struct {
...
 }
 
 // ServiceConfig represent a Wireguard service provider configuration that will be passed to the consumer for establishing a connection.
 type ServiceConfig struct {
 	// LocalPort and RemotePort are needed for NAT hole punching only.
-	LocalPort  int `json:"-"`
-	RemotePort int `json:"-"`
+	LocalPort  int   `json:"-"`
+	RemotePort int   `json:"-"`
+	Ports      []int `json:"ports"`
```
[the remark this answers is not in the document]
soffokl: No, this is local for one side, but remote for another.

nat/traversal/pinger.go
```diff
@@ -41,13 +41,13 @@ var (
 
 // NATProviderPinger pings provider and optionally hands off connection to consumer proxy.
 type NATProviderPinger interface {
-	PingProvider(ip string, providerPort, consumerPort, proxyPort int, stop <-chan struct{}) error
+	PingProvider(params Params, proxyPort int) (*net.UDPConn, error)
```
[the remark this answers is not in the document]
soffokl: This is not a DTO.

services/openvpn/client.go
```diff
@@ -194,8 +211,20 @@ func (c *Client) GetConfig() (connection.ConsumerConfig, error) {
 		return nil, errors.Wrap(err, "failed to get consumer public IP")
 	}
 
+	pool := port.NewPool()
+
+	for i := 0; i < 10; i++ {
+		cp, err := pool.Acquire()
```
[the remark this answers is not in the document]
soffokl: Added.

mobile/mysterium/openvpn_connection_setup.go
```diff
@@ -128,24 +117,45 @@ func (c *openvpnConnection) Statistics() <-chan consumer.SessionStatistics {
...
+			LocalPorts:  c.ports,
+			RemotePorts: sessionConfig.Ports,
+		}
+
+		conn, err := c.natPinger.PingProvider(params, sessionConfig.LocalPort)
 		if err != nil {
 			return err
 		}
+
+		if addr, ok := conn.LocalAddr().(*net.UDPAddr); ok {
```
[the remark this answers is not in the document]
soffokl: I need to return the connection. Only ports are not enough.

services/wireguard/service/service_unix.go
soffokl: Added a check just in case.

nat/traversal/pinger.go
soffokl: params is not a DTO. Local and remote ports are never transferred together. TTL does not need to be transferred.

nat/traversal/pinger.go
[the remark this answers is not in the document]
Waldz: When why to pass function arguments in such complex manner and not just `PingProvider(ip string, remotePorts []int, localPorts []int)`?

mobile/mysterium/openvpn_connection_setup.go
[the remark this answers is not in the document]
Waldz: I dont see returned `conn` being used + duplicated code + leaked component internals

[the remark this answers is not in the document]
Waldz: +`conn` not closed in `openvpn_connection_setup.go`

nat/traversal/pinger.go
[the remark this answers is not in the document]
soffokl: Changed it as you wish.

mobile/mysterium/openvpn_connection_setup.go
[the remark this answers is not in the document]
soffokl: Changed it as you wish.

PR #1733, opened 2020-02-18
Apply DNS policy rules for incoming consumer traffic

Waldz:
Updates: #1597

While PR #1632 blocked all traffic on particular device if DNS policies exist, this puts in action whitelisting part:
- Implemented rule: block all & allow certain DNS hostnames
- Implemented rule: block all & allow certain DNS wildcard hostnames
- To whitelist traffic consumer must announce hostnames through it MUST use provider's DNS proxy. Otherwise all traffic is blocked by default


Waldz:
Repeating PR #1700 from correct fork, to pass full CI

dns/handler_system.go
```diff
@@ -0,0 +1,72 @@
...
+
+func (ph *proxyHandler) ServeDNS(writer dns.ResponseWriter, req *dns.Msg) {
+	for _, addr := range ph.proxyAddrs {
+		resp, _, err := ph.client.Exchange(req, addr)
+		if err != nil {
+			log.Error().Err(err).Msg("Error proxying DNS query to " + addr)
+			continue
+		}
+
+		writer.WriteMsg(resp)
```
anjmao: WriteMsg returns error. Is it ok to return here without checking for error?

dns/handler_whitelist.go
```diff
@@ -0,0 +1,90 @@
...
+		return
+	}
+
+	writer.WriteMsg(resp)
+}
+
+func (wh *whitelistHandler) whitelistByAnswer(response *dns.Msg) error {
+	for _, record := range response.Answer {
+		switch recordValue := record.(type) {
+		case *dns.A:
```
anjmao: What about IPv6 (AAAA) ?

dns/handler_whitelist.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+	return nil
+}
+
+func (wh *whitelistHandler) whitelistByARecord(record *dns.A) error {
+	host := strings.TrimRight(record.Hdr.Name, ".")
+	ip := record.A
+
+	if wh.policies.IsHostAllowed(host) {
+		_, err := wh.trafficBlocker.AllowIPAccess(ip)
```
anjmao: I would allow ip access if host is allowed and is not added. Not sure how will this impact performance if you constantly check and allow the same IP access.

PR #1738, opened 2020-02-19
Add retry for initial state load

### LINKED ISSUE #1737: Payment error while calling accountant
When consumer connects to VPN and fails to call accountant it's got disconnected.

```
2020-02-19T10:47:12.427 ERR core/connection/manager.go:459           > Payment error error="could not calculate amount to promise: could not request consumer data accountant: could not execute request: Get https://testnet-accountant.mysterium.network/api/v2/data/consumer/0xc150a4deeef9088397f491ca85c2a6e55eb1500d: net/http: request canceled (Client.Timeout exceeded while awaiting headers)"
```

Solution could be to add HTTP retry.

vkuznecovas:
Fixes #1737

### DIFF
--- a/session/pingpong/invoice_payer.go
+++ b/session/pingpong/invoice_payer.go
@@ -18,11 +18,13 @@
 package pingpong
 
 import (
+	"context"
 	"math"
 	"strings"
 	"sync"
 	"time"
 
+	"github.com/cenkalti/backoff/v4"
 	"github.com/mysteriumnetwork/node/core/connection"
 	"github.com/mysteriumnetwork/node/eventbus"
 	"github.com/mysteriumnetwork/node/market"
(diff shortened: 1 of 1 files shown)


session/pingpong/invoice_payer.go
```diff
@@ -167,14 +169,44 @@ func (ip *InvoicePayer) getGrandTotalPromised() (uint64, error) {
...
+			log.Debug().Msgf("No previous invoice grand total, assuming zero")
+			return nil
+		}
+		data = d
+		return nil
 	}
+
+	err := backoff.Retry(toRetry, boff)
+	if err != nil {
+		log.Error().Err(err).Msg("could not update balance from accountant")
```
zolia:
```suggestion
		log.Error().Err(err).Msg("could not get balance from accountant")
```

session/pingpong/invoice_payer.go
```diff
@@ -167,14 +169,44 @@ func (ip *InvoicePayer) getGrandTotalPromised() (uint64, error) {
...
+	}
+
 	return data.LatestPromise.Amount, nil
```
zolia: should we return err here?

session/pingpong/invoice_payer.go
```diff
@@ -167,15 +169,45 @@ func (ip *InvoicePayer) getGrandTotalPromised() (uint64, error) {
...
+
+	boff = backoff.WithMaxRetries(eback, 10)
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	go func() {
+		// Since we want to cancel the request if it.Stop is called, we're adding the defer and the select here
+		defer cancel()
+		select {
+		case <-ip.stop:
```
tadaskay: Perhaps it's simpler to add `cancel()` here and skip the explanation?

PR #1764, opened 2020-02-25
Add flag to enable pprof http endpoints

soffokl:


core/node/options.go
```diff
@@ -85,6 +86,7 @@ func GetOptions() *Options {
 		TequilapiPort:    config.GetInt(config.FlagTequilapiPort),
 		TequilapiEnabled: true,
 		BindAddress:      config.GetString(config.FlagBindAddress),
+		PProfEnabled:     config.GetBool(config.FlagPProfEnable),
```
tadaskay: In fact, I only added all flags to node options for compatibility. For new flags, it's sufficient to get it directly from config where needed: `config.GetBool(config.FlagPProfEnable)`

Issue #1784, opened 2020-02-26
Could not retrieve proposals for 0.21.0 & 0.20.3


etherunit:
trying to retrieve proposals list via Mysteriumn CLI
0.22-rc2 list 14 proposals in total
0.21.0, 0.20.3 returns 0 proposals
0.19.0 returns 1383 proposals

daemon output for 0.21.0 and 0.20.3
```
2020-02-26T16:38:23.215 DBG source/core/discovery/repository.go:61   > Retrieving proposals from 2 repositories
panic: not supported

goroutine 373 [running]:
github.com/mysteriumnetwork/node/market.UnsupportedPaymentMethod.GetPrice(...)
	/source/market/payment_method.go:48
github.com/mysteriumnetwork/node/core/discovery/reducer.Price.func1(0x1, 0xc00074f980, 0x13, 0xc000873d78, 0x7, 0x55a2480, 0xc0003b2680, 0xc00074f9c0, 0x1b, 0x55b9e20, ...)
	/source/core/discovery/reducer/proposal.go:70 +0x52
github.com/mysteriumnetwork/node/core/discovery/reducer.And.func1(0x1, 0xc00074f980, 0x13, 0xc000873d78, 0x7, 0x55a2480, 0xc0003b2680, 0xc00074f9c0, 0x1b, 0x55b9e20, ...)
	/source/core/discovery/reducer/and.go:32 +0xb0
github.com/mysteriumnetwork/node/core/discovery/proposal.(*Filter).Matches(0xc01124a540, 0x1, 0xc00074f980, 0x13, 0xc000873d78, 0x7, 0x55a2480, 0xc0003b2680, 0xc00074f9c0, 0x1b, ...)
	/source/core/discovery/proposal/filter.go:58 +0x17e
github.com/mysteriumnetwork/node/core/discovery/brokerdiscovery.(*ProposalStorage).MatchProposals(0xc0009a3d80, 0xc010f2cde8, 0x0, 0x0, 0x0, 0x0, 0x0)
	/source/core/discovery/brokerdiscovery/storage.go:70 +0x1dc
github.com/mysteriumnetwork/node/core/discovery/brokerdiscovery.(*ProposalStorage).FindProposals(0xc0009a3d80, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
	/source/core/discovery/brokerdiscovery/storage.go:79 +0x96
[...]

PR #1787, opened 2020-02-27
Exclude unsupported proposals from message broker

anjmao:
When proposals was added from broker to storage it was not validated if it's supported. Now adding to storage is no op if proposal is not supported. Also removed some panic and added error logging instead.

Fixes https://github.com/mysteriumnetwork/node/issues/1784

### DIFF
--- a/core/discovery/brokerdiscovery/repository_test.go
+++ b/core/discovery/brokerdiscovery/repository_test.go
@@ -18,27 +18,53 @@
 package brokerdiscovery
 
 import (
+	"encoding/json"
 	"testing"
 	"time"
 
 	"github.com/mysteriumnetwork/node/communication/nats"
 	"github.com/mysteriumnetwork/node/eventbus"
 	"github.com/mysteriumnetwork/node/market"
+	"github.com/mysteriumnetwork/node/money"
 	"github.com/stretchr/testify/assert"
 )
+
+func init() {
+	market.RegisterServiceDefinitionUnserializer(
+		"mock_service",
+		func(rawDefinition *json.RawMessage) (market.ServiceDefinition, error) {
+			return mockServiceDefinition{}, nil
+		},
+	)
+	market.RegisterPaymentMethodUnserializer(
+		"mock_payment",
+		func(rawDefinition *json.RawMessage) (market.PaymentMethod, error) {
+			return mockPaymentMethod{}, nil
+		},
+	)
+	market.RegisterContactUnserializer("mock_contact",
+		func(rawMessage *json.RawMessage) (market.ContactDefinition, error) {
+			return mockContact{}, nil
+		},
+	)
+}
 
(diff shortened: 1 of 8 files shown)


core/discovery/brokerdiscovery/storage.go
```diff
@@ -121,12 +121,16 @@ func (s *ProposalStorage) GetProposal(id market.ProposalID) (*market.ServiceProp
...
 }
 
-// AddProposal appends given proposal to storage
+// AddProposal appends given proposal to storage. If proposal is not supported it will not be added.
 func (s *ProposalStorage) AddProposal(proposals ...market.ServiceProposal) {
 	s.mutex.Lock()
 	defer s.mutex.Unlock()
 
 	for _, p := range proposals {
+		if !p.IsSupported() {
```
Waldz: I would not suggest to add logic it to storage level, it's fetcher responsibility to validate proposal before adding.
+Simple fetcher condition before adding and those complex mocked tests won't be needed

Waldz: I would not suggest to add logic it to storage level, it's fetcher responsibility to validate proposal before adding.
+Simple fetcher condition before adding and those complex mocked tests won't be needed

core/discovery/brokerdiscovery/repository_test.go
```diff
@@ -18,27 +18,53 @@
...
 	"time"
 
 	"github.com/mysteriumnetwork/node/communication/nats"
 	"github.com/mysteriumnetwork/node/eventbus"
 	"github.com/mysteriumnetwork/node/market"
+	"github.com/mysteriumnetwork/node/money"
 	"github.com/stretchr/testify/assert"
 )
 
+func init() {
```
Waldz: I don't see a point to mock special payment method in this testcase, because production code should handle unserialisable payment method already

core/discovery/brokerdiscovery/repository_test.go
```diff
@@ -52,14 +78,14 @@ func Test_Subscriber_StartSyncsNewProposals(t *testing.T) {
 	assert.NoError(t, err)
 
 	proposalRegister(connection, `{
-		"proposal": {"provider_id": "0x1"}
+		"proposal": {"provider_id": "0x1", "service_type": "mock_service", "payment_method_type": "mock_payment", "provider_contacts": [{"type":"mock_contact"}]}
```
Waldz: This should be enought

core/discovery/brokerdiscovery/repository_test.go
anjmao: I suggest to try my branch with you suggested changes.

Waldz: ok, looks still needed to test successful proposal add

PR #1788, opened 2020-02-27
Fix new TTL based NAT hole punching for mobile 

soffokl:


mobile/mysterium/openvpn_connection_setup.go
```diff
@@ -35,6 +34,8 @@ import (
 	"github.com/rs/zerolog/log"
 )
 
+const natPunchingMaxTTL = 10
```
anjmao: I suggest to change it to more clear name like `natPunchingPortsLen` etc. In this context TTL is a bit confusing with IP packets TTL.

soffokl: The number of ports is needed to test all TTL values up to the `natPunchingMaxTTL`.

PR #1790, opened 2020-02-27
Give more time to produce session stats

zolia:
Fixes continuously failing e2e for session stats checks


e2e/connection_test.go
```diff
@@ -158,7 +158,7 @@ func consumerConnectFlow(t *testing.T, tequilapi *tequilapi_client.Client, consu
 	assert.Equal(t, "New", se.Status)
 
 	// Wait some time for session to collect stats.
-	time.Sleep(2 * time.Second)
+	time.Sleep(4 * time.Second)
```
Waldz: Lets do "eventually" approach here instead of constant timeout increase

Issue #1751, opened 2020-02-20
session-destroy is not sent


zolia:
Node provider / consumer versions: Mysterium Node 0.21.0-1snapshot-20200220T0755-7e017106 (latest master)

Provider: OSX
Consumer: debian 10

After successful connection consumer fails to send session-destroy. 
Actual session-destroy message is being sent way too late. It should be sent before even attempting to disconnect openvpn session. 

```
2020-02-20T12:50:57.038 DBG consumer/statistics/tracker.go:78        > bytes received 2.2 MB, sent 374.5 kB
2020-02-20T12:50:57.038 DBG consumer/bandwidth/tracker.go:95         > Download speed: 41.1 kbps
2020-02-20T12:50:57.039 DBG consumer/bandwidth/tracker.go:96         > Upload speed: 17.4 kbps
2020-02-20T12:50:57.826 INF core/connection/manager.go:426           > Connection state: Connected → Disconnecting
2020-02-20T12:50:57.864 DBG firewall/iptables/iptables.go:34         > "sudo /sbin/iptables -D OUTPUT -s 88.198.207.96 -j MYST_CONSUMER_KILL_SWITCH" output:

2020-02-20T12:50:57.865 INF logconfig/openvpnlog.go:54               > [openvpn-mgmt] Shutdown
2020-02-20T12:50:57.865 INF logconfig/openvpnlog.go:54               > [openvpn-mgmt] Shutdown finished
2020-02-20T12:50:57.865 DBG logconfig/openvpnlog.go:63               > [openvpn-proc] Stdout: Thu Feb 20 12:50:57 2020 event_wait : Interrupted system call (code=4)
2020-02-20T12:50:57.865 DBG logconfig/openvpnlog.go:63               > [openvpn-proc] Stdout: Thu Feb 20 12:50:57 2020 SIGTERM received, sending exit notification to peer
2020-02-20T12:50:57.902 WRN logconfig/openvpnlog.go:45               > Failed to remove tun device: [sudo ip tuntap delete dev tun0 mode tun] Returned exit error: exit status 1 Cmd output: ioctl(TUNSETIFF): Device or resource busy

2020-02-20T12:50:57.928 DBG firewall/iptables/iptables.go:34         > "sudo /sbin/iptables -D MYST_CONSUMER_KILL_SWITCH -d 88.119.191.24 -j ACCEPT" output:

[...]

PR #1792, opened 2020-02-27
NATS reconnect waits for successful connection

Waldz:
Fixes: #1751

### DIFF
--- a/communication/nats/connection_mock.go
+++ b/communication/nats/connection_mock.go
@@ -158,6 +158,11 @@ func (conn *ConnectionMock) Open() error {
 	return nil
 }
 
+// Reopen restarts the connection
+func (conn *ConnectionMock) Reopen() error {
+	return nil
+}
+
 // Close destructs the connection
 func (conn *ConnectionMock) Close() {
 	conn.queueShutdown <- true
--- a/communication/nats/connector.go
+++ b/communication/nats/connector.go
@@ -89,8 +89,11 @@ func (b *BrokerConnector) ReconnectAll() {
 		go func(id uuid.UUID, conn *ConnectionWrap) {
 			defer wg.Done()
 			log.Info().Msgf("Re-establishing broker connection %v", id)
-			err := conn.Conn.Flush()
-			log.Info().Msgf("Re-establishing broker connection %v DONE (check result=%v)", id, err)
+			if err := conn.Reopen(); err != nil {
+				log.Err(err).Msgf("Re-establishing broker connection %v failed", id)
+			} else {
+				log.Info().Msgf("Re-establishing broker connection %v DONE", id)
+			}
 		}(k, v)
 	}
 	wg.Wait()
--- a/communication/nats/connector.go
+++ b/communication/nats/connector.go
@@ -29,14 +29,14 @@ import (
 
 // BrokerConnector establishes new connections to NATS servers and handles reconnects.
 type BrokerConnector struct {
-	registry map[uuid.UUID]Connection
+	registry map[uuid.UUID]*ConnectionWrap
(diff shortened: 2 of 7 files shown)


communication/nats/connector.go
```diff
@@ -86,10 +86,10 @@ func (b *BrokerConnector) ReconnectAll() {
 	var wg sync.WaitGroup
 	for k, v := range b.registry {
 		wg.Add(1)
-		go func(id uuid.UUID, conn Connection) {
+		go func(id uuid.UUID, conn *ConnectionWrap) {
 			defer wg.Done()
 			log.Info().Msgf("Re-establishing broker connection %v", id)
-			err := conn.Check()
+			err := conn.Conn.Flush()
```
soffokl: `Check` does `Flush` inside with timeout. How these changes should resolve the problem?

communication/nats/connector.go
```diff
@@ -29,14 +29,14 @@ import (
 
 // BrokerConnector establishes new connections to NATS servers and handles reconnects.
 type BrokerConnector struct {
-	registry map[uuid.UUID]Connection
+	registry map[uuid.UUID]*ConnectionWrap
```
soffokl: Could you explain why abstract `Connection` should be replaced by exact implementation?

communication/nats/connector.go
anjmao: Check does flush with 3 seconds timeout. Conn.Flush does it with 60 seconds timeout. I tested this branch on my local pc and it seems work except that now I need to wait much longer to disconnect.

Original issue suggests that session destroy should be sent before VPN tunnel disconnect., but these changes doesn't address that.

communication/nats/connector.go
anjmao: Because conn.Conn.Flush() call is not available on the interface.

communication/nats/connector.go
zolia: There is a case, when we are connected to a node, but node does not forward traffic. In such case reordering session-destroy with actual vpn disconnect does not help. 

My suggestion is to implement session-destroy send with retry and attempt to resend up to 60sec. Since we DO get error on failed send.

anjmao: @zolia Why do we forward broker traffic via VPN? Maybe we could forward broker traffic via default consumer gateway?

communication/nats/connector.go
Waldz: Yes, and it's nats.Connector, which holds exact implementations, not abstract ones

communication/nats/connector.go
Waldz: > My suggestion is to implement session-destroy send with retry and attempt to resend up to 60sec. Since we DO get error on failed send.

I tested with retry, but every message sending could end up being slow then.
Left this approach, because:
- no point to do ReconnectAll() wo checking if it works after recconnect
- it's not slow, it takes as much time, as user gets his internet back
- slow parts are postponed to be done after actual reconnect, non blocking anything

Waldz: After digging why it's slow found that `Flush()` does no reconnect, but rather sends "ping" message and waits go NAT reconnect routine job to kick in. So it "ping" should be send with timeout > 5s (see `nats.PingInterval`).

Anyway, it's a hacky way, so remade it to do real reconnect.

communication/nats/connection_mock.go
```diff
@@ -158,6 +158,11 @@ func (conn *ConnectionMock) Open() error {
 	return nil
 }
 
+// Open restarts the connection
```
soffokl:
```suggestion
// Reopen restarts the connection
```

PR #1817, opened 2020-03-02
Be more tolerant for asymmetric port mapping in NAT hole punching

soffokl:
Now we are not stick to the port provided in negotiation and listening connections from all ports.

### DIFF
--- a/nat/traversal/nat_proxy.go
+++ b/nat/traversal/nat_proxy.go
@@ -21,7 +21,6 @@ import (
 	"fmt"
 	"io"
 	"net"
-	"time"
 
 	"github.com/rs/zerolog/log"
 )
@@ -43,7 +42,6 @@ func newNATProxy() *natProxy {
 }
 
 func (np *natProxy) consumerHandOff(consumerAddr string, remoteConn *net.UDPConn) chan struct{} {
-	time.Sleep(400 * time.Millisecond)
 	stop := make(chan struct{})
 	if np.socketProtect == nil {
 		// shutdown pinger session since openvpn client will connect directly (without natProxy)
--- a/nat/traversal/pinger.go
+++ b/nat/traversal/pinger.go
@@ -161,16 +161,13 @@ func (p *Pinger) PingProvider(ip string, localPorts, remotePorts []int, proxyPor
 	return localPort, remotePort, err
 }
 
-func (p *Pinger) ping(conn *net.UDPConn, ttl int, stop <-chan struct{}) error {
-	// Windows detects that 1 TTL is too low and throws an exception during send
-	i := 0
-
+func (p *Pinger) ping(conn *net.UDPConn, remoteAddr *net.UDPAddr, ttl int, stop <-chan struct{}) error {
 	err := ipv4.NewConn(conn).SetTTL(ttl)
 	if err != nil {
 		return errors.Wrap(err, "pinger setting ttl failed")
 	}
 
-	for {
+	for i := 1; time.Duration(i)*p.pingConfig.Interval < p.pingConfig.Timeout; i++ {
 		select {
 		case <-stop:
 			return nil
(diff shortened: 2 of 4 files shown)


nat/traversal/pinger.go
```diff
@@ -161,58 +161,33 @@ func (p *Pinger) PingProvider(ip string, localPorts, remotePorts []int, proxyPor
...
-	i := 0
-
+func (p *Pinger) ping(conn *net.UDPConn, remoteAddr *net.UDPAddr, ttl int, stop <-chan struct{}) error {
 	err := ipv4.NewConn(conn).SetTTL(ttl)
 	if err != nil {
 		return errors.Wrap(err, "pinger setting ttl failed")
 	}
 
-	for {
+	for i := 1; time.Duration(i)*p.pingConfig.Interval < p.pingConfig.Timeout; i++ {
```
tadaskay: Shouldn't this be `i * Interval`? You're multiplying nanos with a Duration, I'm not sure what's the result :)

tadaskay: Also, the whole condition is only approximate, because the loop code will also consume time. It would be better to simply use a timer, I guess.

nat/traversal/nat_proxy.go
```diff
@@ -43,7 +42,6 @@ func newNATProxy() *natProxy {
 }
 
 func (np *natProxy) consumerHandOff(consumerAddr string, remoteConn *net.UDPConn) chan struct{} {
-	time.Sleep(400 * time.Millisecond)
```
zolia: As commit message states, this was introduced to workaround the case with very slow providers. I tested it on some weak machine with a single core.

nat/traversal/pinger.go
```diff
@@ -330,18 +305,33 @@ func (p *Pinger) multiPing(ip string, localPorts, remotePorts []int, initialTTL
 }
 
 func (p *Pinger) singlePing(remoteIP string, localPort, remotePort, ttl int, stop <-chan struct{}) (*net.UDPConn, error) {
-	conn, err := p.getConnection(remoteIP, remotePort, localPort)
+	conn, err := net.ListenUDP("udp4", &net.UDPAddr{Port: localPort})
```
zolia: This is smart. :) Did you have a change to test with those mobile clients this code already?

nat/traversal/pinger.go
```diff
@@ -161,58 +161,33 @@ func (p *Pinger) PingProvider(ip string, localPorts, remotePorts []int, proxyPor
...
-	conn, err := net.DialUDP("udp", &net.UDPAddr{Port: localPort}, udpAddr)
-	if err != nil {
-		return nil, err
-	}
 
-	log.Info().Msg("Local socket: " + conn.LocalAddr().String())
-
-	return conn, nil
+	err = errors.New("timeout while waiting for ping ack, trying to continue")
+	p.eventPublisher.Publish(event.AppTopicTraversal, event.BuildFailureEvent(StageName, err))
```
Waldz: It' should be one single result event (success/failure), currently it spams with each parallel ping result

nat/traversal/pinger.go
```diff
@@ -234,32 +209,32 @@ func (p *Pinger) BindServicePort(key string, port int) {
...
 			// Add read deadline to prevent possible conn.Read hang when remote peer doesn't send ping ack.
 			conn.SetReadDeadline(time.Now().Add(p.pingConfig.Timeout * 2))
-			n, err := conn.Read(buf)
+			n, raddr, err := conn.ReadFromUDP(buf)
 			// Reset read deadline.
 			conn.SetReadDeadline(time.Time{})
 
 			if err != nil || n == 0 {
-				log.Error().Err(err).Msgf("Failed to read remote peer: %s - attempting to continue", conn.RemoteAddr().String())
+				log.Error().Err(err).Msgf("Failed to read remote peer: %s - attempting to continue", raddr)
```
Waldz: It don't think it's ERR level for each parallel ping result.

nat/traversal/pinger.go
soffokl: Changed it to be more readable. We don't even need `i` here anymore.

nat/traversal/nat_proxy.go
soffokl: We have now whole ping `Timeout` in 10 seconds to finish it. This should not be a problem now.

nat/traversal/pinger.go
soffokl: It works with our mobile client, just tested it. The provider with a cellular internet is not tested yet.

nat/traversal/pinger.go
soffokl: Moved it to the top level only.

nat/traversal/pinger.go
soffokl: Changed it to debug.

nat/traversal/pinger.go
Waldz: Are you sure, events for single ping case are triggered?

soffokl: There are no single pings any more. Entrypoint is `multiPing` even if there is only single port in the slice.

nat/traversal/pinger.go
```diff
@@ -326,22 +298,40 @@ func (p *Pinger) multiPing(ip string, localPorts, remotePorts []int, initialTTL
...
 }
 
 func (p *Pinger) singlePing(remoteIP string, localPort, remotePort, ttl int, stop <-chan struct{}) (*net.UDPConn, error) {
-	conn, err := p.getConnection(remoteIP, remotePort, localPort)
+	conn, err := net.ListenUDP("udp4", &net.UDPAddr{Port: localPort})
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to get connection")
 	}
 
+	log.Info().Msg("Local socket: " + conn.LocalAddr().String())
```
anjmao: Could be 

```go
log.Info().Msg("Local socket: %s", conn.LocalAddr())
```

nat/traversal/pinger_test.go
```diff
@@ -56,15 +56,17 @@ func TestPinger_Provider_Consumer_Ping_Flow(t *testing.T) {
...
-	proxyBuf := make([]byte, 1024)
+	ch := make(chan string)
 	go func() {
 		addr, _ := net.ResolveUDPAddr("udp4", fmt.Sprintf("127.0.0.1:%d", providerProxyPort))
 		conn, err := net.ListenUDP("udp4", addr)
 		assert.NoError(t, err)
 
 		for {
-			_, err := conn.Read(proxyBuf)
+			proxyBuf := make([]byte, 1024)
```
anjmao: Allocate buffer once outside loop.

nat/traversal/nat_proxy.go
anjmao: It would be good to avoid such random sleeps and redesign code. You never how weak machine is :D

Issue #1785, opened 2020-02-26
Better handling for connect errors in Android


anjmao:
Currently when node.Connect returns error in android it is shown as generic `Failed to connect.Please try again` error. In case when there is no enough tokens balance we should instead inform user with correct message like `No enough balance` etc.

PR #1819, opened 2020-03-02
Validate payment balance during connect

anjmao:
Fixes https://github.com/mysteriumnetwork/node/issues/1785


core/connection/manager.go
```diff
@@ -124,11 +129,12 @@ func NewManager(
 	dialogCreator DialogCreator,
 	paymentEngineFactory PaymentEngineFactory,
 	connectionCreator Creator,
-	eventPublisher Publisher,
+	eventPublisher eventbus.Publisher,
```
soffokl: Looks like the interface in this file (line 85) is not used any more:
```go
type Publisher interface {
	Publish(topic string, data interface{})
}
```

PR #1822, opened 2020-03-02
Add monotonic + bootime time package for payments timer tracker

anjmao:
Add specific time package for payments timer tracker to get correct duration between two durations which includes device suspend time.

### DIFF
--- a/session/mbtime/mbtime.go
+++ b/session/mbtime/mbtime.go
@@ -1,0 +1,78 @@
+/*
+ * Copyright (C) 2020 The "MysteriumNetwork/node" Authors.
+ *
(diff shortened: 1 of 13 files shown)


session/mbtime/mbtime.go
```diff
@@ -0,0 +1,79 @@
...
+}
+
+// Sub returns the duration t-u.
+func (t Time) Sub(u Time) time.Duration {
+	return t.ns - u.ns
+}
+
+// String returns time formatted time string.
+func (t Time) String() string {
+	return fmt.Sprintf("%s", t.ns)
```
soffokl: > S1025: should use String() instead of fmt.Sprintf (gosimple)

PR #1827, opened 2020-03-03
Expose proposal prices via tequilapi

tadaskay:
Needed for the desktop app to display proposal pricing.

Exposed `PerTime` duration as `PerSeconds` for easier parsing/calculations for the clients.

### DIFF
--- a/tequilapi/endpoints/proposals.go
+++ b/tequilapi/endpoints/proposals.go
@@ -25,6 +25,7 @@ import (
 	"github.com/mysteriumnetwork/node/core/discovery/proposal"
 	"github.com/mysteriumnetwork/node/core/quality"
 	"github.com/mysteriumnetwork/node/market"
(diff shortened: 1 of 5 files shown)


tequilapi/endpoints/proposals.go
```diff
@@ -101,6 +116,14 @@ func proposalToRes(p market.ServiceProposal) *proposalDTO {
 			},
 		},
 		AccessPolicies: p.AccessPolicies,
+		PaymentMethod: paymentMethodRes{
+			Type:  p.PaymentMethod.GetType(),
```
anjmao: p.PaymentMethod could be nil. For example if I build my own node to send nil payment method all consumers can start failing when fetching proposals (but not sure how handled in other places).

tadaskay: It shouldn't be `nil`. It's deserialized to `market.UnsupportedPaymentMethod`.

Issue #1851, opened 2020-03-06
Incorrect stale balance state on consumer


zolia:
Provider logs: 

```
2020-03-06T11:59:17.074 DBG session/pingpong/price_calculator.go:66  > Calculated price 1041. Time component: 1041, data component: 0
2020-03-06T11:59:17.075 DBG session/pingpong/invoice_tracker.go:448  > Being lenient for the first payment, asking for 832
2020-03-06T11:59:17.085 DBG communication/nats/sender.go:60          > Message "34e66d60-2f23-46b3-a27f-9deaf423ca08.session-invoice" sending: {"payload":{"invoice":{"AgreementID":6094645989002451636,"AgreementTotal":832,"TransactorFee":0,"Hashlock":"f8041944c623bfd2038b438993f763bc0bc1e1d70e8c4460435b6f60dd1b104f","Provider":"0xe5e54c844d8e47f4eb43b58b286ce89e18b4283c"}},"signature":"g0jVpYilImSX3doUwUsfsek+Fk8wPnioDqAqDGE+fTkM8Ea7UG40Fl2rB3u0BU5TnpccnnIlQ8nWmeI2zfzCGwA="}
[...]

PR #1860, opened 2020-03-09
Sync consumer balance every time on startup

Waldz:
Closes: #1851

### DIFF
--- a/session/pingpong/invoice_payer.go
+++ b/session/pingpong/invoice_payer.go
@@ -19,6 +19,7 @@ package pingpong
 
 import (
 	"context"
+	"fmt"
 	"math"
 	"strings"
 	"sync"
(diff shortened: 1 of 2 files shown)


session/pingpong/invoice_payer.go
```diff
@@ -154,24 +159,7 @@ func (ip *InvoicePayer) Start() error {
 	}
 }
 
-const grandTotalKey = "consumer_grand_total"
-
-func (ip *InvoicePayer) getGrandTotalPromised() (uint64, error) {
```
vkuznecovas: don't remove this func, the force-sync is temporary and we will need to operate on balances we have in state

Waldz: So we have `ip.deps.ConsumerTotalsStorage.Get()` for that and I created `recoverGrandTotalPromised()` for syncing it if needed

session/pingpong/invoice_payer.go
```diff
@@ -204,12 +192,16 @@ func (ip *InvoicePayer) recoverGrandTotalPromised() (uint64, error) {
 		return nil
 	}
 
-	err := backoff.Retry(toRetry, boff)
-	if err != nil {
-		log.Error().Err(err).Msg("could not get balance from accountant")
```
zolia: Why to remove more detailed description here? We could say that "Failed after retry".

Waldz: It's moved one layer up

PR #1862, opened 2020-03-09
Initial p2p channel implementation

anjmao:
This PR adds initial p2p implementation in it's own standalone package. It uses H2C (HTTP2 over plain text) with kcp-go UDP transport and AES encryption. For now API accepts preshared private key and secure key exchange will be implemented in separate PR. It's possible (but not restricted) to send data in Protobuf format.

See channel_example_test.go for examples on proposed API.


p2p/channel_test.go
```diff
@@ -0,0 +1,190 @@
...
+	ports := acquirePorts(t, 2)
+	providerPort := ports[0]
+	consumerPort := ports[1]
+	privateKey := GeneratePrivateKey()
+	assert.NoError(t, err)
+
+	t.Run("Test consumer channel creation", func(t *testing.T) {
+		consumer, err = NewChannel(providerPort, privateKey)
+		assert.NoError(t, err)
+		go func() {
```
zolia: This t.Run already runs in separate goroutine  and since no other code besides ListenAndServe(), theres no point to run inside another goroutine.

p2p/channel_test.go
```diff
@@ -0,0 +1,190 @@
...
+			err = provider.ListenAndServe()
+			assert.NoError(t, err)
+		}()
+	})
+
+	t.Run("Test peer join", func(t *testing.T) {
+		assert.NotNil(t, consumer)
+		consumer.JoinPeer(&net.UDPAddr{IP: net.ParseIP("127.0.0.1"), Port: consumerPort})
```
zolia: I would flip the receiver. Now it looks like consumer is joining consumer.

p2p/context.go
```diff
@@ -0,0 +1,57 @@
...
+	OK() error
+}
+
+type defaultContext struct {
+	req         *Message
+	res         *Message
+	publicError error
+}
+
+func (d *defaultContext) Request() *Message {
```
zolia: Comments missing for public methods

p2p/context.go
```diff
@@ -0,0 +1,57 @@
...
+	return d.req
+}
+
+func (d *defaultContext) Error(err error) error {
+	d.publicError = err
+	return nil
```
zolia: Should it return an error?

anjmao: No, it is required only by handler signature to always return error or success with `c.OK()`.

p2p/context.go
anjmao: It 's not required to put comments on internal interface implementation since comments are already added for Context interface, golint allows this too.

p2p/channel_test.go
anjmao: Yes, but it it still waits for t.Run to finish before running another subtest so I can't block it.

Issue #1882, opened 2020-03-13
Trust oracle rules refresh requires node restart


zolia:
Trust oracle rules refresh requires node restart.

PR #1883, opened 2020-03-13
Override old policy rules

zolia:
Newly released policy updates where not being activated and were requiring a restart. 

fixes: #1882

### DIFF
--- a/core/policy/repository.go
+++ b/core/policy/repository.go
@@ -44,7 +44,7 @@ func NewRepository() *Repository {
 	}
 }
 
-// SetPolicyRules set policy ant it's items to repository
+// SetPolicyRules set policy and it's items to repository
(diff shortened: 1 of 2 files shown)


core/policy/repository.go
```diff
@@ -182,9 +182,9 @@ func (r *Repository) IsHostAllowed(host string) bool {
 }
 
 func (r *Repository) findItemFor(policy market.AccessPolicy) (*listItem, error) {
-	for _, item := range r.items {
+	for i, item := range r.items {
 		if item.policy == policy {
-			return &item, nil
+			return &r.items[i], nil
```
anjmao: Isn't `r.items[i]` is the same as `item` ? How did this helped?

zolia: Its not, range produces a copy.

PR #1890, opened 2020-03-17
Disable port mapping for routers with a private IP

### LINKED ISSUE #1850: Disable port-mapping if more than one NAT router detected
`portmap.Interface` provided by `github.com/ethereum/go-ethereum/p2p/nat` returns external IP of a router:

	// This method should return the external (Internet-facing)
	// address of the gateway device.
	ExternalIP() (net.IP, error)

If such IP falls into such IP classes: 
```
192.168.0.0 / 16
172.16.0.0 / 16 
10.0.0.0 / 8 
```

We then consider that given router is behind another router and we should not use port-mapping.

soffokl:
Closes #1850

### DIFF
--- a/nat/mapping/port_mapping.go
+++ b/nat/mapping/port_mapping.go
@@ -18,6 +18,7 @@
 package mapping
 
 import (
+	"net"
 	"time"
 
 	portmap "github.com/ethereum/go-ethereum/p2p/nat"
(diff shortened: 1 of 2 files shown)


nat/mapping/port_mapping.go
```diff
@@ -99,6 +105,25 @@ func (p *portMapper) Map(protocol string, port int, name string) (release func()
...
+func (p *portMapper) routerIPPublic() bool {
+	ip, err := p.config.MapInterface.ExternalIP()
+	if err != nil {
+		log.Warn().Err(err).Msg("Couldn't detect router IP address")
+		return false
+	}
+
+	log.Debug().Msgf("Detected router public IP address: %s", ip)
+
+	for _, s := range []string{"10.0.0.0/8", "172.16.0.0/12", "192.168.0.0/16"} {
```
Waldz: Maybe better to validate against IP oracle external IP instead of hardcoding these values. Some exotic private networks might be skipped here

zolia: There are no exotic private networks. Only these are considered by numeration standards. Everything else is either reserved for such things as multicasting or should be considered public.

Issue #1585, opened 2020-01-29
Reveal R mechanism


vkuznecovas:
At the moment, R's are only revealed for ongoing sessions.

This, however, is fragile as R's are only revealed when the consumer is connected to us. If we want a more resilient mechanism, a job should start when providing a service and periodically check for promises that might need to be revealed and revealing them.

The session pingpong would then also need to be redesigned a little to make the synchronous calls to accountant obsolete.

vkuznecovas:
This also needs to keep in mind the possibility of hashlock missmatches between provider and accountant.

zolia:
We had a grooming session with @vkuznecovas today. My suggestion is to store encrypted R on Accountant:
  - Loss of R looks like a major issue which forces provider to create a new identity. We want to avoid that.
  - We already manage state on Accountant and assume that it cannot be lost.
  - Provider with request_promise message would send its R encrypted with its own private key.
  - Ability to fetch encrypted R also enables synchronization between concurrent request_promise messages since Provider will not exactly which R is to be expected by Accountant.

@chompomonim can you comment on such solution? Should we proceed?

chompomonim:
I think this solution could work. Using an Accountant as your backup service (for R in encrypted form) looks quite interesting.

chompomonim:
I suppose this error is related?

```
Mar 11 10:10:59 myst-2 myst[16462]: 2020-03-11T10:10:59.718 INF market/mysterium/mysterium_api.go:203    > Proposal pinged for node: 0x9137cbfb964275623121385ecb951e484de5f964 service type: wireguard
Mar 11 10:10:59 myst-2 myst[16462]: 2020-03-11T10:10:59.750 DBG session/pingpong/price_calculator.go:66  > Calculated price 52541. Time component: 51866, data component: 675
[...]

PR #1894, opened 2020-03-17
Allow decryption and encryption of arbitrary data with derived keys

vkuznecovas:
Added key derivation from master key on unlock. This will be used for revealing R and maybe things in the future as well.

Prereq for #1585


identity/keystore_filesystem.go
```diff
@@ -18,17 +18,189 @@
...
+
+	hashFunc := sha512.New
+	hkdfDeriver := hkdf.New(hashFunc, k.PrivateKey.D.Bytes(), nil, nil)
+	key := make([]byte, 32)
+	_, err = io.ReadFull(hkdfDeriver, key)
+	return key, err
+}
+
+// EncryptWithDerived takes a derived key for the given address and encrypts the plaintext.
+func (ks *Keystore) EncryptWithDerived(addr common.Address, plaintext []byte) ([]byte, error) {
```
anjmao: I would name it `Encrypt` and  `Decrypt` as WithDerived is more like implementation detail.

identity/keystore_filesystem.go
```diff
@@ -18,17 +18,189 @@
...
+	kjson, err := ks.ethKeystore.Export(a, passphrase, passphrase)
+	if err != nil {
+		return nil, err
+	}
+
+	k, err := ethKs.DecryptKey(kjson, passphrase)
+	if err != nil {
+		return nil, err
+	}
+
```
zolia: I suspect this private key might get left in memory decrypted. 
```suggestion
defer memguard.WipeBytes(k.PrivateKey.D.Bytes())
```

identity/keystore_filesystem.go
```diff
@@ -18,17 +18,189 @@
...
+}
+
+// Unlock unlocks an account.
+func (ks *Keystore) Unlock(a accounts.Account, passphrase string) error {
+	err := ks.ethKeystore.Unlock(a, passphrase)
+	if err != nil {
+		return err
+	}
+
+	derived, err := ks.deriveKey(a, passphrase)
```
zolia: Is this expensive operation? Can we deriveKey just before encrypting / decrypting and destroy (wipe) right after we used it?

identity/keystore_filesystem.go
```diff
@@ -18,17 +18,189 @@
...
+	return key, err
+}
+
+// EncryptWithDerived takes a derived key for the given address and encrypts the plaintext.
+func (ks *Keystore) EncryptWithDerived(addr common.Address, plaintext []byte) ([]byte, error) {
+	key, err := ks.getDerivedKey(addr)
+	if err != nil {
+		return nil, err
+	}
+
```
zolia:
```suggestion
defer memguard.WipeBytes(key)
```

identity/keystore_filesystem.go
vkuznecovas: It's expensive in a sense that we need to store the pk in memory if we want to invoke this. I'd rather save the derived key.

vkuznecovas: It's expensive in a sense that we need to store the pk in memory if we want to invoke this. I'd rather save the derived key.

identity/keystore_filesystem.go
vkuznecovas: It won't. Function vars won't escape to heap, meaning they will stay on the stack, which in turn means they will not get garbage collected since the stack is destroyed once the function returns. I can add the wipe if you prefer though.

zolia: Destruction of a stack does not mean that physical memory will seize to contain this key, it will simply destroy references to it. Thats why wipe should be needed imho.

identity/keystore_filesystem.go
zolia: Agree, if we can rely on memguard doing its job properly.

PR #1899, opened 2020-03-19
Order pinger connections for each peer

anjmao:
Currently nat pinger pings peer concurrently and returns **n**  connections from max 10 possible (10 ports currently). The problem is that consumer and provider can select connection orders like **[1,2]** and **[2,3]** etc.  For such cases connections will not match and we can't use them for later communication.

This PR splits PingPeer into PingConsumerPeer and PingProviderPeer and adds logic to notify consumer which connections should be used.

### DIFF
--- a/nat/traversal/pinger.go
+++ b/nat/traversal/pinger.go
@@ -174,7 +174,7 @@ func (p *Pinger) PingConsumerPeer(ip string, localPorts, remotePorts []int, init
 	stop := make(chan struct{})
 	defer close(stop)
 
-	ch, err := p.multiPing(ip, localPorts, remotePorts, initialTTL, stop)
+	ch, err := p.multiPing(ip, localPorts, remotePorts, initialTTL, n, stop)
 	if err != nil {
 		log.Err(err).Msg("Failed to ping remote peer")
 		return nil, err
@@ -192,6 +192,8 @@ func (p *Pinger) PingConsumerPeer(ip string, localPorts, remotePorts []int, init
 			continue
 		}
 
+		sendMsg(res.conn, msgOK) // Notify peer that we are using this connection.
+
 		pings = append(pings, res)
 		if len(pings) == n {
 			waitAllConnACKSent(pings)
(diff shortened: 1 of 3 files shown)


anjmao:
@soffokl It set's 128 TTL before doing these sync pings https://github.com/mysteriumnetwork/node/blob/master/nat/traversal/pinger.go#L178 so this should work.

PingPeer method itself doesn't describe if this is consumer or provider.  I can split it into two separate methods PingProviderV2 and PingConsumerV2 and implement you suggestions as it will simplify logic a lot. Later will replace old PingProvider and PingConsumer anyway.

soffokl:
@anjmao it sets TTL to 128 when connection established to notify another peer. 
Initial TTL is an argument: https://github.com/mysteriumnetwork/node/blob/master/nat/traversal/pinger.go#L168

The fast hack could be
```
If initialTTL == maxTTL {
   Do not send "OK"
} else {
   Do send "OK"
}
```

zolia:
> Obviously this doesn't work when you are testing everything in 127.0.0.1, because TTL 1 is enough to reach both sides and both peers can receive pings and keep this connection.
> This can be resolved by sending `OK` response explicitly only from one side.

It is ok to assume that all connections might establish. Its one of test cases :) 

As for identification - both sides have to know which connection is being used for which service. So logic behind this could be: 
 - You always assume that 1st session is for communication channel
 - 2nd session for 1st service and etc.
 - It is enough on one side to send IDS: id1, id2, id3
 - other side side sends "OKs" and waits for "OKACK" to be sure this session is alive
 - lowest established ID would be used for communication channel.
 - next established ID would be used for service.

nat/traversal/pinger.go
```diff
@@ -171,28 +180,148 @@ func (p *Pinger) PingPeer(ip string, localPorts, remotePorts []int, initialTTL i
...
-		} else {
+			continue
+		}
+
+		if err := ipv4.NewConn(res.conn).SetTTL(maxTTL); err != nil {
+			log.Warn().Err(res.err).Msg("Failed to set connection TTL")
+			continue
+		}
+
+		if err := sendConnACK(res.conn); err != nil {
```
anjmao: One improvement here can be to move sendConnACK into block after needed amount of pings is collected and send it concurrently waiting for response or timeout.

nat/traversal/pinger.go
```diff
@@ -115,7 +123,7 @@ func (p *Pinger) Stop() {
 
 // PingProvider pings provider determined by destination provided in sessionConfig
 func (p *Pinger) PingProvider(ip string, localPorts, remotePorts []int, proxyPort int) (localPort, remotePort int, err error) {
-	conns, err := p.PingPeer(ip, localPorts, remotePorts, maxTTL, 1)
+	conns, err := p.PingProviderPeer(ip, localPorts, remotePorts, maxTTL, 1)
```
zolia: Is it ok to use maxTTL from the beginning? maxTTL should be set only if we receive message from peer.

Now, provider's router might treat incoming message from consumer as ddos and disable this consumer for some time. Even if provider sends message to this consumer, router keeps this consumer disabled for some time. We need incremental ttl here too.

soffokl: This is a consumer part that sets maxTTL from the very beginning. 
The provider starts listening and sending pings before sending config to the consumer. 
So consumer send first ping when provider router already expecting it.

nat/traversal/pinger.go
```diff
@@ -282,7 +425,7 @@ func (p *Pinger) multiPing(ip string, localPorts, remotePorts []int, initialTTL
 		go func(i int) {
 			conn, err := p.singlePing(ip, localPorts[i], remotePorts[i], initialTTL+i, stop)
 
-			ch <- pingResponse{conn, err}
+			ch <- pingResponse{conn: conn, err: err, id: i}
 
 			wg.Done()
```
zolia: maybe move this to beginning of function and `defer wg.Done()` ?

PR #1903, opened 2020-03-23
P2P messages

soffokl:


core/service/subscription.go
```diff
@@ -0,0 +1,117 @@
...
+			AccountantID:   identity.FromAddress(sr.GetConsumer().GetAccountantID()),
+			PaymentVersion: session.PaymentVersion(sr.GetConsumer().GetPaymentVersion()),
+		}
+
+		paymentVersion := string(session.PaymentVersionV3)
+		session := &session.Session{ID: session.ID(id)}
+		config, err := service.ProvideConfig(string(session.ID), consumerConfig)
+		if err != nil {
+			log.Error().Err(err).Msgf("Cannot get provider config for session %s", string(session.ID))
+			return err
```
anjmao: Internally p2p handler will also log error using log.Error().Err. I suggest instead only return formatted errors and do not log here.

```go
return fmt.Errorf("cannot get provider config for session %s", string(session.ID)")
```

core/service/subscription.go
```diff
@@ -0,0 +1,117 @@
...
+		}
+
+		if config.SessionDestroyCallback != nil {
+			go func() {
+				<-session.Done
+				config.SessionDestroyCallback()
+			}()
+		}
+
+		data, err := json.Marshal(config.SessionServiceConfig)
```
anjmao: Would be nice to put SessionServiceConfig into proto too. SessionResponse Config field can use oneof proto feature.

PR #1904, opened 2020-03-23
Allow for R recovery

### LINKED ISSUE #1585: Reveal R mechanism

vkuznecovas:
The provider will now send an encrypted R to the accountant so that he can store it.

Once there, the encrypted R will be sent back to us if a missmatch occurs.

Upon receiving such an error, a provider will decrypt the R and send it back to accountant.

closes #1585


vkuznecovas:
The required changes are already live on accountant and are backwards compatible.

session/pingpong/accountant_caller.go
```diff
@@ -206,4 +211,27 @@ var accountantCauseToError = map[string]error{
...
+}
+
+var encryptedRRegexp = regexp.MustCompile(`Encrypted recovery data: "(?P<recoverydetails>[0-9a-f]+)"`)
+
+type rRecoveryDetails struct {
+	R           string `json:"r"`
+	AgreementID uint64 `json:"agreement_id"`
+}
+
+func parseErrNeedsRRecoveryDetails(in string) (string, error) {
```
Waldz: Looks bogus. Why parsing from error message is needed at all?

vkuznecovas: Because I don't want to do extra requests.

tadaskay: I think what Waldz meant to ask is: perhaps it's possible to use a more convenient struct than a string?

anjmao: Agree. Maybe returning recovery details in separate field could help?

session/pingpong/accountant_caller.go
```diff
@@ -29,13 +30,65 @@ import (
...
+func (aer *AccountantErrorResponse) UnmarshalJSON(data []byte) error {
+	var s struct {
+		CausedBy     string `json:"cause"`
+		ErrorMessage string `json:"message"`
+		ErrorData    string `json:"data"`
+	}
+
+	err := json.Unmarshal(data, &s)
+	if err != nil {
+		return err
```
zolia: would wrap another context here.

session/pingpong/invoice_tracker.go
```diff
@@ -271,9 +279,25 @@ func (it *InvoiceTracker) requestPromise(r []byte, pm crypto.ExchangeMessage) er
...
 	}
 
+	details := rRecoveryDetails{
+		R:           hex.EncodeToString(r),
+		AgreementID: pm.AgreementID,
+	}
+
+	bytes, err := json.Marshal(details)
+	if err != nil {
+		return errors.Wrap(err, "could not marshal R recovery details")
```
zolia: old wrap?

session/pingpong/invoice_tracker.go
```diff
@@ -486,30 +510,64 @@ func (it *InvoiceTracker) waitForInvoicePayment(hlock []byte) {
...
-		ErrAccountantNotFound,
-		ErrAccountantNoPreviousPromise,
-		ErrAccountantMalformedJSON:
+		stdErr.Is(err, ErrAccountantInternal),
+		stdErr.Is(err, ErrAccountantNotFound),
+		stdErr.Is(err, ErrAccountantNoPreviousPromise),
+		stdErr.Is(err, ErrAccountantMalformedJSON):
 		// these are ignorable, we'll eventually fail
 		if it.incrementAccountantFailureCount() > it.deps.MaxAccountantFailureCount {
 			return err
```
zolia: We should wrap context here.

session/pingpong/invoice_tracker.go
vkuznecovas: This whole component needs to be changed, will do it in a separate pr someday.

session/pingpong/invoice_tracker.go
vkuznecovas: the context is added above.

PR #1905, opened 2020-03-23
Validate latest promise when requesting accountant data

### LINKED ISSUE #1721: Validate promise from accountant, don’t trust blindly
When a consumer is asking for his last promise from Accountant (e.g. in event of losing own state and re-sinking information about his last given premises) he should validate promise signature to ensure that it's his own promise.

vkuznecovas:
Promise will now be checked and the signer must match the requested identity.

Closes #1721

### DIFF
--- a/session/pingpong/accountant_caller.go
+++ b/session/pingpong/accountant_caller.go
@@ -170,6 +170,12 @@ type LatestPromise struct {
 
 // isValid checks if the promise is really issued by the given identity
 func (lp LatestPromise) isValid(id string) error {
+	// if we've not promised anything, that's fine for us.
+	// handles the case when we've just registered the identity.
+	if lp.Amount == 0 {
+		return nil
+	}
+
 	decodedChannelID, err := hex.DecodeString(strings.TrimPrefix(lp.ChannelID, "0x"))
 	if err != nil {
 		return errors.Wrap(err, "could not decode channel ID")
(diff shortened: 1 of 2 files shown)


session/pingpong/accountant_caller.go
```diff
@@ -96,7 +99,20 @@ func (ac *AccountantCaller) GetConsumerData(id string) (ConsumerData, error) {
 	}
 	var resp ConsumerData
 	err = ac.doRequest(req, &resp)
-	return resp, errors.Wrap(err, "could not request consumer data accountant")
+	if err != nil {
+		return ConsumerData{}, errors.Wrap(err, "could not request consumer data from accountant")
+	}
+
+	valid, err := resp.LatestPromise.isValid(id)
```
tadaskay: I suppose it's enough to return err by joining `!valid` to the error case, then there's no need to double-check because the promise is invalid in both cases:
```
err = resp.LatestPromise.isValid(id)
if err != nil {
		return ConsumerData{}, errors.New("accountant sent an invalid promise for the ID")
	}
```

session/pingpong/accountant_caller.go
```diff
@@ -96,7 +99,16 @@ func (ac *AccountantCaller) GetConsumerData(id string) (ConsumerData, error) {
 	}
 	var resp ConsumerData
 	err = ac.doRequest(req, &resp)
-	return resp, errors.Wrap(err, "could not request consumer data accountant")
+	if err != nil {
+		return ConsumerData{}, errors.Wrap(err, "could not request consumer data from accountant")
```
tadaskay: You can use `fmt.Errorf("could not request consumer data from accountant: %w", err)`

Issue #1844, opened 2020-03-04
Event-stream API for consumer state


tadaskay:
### Rationale

1. Lots of polling needed to get the current state
2. Lots of redundant endpoints (connection/ip, connection/location, etc)

### Suggestion

Implement event-stream based API, similar to the one we have for the provider mode (`/events/state`). We might re-use existing https://github.com/mysteriumnetwork/node/blob/master/core/state/state.go#L56 by adding consumer state to it.

tadaskay:
We could also consider graphql: it supports queries and subscriptions using the same model. I think it would fit our use case perfectly.

chompomonim:
Let's do it in a similar way as event-streaming for providers. GraphQL is good idea, but it is expensive to rewrite all tequila and related packages that are using it. And supporting two different APIs doesn't sound as proper solution.

PR #1906, opened 2020-03-23
Use same DTOs for tequilapi server and client (identity)

tadaskay:
The same model will also be used in SSE.

Ref: #1844

### DIFF
--- a/tequilapi/contract/identity.go
+++ b/tequilapi/contract/identity.go
@@ -22,7 +22,7 @@ import (
 	"github.com/mysteriumnetwork/node/tequilapi/validation"
 )
 
-// IdentityDTO holds identity address.
+// IdentityDTO holds identity information.
 // swagger:model IdentityDTO
 type IdentityDTO struct {
 	// identity in Ethereum address format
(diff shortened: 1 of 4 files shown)


tequilapi/contract/identity.go
```diff
@@ -0,0 +1,111 @@
...
+ */
+
+package contract
+
+import (
+	"github.com/mysteriumnetwork/node/identity"
+	"github.com/mysteriumnetwork/node/tequilapi/validation"
+)
+
+// IdentityDTO holds identity address.
```
soffokl:
```suggestion
// IdentityDTO holds identity information.
```

tequilapi/contract/identity.go
```diff
@@ -0,0 +1,111 @@
...
+	}
+	for i, id := range ids {
+		result.Identities[i] = NewIdentityDTO(id)
+	}
+	return result
+}
+
+// IdentityGetCurrentRequest request to get/create an identity.
+// swagger:model CurrentIdentityDTO
+type IdentityGetCurrentRequest struct {
```
soffokl: `IdentityCreateRequest`, `IdentityUnlockRequest` and `IdentityGetCurrentRequest` are the same, do we need all of them?

tadaskay: Merged into `IdentityRequest`. Thanks!

PR #1910, opened 2020-03-24
Publish identities to SSE state

tadaskay:
Ref #1844

Identity list is fetched on the initial state construction and then updated on:

- AppTopicIdentityCreated (re-fetches the whole list, to ensure the same order as the identity manager)
- AppTopicIdentityRegistration
- AppTopicBalanceChanged (Which is also triggered by AppTopicIdentityUnlock)

### DIFF
--- a/core/state/event/event.go
+++ b/core/state/event/event.go
@@ -21,6 +21,7 @@ import (
 	"time"
 
 	"github.com/mysteriumnetwork/node/core/connection"
+	"github.com/mysteriumnetwork/node/identity/registry"
 	"github.com/mysteriumnetwork/node/market"
 )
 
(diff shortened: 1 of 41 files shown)


core/state/event/event.go
```diff
@@ -29,10 +30,18 @@ const AppTopicState = "State change"
...
+	NATStatus  NATStatus        `json:"natStatus"`
+	Services   []ServiceInfo    `json:"serviceInfo"`
+	Sessions   []ServiceSession `json:"sessions"`
+	Consumer   ConsumerState    `json:"consumer"`
+	Identities []Identity       `json:"identities"`
+}
+
+// Identity represents identity and its status.
+type Identity struct {
+	Address            string                      `json:"id"`
```
Waldz: These JSON mapping not needed here, as you're remapping to `contract.IdentityDTO` later.
Is this intermediate structure needed at all?

tequilapi/endpoints/sse_handler.go
```diff
@@ -222,6 +224,15 @@ type consumerConnectionRes struct {
 }
 
 func mapState(event stateEvent.State) stateRes {
+	identitiesRes := make([]contract.IdentityDTO, len(event.Identities))
+	for idx, identity := range event.Identities {
+		identityRes := contract.IdentityDTO{
```
Waldz: confusing place - `contract` package with single model

tadaskay: This is different PR: https://github.com/mysteriumnetwork/node/pull/1906
I've extracted the contract to a separate package for re-use between the server and the client. Not refactoring all at once, but rather following the boy scout rule. What is confusing?

core/state/state.go
```diff
@@ -100,31 +112,45 @@ func NewKeeper(
...
 }
 
+func (k *Keeper) fetchIdentities() []stateEvent.Identity {
+	providedIdentities := k.deps.IdentityProvider.GetIdentities()
+	identities := make([]event.Identity, len(providedIdentities))
+	for idx, providedIdentity := range providedIdentities {
+		status, err := k.deps.IdentityRegistry.GetRegistrationStatus(providedIdentity)
+		if err != nil {
+			log.Warn().Err(err).Msgf("Could not get registration status for %s", providedIdentity.Address)
+			status = registry.Unregistered
```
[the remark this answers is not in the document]
tadaskay: If an error is returned, the result should generally be discarded. What you are saying is an implementation detail of the registry.

core/state/event/event.go
tadaskay: What are you suggesting to replace it with?

Issue #1731, opened 2020-02-18
Add into node and WebUI to show how much node has earned in total


chompomonim:
There are a couple of related tasks.

- [x]  Expose totally earned token amount into Tequilla. (lifetime earnings, count from a latest promise from an accountant).
- [x] Expose total earned amount in WebUI and CLI.
- [x] Show in logs session earned amount. (in consumer as well).

Waldz:
**Balance (to spend)** = BCBalance - (accountantPromised - BCSettled)
```
caller, _ := bindings.NewMystTokenCaller()
BCBalance = caller.BalanceOf()

caller, _ := bindings.NewChannelImplementation()
BCSettled = caller.Accountant().Settled

accountantPromised = boltDB.issued_invoices.AgreementTotal
```

**Earnings** = boltDB.accountant_promises.Promise.Amount
**Earnings (to settle)** = boltDB.accountant_promises.Promise.Amount - BCSettled
```
caller, _ := bindings.NewAccountantImplementationCaller(boltDB.accountant_promises.Promise.ChannelID)
BCSettled := caller.Channels().Settled
```

chompomonim:
I have created separate issue for better "Balance (to spend)" calculation: #1869 1869

PR #1913, opened 2020-03-25
Expose earnings balance to Tequilapi

Waldz:
Updates: #1731 

Tequilapi now haves lifetime earnings and unsettled earnings fields next to identity.
Also E2E to test for simple value transfer from consumer balance to provider earnings.


cmd/commands/cli/command_identities.go
```diff
@@ -107,6 +107,8 @@ func (c *cliApp) getIdentity(actionArgs []string) {
 	info("Channel address:", identityStatus.ChannelAddress)
 	info("Balance:", identityStatus.Balance)
 	info("Balance estimate:", identityStatus.BalanceEstimate)
+	info("Earnings:", identityStatus.Earnings)
```
vkuznecovas: end users will be confused by the earnings and earnings total.

vkuznecovas: I would suggest a better name but can't think of any.

tequilapi/endpoints/identities.go
```diff
@@ -32,37 +32,18 @@ import (
 )
 
 type balanceGetter func(id identity.Identity) uint64
-
 type balanceFetcher func(id string) (pingpong.ConsumerData, error)
+type earningsGetter func(id identity.Identity) uint64
```
tadaskay: Such function types make it hard to navigate to the relevant implementation. I have to go to the constructor, find the matching parameter, then look for the usages of the constructor and determine matching argument (which is usually positional). Sometimes there are more layers, thus multiplying the time needed to navigate to the implementation.

Were it written as:
```go
type earningsGetter interface {
	Balance(id identity.Identity) uint64
}
```
You could use ⌘+⌥+B or ⌘+Click to navigate directly to implementation(s). For easier navigation, I suggest using an interface with a single method.
In case you're wondering, I'll also propose it to the devbook.

e2e/connection_test.go
```diff
@@ -86,8 +99,14 @@ func identityRegistrationFlow(t *testing.T, tequilapi *tequilapi_client.Client,
 		regStatus, err := tequilapi.IdentityRegistrationStatus(id)
 		return regStatus.Registered, err
 	})
+	assert.NoError(t, err)
 
+	idStatus, err := tequilapi.IdentityStatus(id)
 	assert.NoError(t, err)
+	assert.Equal(t, "RegisteredConsumer", idStatus.RegistrationStatus)
+	assert.Equal(t, uint64(690000000), idStatus.Balance)
+	assert.Equal(t, uint64(0), idStatus.Earnings)
```
tadaskay: You could use `assert.Zero(t, ...)`

e2e/connection_test.go
```diff
@@ -181,6 +200,27 @@ func consumerConnectFlow(t *testing.T, tequilapi *tequilapi_client.Client, consu
...
+
+func providerEarnedTokens(t *testing.T, tequilapi *tequilapi_client.Client, id string, earningsExpected uint64) {
+	// Before settlement
+	providerStatus, err := tequilapi.IdentityStatus(id)
+	assert.NoError(t, err)
+	assert.Equal(t, uint64(690000000), providerStatus.Balance)
+	assert.Equal(t, earningsExpected, providerStatus.Earnings)
+	assert.Equal(t, earningsExpected, providerStatus.EarningsTotal)
+	// TODO Compare with sessions stats + proposal price
+	assert.True(t, providerStatus.Earnings > uint64(500), "earnings should be at least 500 but is ", providerStatus.Earnings)
```
tadaskay: If you're providing args, this should be a format string, hence `%d` is missing. As it is now, it will print: 
```
earnings should be at least 500 but is %!(EXTRA int=5)
```

tequilapi/endpoints/identities.go
Waldz: I known, i wrote this before communication. Refactored.

e2e/connection_test.go
Waldz: Added

cmd/commands/cli/command_identities.go
Waldz: Source https://github.com/mysteriumnetwork/node/issues/1731.
ATM can't think off anything better :/

PR #1916, opened 2020-03-25
Expose identity earnings to SSE state

Waldz:
Updates: #1731

Web UI now can interactively show increasing earnings


core/state/state.go
```diff
@@ -369,6 +380,29 @@ func (k *Keeper) consumeBalanceChangedEvent(e interface{}) {
 	go k.announceStateChanges(nil)
 }
 
+func (k *Keeper) consumeEarningsChangedEvent(e interface{}) {
+	k.lock.Lock()
+	defer k.lock.Unlock()
+	evt, ok := e.(pingpong.AppEventEarningsChanged)
+	if !ok {
+		log.Warn().Msg("Received a wrong kind of event for earnings change")
```
soffokl: Should we return here?

core/state/state.go
```diff
@@ -369,6 +380,29 @@ func (k *Keeper) consumeBalanceChangedEvent(e interface{}) {
...
+	}
```
tadaskay:
```suggestion
		log.Warn().Msg("Received a wrong kind of event for earnings change")
		return
	}
```

core/state/state_test.go
```diff
@@ -567,3 +613,21 @@ func interacted(c interactionCounter, times int) func() bool {
 		return c.interactions() == times
 	}
 }
+
+type mockBalanceProvider struct {
+	Balance uint64
+}
+
+// GetBalance returns a pre-defined balance.
+func (mbp *mockBalanceProvider) GetBalance(_ identity.Identity) uint64 {
```
tadaskay: This is identical to `mocks.BalanceProvider`

core/state/state.go
Waldz: Added

core/state/state_test.go
Waldz: I moved mocks next to tests, because keeping them global packages causes circular dependencies: `pingpong -> mocks -> connections -> pingpong`

tadaskay: There is no dependency on `connections` in the mocks package. Either way, mocks should have minimal dependencies.

Waldz: I added similar mock `EarningsProvider.SettlementState(_ identity.Identity) pingpong.SettlementState` and `mocks` package starts to be interdependent.
Lets agree on having mocks next to package we're testing.

tadaskay: Such widely used types as `identity.Identity` or in this case `pingpong.SettlementState` could be in a leaf package with minimal dependencies - that will solve the package cycle. If it's a one-off mock - then it's good to have it next to the test. If it's used multiple times, then the maintenance becomes a burden and it should be in `mocks`.

PR #1919, opened 2020-03-26
Snakify tequilapi

tadaskay:
Changed only tequilapi.
With subsequent PR I'll update the web UI.

### DIFF
--- a/ci/test/test.go
+++ b/ci/test/test.go
@@ -29,7 +29,7 @@ func TestWithCoverage() error {
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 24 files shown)


vkuznecovas:
I'm a bit concerned about third parties. Won't we break our js libraries, afaik someone is depending on that?

zolia:
> I'm a bit concerned about third parties. Won't we break our js libraries, afaik someone is depending on that?

We probably should be making JS release too.

tadaskay:
mysterium-vpn-js will be updated (and included in web UI)

PR #1921, opened 2020-03-26
Fix balance for good

### LINKED ISSUE #1724: Rework balance calculations after topup
Node (consumer) should not just get the balance from an accountant but should watch blockchain for popup events and update his own local state after seeing a popup event.

The only time node should get the balance from an accountant is in case of lost own state.

### LINKED ISSUE #1869: Make accurate way of consumers balance calculation
When in CLI consumer is asking to show balance, this is an ideal time to recalculate "true" balance.

Consumer balance can be calculated this way: 
Balance (to spend) = BCBalance - (accountantPromised - BCSettled)

```
caller, _ := bindings.NewMystTokenCaller()
BCBalance = caller.BalanceOf()

caller, _ := bindings.NewChannelImplementation()
BCSettled = caller.Accountant().Settled

accountantPromised = boltDB.issued_invoices.AgreementTotal
```

We should also store in boltdb last issued promise. If it's not there, we should ask for own latest promise from the accountant. 

Ideally, we could also have "recalculate balance" button in WebUI.

vkuznecovas:
1) Removed local BC client in favor of payments BC client.
2) Simplified the balance calculations quite drastically.

Accountant is now exclusively used to recover grand total promised only.

The balance seems to be way more consistent now.

Fixes #1869
Fixes #1724


tequilapi/endpoints/identities.go
```diff
@@ -273,21 +274,19 @@ func (endpoint *identitiesAPI) Status(resp http.ResponseWriter, _ *http.Request,
...
-	if regStatus.Registered() {
-		consumer, err = endpoint.consumerProvider.GetConsumerData(id.Address)
-		if err != nil {
-			utils.SendError(resp, errors.Wrap(err, "failed to check balance status"), http.StatusInternalServerError)
-			return
-		}
+	balance := endpoint.balanceProvider.ForceBalanceUpdate(id)
+	addr, err := endpoint.channelAddressCalculator.GetChannelAddress(id)
+	if err != nil {
+		utils.SendError(resp, errors.Wrap(err, "failed to calculate channel address"), http.StatusInternalServerError)
```
zolia: we should use %w for new code.

tequilapi/endpoints/identities.go
```diff
@@ -273,21 +275,19 @@ func (endpoint *identitiesAPI) Status(resp http.ResponseWriter, _ *http.Request,
...
 
 	status := contract.IdentityDTO{
 		Address:            address,
 		RegistrationStatus: regStatus.String(),
-		ChannelAddress:     consumer.ChannelID,
-		Balance:            consumer.Balance,
-		BalanceEstimate:    endpoint.balanceProvider.GetBalance(id),
+		ChannelAddress:     addr.Hex(),
+		Balance:            balance,
+		BalanceEstimate:    balance,
```
Waldz: Balance estimate can be dropped now

PR #1936, opened 2020-03-27
Simplify balance further

vkuznecovas:
After discussions, decided to add a struct wrapping the calculations,
as well as emitting a single grand total change event for simplicity.

### DIFF
--- a/session/pingpong/consumer_balance_tracker.go
+++ b/session/pingpong/consumer_balance_tracker.go
@@ -35,7 +35,7 @@ import (
 // TODO: this needs to take into account the saved state.
 type ConsumerBalanceTracker struct {
 	balancesLock sync.Mutex
-	balances     map[identity.Identity]uint64
+	balances     map[identity.Identity]ConsumerBalance
 
 	accountantAddress        common.Address
 	mystSCAddress            common.Address
@@ -57,7 +57,7 @@ func NewConsumerBalanceTracker(
 	consumerGrandTotalGetter consumerGrandTotalGetter,
 ) *ConsumerBalanceTracker {
 	return &ConsumerBalanceTracker{
-		balances:                 make(map[identity.Identity]uint64),
+		balances:                 make(map[identity.Identity]ConsumerBalance),
 		consumerBalanceChecker:   consumerBalanceChecker,
 		mystSCAddress:            mystSCAddress,
 		accountantAddress:        accountantAddress,
@@ -91,11 +91,7 @@ func (cbt *ConsumerBalanceTracker) Subscribe(bus eventbus.Subscriber) error {
 	if err != nil {
 		return err
 	}
-	err = bus.SubscribeAsync(AppTopicExchangeMessage, cbt.handleExchangeMessageEvent)
-	if err != nil {
-		return err
-	}
-	err = bus.SubscribeAsync(AppTopicGrandTotalRecovered, cbt.handleGrandTotalRecovered)
+	err = bus.SubscribeAsync(AppTopicGrandTotalChanged, cbt.handleGrandTotalChanged)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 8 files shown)


session/pingpong/consumer_balance_tracker.go
```diff
@@ -129,8 +125,13 @@ func (cbt *ConsumerBalanceTracker) handleUnlockEvent(id string) {
 	cbt.ForceBalanceUpdate(identity)
 }
 
-func (cbt *ConsumerBalanceTracker) handleGrandTotalRecovered(ev GrandTotalRecovered) {
-	cbt.ForceBalanceUpdate(ev.Identity)
+func (cbt *ConsumerBalanceTracker) handleGrandTotalChanged(ev AppEventGrandTotalChanged) {
+	if _, ok := cbt.balances[ev.ConsumerID]; !ok {
```
anjmao: Missing lock?

session/pingpong/consumer_balance_tracker.go
```diff
@@ -129,8 +125,13 @@ func (cbt *ConsumerBalanceTracker) handleUnlockEvent(id string) {
 	cbt.ForceBalanceUpdate(identity)
 }
 
-func (cbt *ConsumerBalanceTracker) handleGrandTotalRecovered(ev GrandTotalRecovered) {
-	cbt.ForceBalanceUpdate(ev.Identity)
+func (cbt *ConsumerBalanceTracker) handleGrandTotalChanged(ev AppEventGrandTotalChanged) {
```
Waldz: I suggest to pass to what amount it changed in `AppEventGrandTotalChanged `, and you would need to call `consumerGrandTotalGetter.Get()` or couple to this storage at all

session/pingpong/consumer_balance_tracker.go
```diff
@@ -214,34 +216,36 @@ func (cbt *ConsumerBalanceTracker) handleStopEvent() {
...
+		cbt.balances[id] = v
+	} else {
+		cbt.ForceBalanceUpdate(id)
 	}
 
-	current := safeAdd(before, b)
-	cbt.balances[id] = current
-
-	go cbt.publishChangeEvent(id, before, current)
+	go cbt.publishChangeEvent(id, before, cbt.balances[id].GetBalance())
```
Waldz: Maybe publish full struct as it changes

session/pingpong/invoice_payer_test.go
```diff
@@ -689,13 +700,19 @@ func TestInvoicePayer_issueExchangeMessage(t *testing.T) {
...
 	err        error
 	calledWith uint64
 }
 
-func (mcts *mockConsumerTotalsStorage) Store(providerAddress, accountantAddress string, amount uint64) error {
+func (mcts *mockConsumerTotalsStorage) Store(consumerAddress, accountantAddress string, amount uint64) error {
 	mcts.calledWith = amount
+	go mcts.bus.Publish(AppTopicGrandTotalChanged, AppEventGrandTotalChanged{
+		Current:      amount,
+		AccountantID: identity.FromAddress(accountantAddress),
```
Waldz: `Identity` struct is construct of local MYST accounts, no point use this for `accountantAddress`

session/pingpong/consumer_totals_storage.go
```diff
@@ -29,20 +31,34 @@ const consumerTotalStorageBucketName = "consumer_promised_totals"
...
-func NewConsumerTotalsStorage(bolt persistentStorage) *ConsumerTotalsStorage {
+func NewConsumerTotalsStorage(bolt persistentStorage, bus eventbus.Publisher) *ConsumerTotalsStorage {
 	return &ConsumerTotalsStorage{
 		bolt: bolt,
+		bus:  bus,
 	}
 }
 
 // Store stores the given amount as promised for the given channel.
 func (cts *ConsumerTotalsStorage) Store(consumerAddress, accountantAddress string, amount uint64) error {
```
Waldz: Probably can pass `consumerID` your validated value in upper layer, now it recalculated second time

session/pingpong/invoice_payer_test.go
vkuznecovas: There's a point, they format the addresses differently therefore all the addresses should be stored in identity format.

session/pingpong/consumer_totals_storage.go
vkuznecovas: what?

session/pingpong/consumer_balance_tracker.go
vkuznecovas: I'll break a ton of users if i do that, so no, it's staying here :)

session/pingpong/consumer_balance_tracker.go
vkuznecovas: I'm already passing in the amount.

session/pingpong/consumer_balance_tracker.go
vkuznecovas: actually - no, both of the cases are internally locked aftewards, so we're fine here.

Issue #1932, opened 2020-03-26
P2P: Enable channel on consumer / provider sides


anjmao:
1. Check and fill any missing session creation config from provider side (DNSs fields etc.).
2. Pass second punched UDP connection for consumer / provider when needed.
3. Make sure p2p channels are closed properly when no longer used. 
4. Test backwards compatibility (old consumer, new provider), (new consumer, old provider).

Issue #1931, opened 2020-03-26
P2P: Fix nat hole punching


anjmao:
P2P need to at least 2 UDP connections punched. For that we are using new methods pinger.PingConsumerPeer and pinger.PingProviderPeer, which sends ping ack messages in order to synchronise connection orders. For some networks with higher latency it's stable. Need to investigate.

PR #1946, opened 2020-03-30
Enable P2P communication by default

soffokl:
Closes: #1931 
Closes: #1932


mobile/mysterium/openvpn_connection_setup.go
```diff
@@ -153,8 +153,8 @@ func (c *openvpnConnection) Start(options connection.ConnectOptions) error {
 		return err
 	}
 
-	// TODO this backward compatibility check needs to be removed once we will start using port ranges for all peers.
-	if sessionConfig.LocalPort > 0 || len(sessionConfig.Ports) > 0 {
+	// TODO this backward compatibility block needs to be removed once we will fully migrate to the p2p communication.
+	if len(sessionConfig.Ports) > 0 {
```
anjmao: Mobile openvpn also need to use 2 NAT UDP conn, the same as in wireguard_connection_setup `options.ProviderNATConn`

soffokl: It's done in the `NewClientConfigFromSession` which is shared between mobile and desktop.
https://github.com/mysteriumnetwork/node/pull/1946/files#diff-2a60cf8a8b871ed9f1078ce9e47c33f2R103-R106

services/openvpn/service/manager.go
```diff
@@ -264,50 +266,77 @@ func (m *Manager) ProvideConfig(_ string, sessionConfig json.RawMessage) (*sessi
...
-		err := json.Unmarshal(sessionConfig, &consumerConfig)
-		if err != nil {
-			return nil, errors.Wrap(err, "could not parse consumer config")
-		}
-
-		if len(consumerConfig.Ports) == 0 {
-			cp, err := m.natPingerPorts.Acquire()
+	// TODO this backward compatibility block needs to be removed once we will fully migrate to the p2p communication.
+	if conn == nil {
+		// Older clients do not send any sessionConfig, but we should keep back compatibility and not fail in this case.
```
zolia: I think all clients that are now able to connect to our new nodes sends this config. We can remove this check.

services/openvpn/service/manager.go
```diff
@@ -264,50 +266,77 @@ func (m *Manager) ProvideConfig(_ string, sessionConfig json.RawMessage) (*sessi
...
+		if sessionConfig != nil && len(sessionConfig) > 0 && m.natPinger.Valid() {
+			var consumerConfig openvpn_service.ConsumerConfig
+			err := json.Unmarshal(sessionConfig, &consumerConfig)
 			if err != nil {
-				return nil, err
+				return nil, errors.Wrap(err, "could not parse consumer config")
```
zolia: we should use new %w wrap here already.

services/openvpn/service/manager.go
```diff
@@ -264,50 +266,77 @@ func (m *Manager) ProvideConfig(_ string, sessionConfig json.RawMessage) (*sessi
...
+		}
+	} else {
+		localPort, err := net.ResolveUDPAddr("udp4", fmt.Sprintf("127.0.0.1:%d", m.vpnServerPort))
+		if err != nil {
+			return nil, err
 		}
-	}
 
+		openVPNProxy, _ := net.DialUDP("udp", nil, localPort)
+		go copyStreams(openVPNProxy, conn)
```
zolia: It is weird to see this logic inside ProvideConfig.

services/openvpn/service/manager.go
```diff
@@ -264,50 +266,77 @@ func (m *Manager) ProvideConfig(_ string, sessionConfig json.RawMessage) (*sessi
...
+		go copyStreams(conn, openVPNProxy)
+	}
 	return &session.ConfigParams{SessionServiceConfig: vpnConfig, TraversalParams: traversalParams}, nil
 }
 
+func copyStreams(dstConn *net.UDPConn, srcConn *net.UDPConn) {
```
zolia: Can we move this generic copy function outside openvpn service manager? I understand that it used by openvpn only atm.

session/manager.go
```diff
@@ -56,7 +57,7 @@ type publisher interface {
 
 // ConfigProvider is able to handle config negotiations
 type ConfigProvider interface {
-	ProvideConfig(sessionID string, sessionConfig json.RawMessage) (*ConfigParams, error)
+	ProvideConfig(sessionID string, sessionConfig json.RawMessage, conn *net.UDPConn) (*ConfigParams, error)
```
zolia: i would not inject conn here, would pass just whats required: 
i.e.: `remoteConn.LocalAddr().(*net.UDPAddr).Port`
We should handle streams separately, not in config provider.

services/openvpn/service/manager.go
soffokl: Removed it.

services/openvpn/service/manager.go
soffokl: Changed.

services/openvpn/service/manager.go
soffokl: I moved this to the new file. You are right it's used only for OpenVPN transport now. I'd want to have it close to the place where it's used until we will have another need for it.

services/openvpn/service/manager.go
soffokl: Config provider is not just config generation for a long time. It's preparing the environment including and starts everything that is required. 
We probably need to rename it to the better names, but let's do it in another PR.

session/manager.go
soffokl: We cannot just send LocalPort, it depends on transport what to do with a connection.
OpenVPN needs to start proxy to connect multiple clients to the single OpenVPN process, so it reuses that connection.

Wireguard is closing the connection and starts its own handlers on that ports. We even can use both LocalPort and RemotePort to make the connection more secure.

We never know what will be next transport and how it will need to handle the provided connection.

zolia: Ok, then renaming `ProvideConfig` to more sensible naming would do for some time :)

Issue #1952, opened 2020-03-30
Android app crash with nil pointer dereference in AccountantPromiseSettler


anjmao:
Testing master branch. App crashes.
```
2020-03-30 14:05:26.261 26266-26359/network.mysterium.vpn E/GoLog: 2020-03-30T11:05:26.000 INF ext-go/1/src/github.com/mysteriumnetwork/node/firewall/incoming_firewall_noop.go:53 > Allow URL https://api.ipify.org/?format=json access
2020-03-30 14:05:26.261 26266-26359/network.mysterium.vpn E/GoLog: 2020-03-30T11:05:26.000 INF ext-go/1/src/github.com/mysteriumnetwork/node/firewall/outgoing_firewall_noop.go:57 > Allow URL https://testnet-location.mysterium.network/api/v1/location access
2020-03-30 14:05:26.261 26266-26359/network.mysterium.vpn E/GoLog: 2020-03-30T11:05:26.000 INF ext-go/1/src/github.com/mysteriumnetwork/node/firewall/incoming_firewall_noop.go:53 > Allow URL https://testnet-location.mysterium.network/api/v1/location access
2020-03-30 14:05:26.261 26266-26359/network.mysterium.vpn E/GoLog: 2020-03-30T11:05:26.000 DBG ext-go/1/src/github.com/mysteriumnetwork/node/cmd/di.go:817 > Experimental NAT punching enabled, creating a pinger
2020-03-30 14:05:26.261 26266-26359/network.mysterium.vpn E/GoLog: 2020-03-30T11:05:26.000 DBG ext-go/1/src/github.com/mysteriumnetwork/node/config/config.go:185 > Returning default value access-policy.list:<nil>
2020-03-30 14:05:26.262 26266-26359/network.mysterium.vpn E/GoLog: 2020-03-30T11:05:26.000 DBG ext-go/1/src/github.com/mysteriumnetwork/node/config/config.go:185 > Returning default value access-policy.address:<nil>
2020-03-30 14:05:26.262 26266-26359/network.mysterium.vpn E/GoLog: 2020-03-30T11:05:26.000 DBG ext-go/1/src/github.com/mysteriumnetwork/node/config/config.go:185 > Returning default value access-policy.fetch:<nil>
[...]

PR #1954, opened 2020-03-30
Fix android crash on startup

anjmao:
1. Fix for https://github.com/mysteriumnetwork/node/issues/1952
2. Add e2e test for mobile entry point.
3. Remove di_mobile.go and use flag to determine if specific logic is needed when consumer is mobile.


e2e/runner.go
```diff
@@ -172,3 +175,16 @@ func (r *Runner) stopProviderConsumerNodes(providerHost, consumerHost string) er
 	}
 	return nil
 }
+
+// replaceMobileOpenVPNWithNoop replaces go-openvpn with noop package so can compile mobile entry without any C++ dependencies.
+func (r *Runner) replaceMobileOpenVPNToNoop() {
+	if err := sh.Run("sed -i='' 's+github.com/mysteriumnetwork/go-openvpn/openvpn3+github.com/mysteriumnetwork/node/mobile/mysterium/openvpn3+g' ./mobile/mysterium/openvpn_connection_setup.go"); err != nil {
```
anjmao: God forgive me :(

cmd/di.go
```diff
@@ -160,7 +160,7 @@ type Dependencies struct {
 	ConsumerTotalsStorage    *pingpong.ConsumerTotalsStorage
 	AccountantPromiseStorage *pingpong.AccountantPromiseStorage
 	ConsumerBalanceTracker   *pingpong.ConsumerBalanceTracker
-	AccountantPromiseSettler *pingpong.AccountantPromiseSettler
+	AccountantPromiseSettler pingpong.AccountantPromiseSettlerInterface
```
soffokl: Can we remove the `Interface` word from the name? It's already clear that this is an interface, no need to duplicate it in the name. 
```suggestion
	AccountantPromiseSettler pingpong.AccountantPromiseSettler
```

cmd/di_desktop.go
```diff
@@ -62,6 +61,10 @@ import (
 
 // bootstrapServices loads all the components required for running services
 func (di *Dependencies) bootstrapServices(nodeOptions node.Options, servicesOptions config.ServicesOptions) error {
+	if nodeOptions.MobileConsumer {
```
Waldz: What was wrong with build flags approach?
Because now file is di_desktop.go but full of conditional logic, which would be nice to avoid

anjmao: 1. I can't build mobile entry point using `go build ./mobile/mysterium` since di_mobile will not be included but di_mobile doesn't contain any OS specific logic.
2. There are multiple checks already with feature flags, example https://github.com/mysteriumnetwork/node/blob/master/cmd/di_desktop.go#L342. 
3. I can't write e2e test properly. If I add GOOS=android it will require specific C++, but for now I just want to write e2e test which tests that mobile entry point doesn't crash on startup, so if connect is not working at least use can sent feedback.

vkuznecovas: I too am for conditional compiling, if possible.

zolia: Agree with @Waldz here. We should not increase our codebase with platform checks inside if we can build accordingly.

anjmao: 80% of current mobile entrypoint API can be tested without any android specific build tag and most importantly it tests for possible crashes when some dependencies are not added to mobile . If you still want conditional compiling I can close this PR and add a separate task since it will require 3-4 days at least to make it work.

anjmao:
@Waldz @zolia @vkuznecovas I merge it as is for now. If you have ideas how to add android to our e2e tests fast and cheap without waisting time let me know :)

cmd/di_desktop.go
Waldz: So it's just for the test, but not released artefact, ASFAIK.
Is there a way to force a android tag in this test particularly. Maybe adding custom flag like "e2e" is what you need?

Issue #1877, opened 2020-03-11
Any version greater than 0.21.1 fails to load using Docker


christoefar:
When I update Mysterium past version 0.21.1 it fails to start correctly, I have attached a full log from the console but it looks to me to be related to iptables and/or the ipset command. 

[myst.txt](https://github.com/mysteriumnetwork/node/files/4319213/myst.txt)

`[90m2020-03-11T15:41:15.255[0m [31mWRN[0m [1mfirewall/incoming_firewall_iptables.go:62[0m[36m >[0m Error deleting ipset table. [31merror=[0m[31m"ipset cmd error: \"sudo ipset destroy myst-provider-dst-whitelist\": exit status 1 output: ipset v6.32: Kernel error received: Invalid argument\n"
`

`
[90m2020-03-11T15:41:15.265[0m [1m[31mERR[0m[0m [1mcmd/mysterium_node/mysterium_node.go:58 [0m[36m >[0m Failed to execute command: [31merror=[0m[31m"ipset cmd error: \"sudo ipset create myst-provider-dst-whitelist hash:ip --timeout 86400\": exit status 1 output: ipset v6.32: Kernel error received: Invalid argument\n"
`

 - OS: unRAID 6.8.0
 - Docker: If I use the :latest tag it doesn't work, so I have to force :0.21.1-alpine to get the node up and running.

pwnorbpwnd:
Can also confirm the same error and "workaround" on Unraid 6.8.3(Docker v19.03.5).

zolia:
Hello, 

I am not familiar with this OS, but from the error it looks like linux kernel does not have proper ipset support. Most likely ipset kernel module is missing. Try loading ipset module for the kernel or build a custom one compiling it in.

christoefar:
Thanks both for you suggestions.

@pwnorbpwnd I can confirm that with the latest version of unRAID (6.8.3) I get the same problem.

@zolia This would seem to make sense, unRAID has quite a conservative kernel. I will try and get the kernel module loaded and see how I get on.

Thanks again.

robweighill:
Has anyone found a solution to this issue yet? I am still having to run 0.21.1 on my UnRaid docker.

christoefar:
@robweighill since my knowledge on the Linux Kernel is very limited I contacted UnRaid.

They have said the ipset kernel modules can be included in the 6.9 release.

[...]

PR #1955, opened 2020-03-31
Make incoming firewall with ipset tool optional

soffokl:
Closes: #1877


soffokl:
@zolia 
Traffic slicing is options, not everyone needs it.

PR #1970, opened 2020-03-31
Add p2p sent messages logging

anjmao:
Some more logging for p2p messages.

### DIFF
--- a/core/connection/manager.go
+++ b/core/connection/manager.go
@@ -393,7 +393,7 @@ func (manager *connectionManager) createP2PSession(c Connection, p2pChannel p2p.
 		ProposalID: int64(proposal.ID),
 		Config:     config,
 	}
-	log.Info().Msgf("Sending P2P message to %q: %s", p2p.TopicSessionCreate, sessionRequest.String())
+	log.Debug().Msgf("Sending P2P message to %q: %s", p2p.TopicSessionCreate, sessionRequest.String())
 	res, err := p2pChannel.Send(p2p.TopicSessionCreate, p2p.ProtoMessage(sessionRequest))
 	if err != nil {
 		return session.SessionDto{}, session.PaymentInfo{}, fmt.Errorf("could not send p2p session create request: %w", err)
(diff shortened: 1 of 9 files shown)


core/connection/manager.go
```diff
@@ -384,17 +384,17 @@ func (manager *connectionManager) createP2PSession(c Connection, p2pChannel p2p.
...
 			PaymentVersion: string(session.PaymentVersionV3),
 		},
 		ProposalID: int64(proposal.ID),
 		Config:     config,
-	})
-
-	res, err := p2pChannel.Send(p2p.TopicSessionCreate, sessionRequest)
+	}
+	log.Info().Msgf("Sending P2P message to %q: %s", p2p.TopicSessionCreate, sessionRequest.String())
+	res, err := p2pChannel.Send(p2p.TopicSessionCreate, p2p.ProtoMessage(sessionRequest))
```
Waldz: Please do logging inside `p2pChannel.Send()`, will be hard to remember to add this liners before each sending or receiving.
Also code will be less noisy.

anjmao: Send message accepts this struct
```go
type Message struct {
	Data []byte
}
```
It is not coupled to concrete message struct, you just send bytes. If I add it to Send I can't log formatted payload with fields. There are only 5 places there it is needed right now and we will not have much more of these messages.

Also, what if you want to add more context to logging?  I suggest to keep it explicit as is.

Waldz: Well serialisation should happen in transport level also :)

session/pingpong/exchange_messaging.go
```diff
@@ -64,8 +65,9 @@ func (es *ExchangeSender) Send(em crypto.ExchangeMessage) error {
 		AgreementTotal: em.AgreementTotal,
 		Provider:       em.Provider,
 		Signature:      em.Signature,
-	})
-	_, err := es.ch.Send(p2p.TopicPaymentMessage, pMessage)
+	}
+	log.Info().Msgf("Sending P2P message to %q: %s", p2p.TopicPaymentMessage, pMessage.String())
```
Waldz: I suppose all these struct dumps should not be in info level

anjmao: Changed to debug level.

core/connection/manager.go
anjmao: p2p transport layer works with raw bytes, there is no need to for extra abstractions.

Issue #1966, opened 2020-03-31
P2P: Consumer should not create dialog if p2p was established


anjmao:
https://github.com/mysteriumnetwork/node/blob/master/core/connection/manager.go#L188 Add check to not create dialog if `channel != nil`

PR #1975, opened 2020-04-01
Do not create Dialog when P2P communication available

soffokl:
Closes: #1966

### DIFF
--- a/core/connection/manager.go
+++ b/core/connection/manager.go
@@ -190,10 +190,12 @@ func (manager *connectionManager) Connect(consumerID, accountantID identity.Iden
 
 	channel := manager.createP2PChannel(consumerID, providerID, proposal)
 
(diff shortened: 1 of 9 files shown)


core/connection/manager.go
```diff
@@ -282,16 +285,35 @@ func (manager *connectionManager) checkSessionIP(dialog communication.Dialog, se
...
+	sessionStatus := &pb.SessionStatus{
+		ConsumerID: consumerID.Address,
 		SessionID:  string(sessionID),
-		StatusCode: code,
+		Code:       uint32(code),
 		Message:    errDetailsMsg,
-	})
+	}
+
+	log.Debug().Msgf("Sending session status P2P message to %q: %s", p2p.TopicSessionCreate, sessionStatus.String())
```
anjmao, line 309: In log message it should be p2p.TopicSessionStatus too.

soffokl: Fixed it. Thanks.

PR #1991, opened 2020-04-02
Move p2p keep alive pings to consumer and provider managers

anjmao:
P2P pings are will be done in consumer and provider sides instead of internally in channel which allows to disconnect if max error count is reached.


core/connection/manager.go
```diff
@@ -732,6 +739,48 @@ func (manager *connectionManager) getCurrentSession() SessionInfo {
...
+			_, err := channel.Send(p2p.TopicKeepAlive, p2p.ProtoMessage(msg))
+			if err != nil {
+				log.Err(err).Msgf("Failed to send p2p keepalive ping. SessionID: %s", sess.SessionID)
+				errCount++
+				if errCount == manager.params.KeepAlive.MaxSendErrCount {
+					log.Error().Msgf("Max p2p keepalive err count reached, disconnecting. SessionID: %s", sess.SessionID)
+					manager.Disconnect()
+					return
+				}
+			}
```
soffokl, line 779: If we did a successful send, we can reset the error counter, otherwise, we will accumulate single errors over time. 
```suggestion
			} else {
				errCount = 0
			}
```

pb/payment.pb.go
```diff
@@ -1,18 +1,17 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
-// 	protoc-gen-go v1.20.1-devel
+// 	protoc-gen-go v1.20.0-devel
```
soffokl, line 3: Can we use the same latest version, otherwise we will be pushing these changes overwriting each other all the time?

session/manager.go
```diff
@@ -55,6 +58,27 @@ type publisher interface {
...
+
+// Params contains common configuration options for session manager.
+type Params struct {
+	KeepAlive KeepAliveParams
+}
+
+// DefaultParams returns default params.
+func DefaultParams() Params {
+	return Params{
+		KeepAlive: KeepAliveParams{
```
soffokl, line 75: We have the same values in the connection manager. Can we have it defined in a single place?

session/manager.go
```diff
@@ -212,3 +241,44 @@ func (manager *Manager) Destroy(consumerID identity.Identity, sessionID string)
...
+		case <-sess.done:
+			return
+		case <-time.After(manager.params.KeepAlive.SendInterval):
+			msg := &pb.P2PKeepAlivePing{
+				Msg: fmt.Sprintf("provider SessionID: %s", sess.ID),
+			}
+			_, err := channel.Send(p2p.TopicKeepAlive, p2p.ProtoMessage(msg))
+			if err != nil {
+				log.Err(err).Msgf("Failed to send p2p keepalive ping. SessionID: %s", sess.ID)
+				errCount++
```
soffokl, line 275: I'd suggest resetting the error counter on success here too.

core/connection/manager.go
anjmao: Good point, thanks 👍

session/manager.go
anjmao: For now values are the same, but what if you decide to change it for provider side etc. ? They are also in different packages so for simplicity I would leave it as is.

core/connection/manager.go
```diff
@@ -117,7 +133,7 @@ type connectionManager struct {
 	eventPublisher           eventbus.Publisher
 	connectivityStatusSender connectivity.StatusSender
 	ipResolver               ip.Resolver
-	ipCheckParams            IPCheckParams
+	params                   Params
```
Waldz, line 136: `params` - too generic. Are you trying to introduce manager configuration here?

core/connection/manager.go
```diff
@@ -732,6 +738,50 @@ func (manager *connectionManager) getCurrentSession() SessionInfo {
...
+	// Send pings to provider.
+	var errCount int
+	for {
+		select {
+		case <-manager.ctx.Done():
+			return
+		case <-time.After(manager.params.KeepAlive.SendInterval):
+			sess := manager.getCurrentSession()
+			msg := &pb.P2PKeepAlivePing{
+				Msg: fmt.Sprintf("consumser SessionID: %s", sess.SessionID),
```
Waldz, line 767: I am sure we dont need to send free text here

core/connection/manager_test.go
```diff
@@ -109,12 +109,16 @@ func (tc *testContext) SetupTest() {
...
-		MaxAttempts:             3,
-		SleepDurationAfterCheck: 1 * time.Millisecond,
-		Done:                    make(chan struct{}, 1),
+	tc.params = Params{
+		IPCheck: IPCheckParams{
+			MaxAttempts:             3,
+			SleepDurationAfterCheck: 1 * time.Millisecond,
+		},
+		KeepAlive: KeepAliveParams{
+			SendInterval:    100 * time.Millisecond,
```
Waldz, line 118: Too frequent.
Also we will disconnect session after 500ms of temporarily connectivity issues.

core/connection/manager.go
anjmao: Yes, this is connection manager Params. As it lives in connection package I didn't named it ConnectionnParams. The main purpose for now is to set different values for unit tests, but params can be parsed from flags if needed too.

core/connection/manager_test.go
anjmao: This is unit test :)

session/manager.go
```diff
@@ -212,3 +241,44 @@ func (manager *Manager) Destroy(consumerID identity.Identity, sessionID string)
...
+
+	// Send pings to consumer.
+	var errCount int
+	for {
+		select {
+		case <-sess.done:
+			return
+		case <-time.After(manager.params.KeepAlive.SendInterval):
+			msg := &pb.P2PKeepAlivePing{
+				Msg: fmt.Sprintf("provider SessionID: %s", sess.ID),
```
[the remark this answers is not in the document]
anjmao: It doesn't need anything more. Just added Msg string for better debugging.

core/connection/manager.go
```diff
@@ -59,15 +59,31 @@ var (
...
-		SleepDurationAfterCheck: 3 * time.Second,
-		Done:                    make(chan struct{}, 1),
+// KeepAliveParams contains keep alive options.
+type KeepAliveParams struct {
+	SendInterval    time.Duration
+	MaxSendErrCount int
+}
+
+// Params contains common configuration options for connection manager.
+type Params struct {
```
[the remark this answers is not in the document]
anjmao: Any suggestions? Params here means connection manager Params.
BTW. Manager is also too generic.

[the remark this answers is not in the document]
Waldz: I would go with more specific `IPCheckParams` and `KeepaliveParams`.
Or `connection.Config` if you're merging them and making global configuration

session/manager.go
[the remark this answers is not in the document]
Waldz: Same can be achieved by passing just sessionID. I'm for-seeing somebody parsing this string

Issue #1987, opened 2020-04-01
P2P: Provider channel may not be ready yet when consumer sends


anjmao:
This [unit test](https://github.com/mysteriumnetwork/node/blob/master/p2p/dialer_test.go#L71) sometimes fails if consumer sends message before provider side register handlers. It may happen in real world too.

Possible fix would be to extend p2p exchange and send additional message via broker after this func is executed [here](https://github.com/mysteriumnetwork/node/blob/master/p2p/listener.go#L148)

PR #1995, opened 2020-04-03
Synch service startup by waiting for handlers register message

zolia:
On slow providers, also in tests we experienced race condition when channel handlers where not registered at the moment when consumer starts to initiate a connect.

fixes: #1987


p2p/dialer.go
```diff
@@ -110,6 +114,15 @@ func (m *dialer) exchangeConfig(consumerID, providerID identity.Identity, servic
 	}
 	defer brokerConn.Close()
 
+	ready := make(chan bool)
+	defer close(ready)
+	_, err = brokerConn.Subscribe(channelHandlersReadySubject(providerID, serviceType), func(msg *nats_lib.Msg) {
+		if err := m.channelHandlersReady(msg, ready); err != nil {
+			log.Err(err).Msg("handlers ready handler setup failed")
```
soffokl, line 121: Something wrong with this message. I cannot parse it. 
```suggestion
			log.Err(err).Msg("handlers ready handler setup failed")
```

p2p/dialer.go
```diff
@@ -174,6 +187,16 @@ func (m *dialer) exchangeConfig(consumerID, providerID identity.Identity, servic
 		return nil, fmt.Errorf("could not send signed msg: %v", err)
 	}
 
+	// TODO: put this under feature set check when its available
+	// wait until provider confirms that channel handlers are ready
+	select {
+	case <- ready:
+		log.Debug().Msgf("Received handlers ready message from provider")
```
soffokl, line 194:
```suggestion
		log.Debug().Msg("Received handlers ready message from provider")
```

p2p/dialer.go
```diff
@@ -174,6 +187,16 @@ func (m *dialer) exchangeConfig(consumerID, providerID identity.Identity, servic
...
+	// NOTE: We can make this timeout must larger when all providers migrate to handlers ready message
```
soffokl, line 195:
```suggestion
	// NOTE: We can make this timeout much larger when all providers migrate to handlers ready message.
```

p2p/dialer.go
```diff
@@ -174,6 +187,16 @@ func (m *dialer) exchangeConfig(consumerID, providerID identity.Identity, servic
...
+	case <- time.After(1*time.Second):
+		log.Debug().Msgf("Failed to receive handlers ready message from provider - continuing")
```
soffokl, line 197:
```suggestion
		log.Debug().Msg("Failed to receive handlers ready message from provider - continuing")
```

p2p/dialer.go
```diff
@@ -191,3 +214,16 @@ func (m *dialer) sendSignedMsg(brokerConn nats.Connection, subject string, msg [
 	}
 	return reply.Data, nil
 }
+
+func (m *dialer) channelHandlersReady(msg *nats_lib.Msg, ready chan bool) error {
+	var handlersReady pb.P2PChannelHandlersReady
+	if err := proto.Unmarshal(msg.Data, &handlersReady); err != nil {
+		return fmt.Errorf("failed to unmarshall handlers ready message %w", err)
```
soffokl, line 221:
```suggestion
		return fmt.Errorf("failed to unmarshal handlers ready message %w", err)
```

p2p/dialer.go
```diff
@@ -110,6 +114,15 @@ func (m *dialer) exchangeConfig(consumerID, providerID identity.Identity, servic
 	}
 	defer brokerConn.Close()
 
+	ready := make(chan bool)
+	defer close(ready)
+	_, err = brokerConn.Subscribe(channelHandlersReadySubject(providerID, serviceType), func(msg *nats_lib.Msg) {
+		if err := m.channelHandlersReady(msg, ready); err != nil {
```
soffokl, line 120: If you put this defer inside the function, you don't need to pass channel to the `channelHandlersReady`, it will be just closed when `channelHandlersReady` is done.
```suggestion
	_, err = brokerConn.Subscribe(channelHandlersReadySubject(providerID, serviceType), func(msg *nats_lib.Msg) {
		defer close(ready)
		if err := m.channelHandlersReady(msg); err != nil {
```

p2p/dialer.go
```diff
@@ -174,6 +187,16 @@ func (m *dialer) exchangeConfig(consumerID, providerID identity.Identity, servic
 		return nil, fmt.Errorf("could not send signed msg: %v", err)
 	}
 
+	// TODO: put this under feature set check when its available
+	// wait until provider confirms that channel handlers are ready
+	select {
```
anjmao, line 192: We don't have any provider yet with this new feature. I think this TODO can be removed.

p2p/dialer.go
```diff
@@ -174,6 +187,16 @@ func (m *dialer) exchangeConfig(consumerID, providerID identity.Identity, servic
...
+	case <- ready:
+		log.Debug().Msg("Received handlers ready message from provider")
+	// NOTE: We can make this timeout much larger when all providers migrate to handlers ready message
+	case <- time.After(1*time.Second):
```
anjmao, line 196: It can be made larger now, providers are not using it yet. There is one problem. Dial allows to pass timeout so from api perspective if I call

```go
Dial(...., 10 * time.Second)
``` 
I expect that method call to Dial will take no longer that this time.

It can be solved by instead passing context

```go
ctx, cancel := context.WithTimeout(context.Background(), 10 * time.Second)
Dial(ctx) // Internally pass context to method calls to nats
```

But need to check if nats allows to pass context to Request method.

p2p/dialer.go
```diff
@@ -191,3 +214,15 @@ func (m *dialer) sendSignedMsg(brokerConn nats.Connection, subject string, msg [
 	}
 	return reply.Data, nil
 }
+
+func (m *dialer) channelHandlersReady(msg *nats_lib.Msg) error {
+	var handlersReady pb.P2PChannelHandlersReady
+	if err := proto.Unmarshal(msg.Data, &handlersReady); err != nil {
+		return fmt.Errorf("failed to unmarshal handlers ready message %w", err)
```
anjmao, line 221: Add `:` before error  `failed to unmarshal handlers ready message: %w`

p2p/dialer.go
```diff
@@ -110,6 +114,15 @@ func (m *dialer) exchangeConfig(consumerID, providerID identity.Identity, servic
 	}
 	defer brokerConn.Close()
 
+	ready := make(chan bool)
+	_, err = brokerConn.Subscribe(channelHandlersReadySubject(providerID, serviceType), func(msg *nats_lib.Msg) {
+		defer close(ready)
```
soffokl, line 119: I have just realized that multiple calls for p2p endpoint will try to close channel multiple times, which will panic. 
We need to add something like `sync.Once:
```suggestion
	ready := make(chan bool)
	var once sync.Once
	_, err = brokerConn.Subscribe(channelHandlersReadySubject(providerID, serviceType), func(msg *nats_lib.Msg) {
		defer once.Do(func() { close(ready) })
```

communication/nats/connection_mock.go
```diff
@@ -101,6 +104,9 @@ func (conn *ConnectionMock) MessageWait(waitChannel chan interface{}) (interface
 
 // Publish publishes a new message
 func (conn *ConnectionMock) Publish(subject string, payload []byte) error {
+	conn.m.Lock()
```
Waldz, line 107: No point to lock here as lock not used

p2p/dialer.go
zolia: Modified Dial to accept timeout context. It is what you imagine should be?

communication/nats/connection_mock.go
anjmao: I experienced data races related to connection_mock, perhaps this fixed it.

p2p/dialer.go
```diff
@@ -35,7 +41,7 @@ import (
 type Dialer interface {
 	// Dial exchanges p2p configuration via broker, performs NAT pinging if needed
 	// and create p2p channel which is ready for communication.
-	Dial(consumerID, providerID identity.Identity, serviceType string, timeout time.Duration) (Channel, error)
+	Dial(consumerID, providerID identity.Identity, serviceType string, ctx context.Context) (Channel, error)
```
anjmao, line 44: It's a good practice to always pass context as a first parameter.

```go
Dial(ctx context.Context, consumerID, providerID identity.Identity, serviceType string) (Channel, error)
```

You can also change provider listener to use context 👍

p2p/dialer.go
```diff
@@ -110,6 +116,16 @@ func (m *dialer) exchangeConfig(consumerID, providerID identity.Identity, servic
 	}
 	defer brokerConn.Close()
 
+	ready := make(chan bool)
```
anjmao, line 119: It's better to have empty struct chan for such signals since you don't need it's value. If you have bool channel it's confusing and looks like you care if it's true or false.
```go
ready := make(chan struct{})
```

p2p/dialer.go
```diff
@@ -169,11 +185,22 @@ func (m *dialer) exchangeConfig(consumerID, providerID identity.Identity, servic
...
 	}
 
+	// Put this under feature set check when its available
+	// Wait until provider confirms that channel handlers are ready
+	select {
+	case <-ready:
+		log.Debug().Msg("Received handlers ready message from provider")
+	case <-ctx.Done():
+		log.Debug().Msg("Timeout while performing configuration exchange - trying to continue")
+	case <-time.After(5 * time.Second):
```
anjmao, line 200: I would go simpler and return error if context is canceled, it will be strange to continue if you didn't managed to exchange in given time interval.

```go
select {
    case <-ready:
       return &p2pConnectConfig{....}
   case <-ctx.Done():
       return errors.New("timeout waiting for p2p channel config exchange")
}
```
```

communication/nats/connection_mock.go
zolia: Yes, it addresses a race in tests. Reduced a scope a bit.

p2p/dialer.go
zolia: Do you suggest reducing initial Dial timeout? I've added separate for backwards compatibility. If provider does not support this message, not to wait for another 10sec.

p2p/dialer.go
zolia: Lets implement that in separate PR

PR #2001, opened 2020-04-03
Add p2p send timeouts using context

anjmao:
Each p2p send message will need to be handled individually.

Need to think about payments. Now I just set 20 seconds timeout, but if consumer is not available when payment invoice message is sent session will be destroyed. @vkuznecovas


vkuznecovas:
The consumer is allowed to drop a few messages once in a while. We should be fine.

PR #2002, opened 2020-04-04
Skip p2p NAT pinging when provider is not behind NAT

anjmao:
There is no need to ping provider which is not behind NAT. Is such case provider will send the amount of ports equal the amount of needed UDP connections and both peers will create UDP connections manually.

### DIFF
--- a/p2p/dialer.go
+++ b/p2p/dialer.go
@@ -24,6 +24,7 @@ import (
 	"net"
 	"sync"
 
+	"github.com/mysteriumnetwork/node/firewall"
 	nats_lib "github.com/nats-io/go-nats"
 
 	"github.com/mysteriumnetwork/node/communication/nats"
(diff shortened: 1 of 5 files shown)


p2p/dialer.go
```diff
@@ -75,31 +76,36 @@ func (m *dialer) Dial(ctx context.Context, consumerID identity.Identity, service
 		return nil, fmt.Errorf("could not exchange config: %w", err)
 	}
 
-	var remotePort, localPort int
-	var conn0 *net.UDPConn
-	var serviceConn *net.UDPConn
-	if len(config.peerPorts) == 1 {
-		localPort = config.localPorts[0]
-		remotePort = config.peerPorts[0]
+	if _, err := firewall.AllowIPAccess(config.peerPublicIP); err != nil {
```
anjmao, line 79: @zolia I didn't added cleanup of the rule here.
1. The same rule is added during connection, and will be cleaned up on disconnect.
2. All rules are cleaned on node shutdown.

zolia: AllowIPAccess() could probably check if required IP is not already added to avoid duplications. But not needed with this PR imho.

p2p/listener.go
```diff
@@ -160,7 +167,7 @@ func (m *listener) Listen(providerID identity.Identity, serviceType string, chan
 	return err
 }
 
-func (m *listener) providerStartConfigExchange(brokerConn nats.Connection, signerID identity.Identity, msg *nats_lib.Msg) error {
+func (m *listener) providerStartConfigExchange(brokerConn nats.Connection, signerID identity.Identity, msg *nats_lib.Msg, outboundID string) error {
```
zolia, line 170:
```suggestion
func (m *listener) providerStartConfigExchange(brokerConn nats.Connection, signerID identity.Identity, msg *nats_lib.Msg, outboundIP string) error {
```

zolia:
Now we will  at least have a code we were testing :)

PR #2009, opened 2020-04-06
Add invoice send fail count tolerance

anjmao:
As now p2p send always expects peer to send response in case consumer is not available (sleeping etc) we need sone tolerance similar as with exchange message receive.

### DIFF
--- a/session/pingpong/invoice_tracker.go
+++ b/session/pingpong/invoice_tracker.go
@@ -19,7 +19,7 @@ package pingpong
 
 import (
 	"bytes"
-	cryptoRand "crypto/rand"
+	crand "crypto/rand"
 	"encoding/hex"
 	"encoding/json"
 	stdErr "errors"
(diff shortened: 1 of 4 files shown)


session/pingpong/invoice_tracker.go
```diff
@@ -19,6 +19,7 @@ package pingpong
 
 import (
 	"bytes"
+	cryptoRand "crypto/rand"
```
vkuznecovas, line 22: shorten it to crand

session/pingpong/invoice_tracker.go
```diff
@@ -168,14 +176,20 @@ func NewInvoiceTracker(
...
 		invoicesSent:                   make(map[string]sentInvoice),
+		rnd:                            rand.New(rand.NewSource(1)),
 	}
 }
 
 func calculateMaxNotReceivedExchangeMessageCount(chargeLeeway, chargePeriod time.Duration) uint64 {
 	return uint64(math.Round(float64(chargeLeeway) / float64(chargePeriod)))
 }
 
+func calculateMaxSentReceivedExchangeMessageCount(chargeLeeway, chargePeriod time.Duration) uint64 {
```
vkuznecovas, line 189: probably should be calculateMaxNotSentExchangeMessageCount

Issue #2024, opened 2020-04-07
Connection disconnect / cancel is not reliable 


anjmao:
This check [here](https://github.com/mysteriumnetwork/node/blob/master/core/connection/manager.go#L628) skips if current state is not connected and in connect method it's also set to not connected on error but with no cleanup logic [here](https://github.com/mysteriumnetwork/node/blob/master/core/connection/manager.go#L201)

This can lead to consumer not being disconnected and stuck in connected state even if he sees Connect button.

PR #2025, opened 2020-04-07
Disconnect on any error during connect

anjmao:
Fixes https://github.com/mysteriumnetwork/node/issues/2024

### DIFF
--- a/core/connection/manager.go
+++ b/core/connection/manager.go
@@ -198,8 +198,8 @@ func (manager *connectionManager) Connect(consumerID, accountantID identity.Iden
 	manager.setStatus(statusConnecting())
 	defer func() {
 		if err != nil {
-			manager.setStatus(statusNotConnected())
-			manager.publishStateEvent(NotConnected)
+			log.Err(err).Msg("Connect failed, disconnecting")
+			manager.disconnect()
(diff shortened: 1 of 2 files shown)


core/connection/manager.go
```diff
@@ -198,8 +198,8 @@ func (manager *connectionManager) Connect(consumerID, accountantID identity.Iden
 	manager.setStatus(statusConnecting())
 	defer func() {
 		if err != nil {
-			manager.setStatus(statusNotConnected())
-			manager.publishStateEvent(NotConnected)
+			log.Err(err).Msg("Connect failed, disconnecting")
+			err = manager.Disconnect()
```
Waldz, line 202: It will produce bunch of false statuses like `Disconnecting`

zolia: Does it really that matter? We could introduce another state "Cancelling" if we really gonna analyze and have state statistics..

Waldz: Yes it matters, as we dont want strange transitions: `Connecting -> Disconnecting`

UI blinks to state Disconnecting already too much. Disconnecting transition should happen just in flow:
Connected -> Disconnecting

Issue #2011, opened 2020-04-06
P2P: Use proposals Contact when connecting to provider


anjmao:
Instead of connecting directly to broker use proposals contact definition. If proposal doesn't support new p2p definition, fallback to dialog.

zolia:
agree, new consumers will look up which contact type is best for him.

PR #2026, opened 2020-04-07
Add p2p contact

anjmao:
Closes https://github.com/mysteriumnetwork/node/issues/2011

### DIFF
--- a/core/connection/manager.go
+++ b/core/connection/manager.go
@@ -20,6 +20,7 @@ package connection
 import (
 	"context"
 	"encoding/json"
+	stdErrors "errors"
 	"fmt"
 	"sync"
 	"time"
@@ -209,7 +210,19 @@ func (m *connectionManager) Connect(consumerID, accountantID identity.Identity, 
 
 	providerID := identity.FromAddress(proposal.ProviderID)
 
-	channel := m.createP2PChannel(m.ctx, consumerID, providerID, proposal)
+	var channel p2p.Channel
+	if contact, err := p2p.ParseContact(proposal.ProviderContacts); err == nil {
+		channel, err = m.createP2PChannel(m.ctx, consumerID, providerID, proposal.ServiceType, contact)
+		if err != nil {
+			return fmt.Errorf("could not create p2p channel: %w", err)
+		}
+	} else {
+		if stdErrors.Is(err, p2p.ErrContactNotFound) {
+			log.Debug().Msgf("Provider %s doesn't support p2p, will fallback to dialog", providerID.Address)
+		} else {
+			return err
+		}
+	}
 
 	var dialog communication.Dialog
 	if channel == nil {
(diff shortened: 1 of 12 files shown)


core/connection/manager.go
```diff
@@ -209,7 +210,19 @@ func (m *connectionManager) Connect(consumerID, accountantID identity.Identity,
 
 	providerID := identity.FromAddress(proposal.ProviderID)
 
-	channel := m.createP2PChannel(m.ctx, consumerID, providerID, proposal)
+	var channel p2p.Channel
+	if contact, err := p2p.ParseContact(proposal.ProviderContacts); err == nil {
+		channel, err = m.createP2PChannel(m.ctx, consumerID, providerID, proposal.ServiceType, contact)
+		if err != nil {
+			return fmt.Errorf("could not create p2p channel: %w", err)
```
anjmao, line 217: Note that now if proposals supports p2p it will try to use p2p or fail and will not fallback to dialog. Fallback to dialog will happen only if proposal doesn't support p2p.

zolia: Thats fine imho. Most likely it will to setup transport then anyways.

p2p/dialer.go
```diff
@@ -41,14 +41,13 @@ import (
 type Dialer interface {
 	// Dial exchanges p2p configuration via broker, performs NAT pinging if needed
 	// and create p2p channel which is ready for communication.
-	Dial(ctx context.Context, consumerID identity.Identity, serviceType string, providerID identity.Identity) (Channel, error)
+	Dial(ctx context.Context, consumerID identity.Identity, providerID identity.Identity, serviceType string, contactDef ContactDefinition) (Channel, error)
```
soffokl, line 44:
```suggestion
	Dial(ctx context.Context, consumerID, providerID identity.Identity, serviceType string, contactDef ContactDefinition) (Channel, error)
```

p2p/dialer.go
```diff
@@ -65,13 +64,12 @@ type dialer struct {
...
 	verifier       identity.Verifier
 	ipResolver     ip.Resolver
-	brokerAddress  string
 }
 
 // Dial exchanges p2p configuration via broker, performs NAT pinging if needed
 // and create p2p channel which is ready for communication.
-func (m *dialer) Dial(ctx context.Context, consumerID identity.Identity, serviceType string, providerID identity.Identity) (Channel, error) {
-	brokerConn, err := m.broker.Connect(m.brokerAddress)
+func (m *dialer) Dial(ctx context.Context, consumerID identity.Identity, providerID identity.Identity, serviceType string, contactDef ContactDefinition) (Channel, error) {
```
soffokl, line 71:
```suggestion
func (m *dialer) Dial(ctx context.Context, consumerID, providerID identity.Identity, serviceType string, contactDef ContactDefinition) (Channel, error) {
```

p2p/listener.go
```diff
@@ -40,13 +41,16 @@ type Listener interface {
 	// Listen listens for incoming peer connections to establish new p2p channels. Establishes p2p channel and passes it
 	// to channelHandlers
 	Listen(providerID identity.Identity, serviceType string, channelHandler func(ch Channel)) error
+
+	// GetContact returns which is later can be added to proposal contacts definition so consumer can
```
soffokl, line 45:
```suggestion
	// GetContact returns contract which is later can be added to proposal contacts definition so consumer can
```

Issue #2017, opened 2020-04-06
P2P: Add upnp for peer listener


anjmao:
Currently provider peer detects only if it's not behind nat, but doesn't check if upnp can be used.

PR #2030, opened 2020-04-08
Add upnp port mapping support for p2p ports

anjmao:
- [x] Cleanup session on destroy (provider side) and release.
- [x] Do not fail mapping if it already added for myst.

Closes https://github.com/mysteriumnetwork/node/issues/2017

### DIFF
--- a/p2p/listener.go
+++ b/p2p/listener.go
@@ -209,6 +209,16 @@ func (m *listener) providerStartConfigExchange(signerID identity.Identity, msg *
 		return fmt.Errorf("could not prepare ports: %w", err)
 	}
 
+	m.setPendingConfig(p2pConnectConfig{
+		publicIP:         publicIP,
+		localPorts:       localPorts,
+		privateKey:       privateKey,
+		peerPubKey:       peerPubKey,
+		upnpPortsRelease: portsRelease,
+		peerPublicIP:     "",
+		peerPorts:        nil,
+	})
+
 	config := pb.P2PConnectConfig{
 		PublicIP: publicIP,
 		Ports:    intToInt32Slice(localPorts),
@@ -230,16 +240,6 @@ func (m *listener) providerStartConfigExchange(signerID identity.Identity, msg *
 	if err != nil {
 		return fmt.Errorf("could not publish message via broker: %v", err)
 	}
-
-	m.setPendingConfig(p2pConnectConfig{
-		publicIP:         publicIP,
-		localPorts:       localPorts,
-		privateKey:       privateKey,
-		peerPubKey:       peerPubKey,
-		upnpPortsRelease: portsRelease,
-		peerPublicIP:     "",
-		peerPorts:        nil,
-	})
 	return nil
 }
 
--- a/p2p/listener.go
+++ b/p2p/listener.go
@@ -29,6 +29,7 @@ import (
 	"github.com/mysteriumnetwork/node/core/port"
 	"github.com/mysteriumnetwork/node/identity"
 	"github.com/mysteriumnetwork/node/market"
+	"github.com/mysteriumnetwork/node/nat/mapping"
 	"github.com/mysteriumnetwork/node/pb"
 
 	nats_lib "github.com/nats-io/go-nats"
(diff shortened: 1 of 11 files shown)


p2p/listener.go
```diff
@@ -227,10 +231,55 @@ func (m *listener) providerStartConfigExchange(signerID identity.Identity, msg *
...
+	}
+	// Return these ports if provider is not behind NAT.
+	if outboundIP == publicIP {
+		return localPorts, nil, nil
+	}
+
+	// Try to add upnp ports mapping.
+	var portsRelease []func()
+	var portMappingOk bool
+	var portRelease func()
```
anjmao, line 260: @zolia I'm thinking what is the best way to handle cleanup for mapped ports. Maybe port mapper can hold mapped ports state globally and can be called on node shutdown with Close method. This way mapped ports can be handled in once place and we be sure they are closed. Also p2p channel holds the same release func so if session is destroyed gracefully(by triggering from consumer side) it will close p2p channel and release mapped ports. 

In generall I think such generic session cleanup on shutdown can be implemented in separate task https://github.com/mysteriumnetwork/node/issues/2032

session/pingpong/invoice_tracker.go
```diff
@@ -423,7 +423,7 @@ func (it *InvoiceTracker) Start() error {
 	defer it.revealPromise()
 
 	// give the consumer a second to start up his payments before sending the first request
-	firstSend := time.After(time.Second)
+	firstSend := time.After(3 * time.Second)
```
anjmao, line 426: @vkuznecovas 1 second may not be enough for p2p handler to register session-invoice as it happens deep inside session. I was thinking that it could be solved by sending p2p message from consumer to provider notifying that payments are ready, what do you think?

Waldz: Already so many messages, why aren't they enough, "listeners ready" message was especially for that, wasn't it?
So we need just bind to it correctly to remove this sleep

anjmao: Listeners ready message was for provider side. This issue is for consumer side. Provider sends payment-invoice after one second assuming that during this period consumer will set up payments.

p2p/listener.go
zolia: Global state basically IS needed, but not for shutdown cleanup. I think cleanup on session termination should do. Issue here is that destroys do not happen and nodes shuts down or are killed abruptly all the time. That will lead to upnp mapping poisoning on routers and eventually node will start to fail to allocate mappings.

My suggestion is to implement go routine which would poll actual state on a router and would sync it with a node's port mapping global state. This way we would be able to cleanup stale mappings. Question remains if we can make such polling reliably and if we can really check if mapping introduced previously is by our node or not. Maybe comment can hold some ids and be read from router.. need to check for similar stuff.

session/pingpong/invoice_tracker.go
anjmao: Ok, I see that this is mostly a problem for e2e tests.

zolia: Loosing a single invoice is tolerable and should not pose an issue. If it still does - we should fix that.

vkuznecovas: The first session invoice MUST be enforced, this is the only way for the provider to know that the consumer is able to pay, there's no other way to verify. We should be able to drop invoices AFTER the first one, but the first one is critical.

vkuznecovas: #2034

Waldz: If so it should be in session-create response, simple approach to avoid racing

zolia: Mixing payment logic with session create logic looks like a hack. Would like to decouple payments from session creation since both are not strictly each-other dependent. 

Could we somehow make payment dependent services payment dependent? :) That is if payment is not provided - we don't provide (suspent, put on hold) a service until we get a promise. This way, even if we miss first invoice we still could resume a service after invoice retry. Until then only communication channel works. 

Other approach would be to suspend actual service connect until invoice is received and promise set. But that would be a weak point since consumers might want to abuse such provider behaviour.

PR #2038, opened 2020-04-10
Add options to set service prices for provider

soffokl:


services/noop/service_test.go
```diff
@@ -38,20 +37,16 @@ var _ service.Service = NewManager()
 
 func Test_GetProposal(t *testing.T) {
 	country := "LT"
+	paymentMethod := pingpong.NewPaymentMethod(0, 10000)
```
vkuznecovas, line 40: can we make this configurable as well? would make testing payments easier

cmd/commands/cli/command.go
```diff
@@ -702,3 +709,54 @@ func parseStartFlags(serviceType string, args ...string) (service.Options, confi
 
 	return nil, config.ServicesOptions{}, errors.New("service type not found")
 }
+
+func (c *cliApp) servicePrice(argsString string) {
+	const paymentsHelp = "Usage: service price <openvpn|wireguard> [gb|minute] [value]"
```
Waldz, line 714: Quite a strange flow to set a price and then to call another command `service start`. Lets be consistent and pass service create variables while starting it. Same as we do all other service variables `service start openvpn --access-policy.list=mysterium --shaper.enabled --price-minute=3000`

do same as shared configs `access-policy.list`

config/flags_service_openvpn.go
```diff
@@ -46,6 +46,18 @@ var (
 		Usage: "OpenVPN subnet netmask",
 		Value: "255.255.255.0",
 	}
+	// FlagOpenVPNPriceMinute sets the price per minute for provided OpenVPN service.
+	FlagOpenVPNPriceMinute = cli.Uint64Flag{
+		Name:  "openvpn.price-minute",
+		Usage: "Sets the price of the OpenVPN service per minute.",
+		Value: 50000,
```
Waldz, line 53: Human unreadable. IMHO user should input MYST value: 0.0006
@chompomonim, opinions?

cmd/commands/cli/command.go
```diff
@@ -699,3 +706,54 @@ func parseStartFlags(serviceType string, args ...string) (service.Options, confi
...
+
+	if err := config.Current.SaveUserConfig(); err != nil {
+		warnf("Failed to save user config %v\n", err)
+		return
+	}
+
+	priceGB = config.GetUInt64(flagGB)
+	priceMinute = config.GetUInt64(flagMinute)
+
+	infof("New price per GiB: %d\n", priceGB)
```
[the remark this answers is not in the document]
soffokl: This is price per GB, this is not related to human-readable bitsize.

config/flags_service_openvpn.go
soffokl: Isn't 0.0006 more confusing and hard to manipulate?

session/pingpong/factory.go
```diff
@@ -36,23 +36,35 @@ import (
...
+	if pricePerGB > 0 {
+		pricePerGB = gb * accuracy / pricePerGB
+	}
+	if pricePerMinute > 0 {
+		pricePerMinute = uint64(time.Minute) * accuracy / pricePerMinute
+	}
+
+	return PaymentMethod{
+		Price:    money.NewMoney(accuracy, money.CurrencyMyst),
+		Duration: time.Duration(pricePerMinute),
```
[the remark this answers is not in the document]
soffokl: I didn't change it. The price per duration was here all the time. If we need to change it let's do it in another PR to avoid putting everything together.

cmd/commands/cli/command.go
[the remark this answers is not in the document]
Waldz: I'm about money representation here

config/flags_service_openvpn.go
Waldz: User does not know decimal size of our token, why should he bother finding it, understanding how much real MYST is it

soffokl: Changed it to display in MYST.

cmd/commands/cli/command.go
soffokl: Removed it, now it goes through the arguments for service start.

Issue #2021, opened 2020-04-07
P2P: Determine consumer port changes


anjmao:
When consumer switches network and his NAT type changes consumer's source port could change and provider peer need to update it's port. It can be done by reader remote addr inside p2p channel read loop.

PR #2042, opened 2020-04-10
Detect peer remote addr changes

anjmao:
Add proxy UDP conn which can be used to detect peer remote address change.

Closes https://github.com/mysteriumnetwork/node/issues/2021

### DIFF
--- a/localnet/consumer.sh
+++ b/localnet/consumer.sh
@@ -7,6 +7,7 @@ exec /node/build/myst/myst \
   --log-dir= --data-dir=/var/lib/mysterium-node \
   --runtime-dir=/var/run/mysterium-node \
   --tequilapi.address=0.0.0.0 \
+  --log-level=debug \
   --payments.mystscaddress=0x4D1d104AbD4F4351a0c51bE1e9CA0750BbCa1665 \
   --ip-detector=http://ipify:3000/?format=json \
   --location.type=manual \
--- a/p2p/channel.go
+++ b/p2p/channel.go
@@ -25,8 +25,8 @@ import (
 	"net"
 	"net/textproto"
 	"os"
+	"strings"
 	"sync"
-	"time"
 
 	"github.com/rs/zerolog/log"
 	"github.com/xtaci/kcp-go/v5"
@@ -42,6 +42,11 @@ var (
 
 	// ErrHandlerNotFound indicates that peer is not registered handler yet.
 	ErrHandlerNotFound = errors.New("p2p peer handler not found")
+)
+
+const (
+	kcpMTUSize = 1280
+	mtuLimit   = 1500
 )
 
 // ChannelSender is used to send messages.
@@ -80,12 +85,45 @@ type stream struct {
 	resCh chan *transportMsg
 }
 
+type peer struct {
+	sync.RWMutex
+	publicKey  PublicKey
+	remoteAddr *net.UDPAddr
+}
+
+func (p *peer) addr() *net.UDPAddr {
+	p.RLock()
+	defer p.RUnlock()
+
+	return p.remoteAddr
+}
+
+func (p *peer) updateAddr(addr *net.UDPAddr) {
+	p.Lock()
+	defer p.Unlock()
+
+	p.remoteAddr = addr
+}
+
 // transport wraps network primitives for sending and receiving packets.
 type transport struct {
-	*textproto.Writer
-	*textproto.Reader
+	// textReader is used to read p2p text protocol data.
+	textReader *textproto.Reader
+
+	// textWriter is used to marshal messages to underlying text protocol.
+	textWriter *textproto.Writer
+
+	// session is KCP session which wraps UDP connection and adds reliability and ordered messages support.
 	session *kcp.UDPSession
-	conn    *net.UDPConn
+
+	// remoteConn is initial conn which should be created from NAT hole punching or manually. It contains
+	// initial local and remote peer addresses.
+	remoteConn *net.UDPConn
+
+	// proxyConn is used for KCP session as a remote. Since KCP doesn't expose it's data read loop
+	// this is needed to detect remote peer address changes as we can simply use conn.ReadFromUDP and
+	// get updated peer address.
+	proxyConn *net.UDPConn
 }
 
 // channel implements Channel interface.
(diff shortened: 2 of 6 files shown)


p2p/channel.go
```diff
@@ -80,86 +81,218 @@ type stream struct {
...
+		n, addr, err := c.tr.remoteConn.ReadFrom(buf)
+		if err != nil {
+			log.Warn().Err(err).Msgf("Read from remote conn failed")
+			// TODO: Check if exit is ok here.
+			return
+		}
+
+		// Check if peer address is not changed and update it if changed.
+		if addr, ok := addr.(*net.UDPAddr); ok {
+			if !addrEqual(latestPeerAddr, addr) {
```
anjmao, line 251: This could be very insecure. I should only update port but not the whole IP. It need to be checked if addr.IP is the same as initial peer addr.IP.

p2p/channel.go
```diff
@@ -80,86 +81,218 @@ type stream struct {
...
+	localAddr := remoteConn.LocalAddr().(*net.UDPAddr)
+	peerAddr := remoteConn.RemoteAddr().(*net.UDPAddr)
+	log.Debug().Msgf("Creating p2p channel with local port: %d, remote port: %d", localAddr.Port, peerAddr.Port)
 
-	blockCrypt, err := newBlockCrypt(privateKey, peerPubKey)
+	remoteConn.Close()
+	remoteConn, err := net.ListenUDP("udp4", localAddr)
 	if err != nil {
-		return nil, fmt.Errorf("could not create block crypt: %w", err)
+		return nil, err
```
zolia, line 176: wrap an error

p2p/channel.go
```diff
@@ -80,86 +81,218 @@ type stream struct {
...
+	// Create local proxy UDP conn which will receive packets from local KCP UDP conn.
+	// TODO: Pass global port pool.
+	proxyPort, err := port.NewPool().Acquire()
 	if err != nil {
-		return nil, fmt.Errorf("could not create UDP conn: %w", err)
+		return nil, err
+	}
+	proxyConn, err := net.ListenUDP("udp4", &net.UDPAddr{IP: net.ParseIP("127.0.0.1"), Port: proxyPort.Num()})
+	if err != nil {
+		return nil, err
```
zolia, line 187: wrap an err

p2p/channel.go
```diff
@@ -80,86 +81,218 @@ type stream struct {
...
+		// Check if peer address is not changed and update it if changed.
+		if addr, ok := addr.(*net.UDPAddr); ok {
+			if !addrEqual(latestPeerAddr, addr) {
+				log.Debug().Msgf("peer addr changed from %s to %s", latestPeerAddr.String(), addr.String())
+				c.peer.updateAddr(addr)
+				latestPeerAddr = addr
+			}
+		}
+
+		_, err = c.tr.remoteConn.WriteToUDP(buf[:n], c.localSessionAddr)
```
zolia, line 258: shouldn't it be `proxyConn.WriteToUDP` here?

p2p/channel.go
```diff
@@ -80,86 +81,218 @@ type stream struct {
...
-	localPort := punchedConn.LocalAddr().(*net.UDPAddr).Port
-	remotePort := punchedConn.RemoteAddr().(*net.UDPAddr).Port
-	log.Debug().Msgf("Creating p2p channel with local port: %d, remote port: %d", localPort, remotePort)
+func newChannel(remoteConn *net.UDPConn, privateKey PrivateKey, peerPubKey PublicKey) (*channel, error) {
+	localAddr := remoteConn.LocalAddr().(*net.UDPAddr)
+	peerAddr := remoteConn.RemoteAddr().(*net.UDPAddr)
+	log.Debug().Msgf("Creating p2p channel with local port: %d, remote port: %d", localAddr.Port, peerAddr.Port)
 
-	blockCrypt, err := newBlockCrypt(privateKey, peerPubKey)
+	remoteConn.Close()
```
zolia, line 173: This looks a bit awkward. We pass alive conn and close right away. Would extract this close() to other func `remoteConn = updateConn(localAddr)`

p2p/channel.go
```diff
@@ -80,125 +85,251 @@ type stream struct {
...
-	streams          map[uint64]*stream
-	nextStreamID     uint64
-	privateKey       PrivateKey
-	peerPubKey       PublicKey
-	peerAddr         *net.UDPAddr
-	localAddr        *net.UDPAddr
-	blockCrypt       kcp.BlockCrypt
-	stop             chan struct{}
-	sendQueue        chan *transportMsg
+	// tr is transport containing network related connections for p2p to work.
```
soffokl, line 134: These obvious comments for private fields looks excess.

p2p/channel.go
```diff
@@ -209,11 +340,14 @@ func (c *channel) sendLoop() {
 			}
 
 			if debugTransport {
-				log.Debug().Msgf("send to %s: %+v", c.tr.session.RemoteAddr(), msg)
+				fmt.Printf("send to %s: %+v\n", c.tr.session.RemoteAddr(), msg)
```
soffokl, line 343: Should this be a `fmt.Printf`?

p2p/channel.go
```diff
@@ -380,17 +554,18 @@ func newBlockCrypt(privateKey PrivateKey, peerPublicKey PublicKey) (kcp.BlockCry
...
-func listenUDP(punchedConn *net.UDPConn) (*net.UDPConn, error) {
-	punchedConn.Close()
-
-	raddr := punchedConn.RemoteAddr().(*net.UDPAddr)
-	network := "udp4"
-	if raddr.IP.To4() == nil {
-		network = "udp"
+func errNetClose(err error) bool {
+	// Hack. See https://github.com/golang/go/issues/4373 which should expose net close error with 1.15.
+	if strings.Contains(err.Error(), "use of closed network connection") {
```
soffokl, line 559: > S1008: should use 'return strings.Contains(err.Error(), "use of closed network connection")' instead of 'if strings.Contains(err.Error(), "use of closed network connection") { return true }; return false' (gosimple)

p2p/channel.go
```diff
@@ -380,17 +554,18 @@ func newBlockCrypt(privateKey PrivateKey, peerPublicKey PublicKey) (kcp.BlockCry
...
 	}
-	conn, err := net.ListenUDP(network, punchedConn.LocalAddr().(*net.UDPAddr))
-	if err != nil {
-		return nil, fmt.Errorf("could not listen UDP: %w", err)
+	return false
+}
+
+func errPipeClosed(err error) bool {
+	// Hack. We can't check io.ErrPipeClosed as kcp wraps this error with old github.com/pkg/errors.
+	if strings.Contains(err.Error(), "io: read/write on closed pipe") {
```
soffokl, line 567: > S1008: should use 'return strings.Contains(err.Error(), "io: read/write on closed pipe")' instead of 'if strings.Contains(err.Error(), "io: read/write on closed pipe") { return true }; return false' (gosimple)

localnet/consumer.sh
```diff
@@ -7,6 +7,7 @@ exec /node/build/myst/myst \
   --log-dir= --data-dir=/var/lib/mysterium-node \
   --runtime-dir=/var/run/mysterium-node \
   --tequilapi.address=0.0.0.0 \
+  --log-level=trace \
```
tadaskay, line 10: This is excessive. Trace is only meant to be enabled during the debugging.

p2p/channel.go
zolia: imho its good. We don't need just formal documentation. Its sometimes better to have more detailed explanations for more complex code parts.

Issue #1779, opened 2020-02-26
Send session earnings data into monitoring 


chompomonim:
Send earning data together with the session's traffic data into discovery service or into monitoring.

zolia:
We need more clarification here. At the end of the session we need to report to metrics service how much that session did cost? How and where this data will be used. It is good to know such things before starting implementation.

And it says "in progress" who is doing this one? @chompomonim ?

chompomonim:
This tada will be shown in MMN in the list of latest sessions. 

<img width="1409" alt="Screenshot 2020-03-13 at 12 50 18" src="https://user-images.githubusercontent.com/153896/76614686-4261ed80-6529-11ea-9aca-7a07d665165f.png">

I would say that sending jus after session end would be not enough. MMN is showing also on-going session, so there users would like to see earnings together with data about sessions time and provided traffic.

PR #2043, opened 2020-04-10
Send session earnings metric

Waldz:
Updates: #1779


core/connection/interface.go
```diff
@@ -57,7 +58,7 @@ type PromiseIssuerCreator func(issuerID identity.Identity, dialog communication.
 // Manager interface provides methods to manage connection
 type Manager interface {
 	// Connect creates new connection from given consumer to provider, reports error if connection already exists
-	Connect(consumerID, accountantID identity.Identity, proposal market.ServiceProposal, params ConnectParams) error
+	Connect(consumerID identity.Identity, accountantAddress common.Address, proposal market.ServiceProposal, params ConnectParams) error
```
soffokl, line 61: This is very confusing to name it `accountantAddress` because we have CLI argument called `accountant.address` which means `accountant URL address`:
https://github.com/mysteriumnetwork/node/blob/master/config/flags_accountant.go#L27-L31

So the accountant address is an accountant URL address. Can we keep calling it `accountantID`?

consumer/session/session_storage.go
```diff
@@ -43,9 +43,12 @@ type Storer interface {
 	GetAllFrom(bucket string, array interface{}) error
 }
 
+type timeGetter func() time.Time
```
soffokl, line 46: https://github.com/mysteriumnetwork/devbook/blob/master/golang/consumer-interfaces/README.md

consumer/session/session_storage.go
```diff
@@ -114,7 +119,7 @@ func (repo *Storage) consumeSessionSpendingEvent(e pingpongEvent.AppEventInvoice
 		log.Warn().Msg("Received a unknown session update")
 		return
 	}
-	row.Updated = time.Now().UTC()
```
soffokl, line 117: Where is the UTC now?

consumer/session/session_storage_test.go
```diff
@@ -68,49 +70,131 @@ func TestSessionStorageGetAllReturnsError(t *testing.T) {
...
+		},
+		storer.UpdateCalled,
+	)
 }
 
 // StubSessionStorer allows us to get all sessions, save and update them
 type StubSessionStorer struct {
 	SaveError    error
-	SaveCalled   bool
+	SaveCalled   interface{}
```
soffokl, line 184: `SaveCalled` was good for bool. If you storing objects there it's better to use the correct name like `SavedObject`.

consumer/session/session_storage_test.go
```diff
@@ -68,49 +70,131 @@ func TestSessionStorageGetAllReturnsError(t *testing.T) {
...
 	UpdateError  error
-	UpdateCalled bool
+	UpdateCalled interface{}
```
soffokl, line 186: `UpdateCalled` was good for bool. If you storing objects there it's better to use the correct name like `UpdatedObject`.

core/quality/sender.go
```diff
@@ -123,58 +141,63 @@ func (sender *Sender) Subscribe(bus eventbus.Subscriber) error {
...
+	sender.sendEvent(sessionTokensName, sessionTokensContext{
+		Tokens:         e.Invoice.AgreementTotal,
+		sessionContext: session,
 	})
 }
 
 // sendConnStateEvent sends session update events.
 func (sender *Sender) sendConnStateEvent(e connection.AppEventConnectionState) {
-	if !e.SessionInfo.IsActive() {
+	if e.SessionInfo.SessionID == "" {
```
soffokl, line 170: It was more clear before:
```suggestion
	if !e.SessionInfo.IsActive() {
```

core/quality/sender.go
```diff
@@ -224,6 +247,43 @@ func (sender *Sender) sendEvent(eventName string, context interface{}) {
...
+	delete(sender.sessionsActive, context.ID)
+}
+
+func (sender *Sender) recoverSessionContext(sessionID string) (sessionContext, error) {
+	sender.sessionsMu.Lock()
+	defer sender.sessionsMu.Unlock()
+
+	context, found := sender.sessionsActive[sessionID]
+	if !found {
+		return sessionContext{}, fmt.Errorf("unknown session: %v", sessionID)
```
soffokl, line 270:
```suggestion
		return sessionContext{}, fmt.Errorf("unknown session: %w", sessionID)
```

core/quality/sender.go
```diff
@@ -224,6 +247,43 @@ func (sender *Sender) sendEvent(eventName string, context interface{}) {
...
+
+func (sender *Sender) forgetSessionContext(context sessionContext) {
+	sender.sessionsMu.Lock()
+	defer sender.sessionsMu.Unlock()
+
+	delete(sender.sessionsActive, context.ID)
+}
+
+func (sender *Sender) recoverSessionContext(sessionID string) (sessionContext, error) {
+	sender.sessionsMu.Lock()
```
anjmao, line 265: As sessionsMu is RWMutex I probably wanted RLock, RUnlock here?

core/connection/interface.go
tadaskay: I agree that address is a broad term

Waldz: This scope is not related to configuring stage there `accountant.address` is used.

But I reverted this name change `accoutantAddress -> accountantID` , as this accountant unique identifier is already exposed to public contracts as `accountant_id`

consumer/session/session_storage.go
Waldz: Missed. Thanks

consumer/session/session_storage.go
Waldz: It's contract with `time.Now()` function, to mock in in tests

consumer/session/session_storage_test.go
Waldz: Improved

core/quality/sender.go
Waldz: Yes

Issue #1944, opened 2020-03-29
getting session balance timeouts randomly 


zolia:
during that time connectivity is working, no apparent issues. 

daemon output: 

```bash
2020-03-29T14:08:33.712 WRN Users/zolia/golang/pkg/mod/github.com/mysteriumnetwork/payments@v0.0.11-0.20200325083835-89ffae399efa/client/retryable_client.go:78 > retry 3 of 3 error="could not get consumers channel: context deadline exceeded"

2020-03-29T14:09:04.018 ERR session/pingpong/consumer_balance_tracker.go:177 > Could not get consumer channel error="could not get consumers channel: context deadline exceeded"
```

cli output: 

```bash
[SUCCESS] Connected.
» status
[INFO] Status: Connected
[INFO] SID: fa57d410-2eed-40c1-86b8-fc9e8103641d
[INFO] IP: x.x.x.x
[INFO] Location: Vilnius, LT (residential - Telia Lietuva)
[INFO] Proposal: Id: 1 , Provider: 0x2bc8193504a60c1d9746006a9eb3ade5d6b7f51b, Country: LT
(very long time....)
[INFO] Balance: 0
[INFO] Connection duration: 11s
[INFO] Bytes sent: 87494
[INFO] Bytes received: 162472
```

PR #2049, opened 2020-04-14
Reconnect payments client on connect/disconnect.

soffokl:
Closes: #1944

### DIFF
--- a/cmd/di.go
+++ b/cmd/di.go
@@ -25,7 +25,6 @@ import (
 
 	"github.com/ethereum/go-ethereum/accounts/keystore"
 	"github.com/ethereum/go-ethereum/common"
-	"github.com/ethereum/go-ethereum/ethclient"
 	"github.com/mysteriumnetwork/node/communication"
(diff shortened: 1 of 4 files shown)


zolia:
@soffokl tests passes, looks simple change. Does it break something else? Why it is a draft?

soffokl:
@zolia it depends on https://github.com/mysteriumnetwork/payments/pull/53. 
I want to merge payments PR first and use that version in the node.

cmd/di.go
```diff
@@ -842,6 +842,9 @@ func (di *Dependencies) handleHTTPClientConnections() error {
 			di.HTTPClient.Reconnect()
 			di.QualityClient.Reconnect()
 			di.BrokerConnector.ReconnectAll()
+			if err := di.EtherClient.Reconnect(); err != nil {
```
anjmao, line 845: Can we change this main method name from `handleHTTPClientConnections` to something like `handleConnStateChange`

PR #2060, opened 2020-04-15
Add noop service for e2e

anjmao:
Let's see if it helps, but still need to investigate, balance should not become 0 after two connections.


anjmao:
Didn't helped, say no more.

PR #2064, opened 2020-04-15
Make nat traversal more stable

zolia:


p2p/dialer.go
```diff
@@ -106,6 +109,12 @@ func (m *dialer) Dial(ctx context.Context, consumerID, providerID identity.Ident
 			return nil, fmt.Errorf("could not create UDP conn for service: %w", err)
 		}
 	} else {
+		// race condition still happens when consumer starts to ping until provider did not manage to complete required number of pings
+		// this might be provider / consumer performance dependent
+		// make sleep time dependent on pinger interval and wait for 2 ping iterations
+		dur := traversal.DefaultPingConfig().Interval.Milliseconds() * 2
+		log.Debug().Msgf("Sleeping for %v ms - waiting for provider to launch pings", dur)
+		time.Sleep(time.Duration(dur) * time.Millisecond)
```
Waldz, line 117: :(

zolia: Yup, not happy about this too. Wishful thinking does not make it a reality :)

anjmao: Maybe this logic could be moved to provider side here https://github.com/mysteriumnetwork/node/blob/master/p2p/listener.go#L130

Issue #1959, opened 2020-03-31
Expose possibility to get all (even not connectable) nodes via tequila API


chompomonim:
It could be done by adding an additional field in quality oracle, so node could get json with additional data and expose it via tequila API.

PR #2074, opened 2020-04-16
Allow to request all nodes from discovery with monitoring data

soffokl:
Closes: #1959

### DIFF
--- a/core/discovery/apidiscovery/repository.go
+++ b/core/discovery/apidiscovery/repository.go
@@ -41,6 +41,7 @@ func (a *apiRepository) Proposal(id market.ProposalID) (*market.ServiceProposal,
 		NodeKey:         id.ProviderID,
 		ServiceType:     id.ServiceType,
 		AccessPolicyAll: true,
+		IncludeFailed:   true,
 	})
 	if err != nil {
 		return nil, err
--- a/tequilapi/endpoints/proposals.go
+++ b/tequilapi/endpoints/proposals.go
@@ -81,8 +81,6 @@ func NewProposalsEndpoint(proposalRepository proposal.Repository, qualityProvide
 //     schema:
 //       "$ref": "#/definitions/ErrorMessageDTO"
 func (pe *proposalsEndpoint) List(resp http.ResponseWriter, req *http.Request, params httprouter.Params) {
-	fetchConnectCounts := req.URL.Query().Get("fetch_connect_counts")
-
 	upperTimePriceBound, err := parsePriceBound(req, "upper_time_price_bound")
 	if err != nil {
 		utils.SendError(resp, err, http.StatusBadRequest)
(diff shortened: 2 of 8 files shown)


config/flags_node.go
```diff
@@ -170,6 +170,12 @@ var (
 		Usage: "Marks vendor (distributor) of the node for collecting statistics. " +
 			"3rd party vendors may use their own identifier here.",
 	}
+	// FlagUIEnable enables built-in web UI for node.
+	FlagIncludeFailedNodes = cli.BoolFlag{
```
Waldz, line 174: Don't introduce this flag, all filtering "configuration" - is passing Tequilapi parameters. I see you used it only method `mAPI.Proposals()` which is not used.

core/discovery/apidiscovery/repository.go
```diff
@@ -41,6 +42,7 @@ func (a *apiRepository) Proposal(id market.ProposalID) (*market.ServiceProposal,
 		NodeKey:         id.ProviderID,
 		ServiceType:     id.ServiceType,
 		AccessPolicyAll: true,
+		IncludeFailed:   config.GetBool(config.FlagIncludeFailedNodes),
```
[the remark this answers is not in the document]
soffokl: Removed it.

config/flags_node.go
soffokl: Removed it.

tequilapi/endpoints/proposals.go
```diff
@@ -226,6 +237,7 @@ func (pe *proposalsEndpoint) List(resp http.ResponseWriter, req *http.Request, p
 		proposalsRes.Proposals = append(proposalsRes.Proposals, proposalToRes(p))
 	}
 
+	fetchConnectCounts := req.URL.Query().Get("fetch_connect_counts")
```
Waldz, line 240: Probably now `fetch_metrics` is better name for that it does. Or do do granular fetch approach with query:
`&fetch_metric[connect_counts]=false&fetch_metric[monitoring_failed]=true`

tequilapi/endpoints/proposals.go
```diff
@@ -214,6 +224,7 @@ func (pe *proposalsEndpoint) List(resp http.ResponseWriter, req *http.Request, p
 		LowerTimePriceBound: lowerTimePriceBound,
 		UpperTimePriceBound: upperTimePriceBound,
 		ExcludeUnsupported:  true,
+		IncludeFailed:       req.URL.Query().Get("include_failed") == "true",
```
Waldz, line 227: `monitoring_failed=true` or `metrics.monitoring_failed=true` would be better name here as:
- as you would not leak API naming to Tequilapi contract
- it's Tequilapi endpoint , so alligned to Tequilapi proposal DTO

tequilapi/endpoints/proposals.go
```diff
@@ -57,8 +57,20 @@ type serviceDefinitionRes struct {
...
+type qualityMetricsRes struct {
+	ProviderID  string `json:"provider_id"`
+	ServiceType string `json:"service_type"`
+	metricsRes
+}
+
 type metricsRes struct {
-	ConnectCount quality.ConnectCount `json:"connect_count"`
+	ConnectCount     quality.ConnectCount `json:"connect_count"`
+	MonitoringFailed bool                 `json:"monitoring_failed"`
```
Waldz, line 73: BTW, are you introducing this to Tequilapi `client.go` somehow? We are migrating to same DTOs there at `contract` package.

tequilapi/endpoints/proposals.go
soffokl: Renamed.

tequilapi/endpoints/proposals.go
soffokl: Renamed.

tequilapi/endpoints/proposals.go
soffokl: It is in the `contract` package now.

PR #2075, opened 2020-04-16
Add location and ip detection retries

anjmao:
1. Add retries for ip and location detection. 
2. Use separate https://testnet-location.mysterium.network/api/v1/ip endpoint for public ip detection.
3. Send first invoice with fixed value = 1.

### DIFF
--- a/config/flags_location.go
+++ b/config/flags_location.go
@@ -28,7 +28,7 @@ var (
 	FlagIPDetectorURL = cli.StringFlag{
 		Name:  "ip-detector",
 		Usage: "Address (URL form) of IP detection service",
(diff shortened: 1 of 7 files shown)


Waldz:
What is reason we need it? If it's not working, then it not working - unreachable or i don't have connectivity.

Now it will block for like 20s to find out that it does't work. e.g. when we're using this call to measure connection success to provider.

anjmao:
Where will it block?

If Location or especially IP check fails it it's critical. Location oracle is distributed network service to which you call can fail in many cases. Even if you deploy a new location oracle at that time it can fail.

anjmao:
Closing for now, change with invoice 1 didn't solved e2e tests problem and need to investigate.

Issue #2094, opened 2020-04-21
Connection fails on newly registered identity until restart


soffokl:
**Describe the bug**
Currently, we are doing grand total recovery on identity unlock. This does not work correctly for the newly created identity.

**To Reproduce**
1. `identity new`
2. `identity unlock 0x0000...`
3. `identity register 0x0000...`
4. `connect 0x0000... 0x1111... openvpn`

**Expected behavior**
After registration, it should be possible to establish first connection successfully without a restart.

zolia:
Is this reproducible on mobile too? Did not notice this behaviour.

soffokl:
@zolia I didn't test it on mobile.

soffokl:
This could be related to the https://github.com/mysteriumnetwork/node/issues/2052 and https://github.com/mysteriumnetwork/node/issues/2051.

PR #2095, opened 2020-04-21
Try to recover grand total promised if it's not found

soffokl:
Closes: #2094

### DIFF
--- a/session/pingpong/consumer_balance_tracker.go
+++ b/session/pingpong/consumer_balance_tracker.go
@@ -19,6 +19,7 @@ package pingpong
 
 import (
 	"context"
+	"errors"
 	"sync"
 	"time"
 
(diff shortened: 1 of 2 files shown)


session/pingpong/consumer_balance_tracker.go
```diff
@@ -197,6 +197,12 @@ func (cbt *ConsumerBalanceTracker) ForceBalanceUpdate(id identity.Identity) uint
 	}
 
 	grandTotal, err := cbt.consumerGrandTotalsStorage.Get(id, cbt.accountantAddress)
+	if err == ErrNotFound {
+		if err := cbt.recoverGrandTotalPromised(id); err != nil {
+			log.Error().Err(err).Msg("Could not recover Grand Total Promised")
+		}
+		grandTotal, err = cbt.consumerGrandTotalsStorage.Get(id, cbt.accountantAddress)
```
zolia, line 204: What if this one fails too?

soffokl: The error is handled by the next if block. If it's `not found` again we continue assuming `grandTotal` is 0.

session/pingpong/consumer_balance_tracker.go
```diff
@@ -197,6 +197,12 @@ func (cbt *ConsumerBalanceTracker) ForceBalanceUpdate(id identity.Identity) uint
 	}
 
 	grandTotal, err := cbt.consumerGrandTotalsStorage.Get(id, cbt.accountantAddress)
+	if err == ErrNotFound {
```
anjmao, line 200: If possible it would be safer to use `errors.Is(err, ErrNotFound)` as if Get returns wrapped error this code will not get executed.

PR #2101, opened 2020-04-22
Handle 429 from accountant

vkuznecovas:
Accountant will now return 429 on request promise and reveal R.

This PR handles them in the following way:

Accountant caller will retry a request promise and reveal R request if a 429 is encountered. Upon encountering a non 429 error, the error will be bubbled straight away.

### DIFF
--- a/session/pingpong/accountant_caller.go
+++ b/session/pingpong/accountant_caller.go
@@ -18,6 +18,7 @@
 package pingpong
 
 import (
+	"context"
 	"encoding/hex"
 	"encoding/json"
 	"errors"
@@ -27,6 +28,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/cenkalti/backoff/v4"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/mysteriumnetwork/node/requests"
 	"github.com/mysteriumnetwork/payments/crypto"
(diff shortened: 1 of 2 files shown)


session/pingpong/accountant_caller.go
```diff
@@ -119,13 +121,27 @@ func (ac *AccountantCaller) RequestPromise(rp RequestPromise) (crypto.Promise, e
...
-	err = ac.doRequest(req, &res)
-	if err != nil {
-		return res, fmt.Errorf("could not request promise: %w", err)
 
-	}
-	return res, nil
+	return res, backoff.Retry(func() error {
+		err = ac.doRequest(req, &res)
+		if err != nil {
+			// if too many requests, retry
```
anjmao, line 135: Are we sure that after 1.5 second lock will be released? Maybe it would be safer and more clear to block in endpoint so it waits until lock is freed inside hermes (the same as simple locks works). In case it can't finish (let's say lock is not freed for some reason) this request will timeout.

vkuznecovas: This is indented to TRY again, not to ensure success. The accountant will lock for very short periods of time(less than 300ms currently). Besides, this error is not critical at all, as you can skip a few reveals/requests just fine.

PR #2103, opened 2020-04-22
Do not allow to overwrite accountant promise to lower value

vkuznecovas:
The accountant promise has an ever increasing value. Storage now validates this before overwriting.


session/pingpong/accountant_promise_storage.go
```diff
@@ -54,6 +57,15 @@ func (aps *AccountantPromiseStorage) Store(id identity.Identity, accountantID co
 	aps.lock.Lock()
 	defer aps.lock.Unlock()
 
+	previousPromise, err := aps.get(id, accountantID)
+	if err != nil && err != ErrNotFound {
+		return err
+	}
+
+	if previousPromise.Promise.Amount >= promise.Promise.Amount {
```
tadaskay, line 65: According to the error message, this should be `>`.

PR #2111, opened 2020-04-24
Even faster e2e

vkuznecovas:
E2e now starts multiple consumers, these then connect to the provider simultaneously. This mimics the real world scenario more closely.

Had to refactor accountant promise handling flow, as this caused insane races. To fix it, moved all the accountant promise logic to a new component: `accountant_promise_handler.go`. It will handle the promises sequentially and avoid race conditions.

Time savings:
After
`mage teste2ebasic  32.50s user 17.48s system 24% cpu 3:20.12 total `
Before
`mage teste2ebasic  45.29s user 18.08s system 21% cpu 4:58.23 total`


session/pingpong/accountant_promise_handler.go
```diff
@@ -0,0 +1,320 @@
...
+	}
+
+	details := rRecoveryDetails{
+		R:           hex.EncodeToString(er.r),
+		AgreementID: er.em.AgreementID,
+	}
+
+	bytes, err := json.Marshal(details)
+	if err != nil {
+		er.errChan <- fmt.Errorf("could not marshal R recovery details %w", err)
```
anjmao, line 183: `could not marshal R recovery details: %`

session/pingpong/accountant_promise_handler.go
```diff
@@ -0,0 +1,320 @@
...
+		return
+	}
+
+	encrypted, err := aph.deps.Encryption.Encrypt(er.providerID.ToCommonAddress(), bytes)
+	if err != nil {
+		er.errChan <- fmt.Errorf("could not encrypt R %w", err)
```
anjmao, line 189: Add `:` before wrap `could not encrypt R: %w`

session/pingpong/accountant_promise_handler.go
```diff
@@ -0,0 +1,320 @@
...
+	err = bus.SubscribeAsync(servicestate.AppTopicServiceStatus, aph.handleServiceEvent)
+	if err != nil {
+		return fmt.Errorf("could not subscribe to service events %w", err)
+	}
+	return nil
+}
+
+func (aph *AccountantPromiseHandler) handleServiceEvent(ev servicestate.AppEventServiceStatus) {
+	if ev.Status == string(servicestate.Running) {
+		aph.startOnce.Do(
```
anjmao, line 148: AccountantPromiseHandler is created as global instance once during node startup. If I'm correct:
start(ok)
stop(ok)
start(will not work as startOnce is used)
stop(stop will not work also)

session/pingpong/accountant_promise_handler_test.go
```diff
@@ -0,0 +1,223 @@
...
+		deps: AccountantPromiseHandlerDeps{
+			AccountantCaller:         &mockAccountantCaller{},
+			Encryption:               &mockEncryptor{},
+			EventBus:                 eventbus.New(),
+			AccountantPromiseStorage: &mockAccountantPromiseStorage{},
+		},
+		queue: make(chan enqueuedRequest),
+		stop:  make(chan struct{}),
+	}
+	go aph.handleRequests()
```
anjmao, line 42: It could be better to test via public methods only (if possible) when testing such full flow.

session/pingpong/accountant_promise_handler.go
vkuznecovas: That's the intended behaviour. It is only stopped on node exit, and started once the first time we launch the service.

docker-compose.e2e-basic.yml
```diff
@@ -176,7 +186,81 @@ services:
       --openvpn.port=3000
       openvpn,noop,wireguard
 
-  myst-consumer:
+  myst-consumer-noop:
```
Waldz, line 189: Lets have one consumer and create sessions from it. Doing consequent connects/disconnect showed bunch of bugs for us.
How to keep testing that?

Waldz: maybe additional connect after 1st session?

Waldz: Actually connects should happen in same container, to catch race bugs for us.
But we can run those connect tests in parallel

vkuznecovas: I leave this an exercise to the reader. Need a break from e2e for now.

Issue #2066, opened 2020-04-15
Allow to change beneficiary via CLI


chompomonim:
Node should be able to sign the beneficiary changing request and call transactor to make blockchain tx. It should also check if the beneficiary was checked on blockchain and if not, ping transactor once again in some period of time.

There should be possible to set beneficiary during identity registration (using CLI) and during first node start (so we could use proper beneficiary during identity registration).

All related tasks are here: https://github.com/mysteriumnetwork/node/milestone/41

PR #2120, opened 2020-04-27
Allow to set beneficiary address

### LINKED ISSUE #2125: Add fee for changing beneficiary address
The transactor does not charge anything now.

soffokl:
Closes https://github.com/mysteriumnetwork/node/issues/2066
Closes #2125

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -40,7 +40,7 @@ require (
 	github.com/mysteriumnetwork/go-openvpn v0.0.23
 	github.com/mysteriumnetwork/go-wondershaper v1.0.1
 	github.com/mysteriumnetwork/metrics v0.0.3
-	github.com/mysteriumnetwork/payments v0.0.13-0.20200430114543-8b00d4439302
+	github.com/mysteriumnetwork/payments v0.0.13-0.20200506030439-fe21695a9127
 	github.com/nats-io/gnatsd v1.4.1 // indirect
 	github.com/nats-io/go-nats v1.4.0
 	github.com/nats-io/nats-server v1.4.1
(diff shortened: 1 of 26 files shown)


go.mod
```diff
@@ -78,3 +78,5 @@ require (
 )
 
 replace github.com/nats-io/go-nats v1.4.0 => github.com/mysteriumnetwork/nats.go v1.4.1-0.20200303115848-b4a5324c56ed
+
+replace github.com/mysteriumnetwork/payments => ../payments
```
vkuznecovas, line 82: dont think this needs to get merged

session/pingpong/accountant_caller.go
```diff
@@ -202,6 +202,21 @@ func (ac *AccountantCaller) GetConsumerData(id string) (ConsumerData, error) {
 	return resp, nil
 }
 
+// GetProviderBeneficiary gets provider beneficiary address.
+func (ac *AccountantCaller) GetProviderBeneficiary(id string) (string, error) {
```
vkuznecovas, line 206: It would be better to check this on blockchain, as the accountant endpoints are cached, so a change there will take a while to propagate.

vkuznecovas: https://github.com/mysteriumnetwork/payments/blob/b04a3b1325f39dacfb5d01947052a3acb97afe3c/client/client.go#L191

vkuznecovas: This client is already used in node, so we could reuse it for this purpose as well.

go.mod
soffokl: yep. That is why this PR is draft. It will be changed once the payments changes merged.

Issue #2126, opened 2020-04-27
Can't disconnect from non-working provider


tadaskay:
**Describe the bug**
Can't disconnect from a non-working provider.

**To Reproduce**
1) Connect to `0x3eb30a6b70423926b0142aee41f6b6b6299695a3` 
2) Try to disconnect.

Disconnect does not work (node is stuck in disconnecting state)

**Expected behavior**
Node disconnects

**Environment (please complete the following information):**
 - Node version: master
 - OS: macOS
 - Docker: if running on docker, specify docker version and image tag

**Additional context**
If using openvpn, it's possible to kill myst with `-9`. If using wireguard, kill leaves stale DNS configuration - need to reset manually, e.g. `sudo networksetup -setdnsservers wi-fi 1.1.1.1`

zolia:
Looks like packet forwarding on such node is disabled. We get into connected state, but are not able to forward messages. This leaves a node in inconsistent state not able to ack connected state.

zolia:
node in question (0x3eb30a6b70423926b0142aee41f6b6b6299695a3)  is 0.22.5 thus no p2p messaging is supported. P2P disconnect should solve this issue. 

To make consumers resilient and independent from provider version we should still make disconnect on consumer to be robust.

PR #2130, opened 2020-04-27
Do not block connection state event publishing

anjmao:
As we have 10 subscribers for `AppTopicConnectionState` topic which is published during connect if one of the subscribers blocks the whole connect blocks. This was the case for reconnect logic as if there is no internet connection when connected it was not able to reconnect.

Fixes https://github.com/mysteriumnetwork/node/issues/2126

### DIFF
--- a/tequilapi/endpoints/connection_test.go
+++ b/tequilapi/endpoints/connection_test.go
@@ -22,6 +22,7 @@ import (
 	"net/http/httptest"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/julienschmidt/httprouter"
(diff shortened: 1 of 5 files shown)


tequilapi/endpoints/connection_test.go
```diff
@@ -138,13 +142,18 @@ func TestAddRoutesForConnectionAddsRoutes(t *testing.T) {
...
-	fakeState := &mockStateProvider{}
-	fakeState.stateToReturn.Connection.Session = connection.Status{
-		State:     connection.Disconnecting,
-		SessionID: "1",
+	manager := &mockConnectionManager{
+		onStatusReturn: connection.Status{
+			StartedAt:    time.Time{},
+			ConsumerID:   identity.Identity{},
+			AccountantID: common.Address{},
+			State:        "Disconnecting",
```
Waldz, line 150: `connection.Disconnecting`

PR #2136, opened 2020-04-28
Use lightweight keystore by default

anjmao:


vkuznecovas:
Remove the heavyweight all together?

PR #2148, opened 2020-04-29
Alleviate the connection race pains

vkuznecovas:
With the connection statuses being reported rather quickly races might happen.

Eventually will prevent these.

### DIFF
--- a/e2e/connection_test.go
+++ b/e2e/connection_test.go
@@ -134,11 +134,15 @@ func TestConsumerConnectsToProvider(t *testing.T) {
 }
 
 func recheckBalancesWithAccountant(t *testing.T, consumerID string, consumerSpending uint64, serviceType string) {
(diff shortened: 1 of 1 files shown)


e2e/connection_test.go
```diff
@@ -134,11 +134,16 @@ func TestConsumerConnectsToProvider(t *testing.T) {
...
+	var lastAccountant uint64
+	var lastConsumer uint64
+
+	assert.Eventually(t, func() bool {
+		accountantCaller := pingpong.NewAccountantCaller(requests.NewHTTPClient("0.0.0.0", time.Second), "http://accountant:8889/api/v2")
+		accountantData, err := accountantCaller.GetConsumerData(consumerID)
+		assert.NoError(t, err)
+		promised := accountantData.LatestPromise.Amount
+		return promised == consumerSpending
+	}, time.Second*10, time.Millisecond*300, fmt.Sprintf("Consumer reported spending %v  accountant says %v. Service type %v", lastConsumer, lastAccountant, serviceType))
```
soffokl, line 146: Where are the `lastConsumer` and `lastAccountant` updated? I see only declaration and output.

PR #2160, opened 2020-05-04
Fix leaking goroutines in NAT pinger

soffokl:
Closes https://github.com/mysteriumnetwork/node/issues/2159


nat/traversal/pinger.go
```diff
@@ -352,7 +352,7 @@ func sortedConns(pings []pingResponse) []*net.UDPConn {
 
 // waitMsg waits until conn receives given message or timeouts.
 func waitMsg(ctx context.Context, conn *net.UDPConn, msg string) error {
-	ok := make(chan struct{})
```
zolia, line 355: How does this help to fix a leak?

soffokl: It does not block on writing to channel when no one reads from the channel.

anjmao: Yes, good catch here.

PR #2182, opened 2020-05-07
Fix Install script test

soffokl:
https://yeri.be/error-bad-return-status-for-module-build-on-kernel-4-19-0-8-amd64-x86-64

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -188,7 +188,7 @@ install_raspbian() {
 install_debian() {
     # openvpn, etc.
     apt update
-    apt install -y ipset resolvconf openvpn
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -188,7 +188,7 @@ install_raspbian() {
 install_debian() {
     # openvpn, etc.
     apt update
-    apt install -y ipset resolvconf openvpn
+    apt install -y ipset resolvconf openvpn bc
```
Waldz, line 191: what is this utility for?

soffokl: This is build dependency for wireguard.

PR #2193, opened 2020-05-08
Make wireguard connect work without sudo

tadaskay:
1. Move node-supervisor code to node.
2. Add WireGuard remote client package for communication with supervisor.
3. Add ci build step to package supervisor for darwin.

### DIFF
--- a/services/wireguard/endpoint/remoteclient/tun.go
+++ b/services/wireguard/endpoint/remoteclient/tun.go
@@ -1,0 +1,55 @@
+// +build !windows
+
+/*
+ * Copyright (C) 2020 The "MysteriumNetwork/node" Authors.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package remoteclient
+
+import (
+	"fmt"
+	"net"
+	"os/user"
+
+	supervisorclient "github.com/mysteriumnetwork/node/supervisor/client"
+	"github.com/rs/zerolog/log"
+)
+
+// createTUN creates native TUN device for wireguard.
+func createTUN(iface string, subnet net.IPNet) error {
+	currentUser, err := user.Current()
+	if err != nil {
+		return err
+	}
+
+	actualIface, err := supervisorclient.Command("wg-up", "-iface", iface, "-uid", currentUser.Uid)
+	if err != nil {
+		return fmt.Errorf("failed to create wg interface: %w", err)
+	}
+	log.Debug().Msgf("Tunnel interface created: %s", actualIface)
+	if err := assignIP(iface, subnet); err != nil {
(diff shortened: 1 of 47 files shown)


zolia:
What still concerns me is how we will secure such unix socket. Outside applications would be able to access these security lifted facilities pretty much effortlessly. 

One possible way is to create a socketpair and send one socket end to peer using method like this:
https://blog.cloudflare.com/know-your-scm_rights/

Still this does not solve secure initial exchange problem. Somehow supervisor should be willing to communicate with authenticated node only. 

Any ideas how to do that in most simple manner?

anjmao:
@zolia Regarding auth. My idea is to try using sockets `SO_PEERCRED` option which allows to get pid of the caller. It means I should be able to verify if this pid is run by myst node process.

services/wireguard/endpoint/remoteclient/tun.go
```diff
@@ -0,0 +1,56 @@
...
+	}
+	log.Debug().Msgf("Tunnel interface created: %s", actualIface)
+	if err := assignIP(iface, subnet); err != nil {
+		return fmt.Errorf("failed to assign IP address: %w", err)
+	}
+	return nil
+}
+
+func destroyTUN(iface string) error {
+	// call supervisor wg-down -iface iface
```
soffokl, line 50:
```suggestion
```

supervisor/config/config.go
```diff
@@ -0,0 +1,69 @@
...
+
+// Write config file.
+func (c Config) Write() error {
+	if !c.valid() {
+		return errors.New("configuration is not valid")
+	}
+	var out strings.Builder
+	err := toml.NewEncoder(&out).Encode(c)
+	if err != nil {
+		return fmt.Errorf("could not encode cofiguration: %w", err)
```
soffokl, line 50:
```suggestion
		return fmt.Errorf("could not encode configuration: %w", err)
```

supervisor/daemon/myst_darwin.go
```diff
@@ -0,0 +1,115 @@
...
+	pid := cmd.Process.Pid
+	if err := ioutil.WriteFile(pidFile, []byte(strconv.Itoa(pid)), 0700); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func runWithSuccessTimeout(f func() error, timeout time.Duration) error {
+	done := make(chan error)
```
soffokl, line 81: If there is no receiver after timeout it will be blocked forever. 
```suggestion
	done := make(chan error, 1)
```

supervisor/daemon/transport/transport_darwin.go
```diff
@@ -0,0 +1,63 @@
...
+// Conversation is handled by the handlerFunc.
+func Start(handle handlerFunc) error {
+	if err := os.RemoveAll(sock); err != nil {
+		return fmt.Errorf("could not remove sock: %w", err)
+	}
+	l, err := net.Listen("unix", sock)
+	if err != nil {
+		return fmt.Errorf("error listening: %w", err)
+	}
+	if err := os.Chmod(sock, 0777); err != nil {
```
soffokl, line 39: Please add TODO to adjust this 777 permissions. It needs to be changed.

supervisor/daemon/wireguard/monitor.go
```diff
@@ -0,0 +1,71 @@
...
+	"fmt"
+	"sync"
+
+	"github.com/mysteriumnetwork/node/supervisor/daemon/wireguard/wginterface"
+)
+
+// Monitor creates/deletes the WireGuard interfaces and keeps track of them.
+type Monitor struct {
+	interfaces map[string]*wginterface.WgInterface
+	sync.Mutex
```
soffokl, line 30: If you are embedding mutex to the public struct, it can be locked from outside by calling `.Lock()`. This will break internal logic.
```suggestion
	mu sync.Mutex
```

Issue #2190, opened 2020-05-08
Proposal is not visible if max price is set in web UI


anjmao:
1. Set max price in WEB UI for GB and Time.
2. Start provider.
3. From consumer list proposals with default settings for price ranges.
```
payments.consumer.price-pergib-max:11000000
payments.consumer.price-perminute-max:50000
```

This is the proposal from provider.

```
        "payment_method_type": "BYTES_TRANSFERRED_WITH_TIME",
        "payment_method": {
            "price": {
                "amount": 50000,
                "currency": "MYST"
            },
            "duration": 30000000000,
            "bytes": 1073741,
            "type": "BYTES_TRANSFERRED_WITH_TIME"
        },
```

soffokl:
Do you think that the default price limit should be equal to the max allowed price?

tadaskay:
I don’t think it’s a bug.
We have defaults set in desktop for a reason, but you can adjust the slider to the maximum price you are willing to pay.

anjmao:
Yes, so the problem is when only on mobile which should be fixed with https://github.com/mysteriumnetwork/node/issues/2184 but it's quite important when.

Waldz:
pricePerTime() filter:
60*1000*1000*1000) / 30000000000 * 50000 = 100000 = 0,001 MYSTT

pricePerDataTransfer() filter:
(1*1024*1024*1024)  / 1073741 * 50000 = 50000038,370519520070483 = 
0,500000383705195 MYSTT

PR #2198, opened 2020-05-11
Allow max prices up to Android filters

Waldz:
Fixes: #2190


soffokl:
I was thinking it should be fixed in Android, not WebUI, regarding https://github.com/mysteriumnetwork/node/issues/2190

PR #2228, opened 2020-05-13
Destroy remote WireGuard device on close

anjmao:
Disconnect was not working when using supervisor since Wgctrl close doesn't know how to remove tunnel interface.

### DIFF
--- a/services/wireguard/endpoint/remoteclient/client.go
+++ b/services/wireguard/endpoint/remoteclient/client.go
@@ -20,16 +20,20 @@ package remoteclient
 import (
 	"fmt"
 	"net"
+	"sync"
 	"time"
 
 	wg "github.com/mysteriumnetwork/node/services/wireguard"
+	"github.com/mysteriumnetwork/node/utils"
 	"github.com/rs/zerolog/log"
 	"golang.zx2c4.com/wireguard/wgctrl"
 	"golang.zx2c4.com/wireguard/wgctrl/wgtypes"
 )
 
 type client struct {
-	wgc *wgctrl.Client
+	mu    sync.Mutex
+	iface string
+	wgc   *wgctrl.Client
 }
 
 // New create new remote WireGuard client which communicates with supervisor.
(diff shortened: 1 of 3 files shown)


services/wireguard/endpoint/remoteclient/client.go
```diff
@@ -137,6 +142,27 @@ func (c *client) PeerStats(iface string) (*wg.Stats, error) {
...
+func (c *client) Close() (err error) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	var errs []error
+	defer func() {
+		for i := range errs {
+			log.Error().Err(errs[i]).Msg("Failed to close wireguard remote client")
+			if err == nil {
+				err = errs[i]
```
Waldz, line 154: Returning just last error, `ErrorComposite.Err()` would be useful to return multiple errors

zolia: Or just wrap them all with %w

services/wireguard/endpoint/remoteclient/client.go
```diff
@@ -137,6 +142,27 @@ func (c *client) PeerStats(iface string) (*wg.Stats, error) {
...
+
+	if err := c.DestroyDevice(c.iface); err != nil {
+		errs = append(errs, err)
+	}
+
+	if err := c.wgc.Close(); err != nil {
+		errs = append(errs, err)
+	}
+
+	return nil
```
soffokl, line 167: If you are returning `nil` here, defer will not overwrite it.

services/wireguard/endpoint/remoteclient/client.go
```diff
@@ -137,6 +142,27 @@ func (c *client) PeerStats(iface string) (*wg.Stats, error) {
...
 }
 
-func (c *client) Close() error {
-	return c.wgc.Close()
+func (c *client) Close() (err error) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	var errs []error
+	defer func() {
```
soffokl, line 150: Do you need this defer here? Just move the `for` to the end of the function.

Issue #2200, opened 2020-05-11
P2P should map port once


Waldz:
Now server maps P2P port after receiving connection.
Should be mapped once during service start.
Rename to "Myst node P2P port mapping" also.

<img width="979" alt="Screen Shot 2020-05-11 at 18 22 21" src="https://user-images.githubusercontent.com/2713911/81582113-4d65bc80-93b8-11ea-8e63-b52a1d1060dd.png">

zolia:
>Should be mapped once during service start.

No, p2p port mapping happens on session setup since we cannot properly predict which ports we will use.

Currently failed sessions port-mappings are cleaned up only when node shuts down. 
Should be fixed with channel timeout.

PR #2229, opened 2020-05-13
Timeout channels on failed session setup

zolia:
Fixes: #2200

easily reporducible by attempting to connect to your node from client on the same network.﻿Should fix port mappings piling up on routers.

### DIFF
--- a/p2p/channel.go
+++ b/p2p/channel.go
@@ -170,6 +170,9 @@ type channel struct {
 
 	// weather channel saw remote traffic
 	remoteAlive chan struct{}
+
+	// terminate remote aliveness checking only once
+	remoteAliveOnce sync.Once
 }
 
 // newChannel creates new p2p channel with initialized crypto primitives for data encryption
@@ -235,14 +238,15 @@ func newChannel(remoteConn *net.UDPConn, privateKey PrivateKey, peerPubKey Publi
 func (c *channel) remoteReadLoop() {
 	buf := make([]byte, mtuLimit)
 	latestPeerAddr := c.peer.addr()
+
+	go c.checkIfChannelAlive()
+
 	for {
 		select {
 		case <-c.stop:
 			return
 		default:
 		}
-
-		go c.checkIfChannelAlive()
 
 		n, addr, err := c.tr.remoteConn.ReadFrom(buf)
 		if err != nil {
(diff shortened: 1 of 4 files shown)


p2p/channel.go
```diff
@@ -511,6 +519,24 @@ func (c *channel) setUpnpPortsRelease(release []func()) {
 	c.upnpPortsRelease = release
 }
 
+func (c *channel) checkIfChannelAlive() {
+	for {
+		select {
+		case <-c.stop:
+			return
+		case <-time.After(initialTrafficTimeout):
+			if c.remoteAlive {
```
vkuznecovas, line 528: you're gonna race here. Use a channel and add a third case:

```
case <-c.remoteAlive:
   return
```

zolia: With channel is cleaner anyways. Thanx.

p2p/channel.go
```diff
@@ -244,6 +252,10 @@ func (c *channel) remoteReadLoop() {
 			return
 		}
 
+		c.once.Do(func() {
```
soffokl, line 255: You cannot use `once` twice. It's already used in `Close()`.

zolia: You can and it works, since you are performing different action with it.

soffokl: https://play.golang.org/p/Gyyqn8EwQpI

vkuznecovas: confirming it does not work.

p2p/channel.go
```diff
@@ -511,6 +523,22 @@ func (c *channel) setUpnpPortsRelease(release []func()) {
 	c.upnpPortsRelease = release
 }
 
+func (c *channel) checkIfChannelAlive() {
+	for {
```
anjmao, line 527: for loop is not needed here as you want this logic to execute only once

p2p/channel.go
```diff
@@ -236,6 +242,8 @@ func (c *channel) remoteReadLoop() {
 		default:
 		}
 
+		go c.checkIfChannelAlive()
```
anjmao, line 245: checkIfChannelAlive can be moved outside for loop

p2p/channel.go
zolia: True, we figured out why it worked in my case. Actual channel close was not being executed after successful session.

Issue #2034, opened 2020-04-09
Pingpong: first invoice should be critical


vkuznecovas:
We need to make sure that the first session invoice gets sent and paid for by the consumer as soon as the session starts. This is the only way to make sure that the consumer is able to pay for the service.

There are 2 issues:

1) on sending failure of the invoice, the provider will just increase a failure counter, and ignore the error
2) on successful send, the invoice gets handled just like any other invoice, meaning that if consumer does not respond to it, we will just increase a counter and ignore the error.

Suggested changes:

1) Make sure that the invoice is sent AFTER the p2p channel is ready on both ends.
2) Make sure that the first invoice is treated as a special case and we fail early.

zolia:
Moving this to TODO, as its important and its a bug.

anjmao:
Can we start VPN session only after first invoice is sent and received back?

anjmao:
For now easy fix for first invoice would be to check for ErrHandlerNotRegistered and retry each second.

PR #2230, opened 2020-05-14
First invoice should be critical

soffokl:
Closes: https://github.com/mysteriumnetwork/node/issues/2034

### DIFF
--- a/session/manager.go
+++ b/session/manager.go
@@ -20,6 +20,7 @@ package session
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"net"
 	"sync"
 	"time"
@@ -151,7 +152,7 @@ type Manager struct {
 
 // Start starts a session on the provider side for the given consumer.
 // Multiple sessions per peerID is possible in case different services are used
-func (manager *Manager) Start(session *Session, consumerID identity.Identity, consumerInfo ConsumerInfo, proposalID int, config ServiceConfiguration, pingerParams *traversal.Params) (err error) {
+func (manager *Manager) Start(session *Session, consumerID identity.Identity, consumerInfo ConsumerInfo, proposalID int) (err error) {
 	manager.creationLock.Lock()
 	defer manager.creationLock.Unlock()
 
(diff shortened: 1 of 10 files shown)


session/pingpong/invoice_tracker.go
```diff
@@ -297,6 +302,28 @@ func (it *InvoiceTracker) Start() error {
...
+	timeout := time.After(wait)
+
+	for {
+		select {
+		case <-time.After(10 * time.Millisecond):
+			it.invoiceLock.Lock()
+			if it.firstInvoicePaid {
+				return nil
+			}
+			it.invoiceLock.Unlock()
```
vkuznecovas, line 316: if first invocie is paid this lock will not get released.

session/manager.go
```diff
@@ -190,6 +190,11 @@ func (manager *Manager) Start(session *Session, consumerID identity.Identity, co
 		}
 	}()
 
+	log.Info().Msg("Waiting for a first invoice to be paid")
+	if err := engine.WaitFirstInvoice(3 * time.Second); err != nil {
```
anjmao, line 194: I would increase first invoice to at least 10 seconds as some consumers can be slow.

soffokl: Left it for 3 seconds for compatibility time only, to not slow down connection from old nodes too much. I'll add TODO for it.

session/pingpong/invoice_payer.go
```diff
@@ -250,7 +249,6 @@ func (ip *InvoicePayer) issueExchangeMessage(invoice crypto.Invoice) error {
 
 	ip.deps.EventBus.Publish(event.AppTopicInvoicePaid, event.AppEventInvoicePaid{
 		ConsumerID: ip.deps.Identity,
-		SessionID:  ip.deps.SessionID,
```
vkuznecovas, line 253: Why remove this?

soffokl: Returned it back. Setting now SessionID after session established.

Issue #2241, opened 2020-05-15
Ubuntu 20.04 failed to start mysterium node


soffokl:
```
May 15 07:27:30 ubuntu-2gb-nbg1-1 myst[1376]: 2020-05-15T07:27:30.653 ERR cmd/mysterium_node/mysterium_node.go:58  > Failed to execute command:  error="iptables cmd error: \"sudo /sbin/iptables --version\": exit status 1 output: sudo: /sbin/iptables: command not found\n"
May 15 07:27:30 ubuntu-2gb-nbg1-1 systemd[1]: mysterium-node.service: Main process exited, code=exited, status=1/FAILURE

root@ubuntu-2gb-nbg1-1:~# whereis iptables
iptables: /usr/sbin/iptables /usr/share/iptables /usr/share/man/man8/iptables.8.gz
```

vkuznecovas:
Closed by #2266

PR #2242, opened 2020-05-15
Do not use absolute path for system tools

soffokl:
Closes: https://github.com/mysteriumnetwork/node/issues/2241

### DIFF
--- a/bin/package/sudoers/mysterium-node
+++ b/bin/package/sudoers/mysterium-node
@@ -2,6 +2,7 @@
 # let mysterium-node run any command just as root without password
 mysterium-node    ALL=NOPASSWD: /sbin/sysctl
 mysterium-node    ALL=NOPASSWD: /sbin/iptables
+mysterium-node    ALL=NOPASSWD: /usr/sbin/iptables
 mysterium-node    ALL=NOPASSWD: /usr/sbin/ipset
 mysterium-node    ALL=NOPASSWD: /sbin/ipset
 mysterium-node    ALL=NOPASSWD: /sbin/ip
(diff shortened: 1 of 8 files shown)


zolia:
Removing absolute paths from core utilities is considered a security risk. You never know if someone having access for current user's PATH variable would not substitute this utility with their own. 

Basically we are introducing another vulnerability. Ways to fix this: 
  - check for current PATH and analyze its priorities ( /usr/bin/ /usr/sbin/ etc. should go first)
  - retry with another absolute path if first fails
  - check which that utility is on and check that against whitlisted system paths.

soffokl:
@zolia we are sending iptables commands to set rules which is publically available, even if someone replacing iptables in his system with something else, it will just not work, we are not leaking anything private.

soffokl:
And we are doing sudo without checkin absolute path.

zolia:
@soffokl yes, sudoers safeguards this atm, but in general we should not follow such logic since sudoers might get changed by the user. We probably should execute sudo with absolute path too.

3rd parties replacing current user's path is a common exploit. Lets say they substitute iptables which do not add fw rules...

soffokl:
Well if the user can change files across the system we cannot do anything with it. They can even put the file on the path that we are expecting. Users can remove rules after we added it. Users can break everything if they want.

zolia:
No, what I meant by user might change sudoers file is that user might misconfigure it, for example allow execution of broader paths. Malicious user might have just found a way to affect proper user's PATH. In such situation executing with absolute paths would save proper user.

anjmao:
Yes, this looks insecure. It's not about what can current user do or misconfigure. It's about what can attacker do without user knowing about it.

soffokl:
okey

PR #2256, opened 2020-05-18
Make go buildable crossplatform with `mage build`

tadaskay:
This is mostly for making dev life easier under windows, as well as one more step away from shell scripts to mage.

### DIFF
--- a/utils/fileutil/fileutil.go
+++ b/utils/fileutil/fileutil.go
@@ -1,0 +1,125 @@
+/*
+ * Copyright (C) 2020 The "MysteriumNetwork/node" Authors.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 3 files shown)


utils/fileutil/fileutil.go
```diff
@@ -0,0 +1,125 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+// Copyright 2018 The Prometheus Authors
```
zolia, line 18: Its from prometheus?

tadaskay: Yes, but copied over instead of depending on it

PR #2266, opened 2020-05-19
Make iptables path consistent across dists

zolia:
fixes: #2146 and #2241

/usr/sbin/iptables is a preferred way according recent distributions. Ensuring that this path exist.

- [x] make links to iptables in deb post install
- [x] ensure myst finds iptables in docker image

### DIFF
--- a/bin/helpers/prepare-run-env.sh
+++ b/bin/helpers/prepare-run-env.sh
@@ -1,9 +1,33 @@
 #!/usr/bin/env bash
 
-setDefaultRoute() {
+source bin/helpers/functions.sh
+
+function setDefaultRoute {
     GW=$1;
     ip route del default
     ip route add default via ${GW}
+}
+
+function ensure_paths {
+    iptables_path=`which iptables`
+    if [[ ${iptables_path} == "" ]]; then
+      echo "required dependency missing: iptables"
+      exit 1
+    fi
+
+    # validate utility against valid system paths
+    basepath=${iptables_path%/*}
+    echo "iptables basepath detected: ${basepath}"
+    if ! [[ ${basepath} =~ (^/usr/sbin|^/sbin|^/bin|^/usr/bin) ]]; then
+      echo "invalid basepath for dependency - check if system PATH has not been altered"
+      exit 1
+    fi
+
+    iptables_required_path="/usr/sbin/iptables"
+
+    if ! [[ -x ${iptables_required_path} ]]; then
+        ln -s ${iptables_path} ${iptables_required_path}
+    fi
 }
 
 if [ -n "$GATEWAY" ]; then
@@ -30,3 +54,4 @@ if [ ! -d "$OS_DIR_DATA" ]; then
     mkdir -p $OS_DIR_DATA
 fi
 
+ensure_paths
--- a/bin/package/installation/post-install.sh
+++ b/bin/package/installation/post-install.sh
@@ -39,6 +39,24 @@ function install_chkconfig {
         && chkconfig --add mysterium-node
 }
 
+function ensure_paths {
+    iptables_path=`which iptables`
+
+    # validate utility against valid system paths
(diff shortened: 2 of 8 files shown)


install.sh
```diff
@@ -254,4 +276,8 @@ echo "### Installing myst & dependencies"
 install
 echo "### Installing myst & dependencies - done"
 
+echo "### Ensuring sane paths for dependencies"
```
tadaskay, line 279: What about "Normalizing dependency paths" or "Creating dependency symlinks"? Those sound more to the point to me.

install.sh
```diff
@@ -231,6 +231,28 @@ install() {
...
+    if [[ ${iptables_path} == "" ]]; then
+       echo "required dependecy missing: iptables"
+       return
+    fi
+
+    # validate utility against valid system paths
+    basepath=${iptables_path%/*}
+    echo "iptables basepath detected: ${basepath}"
+    if ! [[ ${basepath} =~ (^/usr/sbin|^/sbin|^/bin|^/usr/bin) ]]; then
+      echo "invalid basepath for dependecy - check if system PATH has not been altered"
```
tadaskay, line 245:
```suggestion
      echo "invalid basepath for dependency - check if system PATH has not been altered"
```

install.sh
```diff
@@ -231,6 +231,28 @@ install() {
...
+    if [[ ${iptables_path} == "" ]]; then
+       echo "required dependency missing: iptables"
+       return
+    fi
+
+    # validate utility against valid system paths
+    basepath=${iptables_path%/*}
+    echo "iptables basepath detected: ${basepath}"
+    if ! [[ ${basepath} =~ (^/usr/sbin|^/sbin|^/bin|^/usr/bin) ]]; then
+      echo "invalid basepath for dependecy - check if system PATH has not been altered"
```
tadaskay, line 245:
```suggestion
      echo "invalid basepath for dependency - check if system PATH has not been altered"
```
🦠

bin/helpers/prepare-run-env.sh
```diff
@@ -1,11 +1,35 @@
...
     GW=$1;
     ip route del default
     ip route add default via ${GW}
 }
 
+function ensure_paths {
+    iptables_path=`which iptables`
+    if [[ ${iptables_path} == "" ]]; then
+       echo "required dependency missing: iptables"
+       return
```
soffokl, line 15: Should it be `exit 1` or some error code returned here? Now it just prints output and exit with 0 status.

bin/helpers/prepare-run-env.sh
```diff
@@ -1,11 +1,35 @@
...
+    fi
+
+    # validate utility against valid system paths
+    basepath=${iptables_path%/*}
+    echo "iptables basepath detected: ${basepath}"
+    if ! [[ ${basepath} =~ (^/usr/sbin|^/sbin|^/bin|^/usr/bin) ]]; then
+      echo "invalid basepath for dependency - check if system PATH has not been altered"
+      return
```
soffokl, line 23: The same here.

bin/package/installation/post-install.sh
```diff
@@ -39,6 +39,28 @@ function install_chkconfig {
         && chkconfig --add mysterium-node
 }
 
+function ensure_paths {
+    iptables_path=`which iptables`
+    if [[ ${iptables_path} == "" ]]; then
+       echo "required dependency missing: iptables"
```
soffokl, line 45: It should be enough to have just dependency for the package, no need to recheck it in `post-install` script.

zolia: Added this for cases, when user forces dpkg install. This lets install package without dependencies.

zolia: Maybe you are right. Removed this additional check. User should rely on OS provided depencency policy.

PR #2267, opened 2020-05-20
Kill node in a more reliable way

tadaskay:
Give node 3 seconds to shutdown gracefully, then use the force


tadaskay:
Need to make disconnect faster then :)

PR #2280, opened 2020-05-22
Fix reseting DNS configuratin on MacOS

### LINKED ISSUE #2133: After disconnect DNS is not reset to default
**Current behaviour:**
On macOS I use `1.1.1.1` as my default DNS. After I connect to some node and disconnect my DNS is reset to routers default DNS.

**Expected behaviour**
Reset to previous set DNS.

soffokl:
Closes #2133

### DIFF
--- a/bin/package/config/darwin/update-resolv-conf
+++ b/bin/package/config/darwin/update-resolv-conf
@@ -23,7 +23,7 @@ NETWORKSETUP=/usr/sbin/networksetup
 IFSSAVE=$IFS
 IFS=$'\n'
 # Get adapter list
-for i in `$NETWORKSETUP -listallnetworkservices |grep -v denotes`; do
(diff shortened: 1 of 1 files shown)


bin/package/config/darwin/update-resolv-conf
```diff
@@ -39,44 +39,65 @@ update_all_dns()
...
 {
     for adapter in "${adapters[@]}"
     do
-    echo updating dns for $adapter
-    $NETWORKSETUP -setdnsservers "$adapter" empty
-    $NETWORKSETUP -setsearchdomains "$adapter" empty
+        echo updating dns for $adapter
+
+        dnsservers=""
+        searchdomains=""
```
zolia, line 62: do we need this extra initialization here? We are doing it right below.

soffokl: It's a loop, it preserves value from the previous iteration. Resetting it explicitly.

Issue #2176, opened 2020-05-06
Allow only one active session from consumer


anjmao:
Currently I can create many sessions from the same consumer by connecting and simply disabling network when clicking disconnect. There is no validation on provider side to check if there is already active session for given consumer.

To solve this I suggest that during connect we check if there is active session for given consumer identity and if there is destroy it so we don't have stale sessions left. 

Steps to reproduce:
1. Connect to provider.
2. Disconnect from network.
3. Click disconnect. Now consumer is disconnected but since there was no network he didn't notified provider with session destroy.
4. Repeat from 1.
5. Now provider has 2 sessions.

PR #2285, opened 2020-05-26
Clear stale consumer sessions on new connect

soffokl:
Closes: https://github.com/mysteriumnetwork/node/issues/2176

### DIFF
--- a/session/manager.go
+++ b/session/manager.go
@@ -101,6 +101,7 @@ type PromiseProcessor interface {
 type Storage interface {
 	Add(sessionInstance Session)
 	Find(id ID) (Session, bool)
+	FindBy(opts FindOpts) (ID, bool)
 	Remove(id ID)
 }
 
@@ -158,6 +159,8 @@ func (manager *Manager) Start(session *Session, consumerID identity.Identity, co
 		err = ErrorInvalidProposal
 		return
 	}
+
+	manager.clearStaleSession(consumerID, manager.currentProposal.ServiceType)
 
 	session.ServiceType = manager.currentProposal.ServiceType
 	session.ServiceID = manager.serviceId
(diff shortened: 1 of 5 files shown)


session/manager.go
```diff
@@ -159,6 +160,8 @@ func (manager *Manager) Start(session *Session, consumerID identity.Identity, co
 		return
 	}
 
+	manager.clearStaleSession(consumerID, manager.currentProposal.ServiceType)
```
Waldz, line 163: Part of cleaning happens in routine, maybe everything then

session/manager.go
```diff
@@ -221,13 +224,23 @@ func (manager *Manager) Acknowledge(consumerID identity.Identity, sessionID stri
...
 }
 
+func (manager *Manager) clearStaleSession(consumerID identity.Identity, serviceType string) {
+	sessionID, ok := manager.sessionStorage.FindBy(FindOpts{
+		Peer:        &consumerID,
+		ServiceType: serviceType,
+	})
+	if ok {
+		log.Info().Msgf("Cleaning stale session %s for %s consumer", sessionID, consumerID.Address)
+		go manager.Destroy(consumerID, string(sessionID))
```
Waldz, line 234: This will triggers storage traversal 2 times, reuse found `session` object

session/manager.go
soffokl: No, we need to make sure that we got list of sessions before creating new, otherwise we can clean current session too.

session/manager.go
Waldz: I'm not sure we want to reuse `Destroy()` here, it's logic has different purpose

soffokl: I do not understand this. 
What is storage traversal? 
What logic has `Destroy` except the destroying session?

session/manager.go
anjmao: @Waldz Please answer comment. Thanks.

session/manager.go
anjmao: @Waldz Please answer comment. Thanks.

Waldz: Session storage traversal is happening 2 times with session find methods.

`Destroy` purpose is for public contract. It does 2 things: a) finds session + b) cleans found session.
Your new use-case already has found session and you need just b) clean found session. So:
- either make needed part resuable in `destroySession(session Session)`
- either just call cleaning for found sessions:
```
manager.sessionStorage.Remove(ID(sessionID))
close(session.done)
```

session/manager.go
Waldz: Oh.. i couldn't assume it from reading code and was strange to find that just some part executed in routine. 
Maybe comment here would be beneficial.

session/manager.go
```diff
@@ -221,13 +224,23 @@ func (manager *Manager) Acknowledge(consumerID identity.Identity, sessionID stri
 	return nil
 }
 
+func (manager *Manager) clearStaleSession(consumerID identity.Identity, serviceType string) {
+	sessionID, ok := manager.sessionStorage.FindBy(FindOpts{
```
Waldz, line 228: BTW, should not it be a list of customer stale sessions we should find in storage?

session/manager.go
soffokl: Comment added.

session/manager.go
soffokl: Extracted `destroySession` from `Destroy`.

session/manager.go
soffokl: If you are creating second session, first one will be destroyed. It's impossible to have more than 1 session.

Issue #2233, opened 2020-05-14
Cleanup stale routes with supervisor


zolia:
On client computers might happen such situations as: 
 - computer might hibernate with vpn connection on
 - network might change and node gets killed for whatever reason
 - [place your case here]

This might lead to static route leaks. Meaning there will be stale static route to previously used provider IP and this route will not allow to establish connection to the same provider again. 

We need to cleanup such stale routes. 

Problem here is that we don't know which routes ARE stale. It might be equally active connection. 

We assume that there will only be a single MN session from client computer. 
Supervisor should track witch IPs were used to establish a previous session. When new session comes in, node informs supervisor about it. If supervisor sees a stale route it clears it before confirming new connection to provider.

zolia:
should fix: #2202

PR #2302, opened 2020-05-28
Refactor WireGuard device creation

anjmao:
Refactor WireGuard device creation to setup peer and network in one call `ConfigureDevice`

Interface is simplified from 

```go
type WgClient interface {
	ConfigureDevice(config wg.DeviceConfig) error
	ConfigureRoutes(iface string, ip net.IP) error
	DestroyDevice(name string) error
	AddPeer(iface string, peer wg.Peer) error
	RemovePeer(name string, publicKey string) error
	PeerStats(iface string) (*wg.Stats, error)
	Close() error
}
```

to
```go
type WgClient interface {
	ConfigureDevice(config wg.DeviceConfig) error
	DestroyDevice(name string) error
	PeerStats(iface string) (*wg.Stats, error)
	Close() error
}
```

This also allows to configure wg device in supervisor by passing full wireguard device config as json string. In next PR this will be useful as configuring network on Windows could be implemented the same way as WireGuard-Windows using native win apis instead of powershell and netsh.
```
wg-up -config "device JSON config"
```

PR #2309, opened 2020-06-01
Cleare stale route records

soffokl:
#2233

### DIFF
--- a/cmd/di.go
+++ b/cmd/di.go
@@ -194,6 +194,8 @@ func (di *Dependencies) Bootstrap(nodeOptions node.Options) error {
 		return err
 	}
 
+	netutil.ClearStaleRoutes()
+
 	if err := di.bootstrapNetworkComponents(nodeOptions); err != nil {
 		return err
 	}
@@ -393,6 +395,8 @@ func (di *Dependencies) bootstrapStorage(path string) error {
 	}
 
 	di.Storage = localStorage
+
+	netutil.SetRouteManagerStorage(di.Storage)
 
 	invoiceStorage := pingpong.NewInvoiceStorage(di.Storage)
 	di.ProviderInvoiceStorage = pingpong.NewProviderInvoiceStorage(invoiceStorage)
--- a/cmd/di.go
+++ b/cmd/di.go
@@ -46,6 +46,7 @@ import (
 	"github.com/mysteriumnetwork/node/core/state"
 	"github.com/mysteriumnetwork/node/core/storage/boltdb"
 	"github.com/mysteriumnetwork/node/core/storage/boltdb/migrations/history"
+	"github.com/mysteriumnetwork/node/core/storage/boltdb/migrator"
 	"github.com/mysteriumnetwork/node/eventbus"
 	"github.com/mysteriumnetwork/node/feedback"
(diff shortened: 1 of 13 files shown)


cmd/di.go
```diff
@@ -194,6 +194,8 @@ func (di *Dependencies) Bootstrap(nodeOptions node.Options) error {
 		return err
 	}
 
+	netutil.ClearStaleRoutes()
```
anjmao, line 197: After we merge https://github.com/mysteriumnetwork/node/pull/2302 if would be better to move this to suppervisor command to clear routes.

utils/netutil/network_windows.go
```diff
@@ -32,16 +32,15 @@ func assignIP(iface string, subnet net.IPNet) error {
...
-		return err
-	}
-
+func excludeRoute(ip, gw net.IP) error {
 	out, err := exec.Command("powershell", "-Command", "route add "+ip.String()+"/32 "+gw.String()).CombinedOutput()
 	return errors.Wrap(err, string(out))
 }
 
+func deleteRoute(ip, gw string) error {
+	return cmdutil.SudoExec("route", "delete", ip+"/32", "gw")
```
anjmao, line 41: This looks not correct as gw arg is not used. What is this last param "gw" ?

soffokl: Good catch, we don't need this this `"gw"`, removed it.

cmd/di.go
```diff
@@ -394,6 +396,8 @@ func (di *Dependencies) bootstrapStorage(path string) error {
 
 	di.Storage = localStorage
 
+	netutil.SetRouteManagerStorage(di.Storage)
```
zolia, line 399: Why not to have this in factory? And call ClearStaleRoutes() on routeManager instance?

supervisor/daemon/daemon.go
```diff
@@ -132,7 +143,15 @@ func (d *Daemon) wgDown(args ...string) (err error) {
 	if *interfaceName == "" {
 		return errors.New("-iface is required")
 	}
-	return d.monitor.Down(*interfaceName)
+
+	err = d.monitor.Down(*interfaceName)
+	if err != nil {
+		return err
```
zolia, line 149: wrap an error

utils/netutil/network.go
```diff
@@ -21,25 +21,96 @@ import (
...
+		}
+	}
 }
 
 // ExcludeRoute excludes given IP from VPN tunnel.
 func ExcludeRoute(ip net.IP) error {
-	return excludeRoute(ip)
+	gw, err := gateway.DiscoverGateway()
+	if err != nil {
+		return err
```
zolia, line 89: wrap an error

utils/netutil/network.go
```diff
@@ -21,25 +21,96 @@ import (
...
+	}
+
+	if defaultRouteManager != nil {
```
zolia, line 92: these check would be redundant if proper struct instance is used.

cmd/di.go
soffokl: Routes cleaning is a system-wide component. You never need to have more than one of them. Having it independent allows to access it simpler, without passing it down through all components.

Issue #1212, opened 2019-07-09
Connection "Connecting" state can be stuck forever


soffokl:
```
» connect 0x1396bb371aa4a9aa6fdf4bc20e8ff075f36e0871 0x2ed991d33a828f5cc2b0db9b5cf4eda07762db4f openvpn
[CONNECTING] from: 0x1396bb371aa4a9aa6fdf4bc20e8ff075f36e0871 to: 0x2ed991d33a828f5cc2b0db9b5cf4eda07762db4f
[WARNING] Put http://127.0.0.1:4051/connection: net/http: request canceled (Client.Timeout exceeded while awaiting headers)
1562662229749992000 [Error] [Tequilapi.Client] Put http://127.0.0.1:4051/connection: net/http: request canceled (Client.Timeout exceeded while awaiting headers)
» connect 0x1396bb371aa4a9aa6fdf4bc20e8ff075f36e0871 0x2ed991d33a828f5cc2b0db9b5cf4eda07762db4f openvpn
[CONNECTING] from: 0x1396bb371aa4a9aa6fdf4bc20e8ff075f36e0871 to: 0x2ed991d33a828f5cc2b0db9b5cf4eda07762db4f
[WARNING] server response invalid: 409 Conflict (http://127.0.0.1:4051/connection). Possible error: connection already exists
1562662321534510000 [Error] [Tequilapi.Client] server response invalid: 409 Conflict (http://127.0.0.1:4051/connection). Possible error: connection already exists
» status
[INFO] Status: Connecting
[INFO] SID:
»
```

```
2019-07-09T08:48:49.747624 [Info] [NATS.DialogEstablisher] Connecting to: market.Contact{Type:"nats/v1", Definition:discovery.ContactNATSV1{Topic:"0x2ed991d33a828f5cc2b0db9b5cf4eda07762db4f.openvpn", BrokerAddresses:[]string{"nats://testnet-broker.mysterium.network:4222"}}}
2019-07-09T08:48:49.747644 [Info] [Noop firewall]Allow testnet-broker.mysterium.network access
[...]

PR #2312, opened 2020-06-01
Add timeout for p2p channel dial

soffokl:
Closes: https://github.com/mysteriumnetwork/node/issues/1212

### DIFF
--- a/core/connection/manager.go
+++ b/core/connection/manager.go
@@ -39,6 +39,10 @@ import (
 	"github.com/rs/zerolog/log"
 )
 
+const (
+	p2pDialTimeout = 30 * time.Second
+)
+
 var (
 	// ErrNoConnection error indicates that action applied to manager expects active connection (i.e. disconnect)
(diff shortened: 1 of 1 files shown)


core/connection/manager.go
```diff
@@ -357,7 +361,8 @@ func (m *connectionManager) cleanAfterDisconnect() {
 }
 
 func (m *connectionManager) createP2PChannel(ctx context.Context, consumerID, providerID identity.Identity, serviceType string, contactDef p2p.ContactDefinition) (p2p.Channel, error) {
-	channel, err := m.p2pDialer.Dial(ctx, consumerID, providerID, serviceType, contactDef)
+	timeoutCtx, _ := context.WithTimeout(ctx, p2pDialTimeout)
```
anjmao, line 364: Do not ignore cancel to prevent memory leaks.

```go
timeoutCtx, cancel := context.WithTimeout(ctx, p2pDialTimeout)
defer cancel()
```

Issue #1752, opened 2020-02-20
kills switch is turned on even flag is off


zolia:
```
2020-02-20T13:01:59.915 DBG config/config.go:185                     > Returning default value firewall.killSwitch.always:false
```

```
Chain OUTPUT (policy ACCEPT 0 packets, 0 bytes)
 pkts bytes target     prot opt in     out     source               destination
    0     0 MYST_CONSUMER_KILL_SWITCH  all  --  *      *       10.182.0.2           0.0.0.0/0
```

cvl:
@zolia will it be possible to have this kill-switch in mobile / desktop?

PR #2313, opened 2020-06-01
Make outgoing traffic filtering optional

soffokl:
Closes: https://github.com/mysteriumnetwork/node/issues/1752

### DIFF
--- a/config/flags_network.go
+++ b/config/flags_network.go
@@ -69,6 +69,12 @@ var (
 		Usage: "Enables incoming traffic filtering",
 		Value: false,
 	}
+	// FlagOutgoingFirewall enables outgoing traffic filtering.
+	FlagOutgoingFirewall = cli.BoolFlag{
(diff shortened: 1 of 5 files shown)


config/flags_network.go
```diff
@@ -69,6 +69,12 @@ var (
 		Usage: "Enables incoming traffic filtering",
 		Value: false,
 	}
+	// FlagOutgoingFirewall enables outgoing traffic filtering.
+	FlagOutgoingFirewall = cli.BoolFlag{
+		Name:  "outgoing-firewall",
+		Usage: "Enables outgoing traffic filtering",
+		Value: false,
```
anjmao, line 76: Currently it's on by default, do we want to change this behaviour to false?

soffokl: Yeah, I want to keep it disabled for now, I suspect it leads to some connectivity fails.

Issue #2317, opened 2020-06-01
Add supervisor logging on Windows


anjmao:
Add logging to file on Windows supervisor.

PR #2323, opened 2020-06-02
Supervisor logs

anjmao:
Use zerolog in supervisor which allows to extend WireGuard default logger and pass zerolog global instance which is configured to log to file.

Closes https://github.com/mysteriumnetwork/node/issues/2317

### DIFF
--- a/cmd/supervisor/README.md
+++ b/cmd/supervisor/README.md
@@ -3,7 +3,7 @@
 Supervisor is a system background service that allows seamless installation/running of [Mysterium node](https://github.com/mysteriumnetwork/node) under elevated permissions.
 Clients (e.g. desktop applications) can ask the supervisor to RUN or KILL the node instance via OS dependent mechanism.
 
-Currently, only macOS is supported and it is using unix domain sockets for communication.
+Currently, only macOS and Windows are supported and it is using unix domain sockets or named pipes for communication.
 
 For usage see:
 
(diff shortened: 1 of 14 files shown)


cmd/supervisor/README.md
```diff
@@ -15,13 +15,16 @@ myst_supervisor -help
...
+| kill                              | macOS, Win   |      | ok     | ✅          | Kill myst process gracefully |
+| bye                               | macOS   |      | ok     | ✅            | Kill supervisor |
+| wg-up                             | macOS, Win   | -uid, -config    | ok     | ✅           | Setup WireGuard device with given configuration in JSON string format |
+| wg-down                           | macOS, Win   | -iface     | ok     | ✅           | Destroy WireGuard device |
+| wg-stats                          | macOS, Win   | -iface     | `{"bytes_send": 100, "bytes_received": 200, "last_handshake": "2020-06-02T13:42:55.786Z"}`     | ✅           | Get WireGuard device peer statistics |
+
+
+## Logs
+
+On Windows logs could be found at `C:\Windows\system32\myst_supervisor.log`
```
soffokl, line 28: Is it the right place for logs?

cmd/supervisor/supervisor.go
```diff
@@ -19,43 +19,66 @@ package main
...
 	}
 }
 
+func defaultLogPath() string {
+	if runtime.GOOS == "darwin" || runtime.GOOS == "linux" {
+		return "/var/log/myst_supervisor"
+	}
+
+	// On Windows default log file location will be under C:\Windows\system32\myst_supervisor.log
+	return "./myst_supervisor"
```
tadaskay, line 79: Never write anything to system dirs. It would be the best if we could use eventlog on windows, and not bother with file rotation/location at all. WDYT

Waldz: We already logging to file at `~/.mysterium.log`. What about having this one next to it `~/.supervisor.log`?

anjmao: Changed default log path to `C:\Windows\system32\config\systemprofile\AppData\Local\MystSupervisor\myst_supervisor.log`

supervisor/install/install_darwin.go
```diff
@@ -42,10 +43,6 @@ const plistTpl = `
 	</array>
 	<key>KeepAlive</key>
 	<true/>
-	<key>StandardOutPath</key>
-	<string>{{.LogPath}}</string>
```
Waldz, line 46: Do we really want to remove it?
I guess this logs more, like fatal fails, failed startups or crashes. Or just standard and error outputs?

anjmao: From the documentation it says that this is only for application logs. System logs are stored in different directions.

PR #2324, opened 2020-06-02
Skip services bootstrap for user mode

anjmao:
We do not need to bootstrap services when `-usermode` flag is specified as it will work only for consumer mode.

### DIFF
--- a/cmd/di_desktop.go
+++ b/cmd/di_desktop.go
@@ -54,7 +54,8 @@ import (
 
 // bootstrapServices loads all the components required for running services
 func (di *Dependencies) bootstrapServices(nodeOptions node.Options) error {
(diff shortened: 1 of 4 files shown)


cmd/di_desktop.go
```diff
@@ -55,6 +55,12 @@ import (
 // bootstrapServices loads all the components required for running services
 func (di *Dependencies) bootstrapServices(nodeOptions node.Options) error {
 	if nodeOptions.MobileConsumer {
+		log.Info().Msg("Skipping services bootstrap for mobile consumer")
```
anjmao, line 58: As talked with Tadas it would be better to have one flag ‘-consumer’

PR #2331, opened 2020-06-03
Wait supervisor uninstall fix

anjmao:
Some refactoring for supervisor logging. Added log level and checked all log statements to properly log by level.

When installing new supervisor version we first need to stop and uninstall previous service before installing new version, but previously we didn't waited for service deletion as this operation is async. Using OpenService method allows to handle this case.

### DIFF
--- a/supervisor/daemon/transport/transport_windows.go
+++ b/supervisor/daemon/transport/transport_windows.go
@@ -60,7 +60,7 @@ func (m *managerService) Execute(args []string, r <-chan svc.ChangeRequest, s ch
 		case svc.Continue:
 			s <- svc.Status{State: svc.Running, Accepts: cmdsAccepted}
 		default:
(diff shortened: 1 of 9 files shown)


supervisor/daemon/transport/transport_windows.go
```diff
@@ -60,7 +60,7 @@ func (m *managerService) Execute(args []string, r <-chan svc.ChangeRequest, s ch
 		case svc.Continue:
 			s <- svc.Status{State: svc.Running, Accepts: cmdsAccepted}
 		default:
-			log.Printf("unexpected control request #%d", c)
+			log.Error().Msgf("unexpected control request #%d", c)
```
soffokl, line 63:
```suggestion
			log.Error().Msgf("Unexpected control request #%d", c)
```

Issue #2337, opened 2020-06-04
could not resolve DNS IPs: system DNS is not available: error loading DNS config: open /etc/resolv.conf: The system cannot find the path specified.


Waldz:
After connection get initiated successfully with provider, connection instantly gets terminated

**To Reproduce**
1. Install Windows app
2. Start MysteriumVPN.exe
4. Connect to any provider

**Environment (please complete the following information):**
 - OS: Windows 10 (with Parallers)
 
**Additional context**
```
Get-Content C:\Users\waldz\.mysterium\logs\mysterium-node.log -Wait

2020-06-04T21:44:23.970 DBG source/services/wireguard/endpoint/remoteclient/client.go:61 > Tunnel interface created: myst
2020-06-04T21:44:23.970 INF source/services/wireguard/connection/connection.go:144 > Adding connection peer 194.5.235.128:47014
2020-06-04T21:44:23.970 INF source/services/wireguard/connection/connection.go:146 > Waiting for initial handshake
2020-06-04T21:44:24.180 DBG source/core/connection/dns_option.go:86  > Selecting DNS servers using strategy: system
2020-06-04T21:44:24.180 INF source/services/wireguard/connection/connection.go:200 > Stopping WireGuard connection
2020-06-04T21:44:24.621 INF source/firewall/outgoing_firewall_noop.go:50 > Rule for IP: 194.5.235.128 removed
2020-06-04T21:44:24.621 INF source/core/connection/manager.go:247    > Cancelling connection initiation:  error="could not resolve DNS IPs: system DNS is not available: error loading DNS config: open /etc/resolv.conf: The system cannot find the path specified."
```

PR #2342, opened 2020-06-05
Skip DNS setup on windows consumer

anjmao:
Fixes https://github.com/mysteriumnetwork/node/issues/2337

### DIFF
--- a/services/wireguard/connection/connection.go
+++ b/services/wireguard/connection/connection.go
@@ -21,6 +21,7 @@ import (
 	"context"
 	"encoding/json"
 	"net"
+	"runtime"
 	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


services/wireguard/connection/connection.go
```diff
@@ -148,13 +149,16 @@ func (c *Connection) Start(ctx context.Context, options connection.ConnectOption
...
 
-	dnsIPs, err := options.DNS.ResolveIPs(config.Consumer.DNSIPs)
-	if err != nil {
-		return errors.Wrap(err, "could not resolve DNS IPs")
-	}
-	config.Consumer.DNSIPs = dnsIPs[0]
-	if err := c.dnsManager.Set(c.opts.DNSScriptDir, conn.InterfaceName(), config.Consumer.DNSIPs); err != nil {
-		return errors.Wrap(err, "failed to configure DNS")
+	// TODO: Add DNS support for windows with https://github.com/mysteriumnetwork/node/issues/1840
+	if runtime.GOOS != "windows" {
```
Waldz, line 153: Runtime condition :/
Also dont get this change, we're enforcing `system` DNS in desktop, but `system` DNS configuration on interface will be skipped?

anjmao: This is just a quick temporary fix for windows to be able to connect. I'm solving https://github.com/mysteriumnetwork/node/issues/2337 issue which you registered. Our system dns for windows doesn't work, it fails.

anjmao:
Will add normal DNS support.

Issue #1840, opened 2020-03-04
Provider DNS support for Windows and WireGuard


anjmao:
Add support to use providers DNS when connecting to WireGuard service.

PR #2344, opened 2020-06-08
Add Windows WireGuard DNS support

anjmao:
This PR adds support for setting DNS on Windows. Since on Windows to set DNS admin permissions are needed setting DNS also moved to supervisor when running in usermode.

Closes https://github.com/mysteriumnetwork/node/issues/1840


services/wireguard/connection/dns/dns_windows.go
```diff
@@ -0,0 +1,41 @@
...
+		return fmt.Errorf("could not configure DNS, %s:%v", string(out), err)
+	}
+	return nil
+}
+
+func cleanDNS(cfg Config) error {
+	cmd := fmt.Sprintf("netsh interface ipv4 set dnsservers name=%s source=static address=none validate=no register=both", cfg.IfaceName)
+	out, err := exec.Command("powershell", "-Command", cmd).CombinedOutput()
+	if err != nil {
+		return fmt.Errorf("could not clean DNS, %s:%v", string(out), err)
```
tadaskay, line 38:
```suggestion
		return fmt.Errorf("could not clean DNS, %s: %w", string(out), err)
```

core/connection/dns_option.go
```diff
@@ -91,26 +90,19 @@ func (o *DNSOption) ResolveIPs(providerDNS string) ([]string, error) {
 	case DNSOptionProvider:
 		return selectProviderDNS(providerDNS)
 	case DNSOptionSystem:
-		return selectSystemDNS()
+		return []string{}, nil
```
tadaskay, line 93: Use nil slice?

services/wireguard/connection/dns/dns_windows.go
```diff
@@ -0,0 +1,41 @@
...
+	cmd := fmt.Sprintf("netsh interface ipv4 add dnsservers name=%s address=%s validate=no", cfg.IfaceName, cfg.DNS[0])
+	out, err := exec.Command("powershell", "-Command", cmd).CombinedOutput()
+	if err != nil {
+		return fmt.Errorf("could not configure DNS, %s:%v", string(out), err)
+	}
+	return nil
+}
+
+func cleanDNS(cfg Config) error {
+	cmd := fmt.Sprintf("netsh interface ipv4 set dnsservers name=%s source=static address=none validate=no register=both", cfg.IfaceName)
```
zolia, line 35: This basically means we are setting system DNS servers to none, meaning system is left without DNS servers after disconnect?

anjmao: It clears dns only for this interface.

zolia: Do we need that if interface is going down anyways?

services/wireguard/connection/dns/dns.go
```diff
@@ -0,0 +1,80 @@
...
+}
+
+func (dm *dnsManager) Set(cfg Config) error {
+	dm.mu.Lock()
+	defer dm.mu.Unlock()
+
+	dm.cfg = &cfg
+	// Empty DNS means that system DNS will be used, no need to set anything.
+	if len(cfg.DNS) == 0 {
+		return nil
```
Waldz, line 58: With this change darwin/linux won't call `update-resolv-conf` script anymore in case of `server ` mode. Are you sure it won't brake things?

zolia: Prob not needed on server mode.

PR #2346, opened 2020-06-08
Provide ways to query supervisor version

tadaskay:
Will be used by desktop to query and compare running version vs bundled version to force re-install.
Part of #2142

### DIFF
--- a/supervisor/daemon/daemon.go
+++ b/supervisor/daemon/daemon.go
@@ -28,6 +28,7 @@ import (
 	"strings"
 
 	"github.com/mysteriumnetwork/node/core/storage/boltdb"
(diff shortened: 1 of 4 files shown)


supervisor/daemon/daemon.go
```diff
@@ -71,6 +72,8 @@ func (d *Daemon) dialog(conn io.ReadWriter) {
 		cmd := strings.Split(string(line), " ")
 		op := strings.ToLower(cmd[0])
 		switch op {
+		case commandVersion:
```
anjmao, line 75: We do not pass any ldflags for supervisor binary when building. Is it working for your?

tadaskay: we do in `buildBinaryFor`

Issue #2343, opened 2020-06-08
After update to 0.34.0 OpenVPN service does not start


soffokl:
Only wireguard available. 
Example node `0x8408fbb2280d7af69b80276f3d4252f5580f6b06`

https://testnet-api.mysterium.network/v1/proposals?access_policy=%2A&include_failed=true&service_type=wireguard&node_key=0x8408fbb2280d7af69b80276f3d4252f5580f6b06
```
{"proposals":[{"format":"service-proposal/v1","id":1,"payment_method":{"bytes":5965232,"duration":333333333333,"price":{"amount":50000,"currency":"MYST"},"type":"BYTES_TRANSFERRED_WITH_TIME"},"payment_method_type":"BYTES_TRANSFERRED_WITH_TIME","provider_contacts":[{"definition":{"broker_addresses":["nats://testnet-broker.mysterium.network:4222"]},"type":"nats/p2p/v1"}],"provider_id":"0x8408fbb2280d7af69b80276f3d4252f5580f6b06","service_definition":{"location":{"asn":56478,"city":"England","continent":"EU","country":"GB","isp":"Hyperoptic Ltd","node_type":"residential"},"location_originate":{"asn":56478,"city":"England","continent":"EU","country":"GB","isp":"Hyperoptic Ltd","node_type":"residential"}},"service_type":"wireguard"}]}
```

https://testnet-api.mysterium.network/v1/proposals?access_policy=%2A&include_failed=true&service_type=openvpn&node_key=0x8408fbb2280d7af69b80276f3d4252f5580f6b06
```
{"proposals":[]}
```

PR #2349, opened 2020-06-08
Fix iptables path in sudoers on package update

zolia:
shoud fix: #2343

### DIFF
--- a/bin/package/installation/post-install.sh
+++ b/bin/package/installation/post-install.sh
@@ -120,7 +120,7 @@ fi
 
 # TODO: remove me when all nodes updates
 printf "Ensuring that sudoers config is consistent...\n" \
-    && sed -i -e 's/\/sbin\/iptables/\/usr\/sbin\/iptables/g' /etc/sudoers.d/mysterium-node
+    && sed -i -e 's| /sbin/iptables/| /usr/sbin/iptables|g' /etc/sudoers.d/mysterium-node
 
 # Cleanup old log files (before log file rolling has been fixed)
 printf "Cleaning legacy log files...\n" \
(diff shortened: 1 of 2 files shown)


bin/package/installation/post-install.sh
```diff
@@ -110,14 +110,18 @@ ensure_paths
...
     sed -i -e 's/SERVICE_OPTS="openvpn"/SERVICE_OPTS="openvpn,wireguard"/g' /etc/default/mysterium-node
     sed -i -e 's/--tequilapi.address=0.0.0.0/--tequilapi.address=127.0.0.1/g' /etc/default/mysterium-node
     # Append script-dir for existing installations only if missing
     sed -i -e '/^\(SCRIPT_DIR=\).*/{s//\1\"--script-dir=\/etc\/mysterium-node\"/;:a;n;ba;q}' \
            -e '$aSCRIPT_DIR="--script-dir=/etc/mysterium-node"' /etc/default/mysterium-node
 fi
 
+# TODO: remove me when all nodes updates
+printf "Ensuring that sudoers config is consistent...\n" \
+    && sed -i -e 's/\/sbin\/iptables/\/usr\/sbin\/iptables/g' /etc/sudoers.d/mysterium-node
```
soffokl, line 123: It's possible to use another delimiter for `sed` when you have a lot of `/` in the changes.
And It's better to include space in the expression to avoid recursion updates like `/usr/usr/sbin/iptables`
```suggestion
    && sed -i -e 's| /sbin/iptables| /usr/sbin/iptables|g' /etc/sudoers.d/mysterium-node
```

soffokl:
@zolia https://github.com/mysteriumnetwork/node/blob/master/debian/postinst should be updated too?

bin/package/installation/post-install.sh
zolia: I can match beggining of a line too.

zolia:
> @zolia https://github.com/mysteriumnetwork/node/blob/master/debian/postinst should be updated too?

Yes, added.

PR #2386, opened 2020-06-15
Tests skips network stats

Waldz:
Fixes interactive prompt then running `mage test` locally:
Screen Shot 2020-06-15 at 16 19 36

### DIFF
--- a/pb/payment.pb.go
+++ b/pb/payment.pb.go
@@ -7,11 +7,12 @@
 package pb
 
 import (
+	reflect "reflect"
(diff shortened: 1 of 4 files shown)


pb/payment.pb.go
```diff
@@ -7,11 +7,12 @@
 package pb
 
 import (
+	reflect "reflect"
```
soffokl, line 10: > // Code generated by protoc-gen-go. DO NOT EDIT.

PR #2390, opened 2020-06-18
Support alternative iptables path

soffokl:
Some distributives still have a different path to the `iptables` utility. This prevents normal service work.
This PR allows an alternative path to the `iptables` utility. 

```
root@monitoring:~# cat  /etc/issue
Ubuntu 18.04.4 LTS \n \l
root@monitoring:~# ls -l /usr/sbin/iptables
ls: cannot access '/usr/sbin/iptables': No such file or directory
root@monitoring:~# which iptables
/sbin/iptables
root@monitoring:~#
```

image

### DIFF
--- a/debian/postinst
+++ b/debian/postinst
@@ -39,6 +39,24 @@ function install_chkconfig {
         && chkconfig --add mysterium-node
 }
 
+function ensure_paths {
+    iptables_path=`which iptables`
+
+    # validate utility against valid system paths
+    basepath=${iptables_path%/*}
+    echo "iptables basepath detected: ${basepath}"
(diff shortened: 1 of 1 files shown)


zolia:
These distributions does have different paths, but package post install should ensure that there is a link at least: 
```bash
    if ! [[ -x ${iptables_required_path} ]]; then
        ln -s ${iptables_path} ${iptables_required_path}
    fi
```

If that does not work maybe we should investigate why?

soffokl:
@zolia it was missing in the right postinstall script. I have updated PR to fix it. Please take a look.

PR #2396, opened 2020-06-18
First invoice is required now

soffokl:


session/manager.go
```diff
@@ -192,8 +192,7 @@ func (manager *Manager) Start(session *Session, consumerID identity.Identity, co
 	}()
 
 	log.Info().Msg("Waiting for a first invoice to be paid")
-	// TODO 3 seconds timeout should be increased when all consumers will be able to pay for the first invoice before session start.
-	if err := engine.WaitFirstInvoice(3 * time.Second); err != nil {
+	if err := engine.WaitFirstInvoice(30 * time.Second); err != nil {
 		return fmt.Errorf("first invoice was not paid: %w", err)
```
Waldz, line 196: Session should be clearer here

session/pingpong/invoice_tracker.go
```diff
@@ -359,9 +359,7 @@ func (it *InvoiceTracker) WaitFirstInvoice(wait time.Duration) error {
 				return nil
 			}
 		case <-timeout:
-			// TODO uncomment this error return when all consumers will be able to pay for the first invoice before session start.
-			// return fmt.Errorf("failed waiting for first invoice")
-			return nil
+			return fmt.Errorf("failed waiting for first invoice")
```
Waldz, line 362: In my provider logs I see clear 1st invoice failure, but it never triggers `Payment engine error` flow.
```
:455  > Marking invoice as not sent error="timeout waiting for reply to \"p2p-payment-invoice\": p2p send timeout
```

Checked and reason is that this error is being swallowed here:
Screen Shot 2020-06-18 at 15 03 51

As we don't do retries in `sendFirstInvoice()` this strange swallowing condition should only where `sendInvoice()` is used within retry loop.

session/manager.go
soffokl: Added session destroy.

session/pingpong/invoice_tracker.go
soffokl: Added retries for sending the first invoice.

PR #2405, opened 2020-06-25
Cleanup IPResolver cache on disconnect

zolia:
After changing network on mobile connection would fail if previous connect would not complete. This fixes that.

### DIFF
--- a/trace/trace.go
+++ b/trace/trace.go
@@ -98,7 +98,11 @@ func (t *Tracer) Finish() string {
 
 	var strs []string
 	for _, s := range stages {
-		strs = append(strs, fmt.Sprintf("%q took %s", s.key, s.end.Sub(s.start).String()))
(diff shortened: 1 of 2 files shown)


trace/trace.go
```diff
@@ -98,7 +98,11 @@ func (t *Tracer) Finish() string {
 
 	var strs []string
 	for _, s := range stages {
-		strs = append(strs, fmt.Sprintf("%q took %s", s.key, s.end.Sub(s.start).String()))
+		if s.end.After(time.Time{}) {
```
Waldz, line 101: If stage is unfinished, end time should be `time.Now()`

zolia: No, this fixes exactly that. time.Time{} produces not now(), but Time(0), meaning end is not initialised.

tadaskay: loose ≠ lose

PR #2407, opened 2020-06-25
Restrict supervisor socket to the installing user

tadaskay:
This changes `/var/run/myst.sock` permissions 0777 -> 0700. When installing the supervisor, users must provide `-uid` flag. Uid will be stored in the supervisor config and later used to chown the supervisor socket so that the particular user may talk to it without elevated permissions.


cmd/supervisor/supervisor.go
```diff
@@ -50,11 +59,17 @@ func main() {
...
+		ensureInstallFlags()
 		path, err := thisPath()
 		if err != nil {
 			log.Fatal().Err(err).Msg("Failed to determine supervisor's path")
 		}
-
+		cfg := config.Config{
+			Uid: *svflags.FlagUid,
+		}
+		if err := cfg.Write(); err != nil {
```
Waldz, line 70: What about removing config during uninstall?

tadaskay: As a common practice, configs are not uninstalled.

Issue #2370, opened 2020-06-09
Save sessions stats into node bold db


chompomonim:
- [x] Sessions history for last X sessions
- [x] Store all transferred data (last month) counter in bolt
- [ ] Store amount of all sessions (last month) counter
- [x] Store sessions time (last month) counter
- [ ] Store unique users amount (last month) counter 
- [x] Expose all this info via tequila API

PR #2408, opened 2020-06-25
Capture provided sessions to persistent storage

Waldz:
Closes: #2370

Now each provided session are stored to Bolt bucket `session-history`. History contains all transactions done by my identities - consumed sessions + provided sessions.


tadaskay:
Huge feature branches are bad. You could easily achieve the same in smaller iterations.

PR #2409, opened 2020-06-26
Expose data for mobile filters

anjmao:
Added some missing data for mobile including prices, node type and spent tokens.

### DIFF
--- a/mobile/mysterium/proposals_manager.go
+++ b/mobile/mysterium/proposals_manager.go
@@ -44,7 +44,13 @@ const (
 
 // GetProposalsRequest represents proposals request.
 type GetProposalsRequest struct {
-	Refresh bool
+	ServiceType         string
+	Refresh             bool
+	IncludeFailed       bool
+	UpperTimePriceBound int64
+	LowerTimePriceBound int64
+	UpperGBPriceBound   int64
+	LowerGBPriceBound   int64
 }
 
 // GetProposalRequest represents proposal request.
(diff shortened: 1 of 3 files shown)


mobile/mysterium/proposals_manager.go
```diff
@@ -109,7 +131,21 @@ func (m *proposalsManager) getProposals(req *GetProposalsRequest) ([]byte, error
 	}
 
 	// Get proposals from remote discovery api and store in cache.
-	apiProposals, err := m.getFromRepository()
+	upperTimePriceBound := uint64(req.UpperTimePriceBound)
```
zolia, line 134: Why not to use uint64 for GetProposalsRequest fields ? No conversion would be needed?

mobile/mysterium/proposals_manager.go
```diff
@@ -118,26 +154,12 @@ func (m *proposalsManager) getProposals(req *GetProposalsRequest) ([]byte, error
...
 func (m *proposalsManager) getFromCache() []market.ServiceProposal {
 	return m.cache
 }
 
-func (m *proposalsManager) getFromRepository() ([]market.ServiceProposal, error) {
-	allProposals, err := m.repository.Proposals(m.filter)
+func (m *proposalsManager) getFromRepository(filter *proposal.Filter) ([]market.ServiceProposal, error) {
+	allProposals, err := m.repository.Proposals(filter)
 	if err != nil {
 		return nil, err
```
zolia, line 164: wrap an error

PR #2417, opened 2020-06-29
Print stack trace on feedback error

zolia:
No actual error was being printed when feedback fails. Fixed.


tadaskay:
What is the motivation for this fix? Feedback is called from the API endpoint so the stack trace will always remain the same.

Issue #2411, opened 2020-06-28
Endless connecting during Windows app start


chompomonim:
While running Mysterium desktop app for Windows I'm seeing nothing else but window which is showing loading progress bar.

**To Reproduce**

1. Download Windows app and install it.
2. Launch the app
3. App is hanging by just showing "connecting" message

**Expected behavior**
Launch the app, see a list of nodes to choose from. Chose node and get VPN tunnel. 

**Environment:**
 - OS: 'Window 10 home 2004 may 2020 update' and 'Windows 10 corporative'
 - Desktop app version: 2.1.1

**Screenshot**
![screenshoot](https://user-images.githubusercontent.com/153896/85946248-d1193f80-b94b-11ea-8fc2-c2202f84bb7a.jpg)

**Additional context**

main.log
```
[2020-06-25 19:02:54.129] [info] Checking for update
[2020-06-25 19:02:54.981] [info] Connecting to the supervisor...
[2020-06-25 19:02:55.917] [info] Connecting to the supervisor...
[2020-06-25 19:02:56.418] [info] Connecting to the supervisor...
[2020-06-25 19:02:56.920] [info] Connecting to the supervisor...
[2020-06-25 19:02:57.422] [info] Connecting to the supervisor...
[2020-06-25 19:02:57.434] [info] Update for version 2.1.1 is not available (latest version: 2.1.1, downgrade is disallowed).
[2020-06-25 19:02:57.923] [info] Connecting to the supervisor...
[2020-06-25 19:02:58.425] [info] Connecting to the supervisor...
[2020-06-25 19:02:58.926] [info] Connecting to the supervisor...
[2020-06-25 19:02:59.428] [info] Connecting to the supervisor...
[2020-06-25 19:02:59.821] [info] [sudo-exec]  2020-06-25T19:02:58.000 INF Installing supervisor with options: install.Options{SupervisorPath:"C:\\Program Files\\MysteriumVPN\\resources\\static\\bin\\myst_supervisor.exe"}
2020-06-25T19:02:58.000 INF Checking previous installation
2020-06-25T19:02:58.000 INF Uninstalled previous service
2020-06-25T19:02:59.000 INF Supervisor installed

[2020-06-25 19:02:59.929] [info] Connecting to the supervisor...
[2020-06-25 19:03:00.430] [info] Connecting to the supervisor...
[2020-06-25 19:03:00.933] [info] Connecting to the supervisor...
[...]

PR #2427, opened 2020-07-01
Replace 'Users' group lookup with a well-known SID

tadaskay:
```
// https://support.microsoft.com/en-us/help/243330/well-known-security-identifiers-in-windows-operating-systems
// Looking up by name appears to be unreliable: the name could not be found in localized installations.
// Using a well-known security identifier here instead.
```

Fixes: #2411

### DIFF
--- a/supervisor/daemon/transport/transport_windows.go
+++ b/supervisor/daemon/transport/transport_windows.go
@@ -19,7 +19,6 @@ package transport
 
 import (
 	"fmt"
-	"strings"
 
 	"github.com/rs/zerolog/log"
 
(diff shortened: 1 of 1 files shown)


supervisor/daemon/transport/transport_windows.go
```diff
@@ -72,17 +71,12 @@ func (m *managerService) Execute(args []string, r <-chan svc.ChangeRequest, s ch
...
 
 func listenPipe(handle handlerFunc) error {
-	// TODO: Check these permissions, it would be much more secure to pass user id
-	// during supervisor installation as adding whole Users group is not secure.
-	socketGroup := "Users"
+	// https://support.microsoft.com/en-us/help/243330/well-known-security-identifiers-in-windows-operating-systems
+	// Looking up by name appears to be unreliable: the name could not be found in localized installations.
+	// Using a well-known security identifier here instead.
+	usersSID := "S-1-5-32-545"
 	sddl := "D:P(A;;GA;;;BA)(A;;GA;;;SY)"
```
vkuznecovas, line 78: <3 windows

PR #2430, opened 2020-07-01
Use user independent dir for app data

tadaskay:
During install, app dir is set to current user.
While running service, it is set to systemprofile.
So when running as a service, supervisor couldn't access the config.

Fixes: #2411


Waldz:
Please describe all directories in https://github.com/mysteriumnetwork/mysterium-vpn-desktop#app

PR #2438, opened 2020-07-03
Enable services by default

zolia:
Fixes issue in current master, when no services are being started.

### DIFF
--- a/cmd/commands/service/command.go
+++ b/cmd/commands/service/command.go
@@ -23,6 +23,10 @@ import (
 	"strings"
 	"time"
 
+	"github.com/pkg/errors"
+	"github.com/rs/zerolog/log"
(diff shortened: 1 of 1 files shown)


cmd/commands/service/command.go
```diff
@@ -111,6 +111,11 @@ func (sc *serviceCommand) Run(ctx *cli.Context) (err error) {
 		serviceTypes = strings.Split(ctx.Args().Get(0), ",")
 	}
 
+	if len(serviceTypes) == 0 {
```
Waldz, line 114: This leaves deadcode on line 109.
Just revert PR #2424

zolia: It does not leave it dead if its not a 0

PR #2440, opened 2020-07-07
Revert "Create mysterium node config before starting service"

zolia:
Reverts mysteriumnetwork/node#2437

I suspect this one broke our RPI packaging. Do not have a way to test this job since it does work only on releases. Need to make test release.


zolia:
should be fixed with #2444, no revert needed.

Issue #2446, opened 2020-07-08
Failing to connect to services after turning whitelisting off


zolia:
Node version: 0.36-rc1

```
2020-07-08T13:30:18.481 INF source/nat/event/tracker.go:53           > Got NAT event event={"stage":"port_mapping","successful":true}
2020-07-08T13:30:18.489 DBG source/p2p/listener.go:244               > Sending reply with public key 7f05ea462c47a83cc1424d1e0180c0a114d5ddea9bf109f1bf432a67a707b629 and encrypted config to consumer
2020-07-08T13:30:18.783 DBG source/core/ip/cached_resolver.go:75     > Found cached public IP
2020-07-08T13:30:18.793 DBG source/p2p/listener.go:244               > Sending reply with public key 2be5b70b339d4b09d9252d6f730e0d8331f2915ddd6d3dd2ad03d7488a506415 and encrypted config to consumer
2020-07-08T13:30:19.511 ERR source/p2p/listener.go:124               > Could not handle exchange ack error="could not decrypt peer conn config: could not decrypt config to proto bytes: could not decrypt message"
2020-07-08T13:30:19.516 ERR source/p2p/listener.go:124               > Could not handle exchange ack error="could not decrypt peer conn config: could not decrypt config to proto bytes: could not decrypt message"
```

After turning off whitelisting and restarting services client can not connect to the node with error above.
Restarting RPI completely fixes an issue.

PR #2460, opened 2020-07-14
Make whitelisting work again

soffokl:
Closes: #2446


core/service/subscription.go
```diff
@@ -35,23 +37,23 @@ import (
 	"github.com/rs/zerolog/log"
 )
 
-func subscribeSessionCreate(mng *session.Manager, ch p2p.Channel, service Service, eventPublisher eventbus.Publisher) {
+func subscribeSessionCreate(mng *session.Manager, ch p2p.Channel, service Service, eventPublisher eventbus.Publisher, proposal market.ServiceProposal, policyRules *policy.Repository) {
```
Waldz, line 40: You dont need to pass new arguments, running service instance `service Service` contains those

soffokl: `service Service` is an interface that allows us to `Serve` `Stop` and `ProvideConfig`. These arguments required outside the service instance.

p2p/message.go
```diff
@@ -132,6 +135,7 @@ func (m *transportMsg) writeTo(conn *textproto.Writer) error {
 	header.WriteString(fmt.Sprintf("%s:%d\r\n", headerFieldRequestID, m.id))
 	header.WriteString(fmt.Sprintf("%s:%s\r\n", headerFieldTopic, m.topic))
 	header.WriteString(fmt.Sprintf("%s:%d\r\n", headerStatusCode, m.statusCode))
+	header.WriteString(fmt.Sprintf("%s:%s\r\n", headerMsg, m.msg))
```
zolia, line 138: It might be considered a vulnerability of exposing internal error details to a peer. But good for debugging :) Maybe we should make it configurable and enabled by default? :)

soffokl: Let's make it in as a separate issue if you really think that it could be a vulnerability.

Issue #2372, opened 2020-06-09
Expose stats grouped by day


chompomonim:
- [ ] Earning
- [ ] Sessions amount
- [ ] Data transfered

PR #2465, opened 2020-07-15
Session stats by day API

Waldz:
Closes: #2372

Enrich session history API with aggregated statistics grouped by day

PR #2469, opened 2020-07-17
Session stats by day API

Waldz:
Closes: #2372

Enrich session history API with aggregated statistics grouped by day


Waldz:
Repeating #2465 to `master`

PR #2477, opened 2020-07-20
MMN Claim

ignasbernotas:
This PR introduces these changes:

Removed mymysterium.enabled flag. With this PR, node will not register to MMN unless an API key is in the config.

Setting [mmn] api-key=xxx in config.toml through CLI and TequilAPI

- mmn <api-key> CLI command

TequilAPI endpoints
- [GET] /mmn/report // fetches bounty report from MMN
- [GET] /mmn/api-key
- [POST] /mmn/api-key
- [DELETE] /mmn/api-key

Because TequilAPI client's `parseResponseError` reads the response, the response body was no longer accessible from further down the code i.e. in CLI commands. In my case I had a 422 status code response with validation errors in JSON so parseResponseError parsed it, left the response.Body empty and just logged the error. I had no way to parse the validation errors in the CLI command. So I made a fix using a `io.TeeReader`, it reads the buffer, then after read writes it again, so it becomes available once more.


mmn/mmn.go
```diff
@@ -18,49 +18,73 @@
...
-		connection.AppTopicConnectionSession,
-		m.handleClient,
+	m.node.LocalIP = outboundIp
+
+	return nil
+}
+
+// SubscribeToIdentityUnlockRegisterToMMN subscribes to identity unlock, registers identity in MMN if the API key is set
+func (m *MMN) SubscribeToIdentityUnlockRegisterToMMN(eventBus eventbus.EventBus, isRegistrationEnabled func() bool) error {
+	err := eventBus.SubscribeAsync(
```
vkuznecovas, line 72: just return here

tequilapi/client/client.go
```diff
@@ -514,3 +516,35 @@ func (client *Client) Beneficiary(address string) (res contract.IdentityBenefici
 	err = parseResponseJSON(response, &res)
 	return res, err
 }
+
+// SetMMNApiKey sets MMN's API key in config and registers node to MMN
+func (client *Client) SetMMNApiKey(data contract.MMNApiKeyRequest) error {
+	response, err := client.http.Post("mmn/api-key", data)
```
vkuznecovas, line 522: you're not checking the error here.

ignasbernotas: @vkuznecovas this is because if the request fails with a non-200 (e.g. 422) status code, the error generated is super generic. I need to parse the response of a 422 or 403 status code request to get the validation error messages from MMN. I understand the confusion, but I don't see another way of doing it. Advice is welcome.

tequilapi/client/client.go
```diff
@@ -514,3 +516,35 @@ func (client *Client) Beneficiary(address string) (res contract.IdentityBenefici
...
+
+	if response == nil {
```
ignasbernotas, line 524: @vkuznecovas I do a different check it here. It's dumb, but it works :D

Issue #2543, opened 2020-07-28
[Mac] Could not connect to any provider in 2.5.1


etherunit:
**Describe the bug**
Could not connect to any selected provider. Each connection attempt gives an error.

**To Reproduce**
1. Select any provider
2. Click Connect

**Expected behavior**
Consumer can connect to the selected node.

**Screenshots**
<img width="687" alt="Screenshot 2020-07-28 at 16 55 43" src="https://user-images.githubusercontent.com/60128746/88675504-add0e400-d0f3-11ea-8f2c-5dbedc8b6933.png">
<img width="696" alt="Screenshot 2020-07-28 at 16 56 07" src="https://user-images.githubusercontent.com/60128746/88675532-b1fd0180-d0f3-11ea-97ca-a3b4d7567932.png">

**Environment (please complete the following information):**
 - Node version: Any
 - OS: mac version 10.15.5 (19F101)

**Additional context**

supervisor.log

```
2020-07-27T18:00:58.000 ??? DEBUG: (utun6) 2020/07/27 18:00:58 Routine: encryption worker - stopped
2020-07-27T18:00:58.000 ??? DEBUG: (utun6) 2020/07/27 18:00:58 Routine: handshake worker - stopped
2020-07-27T18:00:58.000 ??? DEBUG: (utun6) 2020/07/27 18:00:58 Routine: handshake worker - stopped
2020-07-27T18:00:58.000 ??? DEBUG: (utun6) 2020/07/27 18:00:58 Routine: decryption worker - stopped
2020-07-27T18:00:58.000 ??? DEBUG: (utun6) 2020/07/27 18:00:58 Routine: decryption worker - stopped
2020-07-27T18:00:58.000 ??? DEBUG: (utun6) 2020/07/27 18:00:58 peer(h3z8…jBhM) - Routine: sequential receiver - stopped
2020-07-27T18:00:58.000 ??? DEBUG: (utun6) 2020/07/27 18:00:58 peer(h3z8…jBhM) - Routine: nonce worker - stopped
2020-07-27T18:00:58.000 ??? DEBUG: (utun6) 2020/07/27 18:00:58 peer(h3z8…jBhM) - Routine: sequential sender - stopped
2020-07-27T18:00:58.000 ??? INFO: (utun6) 2020/07/27 18:00:58 Interface closed
2020-07-27T18:00:58.000 ERR Could not clean DNS error="DNS is already cleaned"
[...]

PR #2550, opened 2020-08-03
Balance tracker now falls back to transactor on unlock if registratio…

vkuznecovas:
…n in progress

Issue is caused with registrations that are stuck in progress with the transactor. This handles this situation better on node side, a fix for transactor is coming soon.

updates #2543

### DIFF
--- a/session/pingpong/consumer_balance_tracker.go
+++ b/session/pingpong/consumer_balance_tracker.go
@@ -43,6 +43,7 @@ type ConsumerBalanceTracker struct {
 	balancesLock sync.Mutex
 	balances     map[identity.Identity]ConsumerBalance
 
+	registry                             registrationStatusProvider
 	accountantAddress                    common.Address
 	mystSCAddress                        common.Address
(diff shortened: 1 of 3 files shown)


session/pingpong/consumer_balance_tracker.go
```diff
@@ -142,7 +144,18 @@ func (cbt *ConsumerBalanceTracker) handleUnlockEvent(id string) {
 	if err != nil {
 		log.Error().Err(err).Msg("Could not recover Grand Total Promised")
 	}
-	cbt.ForceBalanceUpdate(identity)
+
+	status, err := cbt.registry.GetRegistrationStatus(identity)
+	if err != nil {
+		log.Error().Err(err).Msg("Could not recover get registration status")
```
Waldz, line 150: Would be nice to somehow reuse same `handleRegistrationEvent(event registry.AppEventIdentityRegistration)` after force syncing this event.

vkuznecovas: there might not be such events in case a node is restarted.

PR #2551, opened 2020-08-03
Reconnect on darwin after sleep

zolia:
Partially covers #2447, does reconnect on MacOS after wake-up from sleep. It does full reconnect using connect parameters from last session. Checks whether reconnect is required by checking if p2p channel is alive.


core/connection/manager.go
```diff
@@ -647,6 +660,9 @@ func (m *connectionManager) disconnect() {
 	m.statusNotConnected()
 
 	m.cleanAfterDisconnect()
+	//m.cleanupFinishedOnce.Do(func() {
```
vkuznecovas, line 663: remove the dead code

zolia:
Will refactor CheckChannel() with context timeout to make code more elegant.

PR #2564, opened 2020-08-18
Reconnect after wake-up on win

zolia:
Reconnect after receiving wake-up log event on Windows. 
Tested with windows CLI.

Due to super slow connects on windows reconnect takes a long time sometimes hitting 30sec dial timeout, thus increased to 60sec.
Looks like windows is super slow on route cleanup. Which adds up to reconnect timeout. Maybe we should investigate a bit further weather we can improve that.

Plain connect (without previous disconnect) on win depending on a node takes ~18sec which should be not acceptable. This is most likely connects to RPIs.

### DIFF
--- a/sleep/sleep_windows.go
+++ b/sleep/sleep_windows.go
@@ -1,17 +1,34 @@
+/*
+ * Copyright (C) 2020 The "MysteriumNetwork/node" Authors.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
 package sleep
 
-import "C"
 import (
-	"github.com/mysteriumnetwork/gowinlog"
+	winlog "github.com/mysteriumnetwork/gowinlog"
 	"github.com/rs/zerolog/log"
 )
 
-func (n Notifier) Start() {
+//Start starts event log notifier
+func (n *Notifier) Start() {
 	log.Debug().Msg("Register for sleep log events")
 
 	watcher, err := winlog.NewWinLogWatcher()
 	if err != nil {
-		log.Error().Msgf("couldn't create log watcher: %v\n", err)
+		log.Error().Msgf("Couldn't create log watcher: %v\n", err)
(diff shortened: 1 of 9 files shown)


sleep/sleep_windows.go
```diff
@@ -0,0 +1,34 @@
...
+	"github.com/mysteriumnetwork/gowinlog"
+	"github.com/rs/zerolog/log"
+)
+
+func (n Notifier) Start() {
+	log.Debug().Msg("Register for sleep log events")
+
+	watcher, err := winlog.NewWinLogWatcher()
+	if err != nil {
+		log.Error().Msgf("couldn't create log watcher: %v\n", err)
```
soffokl, line 14:
```suggestion
		log.Error().Msgf("Couldn't create log watcher: %v\n", err)
```

sleep/sleep_windows.go
```diff
@@ -0,0 +1,34 @@
...
+	watcher.SubscribeFromNow("System", "*[System[Provider[@Name='Microsoft-Windows-Power-Troubleshooter'] and EventID=1]]")
+	for {
+		select {
+		case <-watcher.Event():
+			n.eventbus.Publish(AppTopicSleepNotification, EventWakeup)
+		case err := <-watcher.Error():
+			log.Error().Msgf("Log watcher error: %v\n", err)
+		}
+	}
+	<-n.stop
```
soffokl, line 27: This `n.stop` should be in the loop and in the `select` block. Otherwise, this loop will never finish.

sleep/sleep_windows.go
```diff
@@ -0,0 +1,34 @@
+package sleep
```
soffokl, line 1: `Copyright` comment is missing.

sleep/sleep_windows.go
```diff
@@ -0,0 +1,34 @@
...
+
+import "C"
```
soffokl, line 3: Do we need this import?

soffokl (approved):

tadaskay (approved):

sleep/sleep_windows.go
```diff
@@ -0,0 +1,55 @@
...
+func (n *Notifier) Start() {
+	log.Debug().Msg("Register for sleep log events")
+
+	watcher, err := winlog.NewWinLogWatcher()
+	if err != nil {
+		log.Error().Msgf("Couldn't create log watcher: %v\n", err)
+		return
+	}
+
+	watcher.SubscribeFromNow("System", "*[System[Provider[@Name='Microsoft-Windows-Power-Troubleshooter'] and EventID=1]]")
```
Waldz, line 35: What is meaning of EventID constant. Some comments or links to documentation would be valuable.

Waldz (approved):

sleep/sleep_windows.go
zolia: Were not able to find proper official documentation. Just conversations on forums. Mostly trial-error approach. 
Event ID == 1 along with 'Power-Troubleshooter' source designates logged event on sleep wake-up.

Issue #2524, opened 2020-07-24
List of last settlements tequilla endpoint 


chompomonim:
Settlements we should get from blockchain or from node cache.

It should support paging, so UI could ask for a limited amount of data.

PR #2565, opened 2020-08-18
Store settlement history and expose it to Tequilapi 

Waldz:
Closes: #2524

Starting to store settlements of your identities with more fields: `providerID`, `accountantID`, `channelAddress`, `beneficiary`.
Also exposing them in paginated Tequilapi endpoint `/transactor/settle/history`.

### DIFF
--- a/session/pingpong/settlement_history_storage.go
+++ b/session/pingpong/settlement_history_storage.go
@@ -61,81 +61,40 @@ func (shs *SettlementHistoryStorage) Store(she SettlementHistoryEntry) error {
 	return shs.bolt.DB().From(settlementHistoryBucket).Save(&she)
 }
 
-// Query executes given query.
-func (shs *SettlementHistoryStorage) Query(query *SettlementHistoryQuery) (err error) {
-	return query.run(shs.bolt.DB().From(settlementHistoryBucket))
+// SettlementHistoryFilter defines all flags for filtering in settlement history storage.
+type SettlementHistoryFilter struct {
+	TimeFrom     *time.Time
+	TimeTo       *time.Time
+	ProviderID   *identity.Identity
+	AccountantID *common.Address
 }
 
-// NewSettlementHistoryQuery creates instance of query.
-func NewSettlementHistoryQuery() *SettlementHistoryQuery {
-	return &SettlementHistoryQuery{}
-}
-
-// SettlementHistoryQuery defines all flags for filtering in settlement history storage.
-type SettlementHistoryQuery struct {
-	Entries []SettlementHistoryEntry
-
-	filterFrom       *time.Time
-	filterTo         *time.Time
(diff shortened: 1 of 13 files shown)


soffokl (approved):

tequilapi/endpoints/transactor.go
```diff
@@ -367,12 +370,38 @@ func (te *transactorEndpoint) SettlementHistory(resp http.ResponseWriter, req *h
...
+	}
+
 	if err := te.settlementHistoryProvider.Query(query.FetchEntries()); err != nil {
 		utils.SendError(resp, err, http.StatusInternalServerError)
 		return
 	}
 
-	response := contract.NewSettlementListResponse(query.Entries)
+	var settlements []pingpong.SettlementHistoryEntry
+	p := paginator.New(adapter.NewSliceAdapter(query.Entries), pageSize)
```
tadaskay, line 397: You're not paging in DB so this has no benefit besides sending a partial result back.
You can page in storm DB, e.g.:
```
err = db.Find("Group", "staff", &users, storm.Limit(10), storm.Skip(10))
```

session/pingpong/settlement_history_storage.go
```diff
@@ -60,17 +61,81 @@ func (shs *SettlementHistoryStorage) Store(she SettlementHistoryEntry) error {
...
+}
+
+// NewSettlementHistoryQuery creates instance of query.
+func NewSettlementHistoryQuery() *SettlementHistoryQuery {
+	return &SettlementHistoryQuery{}
+}
+
+// SettlementHistoryQuery defines all flags for filtering in settlement history storage.
+type SettlementHistoryQuery struct {
+	Entries []SettlementHistoryEntry
```
tadaskay, line 76: What's the benefit of combining a query with the result?

You could simply have query as follows without repeating 'filter':
```go
type SettlementHistoryQuery struct {
	From         *time.Time
	To           *time.Time
	ProviderID   *identity.Identity
	AccountantID *common.Address
}
```

And return `Entries []SettlementHistoryEntry` as a func return value.

tadaskay (changes requested):

session/pingpong/settlement_history_storage.go
Waldz: Simplified to 
```
type settlementHistoryProvider interface {
	List(filter pingpong.SettlementHistoryFilter) ([]pingpong.SettlementHistoryEntry, error)
}
```

tequilapi/endpoints/transactor.go
Waldz: It's 2nd place there we depend on this paginator, didn't write bolt adapter for it because:
- more simple to use it like this
- bolt is traversing&reducing all records in your suggested approach anyways

tadaskay (approved):

PR #2577, opened 2020-08-24
Reduce pinger intervals

zolia:
Should slice off around 1 sec in total time.  - ~300ms when using Publish method for broker.
@soffokl can you test it with long distance connections from your place? Does punching still works reliably?

UPDATE: will not use broker Publish method as we DO require ack from peer before starting to ping Provider. Provider has to start pinging first to avoid DDOS blockage on its side.


vkuznecovas (approved):

Waldz (approved):

soffokl (approved):
I see no problem for connection.

PR #2600, opened 2020-09-07
Unsubscribe sleep event on disconnect

zolia:


soffokl (approved):

zolia:
Will rework to use eventBus only on node startup / shutdown

core/connection/manager.go
```diff
@@ -653,6 +653,8 @@ func (m *connectionManager) Disconnect() error {
 		return ErrNoConnection
 	}
 
+	m.eventBus.Unsubscribe(sleep.AppTopicSleepNotification, m.handleSleepEvent)
```
Waldz, line 656: Avoid subscribing on runtime. To avoid problems like this, or multiple subscriptions, races, memory leaks.

Waldz (changes requested):

Waldz (approved):

PR #2613, opened 2020-09-10
Registration with bounty

vkuznecovas:
No more free registration. Bounties available for free registrations from now on.


vkuznecovas:
Closes mysteriumnetwork/transactor#79

e2e/helpers.go
```diff
@@ -0,0 +1,66 @@
...
+	}, nil
+}
+
+type registrationBounty struct {
+	ID       primitive.ObjectID `bson:"_id" json:"id"`
+	Identity string             `bson:"identity" json:"identity"`
+}
+
+// InsertRegistrationBounty inserts a new entry into the bounty collection
+func (m *Mongo) InsertRegistrationBounty(identity common.Address) error {
```
Waldz, line 57: Why node knows about transactor's internals.
Maybe you can mount DB's directory with some state already (fixtures) OR create bounty via transactor API

Waldz (changes requested):

e2e/helpers.go
vkuznecovas: I can't create a bounty via transactor API. There's no such call and there should be no such call. Fixtures require a bunch of setup - i'll have to add an identity that i'll use for testing as well, since this is done dynamically via node I need a dynamic insert as well.

soffokl (approved):

tadaskay (approved):

PR #2619, opened 2020-09-16
Improve Tequilapi pagination DTOs

Waldz:
Related: https://github.com/mysteriumnetwork/mysterium-vpn-js/pull/50


tequilapi/contract/session.go
```diff
@@ -22,34 +22,34 @@ import (
...
 
-// ListSessionsResponse defines session list representable as json.
-// swagger:model ListSessionsResponse
-type ListSessionsResponse struct {
+// SessionListResponse defines session list representable as json.
+// swagger:model SessionListResponse
+type SessionListResponse struct {
 	Sessions   []SessionDTO               `json:"sessions"`
-	Paging     PagingDTO                  `json:"paging"`
+	Pagination PaginationDTO              `json:"pagination"`
```
tadaskay, line 52: Pagination fields should be at the root level and collection entries should be under a common name, e.g. `entries` or `items` so that it's easier for the client to consume.

tequilapi/contract/session.go
```diff
@@ -22,34 +22,34 @@ import (
...
 	Stats      SessionStatsDTO            `json:"stats"`
 	StatsDaily map[string]SessionStatsDTO `json:"stats_daily"`
```
tadaskay, line 54: These should come from a separate endpoint. Apparently, pagination only works for sessions, not for stats.

tadaskay (dismissed):

tequilapi/contract/session.go
Waldz: Fixed

tequilapi/contract/session.go
Waldz: Not in the scope now, this endpoint returns additional summary of entries in tome

vkuznecovas (approved):

tequilapi/utils/paginator.go
```diff
@@ -0,0 +1,44 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package utils
+
+import (
+	"github.com/vcraescu/go-paginator"
```
mdomasevicius, line 21: Seems like his solution only works half-way for us so I guess we can copy paste what works and not have a dependency?

mdomasevicius (approved):

tequilapi/utils/paginator.go
Waldz: You're right. But we can come back to this issue later

Issue #2658, opened 2020-09-22
Tequilapi logout endpoint


chompomonim:
- [x] Introduce logout endpoint, which clears auth session #2671
- [x] Possibility to check if node still has default password set (logging in is inconvenient) #2672, https://github.com/mysteriumnetwork/mysterium-vpn-js/pull/57
- [ ] Drop username/password concept and leave just password login
- [x] Possibility to reset password then user forgot it (via CLI command) #2704

PR #2672, opened 2020-09-24
Tequilapi authentication token issuing endpoint

Waldz:
Updates: #2658

This allows Web UI to check is default password is still enabled in node without actually creating auth session.
Also will give more flexibility to Tequilapi's users to create other authentication flows.


ignasbernotas (approved):

ui/reverse_proxy.go
```diff
@@ -107,6 +108,43 @@ func ReverseTequilapiProxy(tequilapiAddress string, tequilapiPort int, authentic
...
+	}
+
+	// authenticate from cookie
+	return parseHeaderToken(c)
+}
+
+func parseCookieToken(c *gin.Context) (string, error) {
+	token, err := c.Cookie(auth.JWTCookieName)
+	if err == http.ErrNoCookie {
+		// No error, just no token
```
tomasmik, line 128: You're repeating the same comment twice here.
Also, if we return no error just an empty string if one occurs, can't we just be a little bit more "adventurous" and ignore the error completely? I mean we're not really handling it and if an errors occurs we'll get an empty string either way.

ui/reverse_proxy.go
```diff
@@ -77,16 +78,16 @@ func ReverseTequilapiProxy(tequilapiAddress string, tequilapiPort int, authentic
 			return
 		}
 
-		// authenticate all but the login route
-		if !isTequilapiURL(c.Request.URL.Path, endpoints.TequilapiLoginEndpointPath) {
-			cookieToken, err := c.Cookie(auth.JWTCookieName)
-
+		// authenticate all but the authentication routes
+		if !(isTequilapiURL(c.Request.URL.Path, endpoints.TequilapiAuthenticateEndpointPath) || isTequilapiURL(c.Request.URL.Path, endpoints.TequilapiLoginEndpointPath)) {
+			// authenticate from header
```
tomasmik, line 83: This comment is sort of lying as we authenticate from header or cookie

ui/reverse_proxy.go
```diff
@@ -77,16 +78,16 @@ func ReverseTequilapiProxy(tequilapiAddress string, tequilapiPort int, authentic
 			return
 		}
 
-		// authenticate all but the login route
-		if !isTequilapiURL(c.Request.URL.Path, endpoints.TequilapiLoginEndpointPath) {
-			cookieToken, err := c.Cookie(auth.JWTCookieName)
-
+		// authenticate all but the authentication routes
+		if !(isTequilapiURL(c.Request.URL.Path, endpoints.TequilapiAuthenticateEndpointPath) || isTequilapiURL(c.Request.URL.Path, endpoints.TequilapiLoginEndpointPath)) {
```
tomasmik, line 82: nitpick: could we move this to a self contained func or restructure it some other way? It's kinda hard to read 🤷

tomasmik (approved):
Disregarding a few nitpicks - looks good.

tequilapi/endpoints/auth.go
```diff
@@ -43,31 +43,70 @@ type authenticator interface {
...
+	}
+	err = api.authenticator.CheckCredentials(req.Username, req.Password)
+	if err != nil {
+		utils.SendError(httpRes, err, http.StatusUnauthorized)
+		return
+	}
+
+	jwtToken, err := api.jwtAuthenticator.CreateToken(req.Username)
+	if err != nil {
+		utils.SendError(httpRes, err, http.StatusBadRequest)
```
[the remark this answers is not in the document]
Waldz: you're right, fixed

ui/reverse_proxy.go
Waldz: Yes, if there is not error i return "".
But with `err` I want to recognise user input error here and return 400 in response

vkuznecovas (approved):

ui/reverse_proxy.go
Waldz: Improved

Issue #2666, opened 2020-09-23
Improve node /config/user endpoint


mdomasevicius:
Currently `/config/user` returns overridden variables only. It would be nice if it would return a representation of current configuration.

(default-config + user-config) = current config

PR #2675, opened 2020-09-24
Expose all current values of configuration

Waldz:
Closes: #2666


vkuznecovas (approved):

ignasbernotas (approved):

config/config.go
```diff
@@ -118,6 +118,15 @@ func (cfg *Config) GetUserConfig() map[string]interface{} {
 	return cfg.user
 }
 
+// GetConfig returns current configuration.
+func (cfg *Config) GetConfig() map[string]interface{} {
+	config := make(map[string]interface{}, 0)
```
soffokl, line 123: > S1019: should use make(map[string]interface{}) instead (gosimple)

```suggestion
	config := make(map[string]interface{})
```

soffokl (approved):

config/config.go
Waldz: fixed

Issue #2700, opened 2020-10-04
Ensure proper registration metrics


chompomonim:
Node should report proper registration metrics with additional metadata such as country and operating system.

- [x] ~~More metadata next to identity metric `Event_VersionPayload(os, arch, version, country)`~~ (moved to another issue #2776 )
- [x]  How many successful/failed registrations are done in node: `Event_RegistrationEventPayload(os, arch, version, hermesId, country, event)`
event = "Started/Registered/Failed" (similar to `Event_SessionEventPayload(event)`)

Waldz:
@soffokl did i describe correctly?

soffokl:
The task is about registration metrics. I don't think we should make it complicated. 

We are not sending the registration metric from the node right now. All we have is transaction metrics only, but it's lacking some information. 
The idea here is to send events about identity registration from the node:
```
node_identity_registration{identity="0x00000123", os="androd", version="0.40.0", status="Registration request sent", country="LT"}
node_identity_registration{identity="0x00000321", os="windows", version="0.40.0", status="Registration done", contry="KG"}
```

PR #2758, opened 2020-10-20
Add identity registration metrics

tomasmik:
Addresses: mysteriumnetwork/node#2700


soffokl (approved):

identity/registry/registry_contract.go
```diff
@@ -40,19 +40,23 @@ type registryStorage interface {
...
+	getLocation     getLocation
+
+	stop chan struct{}
+	lock sync.Mutex
+	once sync.Once
 }
 
 // NewIdentityRegistryContract creates identity registry service which uses blockchain for information
-func NewIdentityRegistryContract(ethClient *paymentClient.ReconnectableEthClient, registryAddress, hermesAddress common.Address, registryStorage registryStorage, publisher eventbus.Publisher) (*contractRegistry, error) {
+func NewIdentityRegistryContract(ethClient *paymentClient.ReconnectableEthClient, registryAddress, hermesAddress common.Address, registryStorage registryStorage, publisher eventbus.Publisher, gl getLocation) (*contractRegistry, error) {
```
Waldz, line 59: Dont couple this component to location directly, but rather depend on location event with `Subscribe(eb eventbus.Subscriber) error`

core/quality/morqa_transport.go
```diff
@@ -58,11 +58,30 @@ func mapEventToMetric(event Event) (string, *metrics.Event) {
...
+	return data.Identity, &metrics.Event{
+		Metric: &metrics.Event_RegistrationPayload{
+			RegistrationPayload: &metrics.RegistrationPayload{
+				Version: &metrics.VersionPayload{
+					Version: info.Version,
+					Os:      info.OS,
+					Arch:    info.Arch,
+				},
+				Country:  data.Country,
+				Identity: data.Identity,
```
Waldz, line 78: All metrics are originated with identity, no need to repeat this

identity/registry/registry_contract.go
Waldz: Country is needed for registration metric, not necessarily for all registration subscribers.
Consider appending it only in metric sending.

identity/registry/registry_contract.go
```diff
@@ -210,8 +218,9 @@ func (registry *contractRegistry) subscribeToRegistrationEvent(identity identity
 
 			log.Debug().Msgf("Sending registration success event for %v", identity)
 			registry.publisher.Publish(AppTopicIdentityRegistration, AppEventIdentityRegistration{
-				ID:     identity,
-				Status: status,
+				ID:      identity,
+				Status:  status,
+				Country: registry.getLocation(),
```
Waldz, line 223: `getLocation()` should return all location, so `getLocation().Country`

Waldz (changes requested):

core/quality/morqa_transport.go
tomasmik: From morqa code I assumed that this is something that might get changed or is not ideal.
I think we should keep it here, but we can argue about this 🤷

identity/registry/registry_contract.go
tomasmik: We've talked about this. What I think we should do is remake how location detection works allowing our metrics service to get it any time it wants through the event bus and then remove such dependencies.

What sucks now, is "prefilling" the bus with a location object is kinda bad and hacky.

vkuznecovas (approved):

tadaskay (approved):

core/quality/morqa_transport.go
Waldz: No, it's our agreed design, everyone  metric has `event.TargetId` 
Furthermore, all metrics are signed with identity PK, and it prevents from forging metrics.

identity/registry/registry_contract.go
Waldz: I made node because of 2 concerns:
1. Original node location does not change, so doing request for each registration is excess
2. We have plans location changes event based, so this will be +1 place to refactor

tomasmik: Second point is true, yes this is +1 place to refactor. But I still think that depending on other parts of the code to call detect location just so that we know the location before registering identity is sort of strange and might fail but so be it, I'll make the changes 🤷

core/quality/morqa_transport.go
```diff
@@ -120,7 +139,7 @@ func sessionTokensToMetricsEvent(ctx sessionTokensContext) (string, *metrics.Eve
 		IsProvider: false,
 		Metric: &metrics.Event_SessionTokensPayload{
 			SessionTokensPayload: &metrics.SessionTokensPayload{
-				Tokens: ctx.Tokens.Text(10),
+				Tokens: ctx.Tokens.Uint64(),
```
chompomonim, line 142: What is `Tokens` in this context? If it is related to MYST tokens, then `Uint64` is not enough.

tomasmik: This is sort of weird, I used to be a string, but after generating the new metrics, it turned in to an uint64, I didn't really edit it 🤷 
https://github.com/mysteriumnetwork/metrics/commit/544a1e13d6651b963d257aafb419282c84f434da#diff-6039342a2ba47b726e1134b8ead7c00196fd9cefd9dd925b98e3cf01f21b4e2cR43

Waldz (approved):

PR #2763, opened 2020-10-20
Accept referral token for mobile registration

tadaskay:
Ref: https://github.com/mysteriumnetwork/node/issues/2654


mobile/mysterium/entrypoint.go
```diff
@@ -483,7 +484,11 @@ func (mb *MobileNode) RegisterIdentity(req *RegisterIdentityRequest) error {
 		return errors.Wrap(err, "could not get registration fees")
 	}
 
-	err = mb.transactor.RegisterIdentity(req.IdentityAddress, big.NewInt(0), fees.Fee, "", nil)
+	token := &req.Token
```
tomasmik, line 487: I think that this would read better:
```go
var token *string
if req.Token != "" {
  token = &req.Token
}
```

tomasmik (approved):

PR #2783, opened 2020-10-28
Inject HTTP client to Quality client

Waldz:
Just a minor improvements

### DIFF
--- a/core/quality/mysterium_morqa.go
+++ b/core/quality/mysterium_morqa.go
@@ -26,11 +26,11 @@ import (
 	"sync"
 	"time"
 
-	"github.com/golang/protobuf/proto"
 	"github.com/mysteriumnetwork/metrics"
 	"github.com/mysteriumnetwork/node/identity"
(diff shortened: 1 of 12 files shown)


core/quality/mysterium_morqa.go
```diff
@@ -174,7 +164,9 @@ func (m *MysteriumMORQA) sendMetrics() error {
 		return err
 	}
 
-	defer response.Body.Close()
+	defer func() {
+		_ = response.Body.Close()
```
tomasmik, line 168: Can't we adjust the linter and at least make it ignore defers like these? This is so verbose

tadaskay (approved):

soffokl (approved):

core/quality/mysterium_morqa.go
Waldz: Turns out no need to adjust linter, i just misunderstand linter warning

tomasmik (approved):

PR #2777, opened 2020-10-27
Introduce API client which bypasses DNS lookups

Waldz:
This performs HTTP client requests to by dialling directly to IP instead of doing DNS resolution before dial.

For now changes just for API client (DiscoveryService) in order to conduct tests how it performs.
Planning to introduce failover transport (DNS dial -> IP dial) for all of our services in upcoming PRs.

PR #2792, opened 2020-11-05
HTTP clients with failover transport

Waldz:
Updates: #2777 

This introduces HTTP client failover transport (DNS dial -> IP dial) and runs both in parallel to find earliest successfully established connection.
All our HTTP clients changed to this dialling behaviour.

Lets say `dig api.ipify.org` -> resolves to 50.19.252.36 and 174.129.214.20, so this is diagram how 
`DialerSwarm.DialContext(ctx, "tcp", "api.ipify.org:443")` would act:

```
      call to DialContext()
  ---api.ipify.org:443-----------> 1. ResolveContext("api.ipify.org:443")
                                   2. dials many addrs
                                       use earliest
                                   ---api.ipify.org:443----------->x
                                   ---174.129.214.20:443--------------> conn established
  conn<----------------------------------------------------------------
                                   ---50.19.252.36:443--------->x
                                         any may fail
```

### DIFF
--- a/requests/dialer_swarm.go
+++ b/requests/dialer_swarm.go
@@ -1,0 +1,77 @@
+/*
+ * Copyright (C) 2020 The "MysteriumNetwork/node" Authors.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package requests
+
+import (
+	"context"
+	"net"
+	"time"
+)
+
+// DialerSwarm is a connects to multiple addresses in parallel and first successful connection wins.
+type DialerSwarm struct {
+	resolver *net.Resolver
+	dialer   *net.Dialer
+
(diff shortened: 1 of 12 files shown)


requests/dialer_swarm.go
```diff
@@ -0,0 +1,216 @@
...
+
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	resultCh := make(chan dialResult)
+	err := &ErrorSwarmDial{}
+
+	var active int
+dialLoop:
+	for addrChan != nil || active > 0 {
```
vkuznecovas, line 89: I wish all the strength and courage for the future generations that will be reading this.

Would this not work:

```go

func (ds *DialerSwarm) dialAddrs(ctx context.Context, network string, addrs []string) (net.Conn, error) {
	ctx, cancel := context.WithCancel(ctx)
	defer cancel()

	resultCh := make(chan dialResult)
	for i := range addrs {
		go ds.dialAddr(ctx, network, addrs[i], resultCh)
	}

	var responses int
	for {
		select {
		case v := <-resultCh:
			responses++
			if v.Err != nil || v.Conn == nil {
				if responses == len(addrs) {
					return nil, ErrAllDialsFailed
				}
			} 
			return v.Conn, nil
		case <-ctx.Done():
			return nil, errors.New("cancelled")
		}
	}
}
```

vkuznecovas (changes requested):

requests/dialer_swarm.go
Waldz: It's ideas from `net` and `github.com/libp2p/go-libp2p` packages for better performance.
Anyway simplified for sake of readability.

vkuznecovas (approved):

tomasmik (approved):

mdomasevicius (approved):

requests/dialer_swarm.go
Waldz: Indeed reverted this change, `resultCh` is left unclosed for every dial

PR #2794, opened 2020-11-05
Hermes chain ID - unlock

tomasmik:
This covers the unlock event as long as hermes and chain ID is concerned.

Major changes:
- Identity unlock in tequila now requires chainID
- Identity current now requires chainID (because it also calls unlock and will later register)

We need chainID to update the balance using hermes or the multichain BC. So there is no way to get out of this.

This doesn't build and it's not finished. This covers a small part of the hermes multichain usage we still need to cover other cases like: Register and Grand total change though the amount of changes is getting out of hand.

### DIFF
--- a/session/pingpong/hermes_caller.go
+++ b/session/pingpong/hermes_caller.go
@@ -184,23 +184,28 @@ func (ac *HermesCaller) RevealR(r, provider string, agreementID *big.Int) error 
 }
 
 // GetConsumerData gets consumer data from hermes
-func (ac *HermesCaller) GetConsumerData(id string) (ConsumerData, error) {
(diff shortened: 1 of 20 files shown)


session/pingpong/hermes_caller.go
```diff
@@ -184,23 +184,28 @@ func (ac *HermesCaller) RevealR(r, provider string, agreementID *big.Int) error
 }
 
 // GetConsumerData gets consumer data from hermes
-func (ac *HermesCaller) GetConsumerData(id string) (ConsumerData, error) {
+func (ac *HermesCaller) GetConsumerData(chainID int64, id string) (ConsumerData, error) {
```
tomasmik, line 187: This is from where all the changes started.

vkuznecovas (approved):

PR #2800, opened 2020-11-09
Registration multichain support

tomasmik:
- Remade registry storage, we now store per identity+chainID key
- Added chain ID to the registration workflow

### DIFF
--- a/core/discovery/discovery.go
+++ b/core/discovery/discovery.go
@@ -21,6 +21,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/mysteriumnetwork/node/config"
 	"github.com/mysteriumnetwork/node/eventbus"
 	"github.com/mysteriumnetwork/node/identity"
 	"github.com/mysteriumnetwork/node/identity/registry"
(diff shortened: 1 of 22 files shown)


core/discovery/discovery.go
```diff
@@ -95,6 +97,7 @@ func (d *Discovery) Start(ownIdentity identity.Identity, proposal market.Service
 	d.ownIdentity = ownIdentity
 	d.signer = d.signerCreate(ownIdentity)
 	d.proposal = proposal
+	d.chainID = config.GetInt64(config.FlagChainID)
```
vkuznecovas, line 100: either inject it to the discovery, or just use it from config directly where needed without having it saved in discovery

identity/registry/storage.go
```diff
@@ -77,42 +82,55 @@ func (rss *RegistrationStatusStorage) Store(status StoredRegistrationStatus) err
...
+
+	result := make([]StoredRegistrationStatus, len(list))
+	for i, l := range list {
+		result[i] = l.StoredRegistrationStatus
+	}
+	return result, nil
+}
+
+func (rss *RegistrationStatusStorage) makeKey(identity identity.Identity, chainID int64) string {
+	return fmt.Sprintf("%s%d", identity.Address, chainID)
```
vkuznecovas, line 135: add a separator, makes it easier to read if we ever had to venture into bolt.

vkuznecovas (changes requested):

vkuznecovas (approved):

PR #2804, opened 2020-11-09
Filter status response by chainID

tomasmik:


session/pingpong/consumer_balance_tracker.go
```diff
@@ -340,7 +340,22 @@ func (cbt *ConsumerBalanceTracker) alignWithTransactor(id identity.Identity) {
 		if err != nil {
 			return err
 		}
-		data = resp
+
+		var status *registry.TransactorStatusResponse
+		for _, v := range resp {
+			if v.ChainID != chainID {
```
vkuznecovas, line 346: if you invert this if, the for becomes shorter.

vkuznecovas (changes requested):

session/pingpong/consumer_balance_tracker.go
tomasmik: I know, I like to avoid nesting things in to `ifs` if I can help it. As in this reads better imo, but if you want, i can change it :shrug:

PR #2832, opened 2020-11-18
Mobile API for Pilvytis payment orders

tadaskay:


pilvytis/order_status_tracker.go
```diff
@@ -0,0 +1,128 @@
...
+import (
+	"sync"
+	"time"
+
+	"github.com/mysteriumnetwork/node/eventbus"
+	"github.com/mysteriumnetwork/node/identity"
+	"github.com/rs/zerolog/log"
+)
+
+type OrderSummary struct {
```
vkuznecovas, line 29: you're missing comments, linter will no like you.

vkuznecovas (approved):

pilvytis/order_status_tracker.go
```diff
@@ -0,0 +1,128 @@
...
+	t.startOnce.Do(func() {
+		go trackingFunc()
+	})
+}
+
+func (t *StatusTracker) Pause() {
+	t.stopOnce.Do(func() {
+		log.Info().Msg("Pausing order status tracking")
+		t.stopCh <- struct{}{}
+		t.startOnce = sync.Once{}
```
tomasmik, line 101: Isn't this a race condition? If Pause() and Track() are hit at the same time?

pilvytis/order_status_tracker.go
```diff
@@ -0,0 +1,128 @@
...
+	GetPaymentOrders(id identity.Identity) ([]OrderResponse, error)
+}
+
+type StatusTracker struct {
+	api              orderProvider
+	identityProvider identityProvider
+	eventBus         eventbus.Publisher
+	orders           []OrderSummary
+
+	pauseCondition func(response []OrderResponse) bool
```
tomasmik, line 45: This is not used

pilvytis/order_status_tracker.go
```diff
@@ -0,0 +1,128 @@
...
+		if err != nil {
+			return err
+		}
+		for _, newOrder := range newOrders {
+			order := t.getOrCreate(newOrder.ID, newOrder.Identity)
+			if order.Status != newOrder.Status {
+				order.Status = newOrder.Status
+				t.eventBus.Publish(AppTopicOrderStatusChanged, AppEventOrderStatusChanged{*order})
+			}
+			if newOrder.Status.Incomplete() {
```
tomasmik, line 119: I think if `GetPaymentOrders` returns orders in unexpected order this will make things fail no?
If for example we get `Pending, Pending, Complete` this will stop. Or do we not care about that?

pilvytis/order_status_tracker.go
```diff
@@ -0,0 +1,128 @@
...
+				keepTracking = true
+			}
+		}
+	}
+	if !keepTracking {
+		go t.Pause()
```
tomasmik, line 125: I honestly dont think that the separate goroutine here is needed it just adds additional guess work

pilvytis/service.go
```diff
@@ -0,0 +1,103 @@
...
+}
+
+// Service payment service Pilvytis.
+type Service struct {
+	api              *API
+	eventBus         eventbus.Publisher
+	identityProvider identityProvider
+	tracker          statusTracker
+
+	lock sync.RWMutex
```
tomasmik, line 44: This is not used?

tomasmik (changes requested):

pilvytis/order_status_tracker.go
tadaskay: It will not stop if any of the orders are incomplete. The order does not matter here.

pilvytis/order_status_tracker.go
tadaskay: This is needed because update needs to be non-blocking for stopCh to receive the stop (pause) signal.

pilvytis/order_status_tracker.go
tadaskay: Actually, not reversing, but added `keepTracking = true` to the error case branch.

pilvytis/order_status_tracker.go
```diff
@@ -0,0 +1,138 @@
...
+func (t *StatusTracker) Track() {
+	t.lock.Lock()
+	defer t.lock.Unlock()
+	if t.tracking {
+		return
+	}
+	t.tracking = true
+	go func() {
+		log.Info().Msg("Tracking order statuses...")
+		updateTicker := time.NewTicker(t.updateInterval)
```
tomasmik, line 86: Make sure you stop this.
https://golang.org/pkg/time/#NewTicker

tomasmik (approved):

Issue #2840, opened 2020-11-19
Hide logs in interactive CLI apps


tomasmik:
This issue is related to: https://github.com/mysteriumnetwork/node/issues/2837

Logs, especially debugging level, should not be displayed to the end user as he does not need to see them.
However they should be stored in case there is a need to debug something.

Issue #2837, opened 2020-11-18
User friendly CLI


chompomonim:
For our advanced linux and macOS users we could create easy to use CLI client for consumers.

**Add additional commands for `myst` tool:**
- [x] `myst account register`:  It should create identity, unlock it and show topup address where to topup (if not testnet). If user is running it second time, check if there is identity and don't allow create second one.  (https://github.com/mysteriumnetwork/node/pull/2854 / https://github.com/mysteriumnetwork/node/pull/2867)
- [x] `myst account info` if not registered, suggests to register. Should also show balance of default identity ( https://github.com/mysteriumnetwork/node/pull/2867 )
- [x] `myst account topup [amount of tokens] --currency=BTC`: shows topup instructions and address where to send tokens. ( https://github.com/mysteriumnetwork/node/pull/2867 )
- [x] `myst connection up [node identity]`:  connects into node. Should show message that identity is not registered (or balance is empty). Could do `myst register` automatically in background while in testnet.  ( https://github.com/mysteriumnetwork/node/pull/2862 )
- [x] `myst connection down`:  disconnect from node ( https://github.com/mysteriumnetwork/node/pull/2862 )
- [x] ~~`myst proposals [--all]`:  lists all proposals with additional info (country, IP type, price) (only wireguard if without `--all`)~~
- [x] `myst connection proposals --country=[country] --ip_type=[ip_type]`: list filtered proposals (https://github.com/mysteriumnetwork/node/pull/2859)
- [x] `myst connection info`: shows connection status. ( https://github.com/mysteriumnetwork/node/pull/2865 )
- [x] Make possible to set which identity is default (could be only in advances interactive CLI). ( https://github.com/mysteriumnetwork/node/pull/2870 )
- [x] Update `myst help` command. (This is sort of auto updated as it's being built)
- [x] Don't show terminal logs for interactive commands.  ( https://github.com/mysteriumnetwork/node/pull/2853 )
- [x] Show Terms of Use once, ask to agree and don't show anymore when agreed. ( https://github.com/mysteriumnetwork/node/pull/2894 stuck in review )
- [x] Update CLI documentation in docs-v2 after that. ( https://github.com/mysteriumnetwork/documentation/pull/15 )

On the side:

- have auto un-lock for `connect` type commands.  ( https://github.com/mysteriumnetwork/node/pull/2920 )
- dont require the `0` if no stake is provided in register command  ( https://github.com/mysteriumnetwork/node/pull/2869 )

PR #2842, opened 2020-11-19
Remake logging in end user products

tomasmik:
Now in compiled `myst cli` and future applications, logging to terminal will be disabled.
This is required in order to move along with the user friendly CLI apps.

The end user does not need to see the 50 lines of debugging messages printed when they launch `myst cli`, nor does he need the `zerolog` messages.

Now instead of:

```sh
» identities register 0x5d370ab418179d7de9410899249604d54485b860
2020-11-19T13:47:03.522 ERR tequilapi/client/http_client.go:123      >  error="server response invalid: 500 Internal Server Error (http://127.0.0.1:4050/identities/0x5d370ab418179d7de9410899249604d54485b860/register). Possible error: failed identity registration request: server response invalid: 400 Bad Request (https://testnet-transactor.mysterium.network/api/v1/identity/register){\"error\":{}}"
[WARNING] could not register identity: server response invalid: 500 Internal Server Error (http://127.0.0.1:4050/identities/0x5d370ab418179d7de9410899249604d54485b860/register). Possible error: failed identity registration request: server response invalid: 400 Bad Request (https://testnet-transactor.mysterium.network/api/v1/identity/register){"error":{}}
```

The end user sees this:
```sh
» identities register 0x5d370ab418179d7de9410899249604d54485b867
[WARNING] could not register identity: server response invalid: 500 Internal Server Error (http://127.0.0.1:4050/identities/0x5d370ab418179d7de9410899249604d54485b867/register). Possible error: failed identity registration request: server response invalid: 400 Bad Request (https://testnet-transactor.mysterium.network/api/v1/identity/register){"error":{}}
```

It's not perfect, but better.

Closes: https://github.com/mysteriumnetwork/node/issues/2840

### DIFF
--- a/cmd/commands/cli/command.go
+++ b/cmd/commands/cli/command.go
@@ -44,7 +44,8 @@ import (
 	"github.com/mysteriumnetwork/node/utils"
 )
 
-const cliCommandName = "cli"
+// CliCommandName is the name which is used to call this command
+const CliCommandName = "cli"
 
 const serviceHelp = `service <action> [args]
 	start	<ProviderID> <ServiceType> [options]
@@ -58,7 +59,7 @@ const serviceHelp = `service <action> [args]
 // NewCommand constructs CLI based Mysterium UI with possibility to control quiting
 func NewCommand() *cli.Command {
 	return &cli.Command{
-		Name:   cliCommandName,
+		Name:   CliCommandName,
 		Usage:  "Starts a CLI client with a Tequilapi",
 		Before: clicontext.LoadUserConfigQuietly,
 		Action: func(ctx *cli.Context) error {
--- a/cmd/mysterium_node/mysterium_node.go
+++ b/cmd/mysterium_node/mysterium_node.go
@@ -19,6 +19,8 @@ package main
 
 import (
 	"os"
+	"path/filepath"
+	"sync"
 
 	command_cli "github.com/mysteriumnetwork/node/cmd/commands/cli"
 	"github.com/mysteriumnetwork/node/cmd/commands/daemon"
@@ -29,6 +31,7 @@ import (
 	"github.com/mysteriumnetwork/node/config"
 	"github.com/mysteriumnetwork/node/logconfig"
 	"github.com/mysteriumnetwork/node/metadata"
+	"github.com/rs/zerolog"
 	"github.com/rs/zerolog/log"
 	"github.com/urfave/cli/v2"
 )
@@ -68,16 +71,19 @@ func NewCommand() (*cli.App, error) {
 		versionCommand.Run(ctx)
 	}
 
-	app := cli.NewApp()
+	app, err := newApp()
+	if err != nil {
+		return nil, err
+	}
+
 	app.Usage = "VPN server and client for Mysterium Network https://mysterium.network/"
 	app.Authors = []*cli.Author{
 		{Name: `The "MysteriumNetwork/node" Authors`, Email: "mysterium-dev@mysterium.network"},
 	}
 	app.Version = metadata.VersionAsString()
 	app.Copyright = licenseCopyright
-	if err := config.RegisterFlagsNode(&app.Flags); err != nil {
-		return nil, err
-	}
+	app.Before = readyOnceFunc()
+
 	app.Commands = []*cli.Command{
 		versionCommand,
 		licenseCommand,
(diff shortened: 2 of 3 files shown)


vkuznecovas (changes requested):
This seems way overcomplicated for such a simple objective. 

Add a flag that enables verbose logging to stdout, log to file by default? 
There would be no need to split this for separate commands either.

tomasmik:
> This seems way overcomplicated for such a simple objective.
> 
> Add a flag that enables verbose logging to stdout, log to file by default?
> There would be no need to split this for separate commands either.

My reasoning for the split between different commands is that, I don't see a point in running `node daemon` and not seeing the console logs. Though I can remove that no problem.
The flag though, I was also thinking about it too just wasn't sure if it's of any use.

vkuznecovas:
> > This seems way overcomplicated for such a simple objective.
> > Add a flag that enables verbose logging to stdout, log to file by default?
> > There would be no need to split this for separate commands either.
> 
> My reasoning for the split between different commands is that, I don't see a point in running `node daemon` and not seeing the console logs. Though I can remove that no problem.
> The flag though, I was also thinking about it too just wasn't sure if it's of any use.

That's why I was talking about a totally separate binary. You can approach the problem completely differently here, but we'll still need the CLI as is at least for developer testing.

cmd/commands/cli/command.go
```diff
@@ -58,7 +59,7 @@ const serviceHelp = `service <action> [args]
 // NewCommand constructs CLI based Mysterium UI with possibility to control quiting
 func NewCommand() *cli.Command {
 	return &cli.Command{
-		Name:   cliCommandName,
+		Name:   CliCommandName,
```
mdomasevicius, line 62: Actually according to go bible this should be `CLICommandName` or `cliCommandName` as CLI is an acronym

cmd/mysterium_node/mysterium_node.go
```diff
@@ -89,3 +95,52 @@ func NewCommand() (*cli.App, error) {
...
+				level = zerolog.DebugLevel
+			}
+
+			err = logconfig.ConfigureUI(&logconfig.LogOptions{
+				LogLevel: level,
+				LogHTTP:  false,
+				Filepath: filepath.Join(logDir, cmd),
+			})
+			if err != nil {
+				log.Error().Err(err).Msg("configuring logger for ui")
```
soffokl, line 141:
```suggestion
				log.Error().Err(err).Msg("Failed to configure logger for ui")
```

cmd/mysterium_node/mysterium_node.go
```diff
@@ -89,3 +95,52 @@ func NewCommand() (*cli.App, error) {
...
+
+// uiCommands is a map which consists of all
+// commands are used directly by a user.
+var uiCommands = map[string]struct{}{
+	command_cli.CliCommandName: {},
+}
+
+// readyOnceFunc returns a func which is only run once and can
+// be executed to preconfigure global settings used in the app.
+func readyOnceFunc() cli.BeforeFunc {
```
mdomasevicius, line 112: I think this is a terrible name, `configureCLILogging` even just `configureLogging` is better and would reduce comment docs by a whole line

cmd/mysterium_node/mysterium_node.go
```diff
@@ -89,3 +95,52 @@ func NewCommand() (*cli.App, error) {
...
+				// no log dir.
+				return
+			}
+
+			level, err := zerolog.ParseLevel(ctx.String(config.FlagLogLevel.Name))
+			if err != nil {
+				level = zerolog.DebugLevel
+			}
+
+			err = logconfig.ConfigureUI(&logconfig.LogOptions{
```
mdomasevicius, line 135: How about `logconfig.BootstrapCLI(...)` - package gives away the context and function name can explain the rest?

mdomasevicius:
> That's why I was talking about a totally separate binary. You can approach the problem completely differently here, but we'll still need the CLI as is at least for developer testing.

Separate binary for CLI?

soffokl (commented):
I don't think that we need to have logs for CLI. Even log to file looks like a bad idea.
1) If you are going to write to the same configured file that used for the daemon, it will create a mess in the log file when 2 processes trying to write to the same file at the same time. 
2) CLI should not print logs, only useful output that can be redirected to the pipe for the next command `$ myst cli identity list | myst cli identity register`

If a user wants to have a debug output of the CLI command we can redirect it to the different streams. Debug goes to the STDERR, useful output goes to the STDOUT.

Logging for daemon should go to the STDOUT, since processes that execute it has better log maintenance(systemd, docker, whatever)

Separate binary for CLI will make a delivery process more complicated. Everyone loves golang for the single binary file with no other dependencies. And both binaries reuse the same codebase. I think it's simpler to keep as one file.

logconfig/config.go
```diff
@@ -62,6 +63,30 @@ func Bootstrap() {
 	setGlobalLogger(&logger)
 }
 
+// ConfigureUI configures the logger for user interface CLI
```
soffokl, line 66: UI for CLI looks weird. User interface for command-line interface.

tomasmik:
> I don't think that we need to have logs for CLI. Even log to file looks like a bad idea.
> 
> 1. If you are going to write to the same configured file that used for the daemon, it will create a mess in the log file when 2 processes trying to write to the same file at the same time.

We log to a different file here, which is named according to the command, `cli` in this case and it's only really useful in cases where someone would like to share the logs with us when something is not working as it should.

>Separate binary for CLI will make a delivery process more complicated. Everyone loves golang for the single binary file with no other dependencies. And both binaries reuse the same codebase. I think it's simpler to keep as one file.

I think that the separate binary wasn't about this, current `myst cli` but this: https://github.com/mysteriumnetwork/node/issues/2837

soffokl:
> We log to a different file here, which is named according to the command, cli in this case and it's only really useful in cases where someone would like to share the logs with us when something is not working as it should. Though we can explore writing to stderr

CLI just does calls to TequilAPI, I think it's enough to have just logs from the daemon, it already should include everything. Why do we need to have extra logs for CLI?

tomasmik:
Everyone is on a different page. Let's groom this.

PR #2843, opened 2020-11-20
Remove type matching from config merge

mdomasevicius:
The way I see it - it does not matter what type the config value is stored in a `map[string]interface{}` hierarchy. 
The `config.Config` has receivers for fetching a value in any suitable type. However this introduces a possibility to break config although I have already broken it by consuming tequilapi.

In my case the problem is with json unmarshal converting JSON `number` to `float64` when target type is interface{}

### DIFF
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -143,6 +143,28 @@ func TestConfig_ParseStringSliceFlag(t *testing.T) {
 	}
 }
 
+// this can happen when updating config via tequilapi - json unmarshal
+// translates json number to float64 by default if target type is interface{}
+func TestSimilarTypeMerge(t *testing.T) {
+	// given
+	cfg := NewConfig()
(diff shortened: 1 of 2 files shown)


soffokl (approved):

config/config_test.go
```diff
@@ -143,6 +143,28 @@ func TestConfig_ParseStringSliceFlag(t *testing.T) {
 	}
 }
 
+// this can happen when updating config via tequilapi - json unmarshal
+// translates json number to float64 by default if target type is interface{}
+func TestSimilarTypeMarge(t *testing.T) {
```
mdomasevicius, line 148:
```suggestion
func TestSimilarTypeMerge(t *testing.T) {
```

vkuznecovas (approved):

tadaskay (approved):

PR #2850, opened 2020-11-25
Migrate testnet identity to testnet2 on update DEB

soffokl:


vkuznecovas (changes requested):
Let's migrate keystore only.

soffokl:
@vkuznecovas migrating keystore only now.

vkuznecovas (approved):

PR #2870, opened 2020-11-30
Allow to set default identity

tomasmik:
Added a new sub-command `default`.
For now it only parses a single flag and updates current default identity according to it.

Updates: https://github.com/mysteriumnetwork/node/issues/2837

### DIFF
--- a/cmd/commands/account/command.go
+++ b/cmd/commands/account/command.go
@@ -103,12 +103,36 @@ func NewCommand() *cli.Command {
 					return nil
 				},
 			},
+			{
+				Name:      "set-identity",
+				Usage:     "Sets a new identity for your account which will be used in commands that require it",
+				ArgsUsage: "[IdentityAddress]",
+				Action: func(ctx *cli.Context) error {
+					cmd.setIdentity(ctx)
+					return nil
+				},
+			},
 		},
 	}
 }
 
 type command struct {
 	tequilapi *tequilapi_client.Client
+}
+
+func (c *command) setIdentity(ctx *cli.Context) {
+	givenID := ctx.Args().First()
+	if givenID == "" {
+		clio.Warn("No identity provided")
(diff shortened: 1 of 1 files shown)


cmd/commands/account/command.go
```diff
@@ -103,6 +103,15 @@ func NewCommand() *cli.Command {
 					return nil
 				},
 			},
+			{
+				Name:      "set-identity",
+				Usage:     "Sets a new identity for your account which will be used in commands that require it",
```
mdomasevicius, line 108: Does everyone else has a clear understanding of what is an `account` ?

tomasmik: My explanation for this would be that: commands are grouped under `account` namespace, while it's not a true user account per say but it's sort of similar. So I think that we can refer to your personal configuration as `account` here.

Though then again it's not to later to rename things, change help messages or event drop the `account` namespace and use something different. I'm very open to ideas.

cmd/commands/account/command.go
```diff
@@ -111,6 +120,41 @@ type command struct {
...
+
+	found := false
+	for _, id := range identities {
+		if id.Address == givenID {
+			found = true
+			break
+		}
+	}
+
+	if !found {
```
Waldz, line 144: You dont need to do all this validation, it's performed by `CurrentIdentity()`

Waldz (changes requested):

soffokl (approved):

cmd/commands/account/command.go
tomasmik: I let myself be liberal with the calls because we're calling localhost after all and this is nicer in terms of displaying the actual error message without leaking anything that would confuse a simple non technical user. :shrug:

Waldz (approved):

PR #2894, opened 2020-12-03
Improve terms of use agreement

tomasmik:
Terms are required to be agreed upon for:
* `cli`
* `connection`
* `service`

We can add more commands, if the implementation of this is acceptable, as to now waste time adding something I'll have to drop.

Main changes:
* We only ask for `--agreed-terms-and-conditions` flag once, saving the agreement to user config.
* Other services can check and update terms using the new endpoints.
* Everything is saved in user config, so that we dont have to have tequila started in order to check if user has agreed to them
* We save a single terms version, which if changed, will require the user to re-accept the TOS
* We no longer require to store the date as it's stored in the user config and means absolutely nothing.


config/flags_terms.go
```diff
@@ -0,0 +1,64 @@
...
+	FlagTermsProviderAgree = cli.BoolFlag{
+		Name:  "terms-agreed-provider",
+		Usage: "Defines test network configuration",
+	}
+
+	// FlagTermsVersion is the current agreed TOS version.
+	FlagTermsVersion = cli.StringFlag{
+		Name:  "terms-agreed-version",
+		Usage: "Defines test network configuration",
+	}
```
soffokl: It will be very confusing for users what flag they really need to use to agree.

metadata/terms.go
```diff
@@ -17,6 +17,10 @@
 
 package metadata
 
+// CurrentTermsVersion is the version which has to be agreed
+// with when confirming TOS agreements.
+const CurrentTermsVersion = "0.0.27"
```
soffokl, line 22: Should this be changed at https://github.com/mysteriumnetwork/terms?
I thought that this file is generated and should not be edited.

config/flags_terms.go
```diff
@@ -0,0 +1,64 @@
...
```
soffokl, line 45: It will be very confusing for users what they need to use to agree. 
What is the difference between `agreed-terms-and-conditions` and `agreed-terms-and-conditions`?

config/flags_terms.go
tomasmik: They can pass `--agreed-terms-and-conditions` to agree with either terms as the code works that way. These other flags are only if you wan to pass these specifically and for the ease of use when it comes to parsing the config and saving values to it.

metadata/terms.go
tomasmik: Maybe someone else could have input on this? As far as I can see, this isn't generated, tho I could be wrong.

cmd/commands/service/command.go
```diff
@@ -154,19 +156,46 @@ func (sc *serviceCommand) unlockIdentity(id, passphrase string) string {
...
+	go func() {
+		t := true
+		for i := 0; i < 5; i++ {
+			if err := sc.tequilapi.UpdateTerms(contract.TermsRequest{
+				AgreedProvider: &t,
+				AgreedVersion:  metadata.CurrentTermsVersion,
+			}); err == nil {
+				return
+			}
+			time.Sleep(time.Second * 2)
```
vkuznecovas, line 173: So this stops the world for 10 seconds if I try to exit a node that is having issues with tequila? That's bad. Add something that's cancellable in case of an exit.

cmd/commands/cli/command.go
```diff
@@ -106,18 +108,56 @@ var versionSummary = metadata.VersionAsSummary(metadata.LicenseCopyright(
 	"type 'license --conditions'",
 ))
 
+func (c *cliApp) handleTOS(ctx *cli.Context) error {
+	if ctx.Bool(config.FlagAgreedTermsConditions.Name) {
+		c.acceptTOS(ctx)
+		return nil
+	}
+
+	agreedC := config.GetBool(config.FlagTermsConsumerAgree)
```
vkuznecovas, line 117: do we really have all these different TOS for use? Is it not a single one?

config/flags_terms.go
```diff
@@ -0,0 +1,64 @@
...
+
+var (
+	// FlagAgreedTermsConditions agree with terms & conditions.
+	FlagAgreedTermsConditions = cli.BoolFlag{
+		Name:  "agreed-terms-and-conditions",
+		Usage: "Agree with terms & conditions for consumer, provider or both depending on the command executed",
+	}
+
+	// FlagTermsConsumerAgree marks that consumer TOS has been agreed.
+	FlagTermsConsumerAgree = cli.BoolFlag{
```
vkuznecovas, line 30: I don't understand why we need multiple flags.

metadata/terms.go
vkuznecovas: AFAIK this is generated from the terms and should not be changed manually.

tequilapi/contract/terms.go
```diff
@@ -0,0 +1,73 @@
...
+		AgreedProvider: config.GetBool(config.FlagTermsProviderAgree),
+		AgreedConsumer: config.GetBool(config.FlagTermsConsumerAgree),
+		AgreedVersion:  config.GetString(config.FlagTermsVersion),
+		CurrentVersion: metadata.CurrentTermsVersion,
+	}
+}
+
+// Iterable turns a TermsRequest in to an iterable map which
+// can be mapped directly to a user config.
+func (t *TermsRequest) Iterable() map[string]interface{} {
```
vkuznecovas, line 61: While it is Iterable, ToMap() sounds like a better option here to me.

vkuznecovas (changes requested):

cmd/commands/cli/command.go
tomasmik: Yes that something I would like to discuss and figure out as well. Do we need to make them different? I was sort of looking in to NodeUI/Desktop app, etc. and tried to keep the same consistency.

So lets say you're using only the desktop app and nothing else, well then I guess you need to only agree to the Consumer terms of service, but if you later fire up `myst service` you would have to agree to Provider terms of service aswell.

I would like to hear what @chompomonim  thinks about this, because dropping the 2 different flags and having only 1, would make life a lot easier.

cmd/commands/service/command.go
tomasmik: No one really tracks this go routine, so nothing is really stopped. But you have fair point.

cmd/commands/cli/command.go
Waldz: @tomasmik, we agreed not to have terms for each small command, but rather just for.
And give to @chompomonim to test and feedback.

```
myst service --accepted-terms-of-service
myst connection up --accepted-terms-of-service
```

tomasmik: @Waldz and we dont. in a way. services launch just as you described there.

config/flags_terms.go
Waldz: Tequilapi terms API can have more granular flags, but cli command should be more user friendly and have only `--accepted-terms-of-service`

config/flags_terms.go
```diff
@@ -0,0 +1,64 @@
...
+	// FlagTermsProviderAgree marks that provider TOS has been agreed.
+	FlagTermsProviderAgree = cli.BoolFlag{
+		Name:  "terms-agreed-provider",
+		Usage: "Provider terms agreed",
+	}
+
+	// FlagTermsVersion is the current agreed TOS version.
+	FlagTermsVersion = cli.StringFlag{
+		Name:  "terms-agreed-version",
+		Usage: "Current agreed terms version",
```
Waldz, line 44: Why version flag needed for user to input? it can be saved from `CurrentTermsVersion`

metadata/terms.go
Waldz: As we don't embed terms library to node, but rather show link to https://github.com/mysteriumnetwork/terms
Decided to embed just terms version instead, to known which terms he/she was reading when accepting terms.

Waldz (changes requested):

config/flags_terms.go
tomasmik: @Waldz  I made these flags `hidden` to make it more obvious that the user doesn't have to pass them.

I basically described them because of a single reason, it's easier to read the config if they're described and set.
User only passes the `--accepted-terms-of-service` and we update these flags for him.
While he does have the option to pass these flags himself, he does not have to.

This is done for the reasons stated in the PR description: so that we can parse the config easily, know which flags to check and not have to ask tequila anything. This makes it easier to control and check TOS because we dont have to request them from the API. Sometimes we have to check before the API is even started which makes things even worse.

Another way to do this, would be to define custom config consts somewhere and store them, parsing the config in that regard, though I feel like that would be even more confusing and obscure.

vkuznecovas (approved):

soffokl (approved):

chompomonim (approved):

Issue #2784, opened 2020-10-29
Pilvytis in tequilaAPI and MobileNode


tomasmik:
We need to support Pilvytis API calls in node when it comes to creating new transactions and pinging for a transaction status:

- [x] Extend tequilaAPI
- [x] Extend myst cli
- [ ] Extend mobileNode

PR #2895, opened 2020-12-03
Mobile payment bindings' fixes

tadaskay:
Finalizes mobile bindings for pilvytis.
- Fix gobind-incompatible types (change to JSON serialized as []byte where it's not possible to use compatible types)
- Fix a bug where orders are not updated by the status tracker (wrong reference used)
- Change OrderStatusUpdatedEvent to OrderUpdatedEvent (PayAmount/PayCurrency may come back later as a callback)
- Added error handling for uint64 → int64 shrinking (not likely to be reached, but still a possibility/ gobind supports only signed types https://godoc.org/golang.org/x/mobile/cmd/gobind#hdr-Type_restrictions)

Fixes: #2784

### DIFF
--- a/mobile/mysterium/entrypoint.go
+++ b/mobile/mysterium/entrypoint.go
@@ -22,22 +22,22 @@ import (
 	"encoding/json"
 	"math/big"
 	"path/filepath"
+	"strconv"
 	"time"
 
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/mysteriumnetwork/node/config"
 	"github.com/mysteriumnetwork/node/core/connection/connectionstate"
-	"github.com/mysteriumnetwork/node/pilvytis"
-	"github.com/pkg/errors"
-	"github.com/rs/zerolog/log"
-
 	"github.com/mysteriumnetwork/node/core/port"
 	"github.com/mysteriumnetwork/node/core/state"
 	"github.com/mysteriumnetwork/node/identity/registry"
+	"github.com/mysteriumnetwork/node/pilvytis"
 	wireguard_connection "github.com/mysteriumnetwork/node/services/wireguard/connection"
 	"github.com/mysteriumnetwork/node/session/pingpong"
 	"github.com/mysteriumnetwork/node/session/pingpong/event"
 	"github.com/mysteriumnetwork/payments/crypto"
+	"github.com/pkg/errors"
+	"github.com/rs/zerolog/log"
 
 	"github.com/rs/zerolog"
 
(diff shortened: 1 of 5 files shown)


mobile/mysterium/entrypoint.go
```diff
@@ -622,15 +623,36 @@ func (mb *MobileNode) HealthCheck() *HealthCheckData {
...
-func (mb *MobileNode) RegisterOrderStatusChangeCallback(cb OrderStatusChangedCallback) {
-	_ = mb.eventBus.SubscribeAsync(pilvytis.AppTopicOrderStatusChanged, func(e pilvytis.AppEventOrderStatusChanged) {
-		cb.OnChange(e.ID, string(e.Status))
+// RegisterOrderUpdatedCallback registers OrderStatusChanged callback.
+func (mb *MobileNode) RegisterOrderUpdatedCallback(cb OrderUpdatedCallback) {
+	_ = mb.eventBus.SubscribeAsync(pilvytis.AppTopicOrderUpdated, func(e pilvytis.AppEventOrderUpdated) {
+		payload := OrderUpdatedCallbackPayload{}
+		id, err := shrinkUint64(e.ID)
+		if err != nil {
+			log.Err(err).Send()
```
tomasmik, line 644: Maybe it's worth to also point out the level `Error()` and add some message so it's more clear whats going when reading logs.

pilvytis/order_status_tracker.go
```diff
@@ -69,9 +85,9 @@ func (t *StatusTracker) getOrCreate(id uint64, identityAddress string) *OrderSum
 			return &t.orders[i]
 		}
 	}
-	newEntry := OrderSummary{ID: id, IdentityAddress: identityAddress}
-	t.orders = append(t.orders, newEntry)
-	return &newEntry
+	newOrder := OrderSummary{ID: id, IdentityAddress: identityAddress}
+	t.orders = append(t.orders, newOrder)
+	return &t.orders[len(t.orders)-1]
```
tomasmik, line 90: Can't we just return `&newOrder` here?

pilvytis/order_status_tracker.go
```diff
@@ -124,9 +140,21 @@ func (t *StatusTracker) update() error {
 		}
 		for _, newOrder := range newOrders {
 			order := t.getOrCreate(newOrder.ID, newOrder.Identity)
+			changed := false
 			if order.Status != newOrder.Status {
 				order.Status = newOrder.Status
-				t.eventBus.Publish(AppTopicOrderStatusChanged, AppEventOrderStatusChanged{*order})
+				changed = true
+			}
```
tomasmik, line 147: nitpick: Maybe move these 3 changed checks in to a single func which would then also be tested if you wanted to? Code would also read nicer imo.

mobile/mysterium/entrypoint.go
```diff
@@ -642,20 +664,63 @@ type CreateOrderRequest struct {
 	Lightning       bool
 }
 
+type OrderResponse struct {
```
tomasmik, line 667: Linter will go :angry: if these aren't commented

pilvytis/order_status_tracker.go
```diff
@@ -139,6 +167,24 @@ func (t *StatusTracker) update() error {
 	return nil
 }
 
+func strEqual(s1, s2 *string) bool {
+	if s1 == nil {
```
tomasmik, line 171: Maybe something like this would read better? https://play.golang.org/p/DuMx0-khcMX

tomasmik (approved):

vkuznecovas (approved):

pilvytis/order_status_tracker.go
tadaskay: No :(

mobile/mysterium/entrypoint.go
tadaskay: `log.Err(err)` is a shorthand for `log.Error().Err(err)`

pilvytis/order_status_tracker.go
tadaskay: Done

PR #2896, opened 2020-12-03
Return settlement fees with settlement history

mdomasevicius:


vkuznecovas (approved):

tequilapi/contract/transactor.go
```diff
@@ -153,6 +157,7 @@ func NewSettlementDTO(settlement pingpong.SettlementHistoryEntry) SettlementDTO
 		Beneficiary:    settlement.Beneficiary.Hex(),
 		Amount:         settlement.Amount.Uint64(),
```
vkuznecovas, line 158: this is a bug, you'll exceed the uints rather quick

tequilapi/contract/transactor.go
```diff
@@ -153,6 +157,7 @@ func NewSettlementDTO(settlement pingpong.SettlementHistoryEntry) SettlementDTO
...
 		SettledAt:      settlement.Time.Format(time.RFC3339),
+		Fees:           fees.Uint64(),
```
vkuznecovas, line 160: this is a bug, you'll exceed the uints rather quick

vkuznecovas (changes requested):

tequilapi/contract/transactor.go
vkuznecovas: 19 myst and you overflow

tequilapi/contract/custom_json_types.go
```diff
@@ -0,0 +1,36 @@
+package contract
+
+import (
+	"fmt"
+	"math/big"
+)
+
+type BigInt struct {
+	big.Int
```
tomasmik, line 9: If you use `*big.Int` you'll lose on some of the cool things it [implements](https://golang.org/pkg/math/big/): https://play.golang.org/p/0RrQwVdgGlb

tequilapi/contract/custom_json_types.go
```diff
@@ -0,0 +1,32 @@
...
+	return []byte(b.String()), nil
+}
+
+func (b *BigInt) UnmarshalJSON(raw []byte) error {
+	if string(raw) == "null" {
+		return nil
+	}
+
+	var t big.Int
+	_, ok := t.SetString(string(raw), 10)
```
tomasmik, line 26: You could also do something like this `t, ok := new(big.Int).SetString(..)`.

vkuznecovas (approved):

tomasmik (approved):

PR #2901, opened 2020-12-04
Dns map check

vkuznecovas:
Noticed a mismatch between the given DNS entries and actual IPs so wrote a check if we had any more.

Question is, do we want it to be checked by default by CI?

```
➜  node git:(dns-map-check) ✗ time mage CheckDNSMaps
mage CheckDNSMaps  8.91s user 7.93s system 448% cpu 3.752 total
```

### DIFF
--- a/ci/check/check.go
+++ b/ci/check/check.go
@@ -20,6 +20,7 @@ package check
 import (
 	"errors"
 	"fmt"
+	"net"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


tomasmik (approved):
> Question is, do we want it to be checked by default by CI?

I would say - Yes.

soffokl (approved):

PR #2920, opened 2020-12-09
Improve cli connect command

tomasmik:
Final PR for: https://github.com/mysteriumnetwork/node/pull/2894

- We now actually validate given service.
- Removed the `new` option for consumerID as creating an identity is not enough to actually be able to connect.
- Added a silent auto unlock.

### DIFF
--- a/cmd/commands/cli/command.go
+++ b/cmd/commands/cli/command.go
@@ -335,6 +335,11 @@ func (c *cliApp) connect(argsString string) {
 
 	consumerID, providerID, serviceType := args[0], args[1], args[2]
 
+	if !services.IsTypeValid(serviceType) {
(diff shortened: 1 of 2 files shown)


soffokl (approved):

cmd/commands/cli/command.go
```diff
@@ -745,3 +746,12 @@ func parseStartFlags(serviceType string, args ...string) (services.StartOptions,
 
 	return services.GetStartOptions(serviceType)
 }
+
+func validServiceType(s string) bool {
```
tadaskay, line 750: I would rather include this validation in `services` package - it's more relevant there than in CLI package.

tadaskay (approved):

Issue #2884, opened 2020-12-02
Currency ticker should be configurable


chompomonim:
In testnet we should show everywhere `MYSTT` and in betanet `MYST`. It would be good to have currency ticker in config and use it from there everywhere in user interfaces. 

Fix consistency of displaying currencies in WebUI and Desktop.

Apps should fetch currency identifier (may differ on different networks) and visualise it.

PR #2925, opened 2020-12-10
Show currency depending on the network

tomasmik:
* Console commands and logs will now show `MYSTT` if running in testnet
* Nodes will also report their currency as `MYSTT` in proposals
* Money struct was updated to support these changes, allowing backwards comparability with optional currency param. This allows tests to not break and work as expected.

Improves: https://github.com/mysteriumnetwork/node/issues/2884

### DIFF
--- a/money/money.go
+++ b/money/money.go
@@ -20,6 +20,8 @@ package money
 import (
 	"fmt"
 	"math/big"
+
+	"github.com/mysteriumnetwork/node/config"
 )
 
 // Money holds the currency type and amount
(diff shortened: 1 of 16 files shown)


money/money.go
```diff
@@ -29,13 +31,25 @@ type Money struct {
...
+		m.Currency = currency[0]
+	} else {
+		m.Currency = Currency(config.GetString(config.FlagDefaultCurrency))
+	}
+
+	return m
 }
 
-// String converts struct to string
+// String converts Money struct in to a string
```
soffokl, line 51:
```suggestion
// String converts Money struct into a string
```

soffokl (approved):

money/money.go
```diff
@@ -29,13 +31,25 @@ type Money struct {
...
 
 // NewMoney returns a new instance of Money.
-// The money is a representation of myst in a uint64 form, with the decimal part expanded.
-// This means, that one myst is equivalent to 10 0000 000.
-func NewMoney(amount *big.Int, currency Currency) Money {
-	return Money{amount, currency}
+// Expected `amount` value for 1 myst is equal to 1_000_000_000_000_000_000.
+// It also allows for an optional currency value to be passed,
+// if one is not passed, default config value is used.
+func NewMoney(amount *big.Int, currency ...Currency) Money {
```
tadaskay, line 37: Since the package is `money`, may I suggest simply naming it `money.New()`?

tadaskay (approved):

money/money.go
tomasmik: Of couse. It was like this and I didn't think about changing it. Nice.

Issue #2912, opened 2020-12-09
Session data reports incorrect transfer


soffokl:
In some cases, consumer reports incorrect session data transfer, after some time of missing data:
![image](https://user-images.githubusercontent.com/8612618/101584140-241d1580-3a07-11eb-8931-26408cc0c54b.png)
![image](https://user-images.githubusercontent.com/8612618/101584148-2da67d80-3a07-11eb-9353-8688c9703a93.png)
![image](https://user-images.githubusercontent.com/8612618/101584161-36974f00-3a07-11eb-828f-ede3949769a5.png)

soffokl:
We are now reporting both provider and consumer stats. This will help us to understand such misbehavior. 
![image](https://user-images.githubusercontent.com/8612618/102038902-a7b57880-3df2-11eb-86f6-9f81c63ec02d.png)

soffokl:
Intensive upload from the consumer side could generate too much traffic, which never reaches the provider since the transport is UDP. It's counted by the consumer, but the provider charges only for traffic served by him. We are now showing stats in MMN based on the provider stats, not a consumer. 

https://user-images.githubusercontent.com/8612618/104402361-495df000-5580-11eb-8939-2fed0a5ffd8b.mp4

PR #2929, opened 2020-12-12
Send NAT mapping events for Public IP

soffokl:
Related to the #2912


soffokl:
I'll merge this without review. Let me know if I need to fix something after.

PR #2943, opened 2020-12-18
Payment Order Options gateway

mdomasevicius:


tequilapi/endpoints/pilvytis.go
```diff
@@ -202,11 +203,37 @@ func (e *pilvytisEndpoint) GetPaymentOrderCurrencies(w http.ResponseWriter, r *h
...
+//     schema:
+//       "$ref": "#/definitions/ErrorMessageDTO"
+func (e *pilvytisEndpoint) GetPaymentOrderOptions(w http.ResponseWriter, r *http.Request, params httprouter.Params) {
+	resp, err := e.api.GetPaymentOrderOptions()
+	if err != nil {
+		utils.SendError(w, err, http.StatusInternalServerError)
+		return
+	}
+
+	utils.WriteAsJSON(resp, w)
```
tomasmik, line 228: I think it's typical for tequila to returns objects which are described in the `contract` package.

While sometimes it can be repetitive it sort of protects us from leaking too much information as they act like "public" structs, not only that but they also keep things like  swagger generation in a single place and make developing the tequila `client` or using these endpoints in for example in `cli` a lot cleaner because we dont have to actually import every package in the whole project.

See this for example: https://github.com/mysteriumnetwork/node/pull/2943/files#diff-437c90b69c7d2e457bcc7e0afaacec573c4622bac9d913441ecedd8a323eacd3R176

tomasmik (changes requested):

tequilapi/contract/PaymentOrderOptions.go
```diff
@@ -0,0 +1,17 @@
...
+import "github.com/mysteriumnetwork/node/pilvytis"
+
+// PaymentOrderOptions represents pilvytis payment order options
+// swagger:model PaymentOrderOptions
+type PaymentOrderOptions struct {
+	Minimum   float64   `json:"minimum"`
+	Suggested []float64 `json:"suggested"`
+}
+
+func ToPaymentOrderOptions(poo *pilvytis.PaymentOrderOptions) *PaymentOrderOptions {
```
tomasmik, line 12: Linter might get angry here because no comment provided

tomasmik (approved):

tequilapi/contract/PaymentOrderOptions.go
mdomasevicius: :smirk_cat: :gun:

vkuznecovas (approved):

Issue #2948, opened 2020-12-21
Unexpected config workflow after installing using apt


tomasmik:
Seems that after installing our node from the testnet 2.0 PPA something is not right with the config.
Config is being written to in `/etc/mysterium-node/config.toml ` while on start up it's read from `/root/.mysterium/config.toml`.

This makes it impossible to actually save any flags like `--agreed-terms-and-conditions`

Edit#1: 
After further investigation it seems that this happens because node is indeed writing to `/etc/mysterium-node` but commands are launched by the user so his home dir is picked up and config is read from there.

For example when you agree to TOS, we write the agreement using tequila API (which writes to /etc/...) but we read it using the `LoadUserConfig` function which loads from user home dir.
This is not apparent when developing because you're building and running node and node commands yourself as the same user.

Edit#2:
Two issues  identified:
- [x] TOS agreement during `apt install myst` has no affect later on (e.g. opening webUI requires user to agree again) https://github.com/mysteriumnetwork/node/pull/2954
- [x] Address the double config issue, either combining them in to one, or making CLI tools depens on tequila for config. For example now when agreeing TOS in CLI tool it is done so by editing config of the node via tequilapi (`/etc/mysterium-node/config.toml`) but when launching CLI tool it is read from `~/.mysterium/config.tml` which is a problem https://github.com/mysteriumnetwork/node/pull/2952
- [x] Fix the config show command https://github.com/mysteriumnetwork/node/pull/2955

Edit#3:
More places are missing the tequila API config fix:
- [x] account / connection commands
- reset command https://github.com/mysteriumnetwork/node/issues/2971

PR #2949, opened 2020-12-22
TOS check directly via config

mdomasevicius:
Check - https://github.com/mysteriumnetwork/node/issues/2948

These changes aim to patch faulty workflow as CLI tool reads config from a predefined path but updates it via tequilapi.
I am not sure this is right approach at all but @tomasmik argued that we should have a possibility to launch CLI tool without backing NODE. 

My personal opinion would be to not allow that and follow the common idea that it should only work if the NODE is up. Like with desktop app, webUI and android.

@Waldz you seem to know about CLI philosophy so your opinion is much appreciated.

### DIFF
--- a/cmd/commands/cli/command.go
+++ b/cmd/commands/cli/command.go
@@ -102,6 +102,7 @@ type cliApp struct {
 const redColor = "\033[31m%s\033[0m"
 const identityDefaultPassphrase = ""
 const statusConnected = "Connected"
+const youMustAgree = "You must agree with provider and consumer terms of use in order to use this command"
 
 var versionSummary = metadata.VersionAsSummary(metadata.LicenseCopyright(
 	"type 'license --warranty'",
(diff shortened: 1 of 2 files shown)


config/config.go
```diff
@@ -376,3 +376,7 @@ func GetFloat64(flag cli.Float64Flag) float64 {
 func AppTopicConfig(configKey string) string {
 	return "config:" + configKey
 }
+
+func UpdateTerms() {
```
Waldz, line 380: leftover?

cmd/commands/cli/command.go
```diff
@@ -102,6 +102,7 @@ type cliApp struct {
 const redColor = "\033[31m%s\033[0m"
 const identityDefaultPassphrase = ""
 const statusConnected = "Connected"
+const youMustAgree = "You must agree with provider and consumer terms of use in order to use this command"
```
vkuznecovas, line 105: `
var ErrAgreementNeeded = errors.New("You must agree with provider and consumer terms of use in order to use this command")
`
and return the error where appropriate?

vkuznecovas (changes requested):

PR #2952, opened 2020-12-29
Load CLI config from node

mdomasevicius:
Load config from node via tequilapi for CLI


config/remote/remote_config.go
```diff
@@ -0,0 +1,130 @@
+/*
+ * Copyright (C) 2017 The "MysteriumNetwork/node" Authors.
```
tomasmik, line 2:
```suggestion
 * Copyright (C) 2020 The "MysteriumNetwork/node" Authors.
```

config/remote/remote_config.go
```diff
@@ -0,0 +1,130 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package remote_config
```
tomasmik: I think this should be `remote`.

Or even better, dont nest these packages, and just add a `Remote` object to the `config` package.
Calling something like `confing.NewRemote()` seems fairly reasonable.

config/remote/remote_config.go
```diff
@@ -0,0 +1,130 @@
...
+
+	"github.com/spf13/cast"
+
+	"github.com/mysteriumnetwork/node/tequilapi/client"
+
+	"github.com/mysteriumnetwork/node/config"
+	"github.com/rs/zerolog/log"
+)
+
+var Config = newRemoteConfig()
```
tomasmik, line 35: I am not a big fan of these sort of global variables. Whats the issue with exposing `NewRemoteConfig` and allowing the user to manipulate it how ever he wants to?

It's just much easier to read/edit when things are not hidden in packages. It's also not very static, someone calling `Get` while you're calling `RefreshRemoteConfig` might get nasty.

config/remote/remote_config.go
```diff
@@ -0,0 +1,130 @@
...
+
+type remoteConfig struct {
+	config map[string]interface{}
+}
+
+func newRemoteConfig() *remoteConfig {
+	return &remoteConfig{}
+}
+
+func RefreshRemoteConfig(client *client.Client) error {
```
tomasmik, line 46: Hide the `*client.Client` under an interface and dont import the tequilapi/client here. This would make it more flexible and testable if we ever wanted to do either one of those.

tequilapi/client/client.go
```diff
@@ -734,3 +736,24 @@ func (client *Client) UpdateTerms(obj contract.TermsRequest) error {
...
+
+	bodyBytes, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return nil, err
+	}
+
+	var unmarshaled map[string]interface{}
+	err = json.Unmarshal(bodyBytes, &unmarshaled)
+
+	config := unmarshaled["data"].(map[string]interface{})
```
tomasmik, line 756: I think there are two ways this can panic:
1. If we get an error on json.Unmarshal this will panic as there might be no such thing as `data` resulting in a nil (check for error before trying to cast)
2. If for some reason structure changes and `data` is no longer a `map[string]interface{}` (`config, ok := ..` )

cmd/commands/cli/command.go
```diff
@@ -27,16 +27,18 @@ import (
 	"strings"
 	"time"
 
+	"github.com/mysteriumnetwork/node/cmd/commands"
+
+	remote_config "github.com/mysteriumnetwork/node/config/remote"
```
tomasmik, line 32: Lets not give packages names if we dont have to. If a package requires a rename, we should change the actual name while/if we can.

cmd/commands/connection/command.go
```diff
@@ -22,12 +22,12 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/mysteriumnetwork/node/cmd/commands"
+
```
tomasmik, line 26: Im gonna be a little annoying here but did you configure your linter correctly? This empty line should not be here.

Maybe go to settings and check if `goimports` is enabled?

config/remote/remote_config.go
```diff
@@ -0,0 +1,134 @@
...
+	return rc.GetInt64(flag.Name)
+}
+
+// GetInt64 returns config value as int64.
+func (rc *remoteConfig) GetInt64(key string) int64 {
+	return cast.ToInt64(rc.Get(key))
+}
+
+// TequilAPIAddress - wil resolve default tequilapi address or from flag if one is provided
+func TequilAPIAddress(ctx *cli.Context) string {
```
tomasmik, line 115: Both of these tequila funcs have almost nothing to do with `remote config` so I would argue that this is not the place for them. This seems more or less the part that the cli should do.

cmd/commands/util.go
```diff
@@ -0,0 +1,43 @@
+/*
```
tomasmik, line 1: I am sort of torn about whatever this is the correct place to put this.

I am sort of leaning towards placing this somewhere where the `clio` but I am not sure. Im putting it out there, maybe you'll think of something.

cmd/commands/util.go
```diff
@@ -0,0 +1,43 @@
...
+package commands
+
+import (
+	"fmt"
+
+	"github.com/urfave/cli/v2"
+
+	"github.com/mysteriumnetwork/node/cmd/commands/cli/clio"
+	remote_config "github.com/mysteriumnetwork/node/config/remote"
+	tequilapi_client "github.com/mysteriumnetwork/node/tequilapi/client"
```
tomasmik, line 27:
```suggestion
	"github.com/mysteriumnetwork/node/tequilapi/client"
```

cmd/commands/util.go
```diff
@@ -0,0 +1,43 @@
...
+
+// InitClientAndConfig - initializes and returns a pointer to tequilapi client - also fetches config using it
+func InitClientAndConfig(ctx *cli.Context) (*tequilapi_client.Client, error) {
+	address := remote_config.TequilAPIAddress(ctx)
+	port := remote_config.TequilAPIPort(ctx)
+	client := tequilapi_client.NewClient(address, port)
+
+	err := remote_config.RefreshRemoteConfig(client)
+	if err != nil {
+		clio.Error("failed to connect to node via url:", address+":"+fmt.Sprint(port))
```
tomasmik, line 38: This might just be my personal opinion, but you either log an error or return it. You dont do both, so that the caller doesn't need to be confused whether he actually needs to logs or do something. 🤷🏻

tomasmik (changes requested):

cmd/commands/connection/command.go
mdomasevicius: image

cmd/commands/util.go
mdomasevicius: Your logging configuration eat's up `log.Error()...` at the bottom of the stack for `cli` command at least.

config/remote/remote_config.go
mdomasevicius: It is called a "singleton pattern" and I am not a fan of them too.
But since I wan't to get a stable MVP the idea was to copy past in a similar fashion and then test the solution until we are satisfied it is stable. However maybe this is not a big change.

tequilapi/client/client.go
```diff
@@ -734,3 +736,35 @@ func (client *Client) UpdateTerms(obj contract.TermsRequest) error {
...
+	if err != nil {
+		return nil, err
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode < 200 || resp.StatusCode > 299 {
+		return nil, fmt.Errorf("fetching config failed with status: %d", resp.StatusCode)
+	}
+
+	bodyBytes, err := ioutil.ReadAll(resp.Body)
```
tomasmik, line 752: I think you can [safely reuse](https://play.golang.org/p/FB9g1BBlxPW) the `parseResponseJson` function we already got no?

config/remote/config.go
```diff
@@ -0,0 +1,115 @@
...
+)
+
+// Config - remote config struct
+type Config struct {
+	client *client.Client
+	config map[string]interface{}
+}
+
+// NewRemoteConfig - new remote config instance
+func NewRemoteConfig(client *client.Client) (*Config, error) {
```
tomasmik, line 41:
```suggestion
func NewConfig(client *client.Client) (*Config, error) {
```

cmd/commands/cli/clio/tequilapi.go
```diff
@@ -0,0 +1,63 @@
...
+
+// NewTequilApiClient - initializes and returns a pointer to tequilapi client - also fetches config using it
+func NewTequilApiClient(ctx *cli.Context) (*tequilapi_client.Client, error) {
+	address := TequilAPIAddress(ctx)
+	port := TequilAPIPort(ctx)
+	client := tequilapi_client.NewClient(address, port)
+
+	_, err := client.Healthcheck()
+	if err != nil {
+		Error("failed to connect to node via url:", address+":"+fmt.Sprint(port))
```
tomasmik, line 37:
```suggestion
		Error(fmt.Sprintf("failed to connect to node via url: %s:%d", address, port))
```

tomasmik (approved):
Look good to me. 👍🏻

Waldz (approved):

PR #2954, opened 2020-12-29
Insert ToS agreement when installing deb package

tomasmik:
Updates: #2948

This is a somewhat fragile way to insert the ToS agreement in the config if user is installing the deb package.

While this works any suggestions are welcome.

### DIFF
--- a/cmd/commands/service/command.go
+++ b/cmd/commands/service/command.go
@@ -165,6 +165,7 @@ func (sc *serviceCommand) tryRememberTOS(ctx *cli.Context, errCh chan error) {
 		for i := 0; i < 5; i++ {
 			if err := sc.tequilapi.UpdateTerms(contract.TermsRequest{
 				AgreedProvider: &t,
+				AgreedConsumer: &t,
 				AgreedVersion:  metadata.CurrentTermsVersion,
 			}); err == nil {
 				return


chompomonim (approved):

mdomasevicius (approved):

debian/postinst
```diff
@@ -58,6 +60,24 @@ function ensure_paths {
     fi
 }
 
+function insert_tos_agreement {
+    touch $OS_DIR_CONFIG/config.toml
+
+    # Delete the terms part from config if it exists
+    if grep -Fxq "[terms]" $OS_DIR_CONFIG/config.toml
+    then
+        sed -i -r '/\[terms\]/,/\[/{//!d}' $OS_DIR_CONFIG/config.toml
```
Waldz, line 69: Not sure it's it's good idea to modify terms after each install

Waldz (changes requested):
I suggest to for us to groom how fix issue of double term (sometimes triple, in UI)

tomasmik:
> I suggest to for us to groom how fix issue of double term (sometimes triple, in UI)

Maybe you're misunderstanding the idea behind this. When installing the `.deb` you're asked to `Agree to terms of use` during the installation, which you have to click `yes` or else installation doesn't continue. So if you already agreed during the initial set up, why not just insert it in the config? This is what this pull request tries to solve. See the issue here: https://github.com/mysteriumnetwork/node/issues/2948
The double triple, w/e terms issue is half addressed here: https://github.com/mysteriumnetwork/node/pull/2952 (web ui still needs to be updated).

debian/postinst
tomasmik: We can debate this for sure. We could only insert the `terms` agreement if no such config variable exists.

soffokl (approved):

vkuznecovas (approved):

Waldz (approved):

Issue #2956, opened 2020-12-30
Use existing beneficiary address during auto-upgrade into Testnet 2.0


chompomonim:
In pre 0.40 users were saving beneficiary (bounty payout) addresses in MMN database. In testnet2 for bounty payouts we're going to use beneficiary address stored in blockchain (in hermes smart contract).

**Problem**
Quite significant amount of our exit nodes will use auto-upgrade for their node. At the moment their nodes will be upgraded and registered into new network automatically, but as beneficiary we're using their topup addresses instead of that bounty address from MMN. As result, some of them will forget to change that address and will receive bounties into not accessible address on blockchain.

**Suggested solution**

During node upgrade we should detect that this is existing node and should use his current payout address as beneficiary during identity registration.

PR #2957, opened 2020-12-30
Migrate bounty payout address to BC beneficiary

Waldz:
Closes: #2956

This checks does free auto-registration to BC only if identity used to have payout address for bounty payments.
Also migrates found payout address to identity beneficiary in BC.

### DIFF
--- a/identity/registry/provider_registrar.go
+++ b/identity/registry/provider_registrar.go
@@ -28,6 +28,8 @@ import (
 	"github.com/mysteriumnetwork/node/core/service/servicestate"
 	"github.com/mysteriumnetwork/node/eventbus"
 	"github.com/mysteriumnetwork/node/identity"
+	"github.com/mysteriumnetwork/node/market/mysterium"
+	"github.com/mysteriumnetwork/node/requests"
 	"github.com/pkg/errors"
 	"github.com/rs/zerolog/log"
 )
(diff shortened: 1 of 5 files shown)


identity/registry/provider_registrar.go
```diff
@@ -170,27 +186,41 @@ func (pr *ProviderRegistrar) handleEvent(qe queuedEvent) error {
...
+			log.Info().Msgf("provider %q not eligible for auto registration, will require manual registration", id.Address)
+			return nil
+		}
+	}
+
+	// This part migrates bounty payout address to BC beneficiary
+	payout, err := pr.mysteriumAPI.GetPayoutInfo(id, pr.signerFactory(id))
+	if err != nil {
+		if errHTTP, ok := err.(*requests.ErrorHTTP); ok && errHTTP.Code == 404 {
+			log.Info().Msgf("provider %q not have payout address for auto registration, will require manual registration", id.Address)
```
tomasmik, line 208: Do we have to return here and not continue the registration? If I'm understanding this correctly new nodes will no longer auto register if we do this.

Waldz: This is how i recognise newly created identity, which have to be on-boarded manually via UI

tomasmik: Ye it seems like you can do that and it works. Nice

tomasmik (approved):

chompomonim:
If I understand correctly, you expect that there will be input in api database about payout addresses. I think we may simply copy them from `testnet.api` into `testnet2.api`.

mdomasevicius (approved):

chompomonim (approved):

identity/registry/provider_registrar_test.go
```diff
@@ -222,3 +227,16 @@ func (mt *mockTransactor) RegisterIdentity(id string, stake, fee *big.Int, benef
 func (mt *mockTransactor) CheckIfRegistrationBountyEligible(identity identity.Identity) (bool, error) {
```
chompomonim, line 227: Had we already existing such function before (not mock)? Can't find one in this PR.

Waldz: Yes we do - for free registration (with tokens i guess). But it's enabled for Betanet, but not Testnet2

soffokl (approved):

PR #2969, opened 2021-01-04
Fix account and connection config usage

tomasmik:
Both `account` and `connection` commands now use the tequilaAPI provided config.

Updates: https://github.com/mysteriumnetwork/node/issues/2948

### DIFF
--- a/cmd/commands/connection/command.go
+++ b/cmd/commands/connection/command.go
@@ -24,6 +24,7 @@ import (
 
 	"github.com/mysteriumnetwork/node/cmd/commands/cli/clio"
 	"github.com/mysteriumnetwork/node/config"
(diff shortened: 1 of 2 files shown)


soffokl (approved):

Waldz (approved):

cmd/commands/connection/command.go
```diff
@@ -238,7 +249,7 @@ func (c *command) up(ctx *cli.Context) {
 		DNS:               connection.DNSOptionAuto,
 		DisableKillSwitch: false,
 	}
-	hermesID := config.GetString(config.FlagHermesID)
+	hermesID := c.cfg.GetStringByFlag(config.FlagHermesID)
```
Waldz, line 252: `hermesID` is optional, and defaults to this value in backend

Issue #2971, opened 2021-01-05
Credential reset command not working in production


tomasmik:
Crendential reset command `myst reset --tequilapi` does not working because of the double config.

It's a double problem though. To get information on where the DB is in order to reset credentials we have to have a running node and request the current config, but we can't reset credentials if a node is running.

Possible solution to this would be to add a tequila endpoint which allows to reset credentials.

Edit:
After grooming this, we've decided to move the password to a new file outside the DB.
Required node-ui improvements are continued in: https://github.com/mysteriumnetwork/node/issues/2979

PR #2978, opened 2021-01-06
Node UI password rework

tomasmik:
* Reworked how node password is stored: we now store a password in a separate file on the users machine
* Added a migration to migrate the current password that user has set to be stored in a file (works for testnet and testnet2)
* Fixed the reset command, it now resets users password to default node configured password
* We no longer display debug information when reset command is run

Current config workflow has led to some [unwanted workflows ](https://github.com/mysteriumnetwork/node/pull/2978/files#diff-2626c43c402014222bb0a6bbdd214e37a96687df78dee668b5d20fd304bed007R49)which could be remedied only by improving how we handle the config.
Closes: #2971

### DIFF
--- a/core/auth/credentials.go
+++ b/core/auth/credentials.go
@@ -18,53 +18,52 @@
 package auth
 
 import (
+	"bufio"
 	"errors"
 	"fmt"
+	"os"
+	"path/filepath"
 
 	"github.com/mysteriumnetwork/node/config"
-	"github.com/mysteriumnetwork/node/core/storage"
 	"github.com/rs/zerolog/log"
 	"golang.org/x/crypto/bcrypt"
 )
 
-// Storage for Credentials.
-type Storage interface {
-	GetValue(bucket string, key interface{}, to interface{}) error
-	SetValue(bucket string, key interface{}, to interface{}) error
+// CredentialsManager verifies/sets user credentials for web UI.
+type CredentialsManager struct {
+	fileLocation string
 }
 
-const credentialsDBBucket = "app-credentials"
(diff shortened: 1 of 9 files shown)


core/auth/credentials.go
```diff
@@ -18,93 +18,114 @@
...
+)
 
-// NewCredentials instance.
-func NewCredentials(username, password string, db Storage) *Credentials {
-	return &Credentials{
-		username: username,
-		password: password,
-		db:       db,
+// NewCredentialsManager returns given a password file directory
+// returns a new credentials manager,which can be used to validate or alter
```
soffokl, line 47:
```suggestion
// returns a new credentials manager, which can be used to validate or alter
```

core/node/options_directory.go
```diff
@@ -40,27 +40,44 @@ type OptionsDirectory struct {
...
-		Storage:  filepath.Join(dataDir, networkSubdir, "db"),
+		Storage:  GetOptionsDirecotryDB(networkSubdir),
 		Keystore: filepath.Join(dataDir, "keystore"),
 		Script:   config.GetString(config.FlagScriptDir),
 		Runtime:  config.GetString(config.FlagRuntimeDir),
 	}
 }
 
+// GetOptionsDirecotryDB returns a database directory given a networkSubdir.
+func GetOptionsDirecotryDB(networkSubdir string) string {
```
vkuznecovas, line 76: spelling

vkuznecovas (changes requested):

soffokl (approved):

vkuznecovas (approved):

PR #3007, opened 2021-01-19
Linux supervisor

mdomasevicius:
Enable linux installation & daemon/client transport

### DIFF
--- a/supervisor/client/transport_linux.go
+++ b/supervisor/client/transport_linux.go
@@ -23,7 +23,7 @@ import (
 	"net"
 )
 
-const sock = "/var/run/myst.sock"
+const sock = "/run/myst.sock"
 
 func connect() (io.ReadWriteCloser, error) {
 	conn, err := net.Dial("unix", sock)
--- a/supervisor/client/transport_linux.go
+++ b/supervisor/client/transport_linux.go
@@ -18,10 +18,17 @@
 package client
 
 import (
-	"errors"
+	"fmt"
 	"io"
+	"net"
 )
 
+const sock = "/var/run/myst.sock"
+
 func connect() (io.ReadWriteCloser, error) {
-	return nil, errors.New("not implemented")
+	conn, err := net.Dial("unix", sock)
+	if err != nil {
+		return nil, fmt.Errorf("could not connect to the supervisor socket %s: %w", sock, err)
+	}
+	return conn, nil
 }
--- a/supervisor/daemon/transport/transport_linux.go
+++ b/supervisor/daemon/transport/transport_linux.go
@@ -17,10 +17,56 @@
 
 package transport
 
-import "errors"
+import (
+	"fmt"
+	"net"
+	"os"
+	"strconv"
+
+	"github.com/rs/zerolog/log"
+)
+
+const sock = "/var/run/myst.sock"
 
 // Start starts a listener on a unix domain socket.
 // Conversation is handled by the handlerFunc.
-func Start(handle handlerFunc, _ Options) error {
-	return errors.New("not implemented")
+func Start(handle handlerFunc, options Options) error {
+	if err := os.RemoveAll(sock); err != nil {
+		return fmt.Errorf("could not remove sock: %w", err)
+	}
+	l, err := net.Listen("unix", sock)
+	if err != nil {
+		return fmt.Errorf("error listening: %w", err)
+	}
+	numUid, err := strconv.Atoi(options.Uid)
+	if err != nil {
+		return fmt.Errorf("failed to parse uid %s: %w", options.Uid, err)
+	}
+	if err := os.Chown(sock, numUid, -1); err != nil {
+		return fmt.Errorf("failed to chown supervisor socket to uid %s: %w", options.Uid, err)
+	}
+	if err := os.Chmod(sock, 0700); err != nil {
+		return fmt.Errorf("failed to chmod supervisor socket: %w", err)
+	}
+	defer func() {
+		if err := l.Close(); err != nil {
+			log.Err(err).Msg("Error closing listener")
+		}
+	}()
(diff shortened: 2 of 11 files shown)


vkuznecovas (approved):

supervisor/daemon/transport/transport_linux.go
```diff
@@ -17,10 +17,56 @@
...
+		return fmt.Errorf("error listening: %w", err)
+	}
+	numUid, err := strconv.Atoi(options.Uid)
+	if err != nil {
+		return fmt.Errorf("failed to parse uid %s: %w", options.Uid, err)
+	}
+	if err := os.Chown(sock, numUid, -1); err != nil {
+		return fmt.Errorf("failed to chown supervisor socket to uid %s: %w", options.Uid, err)
+	}
+	if err := os.Chmod(sock, 0700); err != nil {
```
mdomasevicius, line 48: I am not sure what permissions should be applied here, but desktop app does not start unless it's `766` I believe

tomasmik: As I understand desktop app is running in a different user space, so ye `0700` would not work.

supervisor/install/install_linux.go
```diff
@@ -18,15 +18,108 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	t, err := template.New("unit-descriptor").Parse(description)
+	if err != nil {
+		return nil, err
+	}
+	buffer := new(bytes.Buffer)
+	err = t.Execute(buffer, map[string]string{
```
soffokl, line 85: Where this template used? It's just created and never reused or I'm missing something?

supervisor/install/install_linux.go
```diff
@@ -18,15 +18,108 @@
...
+
+func clean(d daemon.Daemon) error {
+	execAndLog(d.Stop)
+	return execAndLog(d.Remove)
+}
+
+func execAndLog(action func() (string, error)) error {
+	output, err := action()
+	if err != nil {
+		log.Info().Msgf("%s\t%s", output, err)
```
soffokl, line 110: Should it be an `Error` or `Warn` if `err != nil`?
```suggestion
		log.Error().Msgf("%s\t%s", output, err)
```

supervisor/install/install_linux.go
```diff
@@ -18,15 +18,108 @@
...
+		return err
+	}
+	log.Info().Msg(output)
+	return nil
 }
 
-// Uninstall installs service for linux. Not implemented yet.
+// Uninstall - remove supervisor daemon
```
soffokl, line 117:
```suggestion
// Uninstall removes supervisor daemon.
```

supervisor/daemon/wireguard/wginterface/interface_linux.go
```diff
@@ -18,20 +18,59 @@
...
+	"github.com/rs/zerolog/log"
+	"golang.zx2c4.com/wireguard/device"
+	"golang.zx2c4.com/wireguard/ipc"
 
 	"golang.zx2c4.com/wireguard/tun"
 )
 
 func createTunnel(requestedInterfaceName string) (tunnel tun.Device, interfaceName string, err error) {
-	return nil, requestedInterfaceName, errors.New("not implemented")
+	log.Info().Msg("createTunnel start")
```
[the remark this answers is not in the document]
mdomasevicius: forgot to remove, these were for debugging

supervisor/install/install_linux.go
```diff
@@ -18,15 +18,108 @@
...
+	log.Info().Msg("Cleaning up previous installation")
+	clean(dmn)
+
+	log.Info().Msg("Installing daemon")
+	err = execAndLog(func() (string, error) { return dmn.Install() })
+	if err != nil {
+		return err
+	}
+
+	err = execAndLog(dmn.Start)
```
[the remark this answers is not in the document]
mdomasevicius: This library prints return unix like output

```
output, err := action()
	if err != nil {
		log.Error().Msgf("%s\t%s", output, err)
		return err
	}
```

e.g.: `Starting Daemon        [ OK ]` 

So I figured that extra "Starting daemon" is not necessary

supervisor/install/install_linux.go
```diff
@@ -18,15 +18,108 @@
...
+	dmn, err := mystSupervisorDaemon(options)
+	if err != nil {
+		return err
+	}
+
+	log.Info().Msg("Cleaning up previous installation")
+	clean(dmn)
+
+	log.Info().Msg("Installing daemon")
+	err = execAndLog(func() (string, error) { return dmn.Install() })
```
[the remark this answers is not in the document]
mdomasevicius:
```console
supervisor/install/install_linux.go:62:22: cannot use dmn.Install (type func(...string) (string, error)) as type func() (string, error) in argument to execAndLog
```

it has different signature

image

supervisor/install/install_linux.go
mdomasevicius: yes, a bug - lost some changes with stashing

supervisor/client/transport_linux.go
```diff
@@ -18,10 +18,17 @@
 package client
 
 import (
-	"errors"
+	"fmt"
 	"io"
+	"net"
 )
 
+const sock = "/var/run/myst.sock"
```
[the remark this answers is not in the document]
mdomasevicius: Should this be aligned with `transport/transport_linux.go` `const sock = "/var/run/myst.sock"` ?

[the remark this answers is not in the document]
soffokl: Yep, it's kind of will work even without it, but `/var/run/` is legacy and just a symlink for compatibility for old systems.

soffokl (approved):

Issue #3022, opened 2021-01-19
Store beneficiary in database


chompomonim:
On Matic we're going to use auto-settlement into node runners "consumer channel" and then withdraw that funds into Mainnet ethereum. So in hermes channel we will store not final beneficiary but "consumer channel". We need to store withdrawal beneficiary separately in database.

if provider is using Matic, then his beneficiary (in Hermes contract) should be his own consumer channel address

When setting the beneficiary, save the beneficiary in the local bolt database. 

On ethereum chain, set the beneficiary with the provided address.
On matic, set the beneficiary to the consumer channel address before sending the settle with beneficiary request.

In all the places where we're exposing beneficiary address we should take it from the database instead of the bc like we do now.

PR #3031, opened 2021-01-20
Handle beneficiary in different chains

tomasmik:
L1 chain workflow didn't change. Introduced a way to handle L2 chain workflows, where we store beneficiary and the database instead.

Closes: https://github.com/mysteriumnetwork/node/issues/3022


metadata/chains.go
```diff
@@ -0,0 +1,52 @@
...
+			return true
+		}
+	}
+
+	return false
+}
+
+// IsL1 returns if a given chains belongs to level 2.
+func (c *chains) IsL2(chainID int64) bool {
+	for _, id := range c.l2 {
```
vkuznecovas, line 45: return !IsL1(chainID). You can then remove the l2 definitions in this file.

tequilapi/endpoints/transactor.go
```diff
@@ -292,8 +298,7 @@ func (te *transactorEndpoint) SettleWithBeneficiary(resp http.ResponseWriter, re
 		return
 	}
 
-	chainID := config.GetInt64(config.FlagChainID)
-	err = te.promiseSettler.SettleWithBeneficiary(chainID, identity.FromAddress(id), common.HexToAddress(req.Beneficiary), common.HexToAddress(req.HermesID))
```
vkuznecovas, line 296: This is incorrect. In both cases, we want to settle with beneficiary. 

In the case of l1 beneficiary settle, we settle with the beneficiary address provided by the user. This automatically serves as a save.

In the case of l2 beneficiary settle, we settle with the generated consumer channel address and save the provided address in the database.

While getting beneficiary addresses, for l2 we want to fetch it from db, for l1 we want to fetch it from BC.

vkuznecovas (changes requested):

tequilapi/endpoints/transactor.go
vkuznecovas: Perhaps we can leave the settlement here, and use the beneficiary handlers that you've made, except change the save method on the l1 handler to just do nothing.

cmd/di.go
```diff
@@ -914,6 +921,20 @@ func (di *Dependencies) bootstrapFirewall(options node.OptionsFirewall) error {
 	return nil
 }
 
+func (di *Dependencies) bootstrapBeneficiaryHandler(options node.Options) error {
+	var err error
+	di.BeneficiaryHandler, err = beneficiary.NewHandler(
```
soffokl, line 926: nitpicking
```suggestion
func (di *Dependencies) bootstrapBeneficiaryHandler(options node.Options) (err error) {
	di.BeneficiaryHandler, err = beneficiary.NewHandler(
```

core/beneficiary/l1_handler.go
```diff
@@ -0,0 +1,53 @@
...
+
+package beneficiary
+
+import (
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/mysteriumnetwork/node/identity"
+)
+
+// l1Handler handles saving beneficiary in L1 chains.
+// Beneficiary is stored and retreived from the blockhain.
```
soffokl, line 26:
```suggestion
// Beneficiary is stored and retrieved from the blockchain.
```

core/beneficiary/l2_handler.go
```diff
@@ -0,0 +1,58 @@
...
+
+import (
+	"fmt"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/mysteriumnetwork/node/identity"
+)
+
+// l2Handler handles beneficiary in L2 chains.
+// Benefiaciry is stored and retreived from the given storage.
```
soffokl, line 28:
```suggestion
// Beneficiary is stored and retrieved from the given storage.
```

core/beneficiary/l2_handler.go
```diff
@@ -0,0 +1,58 @@
...
+		st:      st,
+	}
+}
+
+// SaveBeneficiary saves the given beneficiary in storage.
+func (b *l2Handler) SaveBeneficiary(id identity.Identity, beneficiary common.Address) error {
+	return b.st.SetValue(storageBucket, b.storageKey(id.Address), beneficiary.Hex())
+}
+
+// GetBeneficiary returns an alredy saved beneficiary.
```
soffokl, line 48:
```suggestion
// GetBeneficiary returns an already saved beneficiary.
```

soffokl (approved):

tequilapi/endpoints/transactor.go
tomasmik: You're right I missed that somehow. Totally my bad here.

metadata/chains.go
tomasmik: My main reason for doing both of these, was that if chain is neither l1 or l2, we dont even give a handler for beneficiary because user is operating under a chain ID we dont know how to handle. If you think thats not needed, I can remove it for sure.

core/beneficiary/l2_handler.go
```diff
@@ -0,0 +1,83 @@
...
+		hermesID:              hermesID,
+		registryAddress:       registryAddress,
+		channelImplementation: channelImplmentation,
+	}
+}
+
+// SettleAndSaveBeneficiary settles beneficiary set to users own payments channel address.
+// The given beneficiary is saved to the database and later retrieved from there.
+func (b *l2Handler) SettleAndSaveBeneficiary(id identity.Identity, beneficiary common.Address) error {
+	addr, err := crypto.GenerateChannelAddress(
```
vkuznecovas, line 57: I think it's best to base this off of https://github.com/mysteriumnetwork/node/pull/3027 as I've introduced a new component for both, channel address calculations as well as accessing hermes/registry/channel addresses.

metadata/chains.go
vkuznecovas: the settlewithebeneficiary will return a "unknown chain error" in such a case, so I don't think it's beneficial to have this here.

vkuznecovas (changes requested):

vkuznecovas (approved):

PR #3124, opened 2021-02-17
iOS Framework

mdomasevicius:
This is a partial restoration of iOS packaging.
TODO: Update xgo cross compilation docker image

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -125,6 +125,11 @@ package:windows-amd64:
    tags: [go]
    script: go run mage.go -v PackageWindowsAmd64
 
+package:ios:
+  stage: build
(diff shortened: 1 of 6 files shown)


tomasmik (approved):

.gitlab-ci.yml
```diff
@@ -125,6 +125,11 @@ package:windows-amd64:
    tags: [go]
    script: go run mage.go -v PackageWindowsAmd64
 
+package:ios:
```
vkuznecovas, line 128: I think you can use the tag ios instead of packageIOS, it becomes a standard tag from 1.16 onwards.

vkuznecovas (approved):

PR #3141, opened 2021-02-25
Allow DNS lookups to have a head start

### LINKED ISSUE #3055: Fix DialerSwarm fallback
Currently, the dial is performed in parallel with both dns and IP addresses directly, which means that we are kind of unable to point our DNS and/or move our servers. This needs addressing.

vkuznecovas:
Allow a headstart for DNS-based lookups in the dialer swarm.

The default headstart is 1.5sec.

Closes #3055


tomasmik (approved):
👏🏻

soffokl (approved):

PR #3160, opened 2021-03-04
Account register command now supports tokens

tomasmik:
Added additional flag to allow registering using affiliate code from the `simple cli` commands.

### DIFF
--- a/cmd/commands/account/command.go
+++ b/cmd/commands/account/command.go
@@ -53,6 +53,11 @@ var (
 		Name:  "last-topup",
 		Usage: "Include last top up information",
 	}
+
+	flagToken = cli.StringFlag{
+		Name:  "token",
+		Usage: "Either a referral or affiliate token which can be used when registering",
+	}
 )
 
 // NewCommand function creates license command.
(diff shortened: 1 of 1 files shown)


cmd/commands/account/command.go
```diff
@@ -245,12 +251,18 @@ func (c *command) register() {
...
-	err := c.tequilapi.RegisterIdentity(identity, "", new(big.Int).SetInt64(0), nil)
+func (c *command) registerIdentity(identity, token string) {
+	var validToken *string
+	if token != "" {
+		validToken = &token
+	}
+
+	err := c.tequilapi.RegisterIdentity(identity, "", new(big.Int).SetInt64(0), validToken)
 	if err != nil {
+		fmt.Println(err)
```
soffokl, line 265: Should it be `clio.Error` for some color?

soffokl (approved):

cmd/commands/account/command.go
tomasmik: This was left here because sometimes i forget to remove things after debugging. removed.

vkuznecovas (approved):

PR #3182, opened 2021-03-08
Expose settle with beneficiary transaction info

mdomasevicius:
- a way to show settle with beneficiary status NodeUI
- make settle with beneficiary async since most users reached timeouts


core/beneficiary/saver.go
```diff
@@ -61,23 +63,77 @@ type l1Saver struct {
...
 	}
 }
 
+// SettleWithBeneficiaryState represents the state of settle with beneficiary transaction
+type SettleWithBeneficiaryState string
+
+const bucketBeneficiaryChangeStatus = "beneficiary-change-status"
+const (
+	// PENDING transaction is pending
+	PENDING SettleWithBeneficiaryState = "pending"
```
tomasmik, line 84: Standard library uses camel-case for constants or any other variable except maybe for some vars in `os` package, so I would suggest to keep it like that.

core/beneficiary/saver.go
```diff
@@ -61,23 +63,77 @@ type l1Saver struct {
...
+	var errorMSG string
+	if settleError != nil {
+		errorMSG = settleError.Error()
+	}
+	err := bcs.st.SetValue(bucketBeneficiaryChangeStatus, storageKey(bcs.chainID, id.Address), &BeneficiaryChangeStatus{
+		State: sbs,
+		Error: errorMSG,
+	})
+
+	log.Err(err).Msg("Failed to update BeneficiaryChangeStatus")
```
tomasmik, line 110: Forgot an `if` here?

core/beneficiary/saver.go
```diff
@@ -61,23 +63,77 @@ type l1Saver struct {
...
+// BeneficiaryChangeStatus holds Beneficiary settlement transaction information
+type BeneficiaryChangeStatus struct {
+	State SettleWithBeneficiaryState
+	Error string
+}
+
+func (bcs *beneficiaryChangeKeeper) updateBeneficiaryChangeStatus(id identity.Identity, sbs SettleWithBeneficiaryState, settleError error) {
+	var errorMSG string
+	if settleError != nil {
+		errorMSG = settleError.Error()
```
tomasmik, line 103: Im gonna be annoying here and say that it should be `errorMsg` if we want to be super idiomatic about things 🤷🏻

core/beneficiary/saver.go
```diff
@@ -61,23 +63,77 @@ type l1Saver struct {
...
+	err := bcs.st.SetValue(bucketBeneficiaryChangeStatus, storageKey(bcs.chainID, id.Address), &BeneficiaryChangeStatus{
+		State: sbs,
+		Error: errorMSG,
+	})
+
+	log.Err(err).Msg("Failed to update BeneficiaryChangeStatus")
+}
+
+// BeneficiaryChangeStatus get beneficiary change status for given identity
+func (bcs *beneficiaryChangeKeeper) BeneficiaryChangeStatus(id identity.Identity) *BeneficiaryChangeStatus {
```
tomasmik, line 114: There are few things I would consider:
* Add `Get` or something like that in front of the function name to indicate what exactly it's doing
* Dont return possible `nil` values. Either return `(*BeneficiaryChangeStatus, bool)` or `(*BeneficiaryChangeStatus, error)` and possibly drop loging from here. From general practice, returning `nil` and no error results in unwanted panics :(

core/beneficiary/saver.go
```diff
@@ -61,23 +63,77 @@ type l1Saver struct {
...
 	}
 
+	b.updateBeneficiaryChangeStatus(id, PENDING, nil)
+	settleError := b.set.SettleWithBeneficiary(b.chainID, id, beneficiary, hermesID)
+	b.updateBeneficiaryChangeStatus(id, COMPLETED, settleError)
+	if settleError != nil {
+		return settleError
+	}
+
 	return b.set.SettleWithBeneficiary(b.chainID, id, beneficiary, hermesID)
```
tomasmik, line 137: forgot to remove this?

core/beneficiary/saver.go
```diff
@@ -61,23 +63,77 @@ type l1Saver struct {
...
-		ad:      ad,
+		chainID:                 chainID,
+		set:                     set,
+		ad:                      ad,
+		beneficiaryChangeKeeper: beneficiaryChangeKeeper{st: st, chainID: chainID},
 	}
 }
 
+// SettleWithBeneficiaryState represents the state of settle with beneficiary transaction
+type SettleWithBeneficiaryState string
```
tomasmik, line 79: gonna nag a little: the word `beneficiary` is already in the package name, so when you call this, you'll now call it like `beneficiary.SettleWithBeneficiaryState` I guess something like `beneficiary.SettlementState` would make more sense?

tomasmik (approved):

vkuznecovas (approved):

tadaskay (approved):

Issue #3193, opened 2021-03-09
Import private keys


chompomonim:
Node should have functionality to import encrypted private keys, decrypt them and store in key store. Throw error/message if given identity is not registered.

- [x] Bindings for android to import private keys
- [x] Tequila endpoint which accepts encrypted private keys and password, decrypts it and stores in key store --> will be used later in nodeUI and CLI ( https://github.com/mysteriumnetwork/node/pull/3198 )
- [x] CLI should allow to import encrypted private key from given file. It imported identity is not registered, cli should ask to register it. It could also ask to make this identity as default. https://github.com/mysteriumnetwork/node/pull/3215

Issue #3192, opened 2021-03-09
Export private keys


chompomonim:
Node should have functionality to encrypt private keys with given password and would expose file with encrypted keys.

- [x] Bindings for android to export private keys
- [x] Tequila endpoint which accepts password, encrypt keys and allow to download them (in a secure way) --> will be used later in nodeUI and CLI ( https://github.com/mysteriumnetwork/node/pull/3198 )
- [x] CLI should allow to download private keys encrypted with given password https://github.com/mysteriumnetwork/node/pull/3215

tomasmik:
Exporting privates keys using tequila creates this sort of a security issue. Right now to get access to a key for example on linux running node, you have to have `sudo` permissions. If we add this as a tequila endpoint, anyone can query and export keys.

Problem is that we do not have any passphrases on the keys, which makes it easy to just steal the key and use it. We can't really add passphrases because of how the apps/node work.

My proposed solution is this:
* Mobile: Add bindings and use them to export the key using the `mover` introduced in #3198 . Though we should also ask for some kind of password probably? I guess we dont have one there?
* CLI/Linux: Use the `mover` together with sudo password to export the key
* Desktop apps: Just show the path to the keys and tell the user to copy them out? as we dont really have direct access to node in desktop. ( this is really not a user friendly solution)

We could consider adding the tequila endpoints to export the identity if we had some form of security for these endpoints, like a one time password MMN could issue, or a general password user creates when launching the application/node.

Second part of this - import (https://github.com/mysteriumnetwork/node/issues/3193), could be done easily through tequila as importing keys probably has no security problems.
@chompomonim  ideas?

chompomonim:
> Mobile: Add bindings and use them to export the key using the mover introduced in #3198 . Though we should also ask for some kind of password probably? I guess we dont have one there?

tomasmik:
The biggest issue here is probably the desktop apps. Copying files from directories is just not user friendly. Do you have any ideas? Or we just skip it for now and provide this feature only for cli (which probably is only useful for providers)

PR #3198, opened 2021-03-10
Identity exporting/importing tequilapi

tomasmik:
- We now allow to import or export private keys from keystore.
- To import an identity it must be registered.
- We accept current passphrase (in export) and new passphrase (in import) as optional parameters as no accounts really have them, but it allows for a clear upgrade path in the future if we need it.

Updates: https://github.com/mysteriumnetwork/node/issues/3193 (import)
Updates: https://github.com/mysteriumnetwork/node/issues/3192 (export)


tequilapi/endpoints/identities.go
```diff
@@ -490,4 +597,7 @@ func AddRoutesForIdentities(
 	router.GET("/identities/:id/beneficiary", idmEnd.Beneficiary)
 	router.GET("/identities/:id/referral", idmEnd.GetReferralToken)
 	router.GET("/identities/:id/referral-available", idmEnd.ReferralTokenAvailable)
+
+	router.POST("/identities-export/:id", idmEnd.Export)
```
soffokl, line 601: I'm still not sure how we are going to protect it from being used by anyone.
Tequilapi opened with no auth for anything on the same host. Any script/tool/app can import and export private keys with no password.
I'd as a user prefer to have this secure than convenient for exporting by any app.

tomasmik:
After thinking a bit and having a talk with @soffokl this indeed is not very secure. We should rethink on how this task should be done.

PR #3214, opened 2021-03-16
Migration to set registration status to `unregistered`

tomasmik:


core/storage/boltdb/migrations/registration_state.go
```diff
@@ -0,0 +1,61 @@
+/*
+ * Copyright (C) 2020 The "MysteriumNetwork/node" Authors.
```
soffokl, line 2:
```suggestion
 * Copyright (C) 2021 The "MysteriumNetwork/node" Authors.
```

soffokl (approved):

PR #3215, opened 2021-03-16
Export import keys using CLI

tomasmik:
You can now:
* Export to file/terminal
* Import from file/input

Imported identities are set as default.


cmd/commands/cli/command_identities.go
```diff
@@ -310,3 +320,85 @@ func (c *cliApp) setBeneficiary(actionArgs []string) {
...
+			return err
+		}
+
+		err := write()
+		if err == nil {
+			clio.Success("Identity exported to file:", filepath)
+			return
+		}
+
+		clio.Warn("Export succeeded, file write failed. Will print to screen.")
```
soffokl, line 370: This `Will print to screen` could be a problem. If the user does not expect private keys to be printed to stdout we should not do it. If something failed the user can do a second attempt or print to stdout manually.

cmd/commands/cli/command_identities.go
```diff
@@ -310,3 +320,85 @@ func (c *cliApp) setBeneficiary(actionArgs []string) {
...
+	}
+
+	clio.Success("Private key exported: ")
+	fmt.Println(string(blob))
+}
+
+const usageImportIdentity = "import <passphrase> <key-string/key-file>"
```
soffokl, line 377:
```suggestion
const usageImportIdentity = "import <passphrase> <path/to/key-file>"
```

cmd/commands/cli/command_identities.go
```diff
@@ -310,3 +320,85 @@ func (c *cliApp) setBeneficiary(actionArgs []string) {
...
+	}
+
+	key := actionsArgs[1]
+	passphrase := actionsArgs[0]
+
+	blob := []byte(key)
+	if _, err := os.Stat(key); err == nil {
+		blob, err = ioutil.ReadFile(key)
+		if err != nil {
+			clio.Error(fmt.Sprintf("cant read provided file: %s reason: %s", key, err.Error()))
```
soffokl, line 392:
```suggestion
			clio.Error(fmt.Sprintf("can't read provided file: %s reason: %s", key, err.Error()))
```

soffokl (approved):

cmd/commands/cli/command_identities.go
tomasmik: You can give a string as in paste your key, or a file, so I think im correct here.

vkuznecovas (approved):

PR #3216, opened 2021-03-17
Allow to keep connection after fail to prevent traffic leaks.

soffokl:


tomasmik (approved):

vkuznecovas (approved):

zolia (commented):
Hello,

As I understood from the code - this prevents session cleanup if a payment error is encountered.
But wasn't it all the point to disconnect a session if payments cannot proceed?

Then you put the session in a weird suspend state. What does this mean? Do you simply continue if nothing really happened? This basically breaks the whole point of payments ping-pong.

In which conditions leaking does happen? If network access changes while the client moves (wifi->4g or vice versa) or the user intentionally switches off/on wifi. Those conditions can cause/trigger session reconnect which might result in leaked data while reconnecting.

We need to thoroughly identify all small-time gaps when leaking could happen. TadasV coded leaking prevention for linux by simply allowing only those IPs that should be allowed by the current session, but together with the session reconnect feature it should not work as expected. Similar this can be done for macOS since it has a built-in firewall.

For mobile, the firewalling approach will not work thus the only way is to properly detect network changing events and act upon them minimizing all possible leaking gaps.

Also, there is some time gap when you really don't know that network access change is in progress. This time gap is of interest to us, we need to analyze what really happens during that time. 

First, we should determine that network access started to change, put a session on hold, make sure that reconnect DID happen, and then clean-up/merge an old session.

soffokl:
It prevents session cleanup if any error happened during an active session. This is not only a payment error.

Payment and ping-pong disconnect important only for providers to avoid providing service for free. We are talking here about the consumer session. Some consumers want to avoid unexpected session disconnect for any reason. It is better to have no connection explicitly than return to a direct connection without knowing it. 

If the consumer loses connection to the provider, we are putting this connection to an `On Hold` state. This is just a first step, reconnect without leaks problem is being resolved in the next task https://github.com/mysteriumnetwork/node/issues/3188.

Issue #3247, opened 2021-03-25
Add component that will create and maintain routing table


soffokl:
We are currently adding a couple of route records to exclude some traffic from going through the tunnel. 
This is a static record that never updates and uses a gateway that was set on session creation time. 

Once the user migrates to a different wifi-network his network configuration might be changed and we need to maintain our routing records to keep it workable. 

We need to create a component/worker in the node that will be tracking network configuration and will be maintaining routing rules in the correct form.

Issue #3246, opened 2021-03-25
Add internal DNS cache for cases when consumer loss connection to provider


soffokl:
To fully support smart auto-reconnect without leaking traffic we need to be able to resolve domains when we lose connection to the provider. 
One of the options could be adding a local DNS cache for records to be able to use it when provider DNS service is not available.
This DNS cache could have a configuration to set lists of domain names with the desired TTL:
 * *.mysterium.network - 24h
 * *.ipinfo.io - 1h
 * *.google.com - 10m
 *  default value - 1h

PR #3248, opened 2021-03-25
Reconnect automatically to the same provider on ping loss

soffokl:
Closes https://github.com/mysteriumnetwork/node/issues/3188

Consider this PR as a POC for auto-reconnect.
It's laking some crucial features, that needs to be implemented before releasing it publically:
* https://github.com/mysteriumnetwork/node/issues/3246 support for DNS resolution during the provider fails. 
* https://github.com/mysteriumnetwork/node/issues/3247 add worker that will maintain routing rules.


communication/nats/connection_wrap.go
```diff
@@ -121,3 +132,11 @@ func (c *ConnectionWrap) Close() {
 func (c *ConnectionWrap) Servers() []string {
 	return c.servers
 }
+
+type dialer struct {
+	dialer requests.DialContext
+}
+
+func (d *dialer) Dial(network, address string) (net.Conn, error) {
+	return d.dialer(context.Background(), network, address)
```
vkuznecovas, line 141: won't context.Background() cause infinite hangs?

communication/nats/connector.go
```diff
@@ -21,31 +21,37 @@ import (
...
 
 func (b *BrokerConnector) resolveServers(serverURLs []*url.URL) ([]*url.URL, error) {
-	if b.ResolveContext == nil {
+	if b.resolveContext == nil {
 		return serverURLs, nil
 	}
 
 	for _, serverURL := range serverURLs {
-		addrs, err := b.ResolveContext(context.Background(), "tcp", serverURL.Host)
+		addrs, err := b.resolveContext(context.Background(), "tcp", serverURL.Host)
```
vkuznecovas, line 54: same as above.

vkuznecovas (approved):

mdomasevicius (approved):

core/connection/manager.go
```diff
@@ -408,20 +452,20 @@ func (m *connectionManager) cleanAfterDisconnect() {
...
 	defer tracer.EndStage(trace)
 
-	contactDef, err := p2p.ParseContact(proposal.ProviderContacts)
+	contactDef, err := p2p.ParseContact(opts.Proposal.ProviderContacts)
 	if err != nil {
 		return fmt.Errorf("provider does not support p2p communication: %w", err)
 	}
 
-	timeoutCtx, cancel := context.WithTimeout(ctx, p2pDialTimeout)
+	timeoutCtx, cancel := context.WithTimeout(m.currentCtx(), p2pDialTimeout)
```
alvatar, line 464: This is a bit uncommon (contexts are usually explicitly passed around). What led you to this decision?

core/connection/manager.go
```diff
@@ -507,14 +552,14 @@ func (m *connectionManager) createP2PSession(ctx context.Context, c Connection,
...
-			ConsumerID: consumerID.Address,
+			ConsumerID: opts.ConsumerID.Address,
 			SessionID:  sessionResponse.GetID(),
 		}
 
 		log.Debug().Msgf("Sending P2P message to %q: %s", p2p.TopicSessionDestroy, sessionDestroy.String())
 		ctx, cancel := context.WithTimeout(context.Background(), 1*time.Second)
 		defer cancel()
-		_, err := p2pChannel.Send(ctx, p2p.TopicSessionDestroy, p2p.ProtoMessage(sessionDestroy))
+		_, err := m.channel.Send(ctx, p2p.TopicSessionDestroy, p2p.ProtoMessage(sessionDestroy))
```
alvatar, line 562: I think it would be nice to name this channel in an explicit way even if it's longer. Channels could be many things, so it would provide more context.

mobile/mysterium/wireguard_connection_setup.go
```diff
@@ -196,6 +200,13 @@ func (w *wireguardDeviceImpl) Start(privateKey string, config wireguard.ServiceC
 		return errors.Wrap(err, "could not create tunnel device")
 	}
 
+	oldDevice := w.device
+	defer func() {
+		if oldDevice != nil {
+			oldDevice.Close()
```
alvatar, line 206: I'm confused about this: why is the old device closed at the end, and only if it exists? Could it more direct to just close it as soon as the new one is created and assigned?

mobile/mysterium/wireguard_connection_setup.go
```diff
@@ -259,6 +270,7 @@ func (w *wireguardDeviceImpl) applyConfig(devApi *device.Device, privateKey stri
 			// All traffic through this peer (unfortunately 0.0.0.0/0 didn't work as it was treated as ipv6)
 			AllowedIPs: []string{"0.0.0.0/1", "128.0.0.0/1"},
 		},
+		ReplacePeers: true,
```
alvatar, line 273: What does this do?

requests/dialer_swarm.go
```diff
@@ -243,3 +260,35 @@ func (e *ErrorDial) Error() string {
...
+
+func wrapDialer(dialer *net.Dialer) *dialerWithDNSCache {
+	return &dialerWithDNSCache{
+		dialer: dialer,
+	}
+}
+
+func (wd *dialerWithDNSCache) DialContext(ctx context.Context, network, addr string) (net.Conn, error) {
+	go func() {
+		if !isIP(addr) {
```
alvatar, line 276: Nitpicking: We could avoid creating the goroutine if we do this check first. It's cheap though, so no big deal. But I have no context yet on how many of these dials are being done, so might be worth considering.

requests/resolver.go
```diff
@@ -34,10 +34,28 @@ func NewResolverMap(hostToIP map[string][]string) ResolveContext {
...
 		for _, addrIP := range hostToIP[addrHost] {
 			addrs = append(addrs, net.JoinHostPort(addrIP, addrPort))
 		}
 
-		return addrs, nil
+		return deduplicate(addrs), nil
+	}
+}
+
+func deduplicate(list []string) (result []string) {
```
alvatar, line 50: This could probably go into a common utils package. Do we have such thing?

requests/resolver_cache.go
```diff
@@ -0,0 +1,58 @@
...
+ */
+
+package requests
+
+import "sync"
+
+var defaultResolveCache = NewResolverCache()
+
+type resolverCache struct {
+	mu    sync.Mutex
```
alvatar, line 25: We could consider using a RWMutex to allow for reading from the cache simultaneously. Do you think it could be an improvement?

router/protect.go
```diff
@@ -0,0 +1,67 @@
...
+)
+
+var (
+	mu      sync.RWMutex
+	protect func(fd int) error = nil
+)
+
+// SetProtectFunc sets the callback for using to protect provided connection from going through the tunnel.
+func SetProtectFunc(f func(fd int) error) {
+	mu.Lock()
```
alvatar, line 32: Regarding this synchronization pattern, I would to ask: are we resetting this `protect` function often or is just a one-off?

services/wireguard/endpoint/userspace/client.go
```diff
@@ -22,12 +22,13 @@ import (
...
-	"github.com/mysteriumnetwork/node/services/wireguard/connection/dns"
-	"github.com/mysteriumnetwork/node/services/wireguard/wgcfg"
-	"github.com/mysteriumnetwork/node/utils/netutil"
 	"github.com/pkg/errors"
 	"golang.zx2c4.com/wireguard/device"
 	"golang.zx2c4.com/wireguard/tun"
+
+	"github.com/mysteriumnetwork/node/services/wireguard/connection/dns"
+	"github.com/mysteriumnetwork/node/services/wireguard/wgcfg"
+	"github.com/mysteriumnetwork/node/utils/netutil"
```
alvatar, line 31: Nice that you keep things tidy! :+1:

alvatar (commented):
Great job @soffokl 
Just added comments to understand better, and a few pointers as well on implementation details.

core/connection/manager.go
```diff
@@ -92,7 +92,7 @@ func DefaultConfig() Config {
 			SleepDurationAfterCheck: 3 * time.Second,
 		},
 		KeepAlive: KeepAliveConfig{
-			SendInterval:    20 * time.Second,
+			SendInterval:    5 * time.Second,
```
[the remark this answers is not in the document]
soffokl: This improves the overall reconnect experience to make it faster. These pings are p2p so not a big change in traffic there.

core/connection/manager.go
soffokl: I just moved it from one place to another, we were still passing `m.currentCtx()`.

core/connection/manager.go
soffokl: Good point about naming, but this will affect more changes, let's make it a separate PR.

mobile/mysterium/wireguard_connection_setup.go
soffokl: It does exactly the same. Closes the old device once new created and configured.

mobile/mysterium/wireguard_connection_setup.go
soffokl: It removed the old configuration from the old peers, and replace it with a new configuration. Instead of just adding over and over.

requests/dialer_swarm.go
soffokl: This is very tiny save here. Let's keep it as is until we will have a need for it.

requests/resolver.go
soffokl: Yeah, we have such a package, but there is no point to move it to a common package while it's needed only in one place. We moving it after it starts to be needed in several places.

requests/resolver_cache.go
soffokl: This is a very basic implementation. This is one of the improvements that will be applied later.

router/protect.go
soffokl: This happens only once per session.

PR #3249, opened 2021-03-25
Estimate entertainment times based on MYST amount (MVP)

mdomasevicius:
This is all based on our default node prices. As per discussion with @vkuznecovas fixed pricing is subject to change to organic pricing model, this means that the prices will be not set by providers, so currently prices for estimations are left as constants.

Average data usages per minute are based on results from searches on average mobile data usage for most popular services: youtube, spotify, facebook... also these values are for mid range quality streams.

### DIFF
--- a/consumer/entertainment/estimator.go
+++ b/consumer/entertainment/estimator.go
@@ -1,0 +1,80 @@
+/*
+ * Copyright (C) 2021 The "MysteriumNetwork/node" Authors.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package entertainment
+
+import "math"
+
+const (
+	video720pMBPerMin   = 15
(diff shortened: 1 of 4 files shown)


consumer/entertainment/estimator.go
```diff
@@ -0,0 +1,70 @@
...
+}
+
+// Estimator stores average provider prices to estimate entertainment estimates
+type Estimator struct {
+	pricePerGB  float64
+	pricePerMin float64
+}
+
+// NewEstimator constructor
+func NewEstimator(pricePerGB, pricePerMin float64) *Estimator {
```
vkuznecovas, line 43: I think our current pricing is per GiB rather than GB and you're calculating them further down the line basing on GiB.

vkuznecovas (changes requested):

consumer/entertainment/estimator.go
```diff
@@ -0,0 +1,80 @@
...
+func mib2MB(mibs float64) float64 {
+	return mibs * math.Pow(2, 20) / math.Pow(10, 6)
+}
+
+func mb2MiB(mb float64) float64 {
+	return mb * math.Pow(10, 6) / math.Pow(2, 20)
+}
+
+func (e *Estimator) totalTrafficMiB(amount float64) float64 {
+	return amount / e.pricePerGiB * 1024
```
tomasmik, line 73: `datasize` package already defines numbers like 1024

mobile/mysterium/estimator.go
```diff
@@ -0,0 +1,46 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package mysterium
+
+import "github.com/mysteriumnetwork/node/consumer/entertainment"
+
+// Estimates represent estimated entertainment
+type Estimates struct {
+	VideoMinutes    int64   `json:"video_minutes"`
```
tomasmik, line 24: Do these json flags actually mean somethin when converted to mobile?

tomasmik (approved):

vkuznecovas (approved):

PR #3269, opened 2021-04-06
Resident country config for mobile

mdomasevicius:
@soffokl can you take a look at metrics code? I personally think resident country should be tied to identity although you may have more original ideas.

UPDATE: resident country is required for legal reasons concerning VAT when performing top up's. @chompomonim has more information regarding this but in a nutshell we need to tie identity to country.


soffokl (commented):
@mdomasevicius could you add some context to the description. Do we have an issue created for this? 
For me, it's not clear what we are trying to resolve by this.

tomasmik:
If we're doing this when performing topups, it would probably make sense to also include this country thing in the pilvytis order itself:
https://github.com/mysteriumnetwork/node/blob/master/pilvytis/api.go#L207

identity/resident_country.go
```diff
@@ -0,0 +1,62 @@
...
+		}
+	})
+}
+
+func (l *UnlockEventListener) publishResidentCountry(e AppEventIdentityUnlock) {
+	event := ResidentCountryEvent{
+		ID:      e.ID.Address,
+		Country: config.GetString(config.FlagResidentCountry),
+	}
+	l.eventBus.Publish(AppTopicResidentCountry, event)
```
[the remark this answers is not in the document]
mdomasevicius: I want to separate the manager from direct dependency to config

identity/resident_country.go
```diff
@@ -0,0 +1,77 @@
...
+	eventBus         eventbus.EventBus
+	locationResolver locationProvider
+}
+
+// NewResidentCountry constructor
+func NewResidentCountry(eventBus eventbus.EventBus, locationResolver locationProvider) *ResidentCountry {
+	return &ResidentCountry{eventBus: eventBus, locationResolver: locationResolver}
+}
+
+// Save save country code and fire AppTopicResidentCountry
```
tomasmik, line 48:
```suggestion
// Save country code and fire AppTopicResidentCountry
```

identity/manager.go
```diff
@@ -148,10 +156,13 @@ func (idm *identityManager) Unlock(chainID int64, address string, passphrase str
...
-	go idm.eventBus.Publish(AppTopicIdentityUnlock, AppEventIdentityUnlock{
-		ChainID: chainID,
-		ID:      FromAddress(address),
-	})
+	go func() {
+		idm.eventBus.Publish(AppTopicIdentityUnlock, AppEventIdentityUnlock{
+			ChainID: chainID,
+			ID:      FromAddress(address),
+		})
+		idm.residentCountry.publishResidentCountry(address)
```
tomasmik, line 164: do we still need to publish this on every unlock?

cmd/di.go
```diff
@@ -938,3 +950,19 @@ func (di *Dependencies) handleConnStateChange() error {
...
+	}
+
+	if di.LocationResolver == nil {
+		return errMissingDependency("di.LocationResolver")
+	}
+	di.ResidentCountry = identity.NewResidentCountry(di.EventBus, di.LocationResolver)
+	return nil
+}
+
+func errMissingDependency(dep string) error {
```
tomasmik, line 966: Idea is nice and cool and I dont generally disagree with this, but wont it get too messy if we start doing this for every dependency?

tomasmik (approved):

cmd/di.go
vkuznecovas: it's already pretty messy. The DI part is rather hard to comprehend, a major refactor is in order :)

vkuznecovas (approved):

PR #3347, opened 2021-04-29
Proposal Filter Presets (Mobile) Part 1

mdomasevicius:
This is to unblock mobile team, the actual logic for determining how to filter proposals by preset will come later

### DIFF
--- a/core/discovery/proposal/filter_presets.go
+++ b/core/discovery/proposal/filter_presets.go
@@ -1,0 +1,191 @@
+/*
+ * Copyright (C) 2021 The "MysteriumNetwork/node" Authors.
+ *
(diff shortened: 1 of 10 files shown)


core/discovery/proposal/filter_presets.go
```diff
@@ -0,0 +1,206 @@
...
+package proposal
+
+import (
+	"fmt"
+	"sync"
+
+	"github.com/pkg/errors"
+)
+
+var errMsgBoltNotFound = "not found"
```
tomasmik, line 27: I think it would be better to use `storm.ErrNotFound` or something like that 🤷🏻

tomasmik (approved):

vkuznecovas (approved):

PR #3404, opened 2021-05-18
Fix service start

tadaskay:
https://github.com/mysteriumnetwork/node/pull/3379 added required Price for service start request in tequila. This removes the requirement and also allows using pre-saved config values for service pricing (previous web UI behavior) to make it easier to consume.


tomasmik (approved):
I think CI will mention this, but if it doesn't I think swagger also needs an update now.

mdomasevicius (approved):

vkuznecovas (approved):

PR #3449, opened 2021-06-04
Remove additional `--update` for `apk` in Dockerfile

PeterDaveHello:
There is no need to use `--update` with `--no-cache`, as using `--no-cache` will fetch the index every time and leave no local cache, so the index will always be the latest and without temporary files remains in the image.


soffokl (approved):
Thanks for pointing it.

tomasmik (approved):

Issue #3228, opened 2021-03-18
Export/Import privates keys in desktop app


tomasmik:
After https://github.com/mysteriumnetwork/node/issues/3192 and https://github.com/mysteriumnetwork/node/issues/3193 are complete we now should be able to import and export private keys in our desktop app.

* Access the keystore from the desktop app to export the key (it should be exported in json format and should have a passphrase set)
* Import the keys using tequilapi

tadaskay:
Still relevant

PR #3503, opened 2021-06-17
Export/Import private keys via CLI

### LINKED ISSUE #2448: Export/Import private keys via CLI
It should be possible to import/export private key via CLI. We should not use tequilla for that. Also please ask user to type password for PrivKey backup (at least 12 symbols) as it should be never backoups with empty passwords. User should be able to set backup file destination, by default it could be the path from which the command was run. 

- [x] CLI to export encrypted private key.
- [x] CLI to import private keys (should be endrypted and decrypted).
- [x] Add/update documentation page how to make a privkey backup and how to download it to your machine (if its in docker, rpi or remote server)

Zensey:
Resolves #2448

### DIFF
--- a/cmd/commands/keystore/command_keystore.go
+++ b/cmd/commands/keystore/command_keystore.go
@@ -1,0 +1,277 @@
+package keystore
+
+import (
+	"fmt"
+	"io/ioutil"
+	stdlog "log"
+	"os"
+	"strings"
+
+	"github.com/chzyer/readline"
+	"github.com/ethereum/go-ethereum/accounts/keystore"
+	"github.com/mysteriumnetwork/node/cmd/commands/cli/clio"
+	"github.com/mysteriumnetwork/node/config"
+	"github.com/mysteriumnetwork/node/core/node"
+	"github.com/mysteriumnetwork/node/eventbus"
+	"github.com/mysteriumnetwork/node/identity"
+	"github.com/mysteriumnetwork/node/tequilapi/contract"
+	"github.com/mysteriumnetwork/node/utils"
+	"github.com/rs/zerolog"
+	"github.com/rs/zerolog/log"
+	"github.com/urfave/cli/v2"
+)
+
+func NewCommand() *cli.Command {
+	return &cli.Command{
+		Name:  "keystore",
+		Usage: "keystore CLI for import / export of private keys",
+		Flags: []cli.Flag{},
+		Action: func(ctx *cli.Context) error {
+
+			config.ParseFlagsNode(ctx)
+			cmdCLI := newCliApp()
+			cmdCLI.Run(ctx)
+
+			return nil
+		},
+	}
+}
+
+type cliApp struct {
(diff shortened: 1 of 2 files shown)


tomasmik (changes requested):
Maybe Im a missing something here, but I think we already have these commands:
```
  export <identity> <new_passphrase> [file]
  import <passphrase> <key-string/key-file>
```

Here are the issues related to this:

- Import private keys #3193
- Export private keys #3192

I see in the current issue there talks about some security issues, but I think this is something we already discussed at least a few times. Tequila API does not allow you to export identities, only to import them. So this is the same security risk as someone accessing  your server and copying the keys.

Also this will probably not work in production because of how configuration of node works. Directories of where the keystore and all other files are must be returned from a running node itself or else you're getting different vars because you'll call commands in user space and node is not running there.

Pretty much the only missing part in this whole import/export is this: https://github.com/mysteriumnetwork/node/issues/3228 and documentation https://github.com/mysteriumnetwork/documentation/issues/41

Zensey:
> Maybe Im a missing something here, but I think we already have these commands:
> 
> ```
>   export <identity> <new_passphrase> [file]
>   import <passphrase> <key-string/key-file>
> ```
> 
> Here are the issues related to this:
> 
> * Import private keys #3193
> * Export private keys #3192
> 
> I see in the current issue there talks about some security issues, but I think this is something we already discussed at least a few times. Tequila API does not allow you to export identities, only to import them. So this is the same security risk as someone accessing your server and copying the keys.
> 
> Also this will probably not work in production because of how configuration of node works. Directories of where the keystore and all other files are must be returned from a running node itself or else you're getting different vars because you'll call commands in user space and node is not running there.
> 
> Pretty much the only missing part in this whole import/export is this: #3228 and documentation [mysteriumnetwork/documentation#41](https://github.com/mysteriumnetwork/documentation/issues/41)

yes, it does not work in prduction:
```
[INFO] ksdir: /var/lib/mysterium-node/keystore
[ERROR] Failed to export identity:  identity not found
» 

```

tomasmik:
> > Also this will probably not work in production because of how configuration of node works. Directories of where the keystore and all other files are must be returned from a running node itself or else you're getting different vars because you'll call commands in user space and node is not running there.
> > Pretty much the only missing part in this whole import/export is this: #3228 and documentation [mysteriumnetwork/documentation#41](https://github.com/mysteriumnetwork/documentation/issues/41)
> 
> yes, it does not work in prduction:
> 
> ```
> identities export 0x324e54bcc556ef6c58245e1d3efb7c2cee64511d 123123123123
> [INFO] ksdir: /var/lib/mysterium-node/keystore
> [ERROR] Failed to export identity:  identity not found
> ```
> 
> So we need to add this command to tequilapi ?

Current implementation works just fine. It's just that in production a regular user can't access node files. You have to run the cli with `sudo`. As I said before "Pretty much the only missing part in this whole import/export is this: #3228 and documentation mysteriumnetwork/documentation#41"

Issue #3534, opened 2021-06-22
Disable OpenVPN from the latest docker image


etherunit:
docker pull mysteriumnetwork/myst:latest, running 0.47.2 node. After claiming into MMN I'm still seeing `OpenVPN` and `noop` enabled by default.

Provider: 0xaa3d1e65a1a84055cb619004a3ca59489970cef5
OS: Alpine Linux v3.12(docker)
Architecture:l inux(docker)/amd64
Node version: 0.47.2

```
**Proposals in Discovery**
[
    {
        "provider_id": "0xaa3d1e65a1a84055cb619004a3ca59489970cef5",
        "service_type": "noop",
        "country": "LT",
        "isp": "Telia Lietuva, AB",
        "ip_type": "residential",
        "whitelist": false,
        "price_per_gib": 100000000000000005,
        "price_per_hour": 60000000000000,
        "updated_at": "2021-06-22T19:44:59.481545Z"
    },
    {
        "provider_id": "0xaa3d1e65a1a84055cb619004a3ca59489970cef5",
        "service_type": "wireguard",
        "country": "LT",
        "isp": "Telia Lietuva, AB",
        "ip_type": "residential",
        "whitelist": false,
        "price_per_gib": 100000000000000005,
        "price_per_hour": 60000000000000,
        "updated_at": "2021-06-22T19:44:59.940257Z"
    },
    {
        "provider_id": "0xaa3d1e65a1a84055cb619004a3ca59489970cef5",
        "service_type": "openvpn",
        "country": "LT",
        "isp": "Telia Lietuva, AB",
        "ip_type": "residential",
        "whitelist": false,
        "price_per_gib": 100000000000000005,
        "price_per_hour": 60000000000000,
        "updated_at": "2021-06-22T19:44:59.710732Z"
```

PR #3536, opened 2021-06-23
Start only wireguard service if other not passed

soffokl:
Closes https://github.com/mysteriumnetwork/node/issues/3534

### DIFF
--- a/cmd/commands/service/command.go
+++ b/cmd/commands/service/command.go
@@ -23,6 +23,10 @@ import (
 	"strings"
 	"time"
 
+	"github.com/pkg/errors"
+	"github.com/rs/zerolog/log"
+	"github.com/urfave/cli/v2"
+
 	"github.com/mysteriumnetwork/node/cmd"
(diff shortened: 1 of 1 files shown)


cmd/commands/service/command.go
```diff
@@ -111,6 +113,10 @@ func (sc *serviceCommand) Run(ctx *cli.Context) (err error) {
 	serviceTypes := services.Types()
 	if arg != "" {
 		serviceTypes = strings.Split(arg, ",")
+	} else {
```
tomasmik, line 116: At this point then something like:
```go
serviceTypes := []string{wiregaurd.ServiceType}
if arg != "" {
  serviceTypes = strings.Split(arg, ",")
}
```

Would probably read better 🤷🏻

tomasmik (approved):

vkuznecovas (approved):

etherunit (approved):

PR #3539, opened 2021-06-23
Do not overwrite remote peer address

soffokl:
We are using other reconnect technics now.
This one is not needed and looks insecure now.


tomasmik (approved):

mdomasevicius (approved):
:eyes:

Issue #3542, opened 2021-06-23
Long sessions drain offchain balance without it getting resynced


tomasmik:
Long running sessions drain balance without actually updating it. For apps this is OK because sessions dont last longer than the assigned MYST amount (which ranges from 3 to 10MYST).

Balance gets updated when you get identity from tequila, unlock it, etc. but not when a connection is ongoing or when you're trying to connect.

We should force a periodic balance update in node for offchain targets and check once again on connect no matter what the balance is.

PR #3545, opened 2021-06-24
Less aggresive balance cache in node

tomasmik:
Balance cache is less aggresive allowing for more force syncing to occur.
Changes:
* If balance drops below 1gwei try to force sync on connect.
  * There seem to be cases where balance isn't actually 0, but very close to it and connection creation fails (even though maybe there is some balance on or off chain it was just missed)
* If balance is offchain it now has an expiration time after which we ask hermes for an update.
  * It is configurable because in most cases this will not be needed. This is mainly so that it's easier to sustain long running (many hours or days at a time) connections for super proxy.
* On connect we also force sync a balance if it's offchain and is expired.
  * Useful mainly for monitoring or regular users using node as offchain.

If any of these additional checks fail, we continue as normal. This is intentional as to not make node depend on hermes.

Closes: https://github.com/mysteriumnetwork/node/issues/3542

To make this a better solution, we would need to build some better way for consumer and provider to talk to each other. So that we wouldn't have to check balance on consumer side often or ever.

### DIFF
--- a/session/pingpong/consumer_balance_tracker.go
+++ b/session/pingpong/consumer_balance_tracker.go
@@ -28,6 +28,7 @@ import (
 
 	"github.com/cenkalti/backoff/v4"
 	"github.com/ethereum/go-ethereum/common"
+	"github.com/mysteriumnetwork/node/config"
 	nodevent "github.com/mysteriumnetwork/node/core/node/event"
 	"github.com/mysteriumnetwork/node/eventbus"
 	"github.com/mysteriumnetwork/node/identity"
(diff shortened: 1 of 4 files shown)


soffokl (approved):

session/pingpong/consumer_balance_tracker.go
```diff
@@ -118,6 +120,26 @@ func (cbt *ConsumerBalanceTracker) Subscribe(bus eventbus.Subscriber) error {
 	return bus.SubscribeAsync(identity.AppTopicIdentityUnlock, cbt.handleUnlockEvent)
 }
 
+// BalanceNeedsForceSync returns true if balance needs to be force synced.
+func (cbt *ConsumerBalanceTracker) BalanceNeedsForceSync(chainID int64, id identity.Identity) bool {
```
mdomasevicius, line 124: `BalanceNeedsForceSync` -> `NeedsForceSync` || `NeedsSync` ?

mdomasevicius (approved):

PR #3574, opened 2021-07-02
Move docker config to data dir

soffokl:
This way we will preserve config between container reinstall with a single mounted volume.


tomasmik (approved):
Good change. I like.

eugenegoncharuk (approved):

vkuznecovas (approved):

PR #3636, opened 2021-07-16
Disable local service discovery in docker containers

soffokl:
Many people complain about the following error in the docker containers:
```
2021-07-15T08:53:03.505 ERR ui/server.go:106 > Failed to start local discovery service error="Could not determine host IP addresses for 0195a70a7caf.
```

It does not break anything, but confusing users.


tomasmik (approved):
Yes, finally!

vkuznecovas (approved):

PR #3663, opened 2021-07-23
Make Linux standalone builds static

Snawoot:
This PR makes standalone Linux builds (those which in zip archives on releases page of GitHub) actually standalone as it makes them static and independent from system runtime libraries.

### DIFF
--- a/ci/packages/build.go
+++ b/ci/packages/build.go
@@ -76,10 +76,10 @@ func buildBinary(source, target string) error {
 	if !ok {
 		targetArch = runtime.GOARCH
 	}
-	return buildBinaryFor(source, target, targetOS, targetArch)
+	return buildBinaryFor(source, target, targetOS, targetArch, false)
 }
 
-func buildBinaryFor(source, target, targetOS, targetArch string) error {
+func buildBinaryFor(source, target, targetOS, targetArch string, buildStatic bool) error {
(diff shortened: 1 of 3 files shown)


tadaskay:
We use mage for build and these bash scripts are legacy. Your changes should go to the corresponding mage targets.

tadaskay (changes requested):
See the above comment

Snawoot:
@tadaskay 

Thanks!

Implemented changes for magefiles. These changes only touching standalone builds for Linux because other platforms have CGO dependency.

tadaskay (approved):

soffokl (approved):

ci/packages/build.go
```diff
@@ -76,10 +76,10 @@ func buildBinary(source, target string) error {
 	if !ok {
 		targetArch = runtime.GOARCH
 	}
-	return buildBinaryFor(source, target, targetOS, targetArch)
+	return buildBinaryFor(source, target, targetOS, targetArch, false)
```
mdomasevicius, line 79: Just a thought that any further extensions of `buildBinaryFor` signatures could be done as `buildBinaryFor(..., opts... BuildOptions)` for better readability.

mdomasevicius (approved):

vkuznecovas (approved):

PR #3664, opened 2021-07-23
p2p: verify provider identity

### LINKED ISSUE #3645: Potential MITM on P2P service channel
### Background

Mysterium node uses some flow to establish P2P connection (`p2p.Channel`) to provider node which can be roughly described like this:

1. Provider node listens for request messages on public message broker (`nats://testnet2-broker.mysterium.network:4222`) topics related to it's Provider ID and service pair.
2. Consumer node initiates connection, sending public ECDHE part signed with consumer node identity key.
3. Provider node responds with ephemeral public ECDHE part signed with provider node identity key.
4. Provider and consumer node exchanging P2P connection configuration, authenticated and encrypted with agreed shared key.
5. Provider and Consumer build KCP connection encrypted with agreed shared key.

### Problem

Both peers do not verify if signature made with public key of expected peer identity. [Current identity `Verifier`](https://github.com/mysteriumnetwork/node/blob/dd44549ba5415a9405f00e70861ffa8d2ae45a9d/cmd/di.go#L311-L319) in use [doesn't ensure that signature was made by any known identity key](https://github.com/mysteriumnetwork/node/blob/dd44549ba5415a9405f00e70861ffa8d2ae45a9d/identity/verifier.go#L40-L47), identity just gets discarded. However, there is [another verifier](https://github.com/mysteriumnetwork/node/blob/dd44549ba5415a9405f00e70861ffa8d2ae45a9d/identity/verifier.go#L49-L61) which does identity matching, but it seems it is not used anywhere except integration tests.

Combined with communication via public NATS server, it makes possible to tap into such channels even without actual access to both peers internet connection (i.e. middlebox is not required). This channel is used to pass parameters of actual service connection and altering exchanged parameters may ruin actual VPN service security.

### Mitigation

Obviously, at least Consumer must verify Provider node identity.

It's unclear for me if further interactions has to be cryptographically bound to consumer identity (i.e. disallow service requests and other requests which specify Consumer ID other than ID requested channel). Maybe it's also has to be enforced.

Snawoot:
Closes #3645 

Adds verification of provider's identity from consumer side.

Code organization note: I've made factory of verificator just like it's Signer counterpart.

### DIFF
--- a/p2p/dialer.go
+++ b/p2p/dialer.go
@@ -50,27 +50,27 @@ type Dialer interface {
 }
 
 // NewDialer creates new p2p communication dialer which is used on consumer side.
-func NewDialer(broker brokerConnector, signer identity.SignerFactory, verifier identity.Verifier, ipResolver ip.Resolver, consumerPinger natConsumerPinger, portPool port.ServicePortSupplier, eventBus eventbus.EventBus) Dialer {
+func NewDialer(broker brokerConnector, signer identity.SignerFactory, verifierFactory identity.VerifierFactory, ipResolver ip.Resolver, consumerPinger natConsumerPinger, portPool port.ServicePortSupplier, eventBus eventbus.EventBus) Dialer {
 	return &dialer{
-		broker:         broker,
-		ipResolver:     ipResolver,
-		signer:         signer,
-		verifier:       verifier,
-		portPool:       portPool,
-		consumerPinger: consumerPinger,
-		eventBus:       eventBus,
+		broker:          broker,
+		ipResolver:      ipResolver,
+		signer:          signer,
+		verifierFactory: verifierFactory,
+		portPool:        portPool,
+		consumerPinger:  consumerPinger,
+		eventBus:        eventBus,
 	}
 }
 
 // dialer implements Dialer interface.
 type dialer struct {
-	portPool       port.ServicePortSupplier
-	broker         brokerConnector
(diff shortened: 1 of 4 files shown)


p2p/dialer.go
```diff
@@ -188,7 +188,8 @@ func (m *dialer) startConfigExchange(config *p2pConnectConfig, ctx context.Conte
 	}
 
 	// Parse provider response with public key and encrypted and signed connection config.
-	exchangeMsgReplySignedMsg, err := unpackSignedMsg(m.verifier, exchangeMsgBrokerReply)
+	// TODO: to verifier factory
```
soffokl, line 191: What does this TODO mean? It's not clear to me. Could you explain it more clear?

Snawoot: Sorry, forgot to remove. Now it's gone. Was a mark for me I accidentally left.

soffokl (approved):

vkuznecovas (approved):

mdomasevicius (approved):

vkuznecovas:
It might be a good idea to do this on the provider side as well.

Snawoot:
@vkuznecovas We don't know up front which identity we're expecting. But I think we don't need to. Process which happens here resembles signed DH exchange in TLS, and just like in TLS only one party has to be necessarily verified.

tadaskay (approved):

PR #3702, opened 2021-07-28
wg endpoint: Ensure RAII on configure

### LINKED ISSUE #3560: Network interfaces are not cleaned up after session ends
We need to clean up the network interfaces created for VPN sessions.

Snawoot:
Fix #3560 

Ensures RAII on Wireguard interface configuration routines in a way so caller either will have interface configured or will get an error and will have no need to clean up resources allocated by callee.

Incorrect resource management leads to situations when half-configured interfaces abandoned on interface configuration error.

To reproduce such situation on Linux I use following patch:

```diff
diff --git a/utils/netutil/network_linux.go b/utils/netutil/network_linux.go
index 252bb958..eff3002b 100644
--- a/utils/netutil/network_linux.go
+++ b/utils/netutil/network_linux.go
@@ -20,6 +20,7 @@
 package netutil
 
 import (
+	"errors"
 	"net"
 	"os/exec"
 	"strings"
@@ -37,6 +38,7 @@ func assignIP(iface string, subnet net.IPNet) error {
 }
 
 func excludeRoute(ip, gw net.IP) error {
+	return errors.New("Debugging error")
 	return cmdutil.SudoExec("ip", "route", "add", ip.String(), "via", gw.String())
 }
 
```

### DIFF
--- a/utils/actionstack/actionstack.go
+++ b/utils/actionstack/actionstack.go
@@ -1,0 +1,45 @@
+/*
+ * Copyright (C) 2017 The "MysteriumNetwork/node" Authors.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package actionstack
+
+// Action represents stackable action.
+type Action func()
+
+// ActionStack is a stack of actions which are executed in reverse order
+// they were added
+type ActionStack struct {
+	stack []Action
+}
+
+// NewActionStack creates empty ActionStack
+func NewActionStack() *ActionStack {
+	return new(ActionStack)
+}
+
+// Run executes pushed actions in reverse order (FILO)
+func (a *ActionStack) Run() {
+	for i := len(a.stack) - 1; i >= 0; i-- {
+		a.stack[i]()
+	}
+}
+
+// Push adds new action on top of stack. Last added action will be executed
+// by Run() first.
+func (a *ActionStack) Push(elems ...Action) {
+	a.stack = append(a.stack, elems...)
+}
--- a/utils/actionstack/actionstack.go
+++ b/utils/actionstack/actionstack.go
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2017 The "MysteriumNetwork/node" Authors.
+ * Copyright (C) 2021 The "MysteriumNetwork/node" Authors.
  *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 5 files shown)


Snawoot:
Going to move rollback routines into common utility.

Snawoot:
Ready for review.

utils/actionstack/actionstack.go
```diff
@@ -0,0 +1,45 @@
+/*
+ * Copyright (C) 2017 The "MysteriumNetwork/node" Authors.
```
soffokl, line 2:
```suggestion
 * Copyright (C) 2021 The "MysteriumNetwork/node" Authors.
```

utils/actionstack/actionstack_test.go
```diff
@@ -0,0 +1,49 @@
+/*
+ * Copyright (C) 2017 The "MysteriumNetwork/node" Authors.
```
soffokl, line 2:
```suggestion
 * Copyright (C) 2021 The "MysteriumNetwork/node" Authors.
```

utils/actionstack/actionstack.go
```diff
@@ -0,0 +1,45 @@
...
+func (a *ActionStack) Run() {
+	for i := len(a.stack) - 1; i >= 0; i-- {
+		a.stack[i]()
+	}
+}
+
+// Push adds new action on top of stack. Last added action will be executed
+// by Run() first.
+func (a *ActionStack) Push(elems ...Action) {
+	a.stack = append(a.stack, elems...)
```
soffokl, line 44: If this going to be reused in other places, we need to protect it from race conditions. Pushs can be called from different goroutines at the same time.

utils/actionstack/actionstack.go
```diff
@@ -0,0 +1,45 @@
...
+	stack []Action
+}
+
+// NewActionStack creates empty ActionStack
+func NewActionStack() *ActionStack {
+	return new(ActionStack)
+}
+
+// Run executes pushed actions in reverse order (FILO)
+func (a *ActionStack) Run() {
```
soffokl, line 35: Should we protect `Run` from being executed more than once?

soffokl (approved):

utils/actionstack/actionstack.go
Snawoot: I think we should panic if run more than once or action was added after Run. Going to add it.

utils/actionstack/actionstack.go
Snawoot: I think we should panic if run more than once or action was added after Run. Going to add it.

utils/actionstack/actionstack.go
Snawoot: Not sure what user will expect calling Push in parallel since it doesn't guarantees order, but I'll protect access with mutex.

utils/actionstack/actionstack.go
Snawoot: Applied

utils/actionstack/actionstack.go
Snawoot: @soffokl Applied

utils/actionstack/actionstack.go
vkuznecovas: I'd rather we'd returned an error instead of a panic.

Snawoot: It's essentially same situation when someone tries to take Lock twice - should never happen and if happens something is terribly wrong. But those who want call Run carelessly may wrap it with utils.CallOnce decorator.

vkuznecovas (approved):

Issue #3698, opened 2021-07-28
p2p connection fails due to late STUN responses


Snawoot:
Sometimes my node (versions 0.51.0, 0.53.0) fails to connect to a provider. I see following lines in daemon log:

```
2021-07-27T17:36:11.126 INF home/user/src/node/nat/traversal/pinger.go:169 > NAT pinging to remote peer
2021-07-27T17:36:11.126 INF home/user/src/node/nat/traversal/pinger.go:387 > Local socket: 0.0.0.0:43239
2021-07-27T17:36:11.126 INF home/user/src/node/nat/traversal/pinger.go:387 > Local socket: 0.0.0.0:46206
2021-07-27T17:36:11.130 INF home/user/src/node/nat/traversal/pinger.go:387 > Local socket: 0.0.0.0:44198
2021-07-27T17:36:11.130 INF home/user/src/node/nat/traversal/pinger.go:387 > Local socket: 0.0.0.0:41324
2021-07-27T17:36:11.135 INF home/user/src/node/nat/traversal/pinger.go:387 > Local socket: 0.0.0.0:45237
2021-07-27T17:36:11.135 INF home/user/src/node/nat/traversal/pinger.go:387 > Local socket: 0.0.0.0:45442
2021-07-27T17:36:11.137 INF home/user/src/node/nat/traversal/pinger.go:387 > Local socket: 0.0.0.0:47574
2021-07-27T17:36:11.138 INF home/user/src/node/nat/traversal/pinger.go:387 > Local socket: 0.0.0.0:41613
2021-07-27T17:36:11.139 INF home/user/src/node/nat/traversal/pinger.go:387 > Local socket: 0.0.0.0:43083
2021-07-27T17:36:11.144 INF home/user/src/node/nat/traversal/pinger.go:387 > Local socket: 0.0.0.0:43659
2021-07-27T17:36:11.162 INF home/user/src/node/nat/traversal/pinger.go:387 > Local socket: 0.0.0.0:40122
2021-07-27T17:36:11.175 INF home/user/src/node/nat/traversal/pinger.go:387 > Local socket: 0.0.0.0:41156
2021-07-27T17:36:11.178 INF home/user/src/node/nat/traversal/pinger.go:387 > Local socket: 0.0.0.0:47368
2021-07-27T17:36:11.181 INF home/user/src/node/nat/traversal/pinger.go:327 > Remote peer data received: OK, len: 2, from: 37.208.66.85:61424
[...]

PR #3707, opened 2021-07-29
Receive pings only from fixed expected peer

soffokl:
Closes https://github.com/mysteriumnetwork/node/issues/3698

### DIFF
--- a/nat/traversal/pinger.go
+++ b/nat/traversal/pinger.go
@@ -303,7 +303,7 @@ func (p *Pinger) ping(ctx context.Context, conn *net.UDPConn, remoteAddr *net.UD
 	}
 }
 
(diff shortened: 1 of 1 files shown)


tomasmik (approved):

Snawoot:
@soffokl I think if we'll receive "continuously pinging to " we can be pretty sure we're contacting with right address, no matter which remote address we see. It may be useful for weird cases you were talking about, when address is not that we expect. Can we rely on ping message alone?

soffokl:
yep, right, removed it, we will see if this causes any problem in the future.

Snawoot (approved):

vkuznecovas (approved):

mdomasevicius (approved):

PR #3728, opened 2021-08-03
Remove BC subscriptions

### LINKED ISSUE #3683: Refactor node to avoid using websockets

vkuznecovas:
Closes #3683


core/node/options_payments.go
```diff
@@ -28,8 +28,14 @@ type OptionsPayments struct {
...
 	HermesPromiseSettlingThreshold float64
 	SettlementTimeout              time.Duration
+	SettlementRecheckInterval      time.Duration
 	ConsumerDataLeewayMegabytes    uint64
 	ProviderInvoiceFrequency       time.Duration
 	MaxUnpaidInvoiceValue          *big.Int
 	HermesStatusRecheckInterval    time.Duration
+	BalanceFastPollInterval        time.Duration
+	BalanceFastPollTimeout         time.Duration
+	BalanceLongPollInterval        time.Duration
```
mdomasevicius, line 38: Pattern suggests there should be `BalanceLongPollTimeout` :)

vkuznecovas: the long poll never times out, it's continuous and inifite.

identity/registry/registry_contract.go
```diff
@@ -204,109 +218,78 @@ func (registry *contractRegistry) handleRegistrationEvent(ev IdentityRegistratio
...
+	if err != nil {
+		log.Err(err).Msg("could not check registration status on chain")
+		registry.saveRegistrationStatus(chainID, id, RegistrationError)
+		return
+	}
+	registry.saveRegistrationStatus(chainID, id, status)
+}
+
+func (registry *contractRegistry) saveRegistrationStatus(chainID int64, id string, status RegistrationStatus) {
+	registry.publisher.Publish(AppTopicIdentityRegistration, AppEventIdentityRegistration{
```
tomasmik, line 279: maybe it would be better if we stored first then published? Then we'd know that what we published is the same thing we've stored.

identity/registry/transactor.go
```diff
@@ -444,16 +445,39 @@ type SettleWithBeneficiaryRequest struct {
...
+// QueueResponse represents the queue response from transactor.
+type QueueResponse struct {
+	ID    string `json:"id"`
+	Hash  string `json:"tx_hash"`
+	State string `json:"state"`
+	Error string `json:"error"`
+}
+
+// GetQueueStatus returns the queued transaction status from transactor.
+func (t *Transactor) GetQueueStatus(ID string) (QueueResponse, error) {
```
tomasmik, line 457: This is not used or am I not finding references?

session/pingpong/consumer_balance_tracker.go
```diff
@@ -118,9 +133,46 @@ func (cbt *ConsumerBalanceTracker) Subscribe(bus eventbus.Subscriber) error {
...
+	}
 	return bus.SubscribeAsync(identity.AppTopicIdentityUnlock, cbt.handleUnlockEvent)
 }
 
+func (cbt *ConsumerBalanceTracker) handleOrderUpdatedEvent(ev pevent.AppEventOrderUpdated) {
+	if ev.Status.Incomplete() {
+		return
+	}
+
+	// pilvytis only allows topups on the default chain
```
tomasmik, line 148: Topup requests are created using `FlagChainID` flag. So in theory you can create topup request on a different chains.
Pilvytis also allows a list chain id's to be used for topping up, (it's just that for now there is only one item in the list.)

So i guess at least we could use the flag here and not the default constant.

tomasmik (approved):

identity/registry/transactor.go
vkuznecovas: ur just not finding the ref probably due to interface.

mdomasevicius (approved):

Issue #2004, opened 2020-04-05
Connection cancel doesn't work while connecting


cvl:
1. Call http.put('connection', request, timeout)
2. While still not connected (Connection in progress) -> call http.delete('connection')
3. Instead of canceling the connection in progress, it doesn't do anything, and after a while unwanted connection is established.

You can try provider 0x8d4707b19fa02ba1de6ce09b9796e4e68e260f07 to test this, as it takes long time to establish conn with this one.

cvl:
Note, that in Android node throws CancellationException, which is correct behavior.

cvl:
Another identity (India) which doesn't let to disconnect ever, until app relaunch: 0x24a6571fc0017b2b0c05ab955ded62f2a8e01b3b

chompomonim:
Could you try with the latest macOS app build (freshly installed)? Is this still happening?

cvl:
When I disconnect now, it throws
PUT http://127.0.0.1:4050/connection 500 (Internal Server Error)
Could not connect Request failed with status code 500 (path="connection")


But it behaves correctly (i.e. disconnects).

cvl:
don't

chompomonim:
I believe this is not a problem of desktop app but node's problem itself. We'll recheck scenario described in task description and if it a case, will fix it or create feature request with more details what should be done to achieve wanted experience.

chompomonim:
@cvl could you ellaborate on steps to reproduce?

> 1. Call http.put('connection', request, timeout). 
> 2. While still not connected (Connection in progress) -> call http.delete('connection')

These are tequilapi calls from JavaScript app, right?

cvl:
Yes, but they're being called by tequila api, specifically:
tequilapi.connectionCancel()

Here's how it looks when attempting to disconnect. In this attempt it just hung up in "Connecting" state, have to relaunch the app.

<img width="695" alt="Screenshot 2021-06-24 at 15 33 25" src="https://user-images.githubusercontent.com/100252/123263536-bdbe9700-d501-11eb-8997-0a2711116a24.png">

chompomonim:
Have you checked it only on Mac or this issue is reproducible only on Mac and Windows is ok?

cvl:
Same on Windows

tomasmik:
Tested on linux, disconnect seems to work, but together with the disconnect it also brings down DNS configuration (and maybe something else?) and I'm forced to restart the machine. Will investigate further.

[...]

PR #3733, opened 2021-08-04
NAT compatibility APIs

Snawoot:
This PR introduces `nat_compatibility` filter for proposals in Tequila API and Mobile API. Also it adds new `/nat/type` endpoint which returns type of NAT.

Known issues which are considered acceptable for now:

* NAT probing will produce incorrect result while VPN is on. For this reason list of proposals with filter `nat_compatibility=auto` may not work as expected while VPN is enabled.
* Proposal request with `nat_compatibility=auto` will have increased delay (about 2-3 seconds) because of probing.


mobile/mysterium/proposals_manager.go
```diff
@@ -165,7 +174,16 @@ func (m *proposalsManager) addToCache(proposals []proposal.PricedServiceProposal
 	m.cachedAt = time.Now()
 }
 
-func (m *proposalsManager) getFromRepository(filter *proposal.Filter) ([]proposal.PricedServiceProposal, error) {
+func (m *proposalsManager) getFromRepository(req *GetProposalsRequest) ([]proposal.PricedServiceProposal, error) {
+	filter := req.toFilter()
+	if filter.NATCompatibility == "auto" {
+		natType, err := m.natProber.Probe(context.TODO())
```
soffokl, line 180: `context.TODO()` shows that something incomplete with it.

cmd/natdisco/main.go
```diff
@@ -22,40 +22,43 @@ import (
...
 	"github.com/mysteriumnetwork/node/nat/behavior"
 )
 
+const addressListSeparator = ";"
+
 var (
-	address      = flag.String("server", "stun.stunprotocol.org:3478", "STUN server address")
-	reqTimeout   = flag.Duration("req-timeout", 3*time.Second, "timeout to wait for each STUN server response")
-	totalTimeout = flag.Duration("total-timeout", 15*time.Second, "overall operation deadline")
+	addressList  = flag.String("servers", "stun.mysterium.network:3478;stun.stunprotocol.org:3478;stun.sip.us:3478", "semicolon-separated list of STUN servers")
```
soffokl, line 34: I think we are using comma-separated lists in other places, it will be good to keep it consistent. 
```suggestion
	addressList  = flag.String("servers", "stun.mysterium.network:3478;stun.stunprotocol.org:3478;stun.sip.us:3478", "semicolon-separated list of STUN servers")
```

nat/behavior/race.go
```diff
@@ -0,0 +1,71 @@
...
+
+	lastError := ErrEmptyAddressList
+	for i := 0; i < count; i++ {
+		select {
+		case res := <-results:
+			if res.err != nil {
+				lastError = res.err
+			} else {
+				return res.res, res.err
+			}
```
soffokl, line 64: Is this more readable? 
```suggestion
			if res.err == nil {
				return res.res, nil
			} 
			lastError = res.err
```

tequilapi/endpoints/nat.go
```diff
@@ -50,9 +58,32 @@ func (ne *NATEndpoint) NATStatus(resp http.ResponseWriter, _ *http.Request, _ ht
 	utils.WriteAsJSON(ne.stateProvider.GetState().NATStatus, resp)
 }
 
+// NATType provides NAT type in terms of traversal capabilities
+// swagger:operation GET /nat/type NAT NATTypeDTO
+// ---
+// summary: Shows NAT type in terms of traversal capabilies
```
soffokl, line 64:
```suggestion
// summary: Shows NAT type in terms of traversal capabilities.
```

tequilapi/endpoints/nat.go
```diff
@@ -50,9 +58,32 @@ func (ne *NATEndpoint) NATStatus(resp http.ResponseWriter, _ *http.Request, _ ht
...
+// description: Returns NAT type. May produce invalid result while VPN connection is established
+// responses:
+//   200:
```
soffokl, line 67: Only 200 status is described, but based on code 500 is possible too.

soffokl (approved):

cmd/commands/cli/command.go
```diff
@@ -431,6 +433,29 @@ func (c *cliApp) natStatus() (err error) {
...
 		clio.Infof("NAT traversal status: %q (error: %q)\n", status.Status, status.Error)
 	}
+
+	connStatus, err := c.tequilapi.ConnectionStatus()
+	if err != nil {
+		clio.Warn(err)
+		return
+	}
+
+	if connStatus.Status == statusNotConnected {
```
vkuznecovas, line 443:
```suggestion
	if connStatus.Status != statusNotConnected {
	  return nil
	}
```

Then you can unnest the contents

vkuznecovas (approved):

soffokl:
It will be good to enable the proposal filtering for CLI `proposals` command too.

Snawoot:
@soffokl CLI filtering has own very basic filters and it's filters nothing like filters in API. It just matches substrings. In order to implement it, we'll have to reimplement current CLI filters. I guess it should be done as a separate feature.

tequilapi/endpoints/proposals.go
```diff
@@ -119,6 +125,16 @@ func (pe *proposalsEndpoint) List(resp http.ResponseWriter, req *http.Request, _
 		return float32(f)
 	}()
 
+	natCompatibility := req.URL.Query().Get("nat_compatibility")
+	if natCompatibility == "auto" {
```
tomasmik, line 129: Could we get a constant for this somewhere?

cmd/commands/cli/command.go
```diff
@@ -431,6 +433,30 @@ func (c *cliApp) natStatus() (err error) {
...
 		clio.Infof("NAT traversal status: %q (error: %q)\n", status.Status, status.Error)
 	}
+
+	connStatus, err := c.tequilapi.ConnectionStatus()
+	if err != nil {
+		clio.Warn(err)
+		return
+	}
+
+	if connStatus.Status != statusNotConnected {
```
tomasmik, line 443: I think you can get the same constant from package `connectionstate` -> `connectionstate.NotConnected`
So you dont have to define it again.

cmd/natdisco/main.go
Snawoot: Updated

mobile/mysterium/proposals_manager.go
Snawoot: That's right, getFromRepository function should also have contexts. Only top-level callers should use context.Background. I suspect contexts largely missing from our interfaces. For this reason we have hard to fix problem like this: #2004

tequilapi/endpoints/nat.go
Snawoot: Updated

tequilapi/endpoints/proposals.go
Snawoot: Updated

cmd/commands/cli/command.go
Snawoot: Updated

mdomasevicius (approved):
Nice commit composition

PR #3743, opened 2021-08-05
NAT-related enhancements

Snawoot:
Changes:
* Return error if probe attempt happens while Consumer node is connected
* Use auto NAT compatibility for proposals requests from CLI (requested by @soffokl)

### DIFF
--- a/nat/behavior/service.go
+++ b/nat/behavior/service.go
@@ -19,7 +19,10 @@ package behavior
 
 import (
 	"context"
+	"errors"
 	"time"
+
+	"github.com/mysteriumnetwork/node/core/connection/connectionstate"
 )
 
 // Only servers compatible with RFC 5780 are usable!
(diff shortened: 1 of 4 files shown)


tomasmik (approved):

nat/behavior/service.go
```diff
@@ -31,18 +34,32 @@ var compatibleSTUNServers = []string{
...
+
 // NewNATProber constructs some suitable NATProber without any implementation
 // guarantees.
-func NewNATProber() NATProber {
-	return newConcurrentNATProber(compatibleSTUNServers, concurrentRequestTimeout)
+func NewNATProber(connStatusProvider ConnectionStatusProvider) NATProber {
+	var prober NATProber
+	prober = newConcurrentNATProber(compatibleSTUNServers, concurrentRequestTimeout)
+	prober = newGatedNATProber(connStatusProvider, prober)
+	return prober
```
soffokl, line 59: Can we make it like this? I was confusing me for some time all these assignment to the same var. 
```suggestion
	prober := newConcurrentNATProber(compatibleSTUNServers, concurrentRequestTimeout)
	return newGatedNATProber(connStatusProvider, prober)
```

soffokl (approved):

nat/behavior/service.go
Snawoot: We can, but we have no need to. When we'll need to add yet another "prober" into this chain (e.g. one which hooks on result and sends metrics) it will be more readable if we'll have it like this:

```
+    prober = newSnoopingNATProber(prober, qualitySender)
```

PR #3753, opened 2021-08-10
Port check routine

Snawoot:
This PR implements external UDP port check routine.

Not integrating routine into node in this PR for two reasons:

* Keep size reviewable.
* @soffokl is now already working on changes in P2PListener which I'm about to touch too.

### DIFF
--- a/core/port/reachability.go
+++ b/core/port/reachability.go
@@ -107,7 +107,7 @@ func GloballyReachable(ctx context.Context, port Port, echoServerAddresses []str
 	}
 
 	// Await response
-	ctx1, cl := context.WithTimeout(ctx, timeout)
+	ctx, cl := context.WithTimeout(ctx, timeout)
 	defer cl()
 	responseChan := make(chan struct{})
 
@@ -131,7 +131,7 @@ func GloballyReachable(ctx context.Context, port Port, echoServerAddresses []str
 				select {
 				case responseChan <- struct{}{}:
 					return
-				case <-ctx1.Done():
+				case <-ctx.Done():
 					return
 				}
 			}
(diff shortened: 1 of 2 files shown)


core/port/reachability.go
```diff
@@ -0,0 +1,148 @@
...
+		return false, err
+	}
+	copy(msg[portFieldSize:], probeUUID[:])
+
+	// Send probes. Proceed to listen after first send success.
+	sendResultChan := make(chan error)
+
+	for _, address := range echoServerAddresses {
+		go func(echoServerAddress string) {
+			err := func() error {
```
vkuznecovas, line 75: seems like a perfect candidate to extract to a separate function

core/port/reachability.go
```diff
@@ -0,0 +1,148 @@
...
+			break
+		}
+	}
+
+	if err != nil {
+		return false, fmt.Errorf("every port probe send failed. last error: %w", err)
+	}
+
+	// Await response
+	ctx1, cl := context.WithTimeout(ctx, timeout)
```
vkuznecovas, line 110:
```suggestion
	ctx, cl := context.WithTimeout(ctx, timeout)
```

core/port/reachability.go
```diff
@@ -0,0 +1,148 @@
...
+	defer cl()
+	responseChan := make(chan struct{})
+
+	// Background context-aware receiver
+	go func() {
```
vkuznecovas, line 115: Id also extract this as a separate func, can assign a comprehendable name to it then.

vkuznecovas (approved):

tomasmik (approved):

core/port/reachability.go
Snawoot: Applied

core/port/reachability.go
Snawoot: Applied

PR #3755, opened 2021-08-10
NAT Status v2

mdomasevicius:
Tracks NAT Status of wireguard service through metrics


core/node/options.go
```diff
@@ -82,6 +82,8 @@ type Options struct {
 
 	SwarmDialerDNSHeadstart time.Duration
 	PilvytisAddress         string
+
+	NATStatusTrackerV2 OptionsNATStatusTrackerV2
```
tomasmik, line 86: I think you've missed the part where you set the value in `GetOptions`
Also you probably need to set it for mobile somewhere?

config/flags_node.go
```diff
@@ -270,6 +270,13 @@ var (
 		Name:  "resident-country",
 		Usage: "set resident country. If not set initially a default country will be resolved.",
 	}
+
+	// FlagNATStatusPollInterval nat status poll interval in seconds
+	FlagNATStatusPollInterval = cli.DurationFlag{
+		Name:  "nat-status.poll",
+		Usage: `NAT status poll interval in seconds. Lowest interval of 1 minute is allowed.`,
+		Value: 5 * 60 * time.Second,
```
tomasmik, line 278: Isn't this just `5 * time.Minute`? I mean for time.Duration [it doesnt matter](https://play.golang.org/p/RCitzq3VLmt) how you give the time.
Also the `Usage` comment is a little weird. Why in seconds? I think it'd actually work if given like `3m43s` 🤷🏻

nat/nat_status_v2.go
```diff
@@ -0,0 +1,169 @@
...
+		keeper.startPolling()
+	}
+	return keeper
+}
+
+func (k *StatusTrackerV2) startPolling() {
+	k.updateAndAnnounce()
+	go func() {
+		for {
+			select {
```
tomasmik, line 114: No channels to select from so the `select` is not really needed here

nat/nat_status_v2.go
```diff
@@ -0,0 +1,169 @@
...
+func (k *StatusTrackerV2) updateAndAnnounce() {
+	k.lock.Lock()
+	defer k.lock.Unlock()
+	id, ok := k.currentIdentity.GetUnlockedIdentity()
+	if ok {
+		k.status = resolveNATStatus(k.providerSessions(id.Address))
+	} else {
+		k.status = Pending
+	}
+	k.publisher.Publish(AppTopicNATStatusUpdate, V2NatStatusEvent{Status: k.status})
```
tomasmik, line 131: maybe its not worth to publish the event if the status has not changed? or you could also handle this on the other end.

Snawoot (approved):

nat/nat_status_v2.go
```diff
@@ -0,0 +1,169 @@
...
+}
+
+// StatusForce force update status and return current result
+func (k *StatusTrackerV2) StatusForce() NATStatusV2 {
+	k.lock.Lock()
```
tomasmik, line 136: This will lock mutex then call `updateAndAnnounce` which will lock the mutex again.

core/state/event/event.go
```diff
@@ -39,6 +39,8 @@ const AppTopicState = "State change"
 
 // State represents the node state at the current moment. It's a read only object, used only to display data.
 type State struct {
+	Nat Nat
+	// Deprecated: use state.nat
```
tomasmik, line 43: probably meant event.Nat?

nat/nat_status_v2.go
Snawoot: or even just use time.Sleep

core/node/options.go
mdomasevicius: AFAIK it's useless for mobile/consumers

nat/nat_status_v2.go
```diff
@@ -0,0 +1,168 @@
...
+	}
+	if prevStatus != k.status {
+		k.publisher.Publish(AppTopicNATStatusUpdate, V2NatStatusEvent{Status: k.status})
+	}
+}
+
+// StatusForce force update status and return current result
+func (k *StatusTrackerV2) StatusForce() NATStatusV2 {
+	k.updateAndAnnounce()
+	return k.status
```
tomasmik, line 138: now without the mutexes this could be a race no?

tomasmik (approved):

config/flags_node.go
```diff
@@ -270,6 +270,13 @@ var (
 		Name:  "resident-country",
 		Usage: "set resident country. If not set initially a default country will be resolved.",
 	}
+
+	// FlagNATStatusPollInterval nat status poll interval in seconds
+	FlagNATStatusPollInterval = cli.DurationFlag{
+		Name:  "nat-status.poll",
+		Usage: `NAT status poll interval. (lower than 1 minute wil be set to 1 minute)`,
```
soffokl, line 277:
```suggestion
		Usage: `NAT status poll interval. (lower than 1 minute will be set to 1 minute)`,
```

nat/nat_status_v2.go
```diff
@@ -0,0 +1,167 @@
...
+	}()
+}
+
+func (k *StatusTrackerV2) updateAndAnnounce() {
+	k.lock.Lock()
+	defer k.lock.Unlock()
+	id, ok := k.currentIdentity.GetUnlockedIdentity()
+	prevStatus := k.status
+	if ok {
+		k.status = resolveNATStatus(k.providerSessions(id.Address))
```
soffokl, line 125: Is it really provider `sessions`? I thought we need to track the provider service state. 
What if we will have service stopped, but a historical session available? Will it show `offline` or `online`?

mdomasevicius: `Online` -> `Passed`
`Offline` -> removed

Will only return monitoring agent status. Service status can be already acquired elsewhere.

Issue #3588, opened 2021-07-02
Allow changing traversal priorities


chompomonim:
At the moment traversal mechanism priorities are hard-coded.

1. Check if node is running on **public address**.
2. Try using **UPnP** 
3. **NAT hole punching**
4. Manual port forwarding. 

We could make that priorities configurable via config. The same config could disable some of them (e.g. if `nat_traversal: ["upnp", "port_forwarding"] then nat hole punching and public address checks could be not done at all)

If possible, we could show how many of last 100 sessions were used by which configuration method + their success rate

> NOTE: This would allow make NAT punching as seconds step instead of third because NAT hole punching could be successful in bigger percentage of sessions than UPnP.

Snawoot:
I'm asking to clarify how this task relates to #3585

If we enforce additional check against external services on each forwarded port with UPnP, we may not care about priorities.

soffokl:
This is about giving an option to the user to set the order for the checks. If a user knows that UPnP works for him, he can set this as a first option and not waste time on other options. 
Doing all the checks takes time and these options will give a possibility to make connections faster. 

Most of the users will live with an automatic option though.

PR #3758, opened 2021-08-11
Allow to set traversal methods order

soffokl:
https://github.com/mysteriumnetwork/node/issues/3588

### DIFF
--- a/config/flags_network.go
+++ b/config/flags_network.go
@@ -114,7 +114,7 @@ var (
 		Usage: "Range of UDP listen ports used for connections",
 		Value: "10000:60000",
 	}
-	// FlagTraversalOrder order of NAT traversal methods to be used for providing service.
+	// FlagTraversal order of NAT traversal methods to be used for providing service.
 	FlagTraversal = cli.StringFlag{
 		Name:  "traversal",
 		Usage: "Comma separated order of NAT traversal methods to be used for providing service",
--- a/config/flags_network.go
+++ b/config/flags_network.go
@@ -114,6 +114,12 @@ var (
 		Usage: "Range of UDP listen ports used for connections",
 		Value: "10000:60000",
 	}
+	// FlagTraversalOrder order of NAT traversal methods to be used for providing service.
+	FlagTraversal = cli.StringFlag{
+		Name:  "traversal",
+		Usage: "Comma separated order of NAT traversal methods to be used for providing service",
+		Value: "holepunching,upnp,manual",
+	}
 )
 
 // RegisterFlagsNetwork function register network flags to flag list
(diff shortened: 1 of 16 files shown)


Snawoot (approved):

vkuznecovas (approved):

config/flags_network.go
```diff
@@ -114,6 +114,12 @@ var (
 		Usage: "Range of UDP listen ports used for connections",
 		Value: "10000:60000",
 	}
+	// FlagTraversalOrder order of NAT traversal methods to be used for providing service.
+	FlagTraversal = cli.StringFlag{
+		Name:  "traversal",
+		Usage: "Comma separated order of NAT traversal methods to be used for providing service",
```
tomasmik, line 120: Would it be useful to provide possible values here?

p2p/listener.go
```diff
@@ -304,60 +293,23 @@ func (m *listener) providerStartConfigExchange(providerID identity.Identity, msg
 // required ports count for actual p2p and service connections and fallback to
 // acquiring extra ports for nat pinger if provider is behind nat, port mapping failed
 // and no manual port forwarding is enabled.
-func (m *listener) prepareLocalPorts(id, outboundIP string, tracer *trace.Tracer) (string, []int, []func(), error) {
+func (m *listener) prepareLocalPorts(id string, tracer *trace.Tracer) (string, []int, []func(), nat.StartPorts, error) {
```
tomasmik, line 296: At some point the returns values here should be remade. Adding one more return here is just more confusion.

mdomasevicius: It could be wrapped into struct with meaningful names attached

config/flags_network.go
soffokl: we are including all values as default. Here is how it looks in help:
```
   --traversal value                                      Comma separated order of NAT traversal methods to be used for providing service (default: "holepunching,upnp,manual")
```

p2p/nat/hole_punching.go
```diff
@@ -0,0 +1,66 @@
...
+}
+
+func NewNatHolePunchingPortProvider() PortProvider {
+	udpPortRange, err := port.ParseRange(config.GetString(config.FlagUDPListenPorts))
+	if err != nil {
+		log.Warn().Err(err).Msg("Failed to parse UDP listen port range, using default value")
+	}
+
+	return &natHolePunchingPort{
+		pool:   port.NewFixedRangePool(*udpPortRange),
```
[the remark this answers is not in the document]
soffokl: Yep, faced it with tests too, changed.

p2p/listener.go
soffokl: I'm going to change it with the next couple of PRs. So merging like this now.

PR #3766, opened 2021-08-12
Check if ports are accessible via external service

soffokl:


p2p/nat/upnp.go
```diff
@@ -79,6 +79,11 @@ func (up *upnpPort) PreparePorts() (ports []int, release func(), start StartPort
 		ports = append(ports, p.Num())
 	}
 
+	if err := checkAllPorts(ports); err != nil {
+		log.Debug().Err(err).Msgf("Failed to check manual ports %d globally", ports)
```
Snawoot, line 83:
```suggestion
		log.Debug().Err(err).Msgf("Failed to check UPnP ports %d globally", ports)
```

Snawoot (approved):

config/flags_network.go
```diff
@@ -118,7 +118,13 @@ var (
 	FlagTraversal = cli.StringFlag{
 		Name:  "traversal",
 		Usage: "Comma separated order of NAT traversal methods to be used for providing service",
-		Value: "holepunching,upnp,manual",
+		Value: "manual,upnp,holepunching",
+	}
+	// FlagPortCheckServers list of asymmetric UDP echo servers for checking port availability
+	FlagPortCheckServers = cli.StringFlag{
```
vkuznecovas, line 124: make this hidden, users will have little clue as to what it is and are unlikely to use this, ever. This just adds another flag to an already confusing list :)

vkuznecovas (approved):

p2p/nat/check.go
```diff
@@ -0,0 +1,86 @@
...
+func checkAllPorts(ports []int) error {
+	var addresses []string
+
+	for _, address := range strings.Split(config.GetString(config.FlagPortCheckServers), ",") {
+		address = strings.TrimSpace(address)
+		if address != "" {
+			addresses = append(addresses, address)
+		}
+	}
+
```
tomasmik, line 52: Add a check to see if you even have any addresses to use? Not point to continue if not.

tomasmik (approved):

mdomasevicius (approved):

p2p/nat/check.go
soffokl: We already have this check in a level deeper.

PR #3781, opened 2021-08-13
Remove old NAT status event tracking

soffokl:
We are replacing it with new NAT types values.


tequilapi/endpoints/nat.go
```diff
@@ -45,34 +46,6 @@ func NewNATEndpoint(stateProvider stateProvider, natProber natProber) *NATEndpoi
...
-// swagger:operation GET /v2/nat/status NAT
-// ---
-// summary: Shows NAT status
-// description: NAT status returns the last known NAT traversal status
-// responses:
-//   200:
-//     description: NAT status ("passed"/"failed"/"pending)
-//     schema:
-//       "$ref": "#/definitions/NATStatusDTO"
-func (ne *NATEndpoint) NATStatusV2(resp http.ResponseWriter, _ *http.Request, _ httprouter.Params) {
```
mdomasevicius, line 72: I think you should leave this

mdomasevicius (approved):

PR #3797, opened 2021-08-13
Avoid side effects for configs while merging

mdomasevicius:
After fetching merge config, default config map gets overwritten, this may be a cause of some weird issues

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -137,9 +137,9 @@ func (cfg *Config) GetConfig() map[string]interface{} {
 	cfg.mu.RLock()
 	defer cfg.mu.RUnlock()
 	config := make(map[string]interface{})
-	mergeMaps(cfg.defaults, config, nil)
-	mergeMaps(cfg.user, config, nil)
-	mergeMaps(cfg.cli, config, nil)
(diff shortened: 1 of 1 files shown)


tomasmik (approved):

Snawoot (changes requested):
Here is how this routine works:

1. Shared lock acquired for reading. There will be no changes in default, user and cli maps past this point.
2. Temporary map created.
3. This map gets filled with data from other maps. Submaps linked to temporary map by reference.
4. Temporary map deepcopied to result. No references to shared data past this point.
5. Shared lock released.

Why do we need to deepcopy each map we're copying into result map under read lock if result will be deepcopied either way?

Snawoot (approved):
Alright, now I get it. Submaps are linked into new temporary map by reference and therefore will be updated when temp. map will be used as destination in next merge. Good catch!

mdomasevicius:
> Alright, now I get it. Submaps are linked into new temporary map by reference and therefore will be updated when temp. map will be used as destination in next merge. Good catch!

Actually, I was a bit lazy to go over the recursive code and just made `deep copies` as a quick fix. :)

soffokl (approved):

PR #3806, opened 2021-08-16
Introduce /eth-info endpoint for general eth information

mdomasevicius:


tequilapi/endpoints/eth_info.go
```diff
@@ -0,0 +1,66 @@
...
+// summary: ETH Client Info
+// description: Provides active info about eth client
+// response:
+//	200:
+//		schema:
+//			"$ref": "#/definitions/EthInfoResponse"
+func (e ethInfoEndpoint) EthInfo(resp http.ResponseWriter, _ *http.Request, _ httprouter.Params) {
+	var response contract.EthInfoResponse
+	clientURLOrder := e.ethL2URLProvider.CurrentClientOrder()
+	if len(clientURLOrder) != 0 {
```
tomasmik, line 52: It goes like this:
* go through the list of rpc clients, use them all in current order
  * if one doesn't respond push it to the end of the list
  * if one responds stop and return result

So I would honestly return the whole list and explain they're used in order, as it's possible to use 2nd node in the list if the 1st didnt respond. Unless you think that would just confuse the end user.

tequilapi/endpoints/eth_info_test.go
```diff
@@ -0,0 +1,63 @@
...
+func (f *fakeEthClient) CurrentClientOrder() []string {
+	return f.urls
+}
+
+func Test_EthEndpoints(t *testing.T) {
+	endpoint := newEthInfoEndpoint(newFakeEthClient([]string{"one", "two"}))
+	req := httptest.NewRequest(
+		http.MethodPut,
+		"/irrelevant",
+		strings.NewReader(
```
tomasmik, line 47: I assume this was copied from somewhere else? DOnt think you need there? Also the requests isnt supposed to be `put`

tomasmik (approved):

mdomasevicius:
Decided on different approach

PR #3816, opened 2021-08-17
Change description of traffic shaper bandwidth flag

mdomasevicius:
At least from NodeUI perspective it has always been minimum of 5MiB/s

### DIFF
--- a/config/flags_node.go
+++ b/config/flags_node.go
@@ -107,7 +107,7 @@ var (
 	// FlagShaperBandwidth set the bandwidth limit.
 	FlagShaperBandwidth = cli.Uint64Flag{
 		Name:  "shaper.bandwidth",
-		Usage: "Set the bandwidth limit in bytes",
+		Usage: "Set the bandwidth limit in Kbytes",
 		Value: 5000,
 	}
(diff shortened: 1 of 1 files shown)


tomasmik (approved):

config/flags_node.go
```diff
@@ -108,7 +108,7 @@ var (
 	FlagShaperBandwidth = cli.Uint64Flag{
 		Name:  "shaper.bandwidth",
 		Usage: "Set the bandwidth limit in bytes",
-		Value: 5000,
+		Value: 5_000_000,
```
soffokl, line 111: This value is set in Kbytes based on the code that uses it. 
So the description is not correct, not the value. 
If you want to change it to real bytes underlying code needs to be changed.

soffokl (changes requested):

config/flags_node.go
mdomasevicius: I see, I'll change the description then

vkuznecovas (approved):

soffokl (approved):

Snawoot (approved):

PR #3819, opened 2021-08-18
Use only allowed ports for NAT probe

soffokl:


tomasmik (approved):

Snawoot:
Why do we need this?

soffokl:
I case you forwarded part of your ports manually and want to use this range.

vkuznecovas (approved):

soffokl:
We agreed to do needed changes in the nodeUI, this is not needed now.

PR #3832, opened 2021-08-19
Keep log files in docker container too

soffokl:


Snawoot:
Usual practice for dockerized apps is to write logs to stderr and let container runtime redirect and handle them. Having logs inside container will make writable layer of container bloat.

mdomasevicius (approved):

creotiv:
@Snawoot what about to add env param for that, for debug purpose for example?

soffokl:
We have a feature to report issues from the node via NodeUI. 
Uses send these reports to us with one click to help resolve the problem. 
For this, a node should have access to the logs, and missing logs breaks this feature. 
Node controls these logs rotation to keep it with a reasonable size.

tomasmik (approved):

Issue #3833, opened 2021-08-19
Unconnectable Testnet3 providers. could not send p2p session create request: timeout waiting for reply to "p2p-session-create": p2p send timeout


Waldz:
**Describe the bug**
A significant amount of providers in Testnet3 version are unconnectible.

**To Reproduce**
1. Go to hosting droplet
2. Launch `myst` consumer inside Docker container
3. Connect to mentioned provider  with`myst cli connect 0x.. 0x.. wireguard`
4. See an error

**Environment (please complete the following information):**
 - Node version:
```
docker exec -ti supernode-tunnel-0 myst version
Mysterium Node
  Version: 0.60.0
```
 - OS:
 ```
uname -a
Linux menulis 4.19.0-16-amd64 #1 SMP Debian 4.19.181-1 (2021-03-19) x86_64 GNU/Linux

docker exec -ti supernode-tunnel-0 cat /etc/alpine-release
3.12.7
```
 - Docker (if applicable): specify docker version and image tag
 ```
docker version
Client: Docker Engine - Community
 Version:           20.10.7
 API version:       1.41
 Go version:        go1.13.15
 Git commit:        f0df350
 Built:             Wed Jun  2 11:56:47 2021
 OS/Arch:           linux/amd64
 Context:           default
 Experimental:      true

Server: Docker Engine - Community
 Engine:
  Version:          20.10.7
  API version:      1.41 (minimum version 1.12)
  Go version:       go1.13.15
  Git commit:       b0f5bc3
  Built:            Wed Jun  2 11:54:58 2021
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.4.6
  GitCommit:        d71fcd7d8303cbf684402823e425e9dd2e99285d
 runc:
  Version:          1.0.0-rc95
  GitCommit:        b9ee9c6314599f1b4a7f497e1f1f856fe433d3b7
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
 ```

**Comment by Waldz** [comment]
Provider: 0x57daba353514f004c86edf2e3892740ec4d98c4d
Problem: could not send p2p session create request: timeout waiting for reply to "p2p-session-create": p2p send timeout

Logs:
```
[...]

PR #3835, opened 2021-08-19
Disable keepalive for HTTP transport

### LINKED ISSUE #3834: Disable or restrict HTTP keepalive for Go HTTP transports
Caught following problem after VPN connection, but after a while it disappeared:

```
server response invalid: 503 Service Unavailable (http://supernode-tunnel-1:4050/connection/location). Possible error: failed to execute request: Get https://testnet3-location.mysterium.network/api/v1/location: net/http: request canceled (Client.Timeout exceeded while awaiting headers)
```

It seems built-in golang HTTP client wasn't able to reach that resource. While curl was working fine in same time.

It is possible that http.Transport used by http.Client had established HTTP session before VPN connection was established and reused it in keepalive mode, so it appeared hang. If so, it's pretty easy to fix adjusting transport settings or killing all idle connections of HTTP transport on VPN state change.

Snawoot:
Fix #3834 (from #3833)

### DIFF
--- a/requests/transport.go
+++ b/requests/transport.go
@@ -34,5 +34,6 @@ func NewTransport(dialFunc DialContext) *http.Transport {
 		IdleConnTimeout:       90 * time.Second,
 		TLSHandshakeTimeout:   30 * time.Second,
 		ExpectContinueTimeout: 1 * time.Second,
+		DisableKeepAlives:     true,
 	}
 }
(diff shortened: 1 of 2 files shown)


mdomasevicius (approved):
wow

tomasmik (approved):

Snawoot:
Re-requesting review because tests got broken and exposed another problem in pingpong protocol: HTTP request was reused by backoff retry and all consequesnt requests after first went with drained request body (Request.Body is a one-time io.Reader). In order to fix tests for my PR, I've fixed that as well.

tomasmik (approved):
Nice catch and fix

PR #3850, opened 2021-08-24
Swap http.Handler with Gin Gonic

mdomasevicius:
Since we use Gin in all other projects


tomasmik (commented):
Could we add a flag or something to control gin mode? By default both should be loaded with `release` and have the possibility to switch to `debug` if needed. It now prints logs of debug logs which are probably not needed.

cmd/bootstrap.go
```diff
@@ -21,59 +21,75 @@ package cmd
...
+			tequilapi_endpoints.AddRoutesForProposals(di.ProposalRepository, di.PricingHelper, di.LocationResolver, di.FilterPresetStorage, di.NATProber),
+			tequilapi_endpoints.AddRoutesForService(di.ServicesManager, services.JSONParsersByType, di.ProposalRepository),
+			tequilapi_endpoints.AddRoutesForAccessPolicies(di.HTTPClient, config.GetString(config.FlagAccessPolicyAddress)),
+			tequilapi_endpoints.AddRoutesForNAT(di.StateKeeper, di.NATProber),
+			tequilapi_endpoints.AddRoutesForTransactor(di.IdentityRegistry, di.Transactor, di.HermesPromiseSettler, di.SettlementHistoryStorage, di.AddressProvider),
+			tequilapi_endpoints.AddRoutesForConfig,
+			tequilapi_endpoints.AddRoutesForMMN(di.MMN),
+			tequilapi_endpoints.AddRoutesForFeedback(di.Reporter),
+			tequilapi_endpoints.AddRoutesForConnectivityStatus(di.SessionConnectivityStatusStorage),
+			tequilapi_endpoints.AddRoutesForDocs, // TODO smt wrong
```
tomasmik, line 82:
```suggestion
			tequilapi_endpoints.AddRoutesForDocs,
```
Nothing wrong here anymore

tequilapi/endpoints/auth.go
```diff
@@ -113,7 +117,10 @@ func (api *authenticationAPI) Authenticate(httpRes http.ResponseWriter, httpReq
 //     description: Internal server error
 //     schema:
 //       "$ref": "#/definitions/ErrorMessageDTO"
-func (api *authenticationAPI) Login(httpRes http.ResponseWriter, httpReq *http.Request, _ httprouter.Params) {
+func (api *authenticationAPI) Login(c *gin.Context) {
+	httpReq := c.Request
+	httpRes := c.Writer
```
tomasmik, line 122: Well this seems a bit backwards to me. From a quick search it seems that most of the `SendError` or `WriteAsJSON` are only used in tequila and they all write JSON. So in theory functions in `tequilapi/utils/utils.go` could just accept interface:
```
type JSONErrorWriter interface {
	JSON(code int, body interface{})
}
```

And you could just pass `c` into those funcs. I guess this could also be improvement num 2 🤷🏻

tequilapi/endpoints/connection.go
```diff
@@ -287,13 +292,25 @@ type proposalRepository interface {
...
+	publisher eventbus.Publisher,
+	addressProvider addressProvider,
+) func(*gin.Engine) error {
 	connectionEndpoint := NewConnectionEndpoint(manager, stateProvider, proposalRepository, identityRegistry, publisher, addressProvider)
-	router.GET("/connection", connectionEndpoint.Status)
-	router.PUT("/connection", connectionEndpoint.Create)
-	router.DELETE("/connection", connectionEndpoint.Kill)
-	router.GET("/connection/statistics", connectionEndpoint.GetStatistics)
+	return func(e *gin.Engine) error {
+		connGroup := e.Group("/connection")
```
tomasmik, line 305: I would probably go the other way around this:
* Create an empty group `connGroup := e.Group("")`
* Add each handler like `connGroup.GET("/connection", connectionEndpoint.Status)`

Would be a bit easier to read the actions endpoints as they're next to the func call, but maybe thats just me.

tequilapi/endpoints/connection_test.go
```diff
@@ -350,52 +368,70 @@ func TestPutWithServiceTypeOverridesDefault(t *testing.T) {
...
-			"bytes_received": 2,
-			"throughput_sent": 1000,
-			"throughput_received": 2000,
-			"duration": 0,
-			"tokens_spent": 10001
-		}`,
-		resp.Body.String(),
-	)
-}
+//func TestGetStatisticsEndpointReturnsStatistics(t *testing.T) {
```
tomasmik, line 385: I think you forgot one

tequilapi/endpoints/docs_test.go
```diff
@@ -18,57 +18,52 @@
...
-
-	"github.com/julienschmidt/httprouter"
-	"github.com/stretchr/testify/assert"
 )
 
 func Test_Docs(t *testing.T) {
-	// given
-	router := httprouter.New()
-	AddRoutesForDocs(router)
+	/*	// given
```
tomasmik, line 25: Forgot more here too

tequilapi/endpoints/identities.go
```diff
@@ -646,27 +683,24 @@ func AddRoutesForIdentities(
...
+			identityGroup.GET("/:id", idAPI.Get)
+			identityGroup.GET("/:id/status", idAPI.Get)
+			identityGroup.PUT("/:id/unlock", idAPI.Unlock)
+			identityGroup.GET("/:id/registration", idAPI.RegistrationStatus)
+			identityGroup.GET("/:id/beneficiary", idAPI.Beneficiary)
+			identityGroup.GET("/:id/referral", idAPI.GetReferralToken)
+			identityGroup.GET("/:id/referral-available", idAPI.ReferralTokenAvailable)
+			identityGroup.GET("/:id/payout-address", idAPI.GetPayoutAddress)
+			identityGroup.PUT("/:id/payout-address", idAPI.SavePayoutAddress)
+			identityGroup.PUT("/identities/:id/balance/refresh", idAPI.BalanceRefresh)
```
tomasmik, line 701: Group is `identities` so this produces something like `/identities/identities/:id/balance/refresh`. Probably missed when merged master.

tequilapi/endpoints/transactor.go
```diff
@@ -513,22 +544,32 @@ func (te *transactorEndpoint) TokenRewardAmount(resp http.ResponseWriter, reques
...
+		{
+			transGroup.GET("/fees", te.TransactorFees)
+			transGroup.POST("/settle/sync", te.SettleSync)
+			transGroup.POST("/settle/async", te.SettleAsync)
+			transGroup.GET("/settle/history", te.SettlementHistory)
+			transGroup.POST("/stake/increase/sync", te.SettleIntoStakeSync)
+			transGroup.POST("/stake/increase/async", te.SettleIntoStakeAsync)
+			transGroup.POST("/stake/decrease", te.DecreaseStake)
+			transGroup.POST("/settle/withdraw", te.Withdraw)
+			transGroup.GET("/transactor/token/:token/reward", te.TokenRewardAmount)
```
tomasmik, line 571: This will probably also produce a path something like `/transactor/transactor/token/:token/reward`

tequilapi/http_api_server.go
```diff
@@ -36,18 +67,35 @@ type APIServer interface {
...
+	for _, h := range handlers {
+		err := h(g)
+		if err != nil {
+			return nil, err
+		}
+	}
+
 	server := apiServer{
 		errorChannel: make(chan error, 1),
-		handler:      DisableCaching(ApplyCors(handler, corsPolicy)),
```
tomasmik, line 47: Both `DisableCaching` and `ApplyCors` are mostly useless now. You think it's worth keeping it?

tomasmik (changes requested):

tequilapi/endpoints/auth.go
mdomasevicius: Like it says in the commit description: Quick & Dirty, I would like to throw away those utils and maybe swap them with a dependency to standardize this across all projects that use gin. So I would wish to leave this for another iteration :) this PR is large as it is.

tomasmik: Sure no problem, just pointing an issue to open after this is merged for the future.

tomasmik (approved):

soffokl (approved):

Issue #3774, opened 2021-08-12
Add contexts to VPN connection establishing routines


Snawoot:
Currently all code establishing connection doesn't accept go contexts (`context.Context`). It prevents proper implementation of connection cancel (#2004).

We need to refactor all code responsible for connection establishing to meet these two conditions:

* All methods and routines on the way to establishing connection must accept cancellable contexts and fail as soon as possible when context was cancelled/expired.
* On error (no matter caused by cancelled context or not) routines changing OS state must do a proper rollback. For example see use of `utils/actionstack` in wireguard connection service implementation for Linux.

PR #3851, opened 2021-08-24
handle context cancel on wireguard handshake

Guillembonet:
#3774


services/wireguard/connection/connection.go
```diff
@@ -165,7 +165,7 @@ func (c *Connection) start(ctx context.Context, start startConn, options connect
 	log.Info().Msgf("Adding connection peer %s", config.Provider.Endpoint.String())
 
 	log.Info().Msg("Waiting for initial handshake")
-	if err := c.handshakeWaiter.Wait(conn.PeerStats, c.opts.HandshakeTimeout, c.done); err != nil {
+	if err := c.handshakeWaiter.Wait(ctx, conn.PeerStats, c.opts.HandshakeTimeout, c.done); err != nil {
```
tomasmik, line 168: This will now return an error if context is canceled, but will it get handled by:
```
	defer func() {
		if err != nil {
			c.Stop()
		}
	}()
```

Which is above this code? I think that it wont because `err` here is a local variable to the `if` statement. Not sure about the consequences of not calling `c.Stop` on error though, maybe @soffokl  can comment on that.

Guillembonet: in the logs, I see that the Stop function gets called when the connection is cancelled, so it seems to work.

tomasmik (approved):

mdomasevicius (approved):

soffokl:
I think the initial task is a bit wider than just cancel the wireguard handshake wait. 
It is about making the connection establishing process easier to use. 
Connection start functions should have this context too. Device and route configuration should support it too. 
Is this correct @Snawoot?

services/wireguard/connection/connection.go
soffokl: The stop you see probably called outside. This shadowed `err` in the if will not trigger the condition in the defer. 
this might be fixed by: 
```suggestion
	if err = c.handshakeWaiter.Wait(ctx, conn.PeerStats, c.opts.HandshakeTimeout, c.done); err != nil {
```

Guillembonet:
@soffokl Yes, I think the issue is not resolved yet, but I thought I would divide it into multiple PRs if that's fine

services/wireguard/connection/connection.go
Guillembonet: And also change all the `err:=` to `err=` in previous lines too (except the first one), right?

soffokl: The first one can be changed too, you already have it declared in the function.

services/wireguard/connection/handshake.go
```diff
@@ -61,6 +62,8 @@ func (h *handshakeWaiter) Wait(statsFetch func() (*wgcfg.Stats, error), timeout
 			return errors.New("failed to receive initial handshake")
 		case <-stop:
 			return errors.New("stop received")
+		case <-ctx.Done():
+			return errors.New("cancel received")
```
Snawoot, line 66: Usual practice is to return error from ctx.Err()

Snawoot (approved):

Issue #3884, opened 2021-08-31
Withdrawal command not visible in cli


vkuznecovas:
Withdrawal command is not visible when using help, or tabbing for autosuggest.

PR #3887, opened 2021-09-01
Fixed withdraw bug in cli

Guillembonet:
Withdraw option is now showing under help and works with autocompletion

Closes: https://github.com/mysteriumnetwork/node/issues/3884

Signed-off-by: Guillem <guillem@mysterium.network>


vkuznecovas (changes requested):
Needs to be added for the `help` command as well.

Guillembonet:
> Needs to be added for the `help` command as well.

It is also showing under help with this one-line change

tomasmik:
> > Needs to be added for the `help` command as well.
> 
> It is also showing under help with this one-line change

Help will print the autocomplete tree, so I think we're good here.

vkuznecovas (approved):

Issue #3880, opened 2021-08-30
Payment order bugs


tomasmik:
Payment orders in node have two bugs:
* [x] After order is confirmed balance refresh thread never exists ( Fixed by: https://github.com/mysteriumnetwork/node/pull/3879 )
* [ ] After a user has a single paid order, any order creation will spawn a balance update thread because node doesn't care about order history

Both of these bugs result in a never ending balance update thread.

PR #3889, opened 2021-09-01
Update order tracking logic

tomasmik:
Order tracking was re-written. Major changes include:
* Fixed a bug where pilvytis would be spammed by node non stop
* Fixed a bug where orders created using tequila would not start force resync
* Extended the time for balance sync while also extending the amount of time we wait between syncs
* Reduced the code confusion and removed parts that are not needed in my opinion.
  * Current workflow will stop spamming if there are no new orders which are completed
  * We track failures to call pilvytis - if any occur we assume we might have orders that we are missing
  * If we find a paid order which we havent seen before, we initiate sync
  * All orders are now loaded on identity unlock and then pilvytis is no longer called (until new orders are created)

Closes: https://github.com/mysteriumnetwork/node/issues/3880

This could further be improved if node would be aware of orders that are not only paid but also executed and sent by topperupper. That would reduce BC calls needed by a lot.


pilvytis/order_status_tracker.go
```diff
@@ -73,89 +55,145 @@ func NewStatusTracker(api orderProvider, identityProvider identityProvider, even
...
+
+// UpdateOrdersFor sends a notification to the main running thread to
+// sync orders for the given identity.
+func (t *StatusTracker) UpdateOrdersFor(id identity.Identity) {
+	t.forceSync <- id
+}
+
+// Stop stops the status tracker
+func (t *StatusTracker) Stop() {
+	close(t.stopCh)
```
vkuznecovas, line 108: do a once here, we're notorious for stopping everything twice.

pilvytis/order_status_tracker.go
```diff
@@ -73,89 +55,145 @@ func NewStatusTracker(api orderProvider, identityProvider identityProvider, even
...
+		return
+	}
+
+	updated := make(map[uint64]OrderSummary)
+	for _, no := range newOrders {
+		old, ok := old[no.ID]
+		if !ok {
+			updated[no.ID] = no
+			// If the entry is new but already completed, send an update about it
+			if !no.Status.Incomplete() {
```
vkuznecovas, line 158: more readable imho
```suggestion
			if no.Status.Incomplete() {
			   continue
			}
```

vkuznecovas (approved):

pilvytis/order_status_tracker.go
tomasmik: thats very fair

Guillembonet (approved):

PR #3897, opened 2021-09-02
p2p: add fuzzing test for message serializer

Snawoot:
This PR adds test which always fails. It is intended to illustrate and prove critical bug which affects P2P channel communication and corrupts messages containing binary data. This PR is not intended to be merged now.

It turns out, our transport messages are serialized with net/textproto which is only suitable for text protocols. Due to its text nature, following distortions will occur:

- If message last byte is `0x0a`, it will be stripped.
- If message last byte is `0x0d`, it will be stripped.
- If message contains sequence `0x0d0a`, it will be replaced with `0x0a` byte.

Fuzzing experiment indicate that if message contains, for example, 65 bytes of uniformly distributed binary data (like sec256k1 ethereum signature), there is ~1% chance for message to arrive corrupted on the other end.

Also, as [noted in documentation](https://pkg.go.dev/net/textproto#NewReader) for "net/textproto" module, denial of service attack is possible if remote peer is able to exhaust node memory within time period limited by channel message read timeout.

Fix of this bug will require to change P2P message serialization format, rendering new protocol incompatible with old one. Therefore, new protocol must be negotiated out of band, with some additional attribute in providers' proposals.

### DIFF
--- a/p2p/message_test.go
+++ b/p2p/message_test.go
@@ -1,0 +1,55 @@
+package p2p
+
+import (
+	"bufio"
+	"bytes"
+	"crypto/rand"
+	"net/textproto"
+	"testing"
+)
+
+// Size of Ethereum signature, for instance
+const randomDataSize = 65
+const fuzzingIterations = 1_000_000
+
+func randomData() ([]byte, error) {
+	data := make([]byte, randomDataSize)
+	_, err := rand.Read(data)
+	return data, err
+}
+
+func roundTrip(t *testing.T) {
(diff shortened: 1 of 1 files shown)


mdomasevicius:
Will be great opportunity to test `compatibility` field on our proposals

```
curl -s 'https://testnet3-discovery.mysterium.network/api/v3/proposals' | jq '.[0].compatibility'
```

vkuznecovas (approved):

Issue #3815, opened 2021-08-17
Fiat payment provider support for payment orders


tadaskay:
API documentation https://developers.cardinity.com/api/v1/#payments

Needed functionality:

- [x] Create card payment (including finalizing payment)
- [x] Support 3DSecure v1/v2 flow
- [x] Get payments by consumer ID
- [x] After payment is confirmed, the consumer channel should be topped up the same way as it is done in Coingate case
- [ ] Handle callbacks with redirects
- [ ] Handle failed to save data: probably need to thread or something similar to ping cardinity and ask about transfers.

PR #3927, opened 2021-09-09
Support payment gateways

tomasmik:
Some unwanted code duplication or hacks introduced but can easily be removed once clients upgrade to the new workflow.

Main changes:
* Introduced v2 api for payment orders
  * Create payment orders per gateway
  * Return gateways and their data instead of currency and order suggestions endpoints
 * Order tracker now tracks both v2 and v1 orders with some temporary hacks
 * Added mobile bindings
 * Remade CLI to support v2 payments only. Better suggestions for caller data parsing are welcome.

Closes: https://github.com/mysteriumnetwork/pilvytis/issues/301
Relates to: https://github.com/mysteriumnetwork/node/issues/3815


pilvytis/api.go
```diff
@@ -223,6 +234,135 @@ func (a *API) GetMystExchangeRateFor(curr string) (float64, error) {
...
+	ID     string             `json:"id"`
+	Status PaymentOrderStatus `json:"status"`
+
+	Identity       string `json:"identity"`
+	ChainID        int64  `json:"chain_id"`
+	ChannelAddress string `json:"channel_address"`
+
+	GatewayName string `json:"gateway_name"`
+
+	ReceiveMYST float64 `json:"receive_myst"`
```
Snawoot, line 266: I think we are asking for a problem with floats for monetary values.

tomasmik: Can't argue with that. I am more or less copying the old implementation. Thing is a person can request 10.1231231 MYST and that is sort of valid. I guess we could go with `string` or `decimal` of some sort.

Snawoot: Well, if it already works, let's leave it as is.

Snawoot (approved):

soffokl (approved):

tadaskay (approved):

PR #3946, opened 2021-09-14
Revert and adapt legacy CORS for tequilapi

mdomasevicius:
This is legacy implementation for HTTP router reverted and adapted for gin gonic as middleware


vkuznecovas (approved):

tequilapi/http_api_server.go
```diff
@@ -80,8 +55,12 @@ func NewServer(
 ) (APIServer, error) {
 	gin.SetMode(modeFromOptions(nodeOptions))
 	g := gin.New()
+	g.Use(middlewares.ApplyCORSMiddleware(middlewares.RegexpCorsPolicy{
+		DefaultTrustedOrigin:  "https://mysterium.network",
```
creotiv, line 59: do we use this anywhere?

mdomasevicius: I don't **think** so. According to what I could gather there were some experimental features related to `wallet` (as the next line `AllowedOriginSuffixes: []string{"wallet(-dev)?\\.mysterium\\.network$", "localhost(:\\d+)?$"},` suggests). Maybe someone can comment on this as this was implemented way before my time on this project.

creotiv: If we dont' use it anywhere we should remove this and just block all cross-site calls.

tomasmik (approved):

tadaskay (approved):

Issue #3944, opened 2021-09-13
Token withdrawal flow in NodeUI


chompomonim:
In MainNet node runners will be able to settle collected funds into blockchain wallet (on Polygon or Ethereum).

- [x] Withdrawal widget in NodeUI dashboard

<img width="1680" alt="Dashboard - withdrawal" src="https://user-images.githubusercontent.com/153896/133096956-c4ac0438-8d6e-42f3-80aa-7bdfe478aead.png">

- [x] Popup with withdrawal information

![Withdrawal popup](https://user-images.githubusercontent.com/153896/133535119-8c33527c-1a1d-4e2c-b947-8908b98825d4.png)
![Group 2862348](https://user-images.githubusercontent.com/153896/133535322-e4c6dbec-6f70-4cc0-8ff9-aeb502c9afeb.png)

- [x] Possibility to select chain

Default should be "L2: Polygon-Matic (smaller fees, faster transactions)", secondary option is "L1: Ethereum (bigger fees, slower transactions)

chompomonim:
Will register separate issue of not correct amount being withdrawn.

PR #3961, opened 2021-09-16
Emit settlement/withdrawal events and handle those on consumer balance tracker

vkuznecovas:
updates #3944

### DIFF
--- a/session/pingpong/consumer_balance_tracker.go
+++ b/session/pingpong/consumer_balance_tracker.go
@@ -140,7 +140,25 @@ func (cbt *ConsumerBalanceTracker) Subscribe(bus eventbus.Subscriber) error {
 	if err != nil {
 		return err
 	}
+	err = bus.SubscribeAsync(event.AppTopicSettlementComplete, cbt.handleSettlementComplete)
+	if err != nil {
+		return err
+	}
+	err = bus.SubscribeAsync(event.AppTopicWithdrawalRequested, cbt.handleWithdrawalRequested)
+	if err != nil {
+		return err
+	}
 	return bus.SubscribeAsync(identity.AppTopicIdentityUnlock, cbt.handleUnlockEvent)
+}
+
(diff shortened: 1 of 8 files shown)


session/pingpong/consumer_balance_tracker.go
```diff
@@ -140,9 +140,27 @@ func (cbt *ConsumerBalanceTracker) Subscribe(bus eventbus.Subscriber) error {
...
+	err = bus.SubscribeAsync(event.AppTopicWithdrawalRequested, cbt.handleWithdrawalRequested)
+	if err != nil {
+		return err
+	}
 	return bus.SubscribeAsync(identity.AppTopicIdentityUnlock, cbt.handleUnlockEvent)
 }
 
+// settlements increase balance on the chain they are settled on.
+func (cbt *ConsumerBalanceTracker) handleSettlementComplete(ev event.AppEventSettlementComplete) {
+	go cbt.aggressiveSync(ev.ChainID, ev.ProviderID, cbt.cfg.FastSync.Timeout, cbt.cfg.FastSync.Interval)
```
tomasmik, line 156: At this point it seems like it'd be really to not be able to hit `cbt.aggressiveSync` or `cbt.periodicSync` multiple times. As in if you called it once and it's running - you can't call it again (or calls are ignored) until it finishes. Lots of different places call it and it will start to overlap which will cost us requests for nothing in the future or if something goes wrong 🤷🏻

tomasmik (approved):

session/pingpong/consumer_balance_tracker.go
vkuznecovas: k, will fix

vkuznecovas: Added a balance syncer, that will only re-launch the sync job in case the params match.

tomasmik (approved):

Snawoot (approved):

PR #3969, opened 2021-09-20
pricer: correct clock skew

### LINKED ISSUE #3958: Price consensus depends on correct system clock both on consumer and provider side
**Describe the bug**

One of nose runners:

> i'm seeing a lot of these in my logs:
Sep 10 10:24:53 nas myst[2985920]: 2021-09-10T10:24:53.289 ERR p2p/channel.go:416 > Handler "p2p-session-create" internal error error="cannot start session: : consumer asking for invalid price"

That sort of issues can be caused by incorrect system clock.

If node perception of UTC time falls behind real UTC time, it will pull prices only once on startup and in 10 minutes prices will become completely invalid because from node's point of view these prices are from future and will be valid for long.

If node perception of UTC time ahead of real UTC time, it'll always pull prices on each GetPrices() invocation because they will always appear outdated.

Such cases may happen pretty often in following scenarios:
* user has incorrect timezone setting, but correct time matching his local time
* user adjusts daylight saving time manually adding or subtracting 1 hour
* provider's RPi without RTC clock just started and struggling to synchronize clock timely

It's easy to demonstrate such thing happens in the wild because later case (clock ahead real time) leaves trace of frequent requests to pricer on discovery: https://gist.github.com/Snawoot/5c6ea799347ecb89f6c46fb250a7a54c
I think we should redesign price consensus logic: either we rely on discovery server time, or rely solely on delta of wall clock observed by node 

**To Reproduce**
1. Set system clock 1 hour behind
2. Run MysteriumVPN app
3. Wait 10 minutes
4. Try to connect to residential node in country where prices do not tend to be always constant.

**Expected behavior**
Connection established.

**Actual behavior**
Connection failed with "remote error" reason.

**Screenshots**

Screenshot_win10-clean_2021-09-16_01:47:33

**Environment (please complete the following information):**
 - Node version: 0.62.1
 - OS: Windows 10
 - Desktop app version (if applicable): 6.5.1

**Additional context**

* Excerpt from access log of discovery service: https://gist.github.com/Snawoot/5c6ea799347ecb89f6c46fb250a7a54c
* Consumer node log: [mysterium-node(7).log](https://github.com/mysteriumnetwork/node/files/7176396/mysterium-node.7.log)

Snawoot:
Fix #3958

### DIFF
--- a/docker-compose.e2e-basic.yml
+++ b/docker-compose.e2e-basic.yml
@@ -54,7 +54,7 @@ services:
       retries: 5
 
   discovery:
-    image: mysteriumnetwork/discovery:2.0.0
+    image: mysteriumnetwork/discovery:2.2.0
     expose:
       - 8080
     environment:
--- a/docker-compose.e2e-compatibility.yml
+++ b/docker-compose.e2e-compatibility.yml
@@ -49,7 +49,7 @@ services:
       retries: 5
 
   discovery:
-    image: mysteriumnetwork/discovery:matic-rc3
+    image: mysteriumnetwork/discovery:2.2.0
     depends_on:
       - db
       - http-mock
--- a/docker-compose.e2e-traversal.yml
+++ b/docker-compose.e2e-traversal.yml
@@ -101,7 +101,7 @@ services:
         ipv4_address: 172.31.0.201
 
   discovery:
-    image: mysteriumnetwork/discovery:2.0.0
+    image: mysteriumnetwork/discovery:2.2.0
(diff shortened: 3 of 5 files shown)


vkuznecovas (approved):

vkuznecovas:
You'll probably want to bump discovery version in e2e

Snawoot:
@vkuznecovas right, thanks!

tomasmik (approved):

PR #3977, opened 2021-09-22
fix #3976

### LINKED ISSUE #3976: Second withdrawal fails
Withdrawing first time works fine, fails on the second attempt.

vkuznecovas:


session/pingpong/hermes_promise_settler.go
```diff
@@ -721,24 +722,22 @@ func (aps *hermesPromiseSettler) generateAgreementID() *big.Int {
 	return new(big.Int).SetBytes(agreementID)
 }
 
-func (aps *hermesPromiseSettler) getWithdrawalAmountViaHermes(chainID int64, hermesID, identity common.Address) (*big.Int, error) {
+func (aps *hermesPromiseSettler) getHermesData(chainID int64, hermesID, identity common.Address) (ConsumerData, error) {
```
Snawoot, line 725: Structure `ConsumerData` is quite large. Can't we operate with pointer to it?

Snawoot (changes requested):
ConsumerData structure is quite large. Can't we operate with pointers

session/pingpong/hermes_promise_settler.go
vkuznecovas: Overoptimizing the very rare call, but we can. Done.

Snawoot (approved):

soffokl (approved):

tomasmik (approved):

PR #3989, opened 2021-09-24
no CORS

Snawoot:
Localhost is not trusted origin either - daemons like IPFS and other gateways use it.

### DIFF
--- a/tequilapi/http_api_server.go
+++ b/tequilapi/http_api_server.go
@@ -21,6 +21,7 @@ import (
 	"net"
 	"net/http"
 	"strings"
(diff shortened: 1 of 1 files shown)


soffokl (approved):

mdomasevicius (approved):

tomasmik (approved):

tadaskay:
So basically we can't use node API from a browser anymore?

tadaskay:
FYI @creotiv https://github.com/creotiv/dvpn-tutorial-mysterium-network will have to be redone via IPC

Issue #3967, opened 2021-09-20
Cannot listen on UDP port :: error="listen udp :: bind: address already in use"


etherunit:
**Describe the bug**

Provider id: 0x3712b4009a8c3372f370c79f766a306683e42d53

Service is still responsive - serving **multiple connections** at once, but provider is seeing these warnings in the logs.

```
Sep 16 16:28:39 nas myst[1024972]: 2021-09-16T16:28:39.179 INF core/port/availability.go:37             > Cannot listen on UDP port 32936 error="listen udp :32936: bind: address already in use"
Sep 16 16:28:39 nas myst[1024972]: 2021-09-16T16:28:39.179 INF core/port/availability.go:37             > Cannot listen on UDP port 10000 error="listen udp :10000: bind: address already in use"
Sep 16 16:28:39 nas myst[1024972]: 2021-09-16T16:28:39.179 INF core/port/availability.go:37             > Cannot listen on UDP port 10001 error="listen udp :10001: bind: address already in use"
Sep 16 16:28:39 nas myst[1024972]: 2021-09-16T16:28:39.179 INF core/port/availability.go:37             > Cannot listen on UDP port 10002 error="listen udp :10002: bind: address already in use"
Sep 16 16:28:39 nas myst[1024972]: 2021-09-16T16:28:39.180 INF core/port/availability.go:37             > Cannot listen on UDP port 10003 error="listen udp :10003: bind: address already in use"
Sep 16 16:28:39 nas myst[1024972]: 2021-09-16T16:28:39.180 INF core/port/availability.go:37             > Cannot listen on UDP port 10004 error="listen udp :10004: bind: address already in use"
Sep 16 16:28:39 nas myst[1024972]: 2021-09-16T16:28:39.180 INF core/port/availability.go:37             > Cannot listen on UDP port 10005 error="listen udp :10005: bind: address already in use"
Sep 16 16:28:39 nas myst[1024972]: 2021-09-16T16:28:39.180 INF core/port/availability.go:37             > Cannot listen on UDP port 10006 error="listen udp :10006: bind: address already in use"
Sep 16 16:28:39 nas myst[1024972]: 2021-09-16T16:28:39.180 INF core/port/availability.go:37             > Cannot listen on UDP port 10007 error="listen udp :10007: bind: address already in use"
[...]

PR #3990, opened 2021-09-24
Port Pool imrovements

Snawoot:
Fix https://github.com/mysteriumnetwork/node/issues/3967

This PR contains following changes:

* Safe use of random number generator in port pool (concurrent usage of rand.NewSource instances is not allowed! [[1]](https://pkg.go.dev/math/rand#pkg-overview))
* Also switched to concurrency-safe generator in other parts of node
* Switched payments to CSPRNG
* Revamped port distribution logic of static range port pool. Previously it was tending to seek from start of pool on any collision. This way there was high probability that eventually it will try own ports previously distributed from the beginning of range. Now it always picks random port and circles over entire pool range from random offset position.
* Added more test cases for port pool to prove concurrent connections is not a problem (special for @creotiv)
* Avoid collisions in AcquireMulti result


soffokl (approved):

session/pingpong/invoice_tracker.go
```diff
@@ -232,7 +232,10 @@ func (it *InvoiceTracker) listenForExchangeMessages() error {
 
 func (it *InvoiceTracker) generateAgreementID() {
 	agreementID := make([]byte, 32)
-	crand.Read(agreementID)
+	_, err := crand.Read(agreementID)
+	if err != nil {
+		panic(err)
```
mdomasevicius, line 237: nothing useful to log before panic?

core/port/pool.go
```diff
@@ -68,14 +68,26 @@ func (pool *Pool) seekAvailablePort() (int, error) {
 
 // AcquireMultiple returns n unused ports from pool's range.
 func (pool *Pool) AcquireMultiple(n int) (ports []Port, err error) {
-	for i := 0; i < n; i++ {
+	if n > pool.capacity {
+		return nil, errors.New("requested more ports than pool capacity")
```
mdomasevicius, line 72:
```suggestion
		return nil, fmt.Errorf("requested more ports (%d) than pool capacity (%d)", n, pool.capacity)
```

mdomasevicius (approved):

tomasmik (approved):

session/pingpong/invoice_tracker.go
Snawoot: nope, except panic error itself. I don't expect this happening anyway, just a measure of precaution.

PR #3991, opened 2021-09-25
nat traversal: concurrency fixes

Snawoot:
This PR contains following changes:

* Major rewrite of concurrency control in nat pinger to fix goroutine leaks and associated resources
* Fixes for race conditions which may produce dangling sockets
* Strict ping acknowledgements to avoid consumer and provider picking different ports because of packet reordering (happens even in local tests)
* Stabiized tests because of fixed bugs

### DIFF
--- a/nat/traversal/pinger.go
+++ b/nat/traversal/pinger.go
@@ -42,19 +42,23 @@ const StageName = "hole_punching"
 const (
 	bufferLen = 2048 * 1024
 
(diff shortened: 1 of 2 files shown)


soffokl (approved):

mdomasevicius (approved):
typo in:
> Stabiized tests because of fixed bugs

tomasmik (approved):

PR #3995, opened 2021-09-27
fixed electron problem with CORS

creotiv:
So as i found problem was with `cache-control` header


tomasmik (approved):

mdomasevicius (approved):

soffokl (approved):

Snawoot (changes requested):
This change opens CORS with localhost origins, which are not trusted, and reopens security issue.

PR #4004, opened 2021-09-28
Check balance on update if register in progress

Guillembonet:
Signed-off-by: Guillem <guillem@mysterium.network>

### DIFF
--- a/session/pingpong/consumer_balance_tracker.go
+++ b/session/pingpong/consumer_balance_tracker.go
@@ -46,11 +46,21 @@ func newBalanceKey(chainID int64, id identity.Identity) balanceKey {
 	return balanceKey(fmt.Sprintf("%v_%v", id.Address, chainID))
 }
 
+type balances struct {
+	sync.Mutex
+	valuesMap map[balanceKey]ConsumerBalance
+}
+
+type transactorBounties struct {
+	sync.Mutex
+	valuesMap map[balanceKey]*big.Int
+}
+
 // ConsumerBalanceTracker keeps track of consumer balances.
 // TODO: this needs to take into account the saved state.
 type ConsumerBalanceTracker struct {
-	balancesLock sync.Mutex
-	balances     map[balanceKey]ConsumerBalance
+	balances           balances
+	transactorBounties transactorBounties
 
 	addressProvider                      addressProvider
 	registry                             registrationStatusProvider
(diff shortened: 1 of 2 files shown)


session/pingpong/consumer_balance_tracker.go
```diff
@@ -49,8 +49,10 @@ func newBalanceKey(chainID int64, id identity.Identity) balanceKey {
 // ConsumerBalanceTracker keeps track of consumer balances.
 // TODO: this needs to take into account the saved state.
 type ConsumerBalanceTracker struct {
-	balancesLock sync.Mutex
-	balances     map[balanceKey]ConsumerBalance
+	balancesLock           sync.Mutex
+	balances               map[balanceKey]ConsumerBalance
+	transactorBountiesLock sync.Mutex
```
tomasmik, line 54: Dont think you're supposed to do that.
* [gobyexample (mutexes)](https://gobyexample.com/mutexes)
* [tour of go](https://tour.golang.org/concurrency/9)

Point is: you already got a mutex, an additional one - doesn't really do anything.

session/pingpong/consumer_balance_tracker_test.go
```diff
@@ -255,10 +255,8 @@ func TestConsumerBalanceTracker_FallsBackToTransactorIfInProgress(t *testing.T)
 		bus: bus,
 	}
 	bc := mockConsumerBalanceChecker{
-		channelToReturn: client.ConsumerChannel{
-			Balance: initialBalance,
-			Settled: big.NewInt(0),
-		},
+		errToReturn:         errors.New("No contract deployed"),
```
Snawoot, line 258: Why that change was made? It stabilizes test, but it seems it just hides the problem.

Snawoot (changes requested):

session/pingpong/consumer_balance_tracker_test.go
Guillembonet: If registration is in progress it means no channel SC is deployed in that address, so this call will fail

Guillembonet: If it succeded like it said it would mean registration is complete

session/pingpong/consumer_balance_tracker.go
Guillembonet: golang noob, checking it now

Snawoot (approved):

session/pingpong/consumer_balance_tracker_test.go
Snawoot: Ok, cleared

tomasmik (approved):

PR #4007, opened 2021-09-28
Check response status when calling MMN

tomasmik:
- Close body
- Check response code


mmn/client.go
```diff
@@ -91,10 +91,17 @@ func (m *client) UpdateBeneficiary(data *UpdateBeneficiaryRequest) error {
...
 		return err
 	}
+	defer resp.Body.Close()
+
 	body, err := ioutil.ReadAll(resp.Body)
 	if err != nil {
 		return err
 	}
+
+	if resp.StatusCode < 199 || resp.StatusCode > 299 {
```
Snawoot, line 101:
```suggestion
	if resp.StatusCode < 200 || resp.StatusCode > 299 {
```

Snawoot: Don't want 199 to be considered OK status

mmn/client.go
```diff
@@ -121,11 +128,17 @@ func (m *client) GetBeneficiary(identityStr string) (string, error) {
...
 		return "", err
 	}
+	defer resp.Body.Close()
+
 	body, err := ioutil.ReadAll(resp.Body)
 	if err != nil {
 		return "", err
 	}
 
+	if resp.StatusCode < 199 || resp.StatusCode > 299 {
```
Snawoot, line 138:
```suggestion
	if resp.StatusCode < 200 || resp.StatusCode > 299 {
```

Same

Snawoot (approved):

mmn/client.go
tomasmik: oops

soffokl (approved):

Issue #3870, opened 2021-08-26
Smart (re)connect


chompomonim:
### **There are 2 big tasks that are required for smart-reconnect:**
**1) a way to find a node for initial connection also known as smart-filters.
2) a real reconnect to a new node once we lost connection to the currently used.** 

### Most of the work on the 1st part should be done on the discovery service:
1) Discovery service should support an endpoint to provide an ordered list of suggested nodes by smart-filters.
2) Discovery service should be able to build such ordered lists based on quality data: biggest bandwidth, lowest latency, lowest node load, etc. 
3) Node should support sending these smart-filter requests and be able to establish connections based on the ordered list.

### The 2nd part of them is to finish proper reconnect once the connection is lost without leaking any traffic:
1) We had an initial POC implementation for reconnecting, but it has a lot of problems which is not been solved.
2) It's a platform-dependent solution, and most likely we will need to have platform-specific implementation. This requires additional research. 

_POC of reconnect was disabled by https://github.com/mysteriumnetwork/node/pull/3875
We need to resolve routing maintenance issue to make sure we can use: https://github.com/mysteriumnetwork/node/issues/3247_

soffokl:
### Starting a deamon to support Smart (re)connect

Auto-reconnect is currently disabled by default, and to start using it the following flags needs to be set:
```
myst --keep-connected-on-fail=true --auto-reconnect=true daemon
```

The same options can be set via configuration file by adding the following lines there:
```
auto-reconnect = "true"
keep-connected-on-fail = "true"
```

### Starting connection with a smart filters

A connection could be established now even without specifying exact parameters:
```
myst connection up
```
It will take into account all providers from all countries with a default soring by quality. It will try to establish a connection to the provider with the highest quality.

There are a couple of options that can be set to make the filter with more specific requirements:
- `--country` is a two letter (ISO 3166-1 alpha-2) country code to filter proposals from exact country only;
- `--location-type=` is a provider node location types to filter by eg. 'hosting', 'residential', 'mobile' etc.
- `--sort` Proposal sorting type. One of: `quality`, `bandwidth`, `latency` or `price` (`quality` is a default value); 

All of this options can be combined with any other: 
```
myst connection up --location-type=hosting --country=FI --sort=price
```

```
myst connection up --location-type=residential --sort=latency
```

```
myst connection up --country=LT --sort=bandwidth
```

```
myst connection up --country=US
```

Beside it exact nodes can be specified to establish a connection to it:

```
[...]

PR #4024, opened 2021-10-04
Allow reconnect active connection to multiple providers

soffokl:
This is an essential part of the smart-reconnect. 
Closes https://github.com/mysteriumnetwork/node/issues/3247 and part of the https://github.com/mysteriumnetwork/node/issues/3870 

The new flag was introduced to explicitly enable reconnect: `--auto-reconnect=true` (disabled by default).

Currently it should reconnect if both flags `--auto-reconnect=true` and `--keep-connected-on-fail=true` set and multiple providers passed for connection create:
`myst connection up 0xe948dae2ce1faf719ba1091d8c6664a46bab073d,0x070a908615c53ff0aa69c4e802142791c7d9ef09`


core/connection/manager.go
```diff
@@ -834,7 +853,14 @@ func (m *connectionManager) setupTrafficBlock(disableKillSwitch bool) error {
 }
 
 func (m *connectionManager) reconnectOnHold(state connectionstate.AppEventConnectionState) {
-	if state.State == connectionstate.StateOnHold {
+	if state.State == connectionstate.StateOnHold && config.GetBool(config.FlagAutoReconnect) {
```
vkuznecovas, line 856: can we unnest this and just return if state != on hold or not auto reconnect?

router/router.go
```diff
@@ -18,44 +18,210 @@
...
+
+func (m *manager) ExcludeIP(ip net.IP) error {
+	m.ensureStarted()
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	new := true
+
+	for i, rule := range m.rules {
+		if rule.ip.Equal(ip) {
```
vkuznecovas, line 80:
```suggestion
		if !rule.ip.Equal(ip) { 
		  continue
		}
		...
```

router/router.go
```diff
@@ -18,44 +18,210 @@
...
+
+	return nil
+}
+
+func (m *manager) RemoveExcludedIP(ip net.IP) error {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	for i, rule := range m.rules {
+		if rule.ip.Equal(ip) {
```
vkuznecovas, line 110:
```suggestion
		if !rule.ip.Equal(ip) { 
		  continue
		}
		...
```

router/router.go
```diff
@@ -18,44 +18,210 @@
...
 		}
+	}
+}
 
+func (m *manager) Stop() {
+	if err := m.Clean(); err != nil {
+		log.Error().Err(err).Msg("Failed to clean routing rules")
 	}
+
+	close(m.stop)
```
vkuznecovas, line 152: wrap this in a once, since we're always stopping twice :D

vkuznecovas (approved):

Snawoot:
Doesn't work for some reason:

```
» connect 0xa3caf88de5ac00d00876039e7f563416b53d0a78 0x18730b1a9864b959a6edd1b7e715ddef5f835b00,0xcdfdba200b2455c503406deae7809d90b886d4c9 wireguard
[CONNECTING] from: 0xa3caf88de5ac00d00876039e7f563416b53d0a78 to: 0x18730b1a9864b959a6edd1b7e715ddef5f835b00,0xcdfdba200b2455c503406deae7809d90b886d4c9
[ERROR] Server response invalid: 500 Internal Server Error (http://127.0.0.1:4050/connection). Possible error: failed to lookup proposal: ErrorCollection: proposal does not exist: {ServiceType:wireguard ProviderID:0x18730b1a9864b959a6edd1b7e715ddef5f835b00,0xcdfdba200b2455c503406deae7809d90b886d4c9 ID:0}
```

soffokl:
@Snawoot you read it wrong. It should be:
```
myst connection up 0x18730b1a9864b959a6edd1b7e715ddef5f835b00,0xcdfdba200b2455c503406deae7809d90b886d4c9
```

It's not added to the interactive CLI yet.

Snawoot (approved):

PR #4198, opened 2021-11-03
Close #4150

### LINKED ISSUE #4150: Decrease spam of logs
During a session, every second or so one log line like this gets printed:
`2021-10-25T11:20:47.649 DBG config/config.go:224 > Returning default value launcher.ver:`

vkuznecovas:
Move default value spam to trace.

Closes #4150


tomasmik (approved):
👏🏻

soffokl (approved):

Issue #4234, opened 2021-11-08
External IP resolver sometimes returns IPv6


Snawoot:
**Describe the bug**

External IP resolver sometimes detects IPv6 address of node instead of IPv4 one. It causes provider node to communicate IPv6 address to consumer nodes and, in general, fail P2P connection.

**To Reproduce**

Reproduces randomly when following conditions met:

* Provider has IPv6 enabled.
* Provider randomly picked external ip resolver fallback URL which supports IPv6 and returns it if queried via IPv6.

**Expected behavior**

IPv4 address always used.

**Environment (please complete the following information):**
 - Node version: 0.6*.*
 - OS: Linux
 - Desktop app version (if applicable): N/A
 - Docker (if applicable): N/A

**Additional context**

There are few ways to fix it:

1. Just exclude IP resolver services which have IPv6.
2. Make HTTP client used by external IP resolver use only IPv4 for dials.
3. Switch to STUN completely.

I suggest option (2) or option (3).

PR #4248, opened 2021-11-11
Disable ipv6 http requests

soffokl:
Closes https://github.com/mysteriumnetwork/node/issues/4234


tomasmik (approved):

Snawoot:
@soffokl Will it cause dialer to fallback to tcp4 or it will just fail?

soffokl:
@Snawoot yeah, it skips IPv6 and goes to IPv4.

Snawoot (approved):

PR #4302, opened 2021-11-19
Make debugging payments in sessions easier

tomasmik:
Using this you can start two nodes. 1 in client and 1 in provider mode with `noop` service both on the same machine.
You can then connect to your own node and pump MYST really fast.

Closes: https://github.com/mysteriumnetwork/node/issues/4140

### DIFF
--- a/config/flags_payments.go
+++ b/config/flags_payments.go
@@ -139,6 +139,13 @@ var (
 		Value:  metadata.Testnet3Definition.Testnet3HermesURL,
 		Hidden: true,
 	}
+	// FlagPaymentsDuringSessionDebug sets if we're in debug more for the payments done in a VPN session.
+	FlagPaymentsDuringSessionDebug = cli.BoolFlag{
+		Name:   "payments.during-session-debug",
+		Usage:  "Set debug mode for payments made during a session, it will bypass any price validation and allow absurd prices during sessions",
+		Value:  false,
+		Hidden: true,
+	}
 )
 
 // RegisterFlagsPayments function register payments flags to flag list.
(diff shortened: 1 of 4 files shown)


config/flags_payments.go
```diff
@@ -139,6 +139,13 @@ var (
 		Value:  metadata.Testnet3Definition.Testnet3HermesURL,
 		Hidden: true,
 	}
+	// FlagPaymentsDuringSessionDebug sets if we're in debug more for the payments done in a VPN session.
+	FlagPaymentsDuringSessionDebug = cli.BoolFlag{
+		Name:   "payments.during-session-debug",
+		Usage:  "Set debug mode for payments made during a session, it will bypass any price validation and allow absurd prices during sessions",
```
mdomasevicius, line 145: may users try to abuse it?

core/connection/manager.go
```diff
@@ -324,10 +325,21 @@ func (m *connectionManager) autoReconnect() (err error) {
...
+	p := market.Price{
 		PricePerHour: proposal.Price.PricePerHour,
 		PricePerGiB:  proposal.Price.PricePerGiB,
 	}
+
+	if config.GetBool(config.FlagPaymentsDuringSessionDebug) {
+		log.Info().Msg("Payments debug bas been enabled, will use absurd amounts for the proposal price")
+		p = market.Price{
+			PricePerHour: new(big.Int).SetInt64(5000000000000000000),
+			PricePerGiB:  new(big.Int).SetInt64(5000000000000000000),
```
mdomasevicius, line 337: Maybe `payments.during-session-debug-amount` flag?

config/flags_payments.go
tomasmik: Does it make a difference if it's us editing the code or them? Also not much you can abuse in mainnet really, you need real money to make money. So not sure how you would abuse this.

core/connection/manager.go
tomasmik: Was thinking about that, wasn't sure if it's worth it, but sure.

soffokl (approved):

PR #4488, opened 2021-12-16
Add endpoint to get number of proposals per country

soffokl:


tomasmik (approved):

core/discovery/dhtdiscovery/repository.go
```diff
@@ -47,6 +47,11 @@ func (r *Repository) Proposals(filter *proposal.Filter) ([]market.ServiceProposa
 	return []market.ServiceProposal{}, nil
 }
 
+// Countries returns proposals per country matching the filter.
+func (r *Repository) Countries(filter *proposal.Filter) (map[string]int, error) {
+	return nil, nil
```
mdomasevicius, line 52: just looks like you should return an **empty map** with nil error

soffokl: it's a DHT discovery that we never had, it's just a placeholder.

mdomasevicius (approved):

PR #4508, opened 2021-12-19
Fix path for API key address in Myst CLI

otkd:
Signed-off-by: otkd <7527203+otkd@users.noreply.github.com>

### DIFF
--- a/cmd/commands/cli/command.go
+++ b/cmd/commands/cli/command.go
@@ -346,7 +346,7 @@ func (c *cliApp) connect(args []string) (err error) {
 }
 
 func (c *cliApp) mmnApiKey(args []string) (err error) {
-	profileUrl := c.config.GetStringByFlag(config.FlagMMNAddress) + "user/profile"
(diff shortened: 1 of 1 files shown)


cmd/commands/cli/command.go
```diff
@@ -346,7 +346,7 @@ func (c *cliApp) connect(args []string) (err error) {
 }
 
 func (c *cliApp) mmnApiKey(args []string) (err error) {
-	profileUrl := c.config.GetStringByFlag(config.FlagMMNAddress) + "user/profile"
+	profileUrl := c.config.GetStringByFlag(config.FlagMMNAddress) + "/me"
```
tomasmik, line 349:
```suggestion
	profileUrl := strings.TrimSuffix(c.config.GetStringByFlag(config.FlagMMNAddress), "/") + "/me"
```

Would be nice to make sure we never put `//me` in the address

tomasmik (approved):

PR #4533, opened 2021-12-23
Bring back the ability to settle with beneficiary

tomasmik:
* Web UI can now again use Tequila API to settle with beneficiary and change it's address
* CLI was extended to support this capatability too.

Example of usage and test results:
```
» identities beneficiary-set 0x05c82eea746311bb965fef708684c02da671b381 0x21CA5D6417A08ab6692336ec16244FE966720F84
[SUCCESS] New beneficiary address set
» identities beneficiary-status 0x05c82eea746311bb965fef708684c02da671b381
[INFO] Current beneficiary: 0x21CA5D6417A08ab6692336ec16244FE966720F84
[INFO] Last change request information:
[INFO] Change to: 0x21CA5D6417A08ab6692336ec16244FE966720F84
[INFO] State: completed
```

Note: This completely destroys the old workflow of saving beneficiary in to the database or block chain depending on what block chain you use, this also affect the old `GET /beneficiary` endpoint as it now only pulls from BC. These old workflows should not affect the UI and are mainly no longer needed in mainnet anyway.


cmd/commands/cli/command_identities.go
```diff
@@ -304,6 +311,91 @@ func (c *cliApp) setPayoutAddress(args []string) error {
...
+	if err != nil {
+		return err
+	}
+
+	err = c.tequilapi.SettleWithBeneficiary(address, benef, hermesID)
+	if err != nil {
+		return err
+	}
+
+	timeout := time.After(30 * time.Second)
```
Guillembonet, line 334: is beneficiary stored on the bc? If so shouldn't we also exit when it is in progress? waiting 30 secs seems a bit pointless no?

cmd/commands/cli/command_identities.go
```diff
@@ -304,6 +311,91 @@ func (c *cliApp) setPayoutAddress(args []string) error {
...
+				clio.Success("New beneficiary address set")
+				return nil
+			}
+		}
+	}
+}
+
+const usageSetBeneficiaryStatus = "beneficiary-status <identity>"
+
+func (c *cliApp) setBeneficiaryStatus(actionArgs []string) error {
```
Guillembonet, line 370: setBeneficiaryStatus -> getBeneficiaryStatus ?

Guillembonet (approved):
looks fine to me overall! see the few comments

cmd/commands/cli/command_identities.go
tomasmik: Tranasction is made on polygon. Considering minimal transactor congestion it should complete in under 15 seconds.

Snawoot (approved):

PR #4548, opened 2021-12-30
NodeUI version management

mdomasevicius:
This is experimental MVP of POC that will be tested internally meanwhile.

## Motivation:
POs and Testers want to have complete node release with nodeUI release to test features. Creating a full release of both eats up time, especially if it's something trivial. Keeping latter in mind it would be also big time saver for node runners to just download a new release of nodeUI and run it with some bug fixes or even features.

## TODO (if this will actually be useful)
- NodeUI reset to **bundled** version via CLI if switching to botched release
- Perhaps better downloading solution
- Maybe make NodeUI 100% portable and just ship downloader with Node (more lightweight node binary, less (and less complicated) pipelines)
- Node compatibility with NodeUI improvements
- Better Github API rate limit handling


ui/versionmanager/version_downloader.go
```diff
@@ -0,0 +1,194 @@
...
+		return
+	}
+
+	close(doneChan)
+
+	if opts.Callback != nil {
+		log.Info().Msg("executing post download callback")
+		err := opts.Callback(opts)
+		if err != nil {
+			d.update(d.status.transitionWithErr(failed, err))
```
tomasmik, line 157: Well the download didnt fail I think, just that callback, so is this really correct then?

ui/versionmanager/version_manager.go
```diff
@@ -0,0 +1,289 @@
...
+		if a.Name == compatibilityAssetName {
+			return a.BrowserDownloadUrl
+		}
+	}
+	return ""
+}
+
+// TODO think about sending SSE to inform to nodeUI that a version has been downloaded
+
+// Download download and untar node UI dist
```
tomasmik, line 160:
```suggestion
// Download and untar node UI dist
```

ui/versionmanager/version_manager.go
```diff
@@ -0,0 +1,289 @@
...
+func (vm *VersionManager) Download(versionName string) error {
+	assetURL, err := vm.github.nodeUIDownloadURL(versionName)
+	if err != nil {
+		return err
+	}
+
+	err = os.MkdirAll(vm.versionConfig.uiDistPath(versionName), 0700)
```
tomasmik, line 167: `DownloadNodeUI` Might be already downloading this version or it might exist, but you will still attempt to create a dir for it now.

tomasmik (changes requested):

ui/versionmanager/version_manager.go
mdomasevicius: Does not cause error if path already exists

tomasmik (approved):

PR #4625, opened 2022-01-17
Maximum withdrawal amount is 99 MYST

tomasmik:
Relates to: https://github.com/mysteriumnetwork/transactor/issues/349

### DIFF
--- a/cmd/commands/cli/command_identities.go
+++ b/cmd/commands/cli/command_identities.go
@@ -440,6 +440,10 @@ func (c *cliApp) withdraw(args []string) error {
 		if err != nil {
 			return fmt.Errorf("%v is not a valid number", args[3])
 		}
+		if amf > 99 {
+			return errors.New("max withdrawal amount is 99 MYST")
+		}
+
 		amount = crypto.FloatToBigMyst(amf)
 	}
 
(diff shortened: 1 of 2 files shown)


Snawoot:
What is expected outcome of such limitation? If provider has a friendly transactor w/o such limitation, they will be able to bypass it either way.

tomasmik:
> What is expected outcome of such limitation? If provider has a friendly transactor w/o such limitation, they will be able to bypass it either way.

How? Transactor will also not send a transaction, so there will be nothing to copy. PR for that is coming later.
Also, even if you could bypass this with your own transactor, you're only hurting yourself, so why would you bother doing that.

Snawoot:
@tomasmik

If it's not in the best interest of user, why limit this at all? What is so painful in withdrawal of, let's say, 150 MYST?

tomasmik:
> @tomasmik
> 
> If it's not in the best interest of user, why limit this at all? What is so painful in withdrawal of, let's say, 150 MYST?

Smart contracts actually hard cap the amount, so it wouldn't work anyway.

Snawoot (approved):

soffokl (approved):

PR #4638, opened 2022-01-19
Feedback changes

Guillembonet:
Added endpoint in tequilapi to use the new intercom report feature on the feedback service.


feedback/reporter.go
```diff
@@ -106,6 +112,40 @@ func (r *Reporter) NewIssue(report UserReport) (result *client.CreateGithubIssue
...
 }
 
+// NewIntercomIssue sends node logs, Identity and UserReport to intercom
+func (r *Reporter) NewIntercomIssue(report UserReport) (result *client.CreateIntercomIssueResult, err error) {
+	if err := report.Validate(); err != nil {
+		return nil, err
+	}
+
+	nodeID := r.currentIdentity()
+	location, err := r.locationResolver.DetectLocation()
```
tomasmik, line 122: As I understand this is only for provider, but maybe `GetOrigin` would still be more correct.

tomasmik (approved):

PR #4680, opened 2022-01-28
Allow to establish multiple connections with a proxy mode

soffokl:


services/wireguard/endpoint/proxyclient/proxy.go
```diff
@@ -0,0 +1,79 @@
...
+			return err
+		}
+		go func() {
+			s.handle(c)
+			c.Close()
+		}()
+	}
+}
+
+func (s *proxyServer) handle(c net.Conn) {
```
Snawoot, line 55: Does it support plain HTTP proxy requests, not just CONNECT?

soffokl: nope, I didn't move that part since it's was not required.

Snawoot (approved):

PR #4697, opened 2022-02-01
Pass identity balance as string

mdomasevicius:
Provide a generic `Token` object for returning ethereum currency


Snawoot (approved):

tadaskay (approved):
Noice

tequilapi/contract/tokens.go
```diff
@@ -0,0 +1,52 @@
...
+	if amount == nil {
+		return TokensZero
+	}
+
+	ethers := crypto.BigMystToDecimal(amount)
+
+	return Tokens{
+		Wei:   amount.String(),
+		Ether: ethers.String(),
+		Human: ethers.Truncate(HumanPrecision).String(),
```
tomasmik, line 50: Imo the human value here is not something backend should be concerned about. It's not like API should ever consume it in order to specify some amount of tokens and if we want to display it nicely, apps should just convert it to some notation they think is correct. 🤷🏻 

When it comes to any interaction with the API we should only really accept `Wei`.

mdomasevicius: @tadaskay your position?

tadaskay: We already have this in android API and it has helped a lot to avoid cumbersome logic on client side. Because API is an interface, the most important thing is how easy it is to use it from the consumer point of view. I shouldn't need to deal with the monstrosity where 1 token is presented as a number with 18 zeroes. So thanks for opinion, but this PR solves this long lasting issue.

Issue #4631, opened 2022-01-18
Withdrawals still disappear sometimes


tomasmik:
Relates to: https://github.com/mysteriumnetwork/node/issues/4613 and other similar issues.

It appears that hermes fails to settle the promise sometimes and money never reaches the person trying to withdraw the money. There was hope that transaction sending rework would remedy this complete, but it does not.

For now the current fix is just to withdraw again and you will receive the money, but that is a bad solution.

PR #4762, opened 2022-02-09
Slightly better withdrawals

tomasmik:
* Stop lying to people that withdrawal was a success even though it was not
* Actually save the fact the R is revealed in hermes
* [Transactor is no longer a mister](https://github.com/mysteriumnetwork/node/pull/4762/files#diff-d5b95c0a9601141004e9c61dd3850824e32850fc7b9755083e41c44643a7d2b5R588)

Updates: https://github.com/mysteriumnetwork/node/issues/4631

### DIFF
--- a/session/pingpong/hermes_promise_settler.go
+++ b/session/pingpong/hermes_promise_settler.go
@@ -517,6 +517,13 @@ func (aps *hermesPromiseSettler) Withdraw(
 		return err
 	}
 
+	currentPromise, err := aps.promiseStorage.Get(toChainID, chid)
+	if err != nil {
+		if !errors.Is(err, ErrNotFound) {
+			return err
+		}
+	}
+
 	// 2. issue a self promise
 	promisedAmount := data.LatestPromise.Amount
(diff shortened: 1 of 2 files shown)


session/pingpong/hermes_promise_handler.go
```diff
@@ -332,6 +332,9 @@ func (aph *HermesPromiseHandler) requestPromise(er enqueuedRequest) {
 	})
 
 	err = aph.revealR(ap)
+	if err != nil {
```
Guillembonet, line 335: this looks like its making the next function useless

Guillembonet (changes requested):
I don't really understand how can the `ErrNeedsRRecovery` happen so I'd be nice if you can explain or I can dig more into that tomorrow (anyway the thing I comment stands). The other changes look nice. Also, I will miss the `mister` :(

session/pingpong/hermes_promise_handler.go
tomasmik: I guess the log should be in the handle func, good catch.

Guillembonet (approved):

PR #4766, opened 2022-02-10
Retry withdrawal if it's still valid and possible

tomasmik:
Rely on transactor response codes being correct and retry
a withdrawal if it failed but the request itself was valid.

While retrying a few times for 5 minutes, doesn't completely solve our
issue, this and the previous withdrawal PR should remedy 90% or more of
the withdrawal problems being reported.

Solving this completely is rather difficult because money is actually
first taken from a provider and a payment promise is issued which has to be
settled with transactor we would have to somehow track this promise
and make sure we're resending it, but as it's mentioned in the comment
block in code, sending a settlement after 5 minutes will produce different 
results then shown in the UI as fees will change and sometimes they will change dramatically.

Small issue which was not resolved yet: Every other withdrawal _might_
fail with no money being transferred to beneficiary or taken from provider. That is because we've stopped
lying to the UI and actually report failures now. This is not a huge deal as no real value transfer occurs.

Closes: https://github.com/mysteriumnetwork/node/issues/4631

### DIFF
--- a/session/pingpong/hermes_promise_settler.go
+++ b/session/pingpong/hermes_promise_settler.go
@@ -37,6 +37,7 @@ import (
 	"github.com/mysteriumnetwork/node/eventbus"
 	"github.com/mysteriumnetwork/node/identity"
 	"github.com/mysteriumnetwork/node/identity/registry"
+	"github.com/mysteriumnetwork/node/requests"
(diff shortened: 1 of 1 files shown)


Guillembonet (approved):

soffokl (approved):
👏 👏 👏

PR #4795, opened 2022-02-19
Resend promise request after R recovery

tomasmik:
This solves withdrawal failing prematurely. If we recovered R after
requesting a promise, we can request again. Withdrawals now work without errors.

Closes: https://github.com/mysteriumnetwork/node/issues/4788


soffokl (approved):

Guillembonet (approved):
😍

PR #4822, opened 2022-02-28
Pass project id to gateway orders

mdomasevicius:
Signed-off-by: Mantas Domaševičius <mantas@mysterium.network>


mobile/mysterium/payment_orders.go
```diff
@@ -145,12 +145,13 @@ func (mb *MobileNode) CreatePaymentGatewayOrder(req *CreatePaymentGatewayOrderRe
 	}
 
 	order, err := mb.pilvytisOrderIssuer.CreatePaymentGatewayOrder(
-		identity.FromAddress(req.IdentityAddress),
-		req.Gateway,
-		req.MystAmount,
-		req.PayCurrency,
-		req.Country,
-		req.GatewayCallerData,
+		pilvytis.GatewayOrderRequest{Identity: identity.FromAddress(req.IdentityAddress),
```
tomasmik, line 148: This should now probably send some projectID no?

tomasmik (changes requested):
It would probably also makes sense to update CLI to send some kind of project ID

mobile/mysterium/payment_orders.go
mdomasevicius: projectID is defaulted to `mysteriumvpn` which is reserved for desktop and mobile :)

tomasmik (approved):

Issue #4459, opened 2021-12-09
Status of Withdrawals, transaction details


isaackielma:
**Is your feature request related to a problem? Please describe.**

If for some reason it shows on CLI withdraw succeeded, but it does not reflect on balance. Would be nice if there was an easy way to track. Or any info to be able to check on blockchain explorer.

<img width="1290" alt="Screenshot 2021-12-09 at 18 21 55" src="https://user-images.githubusercontent.com/84313603/145453725-f478d736-b420-4eac-80f5-6eef8972bfc3.png">

When we see "withdrawal succeeded" for a mainnet CLI withdraw, there should be some public way to see the status of this, similar to a blockchain explorer, so we can see any issues or if the transaction fails or stuck some where

Donatas-MN:
Balance issue is now fixed. Regarding transaction hash, it's already available in the UI, but we will also add it to CLI.

tomasmik:
TequilaAPI already exposes an endpoint `/identities/settle/history` which is also used by the NodeUI.

- [ ] Add a new command `identities last-withdrawal <IDENTITY>` which would browse that endpoint and return you some information on your last withdrawal.

n10ty:
Added command ```identities last-withdrawal <identity>```
Example of response:
```
[INFO] Tx hash:  0x01c3fae5848950f178b04061ed899fdccda77693d2099139769425ad1a2bf178
[INFO] DateTime:  2022-03-10T13:08:21Z
[INFO] Beneficiary:  0x40a307f2e7C62Ad833349C8707BE907e523ec431
[INFO] Amount:  0.077056
[INFO] Fees:  0.022944
[INFO] Blockchain explorer Url:  https://polygonscan.com/tx/0x01c3fae5848950f178b04061ed899fdccda77693d2099139769425ad1a2bf178
[INFO] Error: none
```

PR #4864, opened 2022-03-10
#4459 Cli: latest withdrawal details

n10ty:
#4459 Add info about last transaction to cli


cmd/commands/cli/command_identities.go
```diff
@@ -569,3 +572,34 @@ func (c *cliApp) importIdentity(actionsArgs []string) (err error) {
...
+	}
+
+	lastWithdrawal := history.Items[0]
+
+	clio.Info("Tx hash: ", lastWithdrawal.TxHash)
+	clio.Info("DateTime: ", lastWithdrawal.SettledAt)
+	clio.Info("Amount: ", money.New(lastWithdrawal.Amount))
+	clio.Info("Fees: ", money.New(lastWithdrawal.Fees))
+	clio.Info("Blockchain explorer Url: ", lastWithdrawal.BlockExplorerURL)
+	if lastWithdrawal.Error != "" {
```
tomasmik, line 601: Maybe it's just me, but I would rather always print the Error field but let it be `null` or something like that if no error exists. It would be easier to parse if someone is using the CLI in a script of some sort 🤷🏻 

Though this I guess would be up for debate.

n10ty: I agree

tomasmik (approved):
lgtm, just squash those commits so that everything is verified

Guillembonet (approved):
nice, as Tomas said, squash those commits :)

PR #4895, opened 2022-03-18
Go 1.18

tadaskay:


tomasmik (approved):

Snawoot (changes requested):
We need xgo and xgomobile also updated before that.

Snawoot (approved):

PR #4921, opened 2022-03-24
Calculate channel impl instead of using hardcoded during sessions

tomasmik:
* New providers have compatibility set to 2
* New providers will now accept a connection from consumers registered with any arbitrary Hermes

Requires: https://github.com/mysteriumnetwork/payments/pull/189


Guillembonet (approved):
nice

soffokl (approved):

PR #4923, opened 2022-03-24
Remove Transactor Referral Program API

n10ty:
Referral API is not supported anymore. There are 0 referral campaigns. Instead of it, there is an Affiliate program.


tequilapi/endpoints/transactor.go
```diff
@@ -586,44 +583,6 @@ func (te *transactorEndpoint) SettleIntoStakeAsync(c *gin.Context) {
...
-// responses:
-//   200:
-//     description: Token Reward
-//     schema:
-//       "$ref": "#/definitions/TokenRewardAmount"
-//   500:
-//     description: Internal server error
-//     schema:
-//       "$ref": "#/definitions/ErrorMessageDTO"
-func (te *transactorEndpoint) TokenRewardAmount(c *gin.Context) {
```
tomasmik, line 607: This is probably not safe to remove. @Guillembonet could maybe comment more.

I am pretty sure that apps use this to call transactor which proxies requests to affiliator to check for token rewards. This was in a way used to any token, affiliate, referral, whatever.

tomasmik (changes requested):

tequilapi/endpoints/transactor.go
Guillembonet: yep not removable, let's talk on slack

n10ty: You are right, thanks. This endpoint also related to Affiliate program.

PR #4927, opened 2022-03-25
Add uid into eventbus wrapper to mitigate unintended unsubscribe

soffokl:


Guillembonet (approved):
I think it should work! nice code!

tomasmik (approved):

PR #4941, opened 2022-03-28
Referral api changes

n10ty:
* Remove Referral API from cli and TequilaAPI: GetTokenReward, GetReferralToken, ReferralTokenAvailable
* Add Affiliator to Node
* Do TokenRewardAmount request directly to Affiliator, not Transactor


metadata/network.go
```diff
@@ -66,6 +67,7 @@ var MainnetDefinition = NetworkDefinition{
 	AccessPolicyOracleAddress: "https://trust.mysterium.network/api/v1/access-policies/",
 	BrokerAddresses:           []string{"nats://broker.mysterium.network"},
 	TransactorAddress:         "https://transactor.mysterium.network/api/v1",
+	AffiliatorAddress:         "https://affiliator.mysterium.network/api/v1",
```
Guillembonet, line 70: not sure about it but I think it would be nice to use just "https://affiliator.mysterium.network" in order to be able to call both v1 and v2 endpoints in case we have some in the future, maybe thats a bad choice for some reason

tequilapi/endpoints/transactor.go
```diff
@@ -816,7 +817,7 @@ func AddRoutesForTransactor(
 			transGroup.POST("/stake/increase/async", te.SettleIntoStakeAsync)
 			transGroup.POST("/stake/decrease", te.DecreaseStake)
 			transGroup.POST("/settle/withdraw", te.Withdraw)
-			transGroup.GET("/token/:token/reward", te.TokenRewardAmount)
+			transGroup.GET("/token/:token/reward", a.TokenRewardAmount)
```
Guillembonet, line 820: this looks like it repeated, I'd say we should just keep `/affiliator` and bump the version, but maybe it's not such a good idea if we prefer to keep backwards compatibility

Guillembonet (approved):
Looks good to me, just 2 comments. Also some parts of the code I'm not very familiar with so whoever reviews next do it carefully pls

tequilapi/endpoints/transactor.go
n10ty: Backward compatibility requires to leave this endpoint in "transactor" namespace

tomasmik (approved):

PR #4977, opened 2022-04-06
Return whether channel is set as beneficiary or not

Guillembonet:
To do this we need the list of all hermeses, it will be fetched from observer if possible or from parameters if it fails or no url is set.


config/flags_observer.go
```diff
@@ -0,0 +1,40 @@
...
+
+package config
+
+import (
+	"github.com/mysteriumnetwork/node/metadata"
+	"github.com/urfave/cli/v2"
+)
+
+// FlagObserverAddress address of Observer service.
+var FlagObserverAddress = cli.StringFlag{
```
tomasmik, line 26: This could just be a part of `payments` flags 🤷🏻

tequilapi/endpoints/identities.go
```diff
@@ -441,14 +444,32 @@ func (ia *identitiesAPI) RegistrationStatus(c *gin.Context) {
...
+		utils.ForwardError(c, err, apierror.Internal("Failed to get list of known hermeses", contract.ErrCodeBeneficiaryGet))
+	}
+	for _, h := range hermeses {
+		channelAddr, err := ia.addressProvider.GetHermesChannelAddress(chainID, identity, h)
+		if err != nil {
+			log.Err(err).Msgf("could not generate channel address for chain %d and hermes %s", chainID, h.Hex())
+			utils.ForwardError(c, err, apierror.Internal("Failed to generate channel address", contract.ErrCodeBeneficiaryGet))
+		}
+		if channelAddr == beneficiary {
+			isChannel = true
```
tomasmik, line 466: probably a break here

tequilapi/endpoints/identities.go
```diff
@@ -441,14 +444,32 @@ func (ia *identitiesAPI) RegistrationStatus(c *gin.Context) {
...
+	isChannel := false
+	hermeses, err := ia.addressProvider.GetKnownHermeses(chainID)
+	if err != nil {
+		utils.ForwardError(c, err, apierror.Internal("Failed to get list of known hermeses", contract.ErrCodeBeneficiaryGet))
+	}
+	for _, h := range hermeses {
+		channelAddr, err := ia.addressProvider.GetHermesChannelAddress(chainID, identity, h)
+		if err != nil {
+			log.Err(err).Msgf("could not generate channel address for chain %d and hermes %s", chainID, h.Hex())
+			utils.ForwardError(c, err, apierror.Internal("Failed to generate channel address", contract.ErrCodeBeneficiaryGet))
```
tomasmik, line 463: probably a return here and above ^

tomasmik (approved):

PR #4992, opened 2022-04-10
Supervisor permissions update

otkd:
Permissions changed to be consistent with standard


supervisor/config/config.go
```diff
@@ -53,7 +53,7 @@ func (c Config) Write() error {
 	if err != nil {
 		return fmt.Errorf("could not encode configuration: %w", err)
 	}
-	if err := ioutil.WriteFile(confPath, []byte(out.String()), 0700); err != nil {
+	if err := ioutil.WriteFile(confPath, []byte(out.String()), 0644); err != nil {
```
Snawoot, line 56: This one will likely panic on windows because modes other than 0600 and 0400 are not supported there.

supervisor/daemon/transport/transport_darwin.go
```diff
@@ -42,10 +42,7 @@ func Start(handle handlerFunc, options Options) error {
 	if err != nil {
 		return fmt.Errorf("failed to parse uid %s: %w", options.Uid, err)
 	}
-	if err := os.Chown(sock, numUid, -1); err != nil {
-		return fmt.Errorf("failed to chown supervisor socket to uid %s: %w", options.Uid, err)
-	}
-	if err := os.Chmod(sock, 0700); err != nil {
+	if err := os.Chmod(sock, 0666); err != nil {
```
Snawoot, line 45: same

Snawoot (changes requested):
I don't see which value adds this PR, please provide details. Also please check if it works on Windows.

supervisor/config/config.go
otkd: [Doesn't seem like Windows version is implemented](https://github.com/mysteriumnetwork/node/blob/master/supervisor/config/const_windows.go), also the change is from 0700 which would also panic?

supervisor/daemon/transport/transport_darwin.go
otkd: This is for macOS(darwin)

otkd:
> I don't see which value adds this PR, please provide details. Also please check if it works on Windows.

I just checked again and can't seem to find any changes which would apply to Windows? 

Otherwise, this should make supervisor on Mac more consistent with Apple's filesystem changes in recent releases of macOS.

supervisor/config/config.go
Snawoot: No, 0700 should be fine. Windows version of supervisor is actively used.

otkd: > No, 0700 should be fine. Windows version of supervisor is actively used.

I've [split the config](https://github.com/mysteriumnetwork/node/pull/4992/commits/716f1dd5ec692ea5d106d32a6e23aec1510546c9) files leaving 0700 on Windows and 0644 on Mac & Linux

Snawoot:
@otkd what is the reason of changing socket permissions on MacOS to 0666? it will allow any user to control the daemon.

otkd:
> @otkd what is the reason of changing socket permissions on MacOS to 0666? it will allow any user to control the daemon.

@Snawoot that is the idea, since the daemon is globally installed i.e in `/Library/LaunchDaemons` vs `~/Library/LaunchDaemons` the socket is owned by root.

Example:
Screenshot 2022-05-02 at 11 53 57@2x

PR #4996, opened 2022-04-11
Bump transactor 2.6.5 -> 2.7.0

n10ty:
Signed-off-by: AndreyM <andrey@mysterium.network>


tomasmik (approved):

Guillembonet (commented):
I think it needs parameters as well and maybe observer as a mocked service, I'll do this now too

Guillembonet (approved):
I had to change tests a bit also, seems fine now

PR #5001, opened 2022-04-12
Add OpenChannel and ChannelStatus methods to Transactor API

n10ty:
Signed-off-by: AndreyM <andrey@mysterium.network>


identity/registry/transactor.go
```diff
@@ -397,6 +397,126 @@ func (t *Transactor) signRegistrationRequest(signer identity.Signer, regReq Iden
...
+)
+
+// ChannelStatusResponse represents response with channel status
+type ChannelStatusResponse struct {
+	Status ChannelStatus `json:"status"`
+}
+
+// ChannelStatus check the status of the channel
+func (t *Transactor) ChannelStatus(chainID int64, id, hermesID, registryAddress string) (ChannelStatusResponse, error) {
+	endpoint := "channel/status"
```
Guillembonet, line 498: setting this just for using it below seems a bit useless, same above

tequilapi/contract/channel.go
```diff
@@ -59,3 +59,40 @@ type PaymentChannelDTO struct {
 	// Beneficiary - eth wallet address
 	Beneficiary string `json:"beneficiary"`
 }
+
+// OpenChannelRequest represents request to open payment channel
+// swagger:model OpenChannelRequest
+type OpenChannelRequest struct {
```
Guillembonet, line 65: isn't this defined in the previous file too?

Guillembonet (approved):
some minor things but looks nice IMO, maybe we could add tests for it? not sure how painful and useful would they be

PR #5004, opened 2022-04-12
make sure beneficiary is set when settling

Guillembonet:
Also changed default Hermes, we probably want to do this sometime too

This PR also contains rebase of the current master


tomasmik (approved):
Looks fine, just not sure when we release this. Probably work it out with node team.

PR #5013, opened 2022-04-13
Hermes migration consumer

n10ty:


consumer/migration/hermes_migrator.go
```diff
@@ -0,0 +1,229 @@
...
+	oldBalance, err := m.getBalance(chainID, oldHermes.Hex(), id)
+	if err != nil {
+		return fmt.Errorf("error during getting balance: %w", err)
+	}
+	if crypto.FloatToBigMyst(oldBalanceMigrationMinimumMyst).Cmp(oldBalance) > 0 {
+		log.Debug().Msgf("Not enough balance for migration (id: %s, old balance: %.2f)", id, crypto.BigMystToFloat(oldBalance))
+		return nil
+	}
+
+	statusResponse, err := m.transactor.ChannelStatus(chainID, id, activeHermes.Hex(), registryAddress.Hex())
```
Guillembonet, line 101: maybe we can check directly in the blockchain instead of offloading this check to transactor

consumer/migration/hermes_migrator.go
```diff
@@ -0,0 +1,229 @@
...
+
+	registryAddress, err := m.addressProvider.GetRegistryAddress(chainID)
+	if err != nil {
+		return fmt.Errorf("could not get registry address: %w", err)
+	}
+	oldBalance, err := m.getBalance(chainID, oldHermes.Hex(), id)
+	if err != nil {
+		return fmt.Errorf("error during getting balance: %w", err)
+	}
+	if crypto.FloatToBigMyst(oldBalanceMigrationMinimumMyst).Cmp(oldBalance) > 0 {
```
Guillembonet, line 96: I think we should also open channel if there is no balance (and then don't withdraw)

Guillembonet, line 96: I think we should also open channel if there is no balance

consumer/migration/hermes_migrator.go
```diff
@@ -0,0 +1,229 @@
...
+	}
+	knownHermeses, err := m.addressProvider.GetKnownHermeses(chainID)
+	if err != nil {
+		return fmt.Errorf("could not get hermes address: %w", err)
+	}
+	oldHermeses := getOldHermeses(knownHermeses, activeHermes)
+	if len(oldHermeses) != 1 {
+		return nil
+	}
+	oldHermes := oldHermeses[0]
```
Guillembonet, line 86: here I would loop all of them if its not too hard, usually there should only be one to migrate from anyway

Guillembonet, line 86: here I would loop all of them if its not too hard, usually there should only be one to migrate from anyway

consumer/migration/hermes_migrator.go
```diff
@@ -0,0 +1,229 @@
...
+	oldBalance, err := m.getBalance(chainID, oldHermes.Hex(), id)
+	if err != nil {
+		return false, fmt.Errorf("error during getting balance: %w", err)
+	}
+	newBalance, err := m.getBalance(chainID, activeHermes.Hex(), id)
+	if err != nil {
+		return false, fmt.Errorf("error during getting balance: %w", err)
+	}
+
+	return crypto.FloatToBigMyst(oldBalanceMigrationMinimumMyst).Cmp(oldBalance) < 0 && newBalance.Cmp(new(big.Int)) == 0, nil
```
Guillembonet, line 172: I think it's in the notes, but do we not want to migrate if new channel has money but old also has?

Guillembonet, line 172: I think it's in the notes, but do we not want to migrate if new channel has money but old also has?

Guillembonet (commented):
overall its looking nice

tequilapi/endpoints/identities.go
```diff
@@ -599,6 +601,80 @@ func (ia *identitiesAPI) SavePayoutAddress(c *gin.Context) {
...
+	if err != nil {
+		c.Error(apierror.Internal(err.Error(), contract.ErrCodeHermesMigration))
+		log.Err(err).Msgf("could not migrate identity %s", id)
+		return
+	}
+
+	c.Status(http.StatusOK)
+}
+
+// swagger:operation GEY /identities/:id/migrate-hermes/status
```
tomasmik, line 640: This endpoint makes it look like you will return a status for a migration - like is the migration going ok, but rather we're returning _if the migration is required._

tomasmik (commented):
Looks pretty good, but it's missing mobile bindings: https://github.com/mysteriumnetwork/node/tree/master/mobile/mysterium

Also about old hermes enumeration. It's a bit tricky imo.. But maybe we should actually enumerate all hermeses for balance checks etc. Another thing we could do is call blockchain to get hermes version and assume something like:

If our current hermes version is v3 we should only check/migrate hermes v2 and ignore all previous hermeses. That makes sense in my head at least.

consumer/migration/hermes_migrator.go
n10ty: This is a specific Transactor endpoint to check channel status

consumer/migration/hermes_migrator.go
n10ty: I don't think supporting several hermeses make sense here. @tomasmik what do you think?

consumer/migration/hermes_migrator.go
n10ty: If we migrate once and money appears in old channel again - I don't think we need do it again

consumer/migration/hermes_migrator.go
n10ty: To overcomplicate code just for hypotetical case where we have 3 or more hermeses and migrate from any to new - I don't think is a good idea

consumer/migration/hermes_migrator.go
```diff
@@ -0,0 +1,229 @@
...
+	for {
+		select {
+		case <-timer.C:
+			return fmt.Errorf("channel opening timeout for id: %s", id)
+		case <-time.After(5 * time.Second):
+			statusResponse, err := m.transactor.ChannelStatus(chainID, id.Hex(), hermesID.Hex(), registryAddress.Hex())
+			if err != nil {
+				log.Debug().Msgf("Hermes migration error: open channel failed %s", err.Error())
+				return err
+			} else if statusResponse.Status == registry.ChannelStatusFail || statusResponse.Status == registry.ChannelStatusOpen {
```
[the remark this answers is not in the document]
n10ty: Well, idea is Transactor only fail to open channel if there is someone else, sent this transaction before (to take reward). Correct me if I wrong here, I remember Depot Delivery works in this way

tequilapi/endpoints/identities.go
n10ty: In wide sense it is. We just don't have "in progress" status, but we have kind of "to do" and "finished".

tomasmik (approved):

PR #5036, opened 2022-04-19
Allow free registration if paid in pilvytis

Guillembonet:
If pilvytis call fails (pilvytis is down) it will just print a log and assume the order is not paid.


tequilapi/endpoints/pilvytis.go
```diff
@@ -455,6 +456,38 @@ func (e *pilvytisEndpoint) CreatePaymentGatewayOrder(c *gin.Context) {
...
+//       items:
+//         "$ref": "#/definitions/RegistrationPaymentResponse"
+//   500:
+//     description: Internal server error
+//     schema:
+//       "$ref": "#/definitions/APIError"
+func (e *pilvytisEndpoint) GetRegistrationPaymentStatus(c *gin.Context) {
+	resp, err := e.api.GetRegistrationPaymentStatus(identity.FromAddress(c.Param("id")))
+	if err != nil {
+		utils.ForwardError(c, err, apierror.Internal("Failed to list orders", contract.ErrCodePaymentList))
```
tomasmik, line 484: probably incorrect error here?

tequilapi/endpoints/transactor.go
```diff
@@ -320,7 +328,14 @@ func (te *transactorEndpoint) RegisterIdentity(c *gin.Context) {
 	}
 
 	regFee := big.NewInt(0)
-	if req.ReferralToken == nil {
+	canRegisterForFree := false
+	resp, err := te.pilvytis.GetRegistrationPaymentStatus(id)
```
tomasmik, line 332: move to this a seperate method and get rid of `if/else if`?

tomasmik (approved):

PR #5057, opened 2022-04-26
Increase invoice timers during session

tomasmik:
Very basic increase logic for invoices being sent.

New charge period or charge limit is calculated by dividing the current
limit from 3 and adding it on top. 3 is a magic number I came up with,
open to suggestions here.

What this solves:

Currently most of our sessions are long running superproxy sessions, if
we increase the time between invoices for long running or heavy traffic
usage sessions we will decrease the need to hit hermes as much.
This also ensures that at the session start we do more invoices further
decreasing underpaid sessions.

For long running sessions with no traffic, current approach reduces
invoices up to 5 times. For ones sending traffic 10.


Guillembonet:
I haven't checked the code but I think its a nice idea, charge less often for long-running sessions since there is more trust in place + the amount you could lose is much less relatively to the session.

n10ty:
Good job 👍

soffokl (approved):
🚀

PR #5058, opened 2022-04-26
node with no blockchain as provider

Guillembonet:
Node now has fallback services to provide the info required to serve traffic without blockchain.

### DIFF
--- a/session/pingpong/hermes_url_getter.go
+++ b/session/pingpong/hermes_url_getter.go
@@ -23,6 +23,8 @@ import (
 	"sync"
 
 	"github.com/ethereum/go-ethereum/common"
+	"github.com/mysteriumnetwork/node/observer"
+	"github.com/rs/zerolog/log"
 )
 
 // HermesURLGetter allows for fetching and storing of hermes urls.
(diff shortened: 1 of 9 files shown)


session/pingpong/hermes_url_getter.go
```diff
@@ -84,19 +93,36 @@ func (hug *HermesURLGetter) GetHermesURL(chainID int64, address common.Address)
...
+}
+
+func (hug *HermesURLGetter) getHermesURLObserver(chainID int64, address common.Address) (string, error) {
+	hermesData, err := hug.observer.GetHermesData(chainID, address)
 	if err != nil {
 		return "", err
 	}
-	hug.loadedAddresses[chainID][address] = add
-	return add, nil
+	log.Debug().Msgf("returned url from observer %s=%s", address.Hex(), hermesData.URL)
```
tomasmik, line 126: Lets use `Fields()` or something instead of dynamic log message. Easier to grep if needed.

tomasmik (approved):

session/pingpong/hermes_url_getter.go
Guillembonet: it was leftover from debugging, I'll delete it!

PR #5070, opened 2022-04-28
remove useless code from old registration flows

Guillembonet:
Removed stuff related to:
- Automatic provider registration (we now require beneficiary)
- Free registrations
- Provider registration being any different

We can also remove referrals code since its not used now.

Old node UI version in the repo still uses some of the calls deleted in this PR, we need to upgrade it first. @mdomasevicius 


tequilapi/endpoints/transactor.go
```diff
@@ -658,64 +656,6 @@ func (te *transactorEndpoint) ChainSummary(c *gin.Context) {
...
-// responses:
-//   200:
-//     description: Eligibility response
-//     schema:
-//       "$ref": "#/definitions/EligibilityResponse"
-//   500:
-//     description: Internal server error
-//     schema:
-//       "$ref": "#/definitions/APIError"
-func (te *transactorEndpoint) FreeRegistrationEligibility(c *gin.Context) {
```
tomasmik, line 685: I dont really see a big reason to remove this. 🤷🏻  Could prove to be useful again in the future as well. Would be interesting to hear what other people think. @soffokl @Snawoot @n10ty 

This is also a breaking change which not only breaks webUI but probably also apps.

Guillembonet: also @mdomasevicius will be interested in this discussion I think

Guillembonet: Also, keep in mind that one of the endpoints is for providers, and therefore only place it can break is node UI and it's not used anymore there.

mdomasevicius: @tomasmik you can emulate free registration with `/registration-payment` endpoint

Guillembonet:
no consensus

Issue #5067, opened 2022-04-28
Allow to configure fee options for auto settlements


tomasmik:
Problem:

Settlements are now automatic meaning that user has less control over when to settle. If fees in the network spike user could loose most of his settlement just for the fees.

Solution:

Allow a node runner to configure some threshold for an acceptable fee loss when settling. This could be as simple as a flag on start up, so that it's controlled using the node config. NodeUI could also change this setting then as it's able to update config.

You should also ignore this configuration if the settlement is not automatic, meaning it was issued by a request to tequilaAPI.

PR #5113, opened 2022-05-12
Smart settle

Guillembonet:
Closes: #5067

Current problems:
- We settle at 5 MYST always and maybe fees are huge when we do, it's a matter of luck.
- You might force settle from frontend and realize you paid 10x more fees that what was displayed in UI if you are unlucky and do so just before a gas spike.
- We call transactor `/fee` endpoint a lot. In the last 3 minutes there are 1400 calls yet coming only from 550 different nodes.

Proposed solutions:
- Automatic settlements will be smarter now and work as follows:
  - We have 3 values: minimum to settle(5), max fee percentage(5%), and max unsettled(20)
  - If unsettled is below minimum we don't settle
  - If unsettled is above the maximum we always auto-settle
  - If it's somewhere in between we automatically settle if the transaction fees are less than the percentage (i.e. if we have 6 MYST unsettled we will settle as long as the fees are less than 0.3 MYST)
- ~~Now all settling functions allow an additional maxFee param so we can add it as optional parameter in tequilapi to prevent the user from finding himself on the 2nd problem mentioned above. (not in tequilapi yet, if we won't want to do this I can simplify the code before merging)~~
- Transactor fees are now cached for ~~3 minutes~~ 30 seconds (by default) instead of calling every time. Keep in mind that the smart settlement check requires a call there on every hermes promise received (only if we have more unsettled than 5), so with both changes, we might still call transactor `/fee` endpoint quite a lot but I think it should be less. We can discuss whether we should increase this cache amount a bit more (more than 10 mins will for sure be bad).

~~Additionally, I removed some old logic in the settler which auto-settled if the unsettled amount was more than 10% of the current balance (only if we had stake != 0 so it wasn't used now). I'm not sure if this is still useful now, it doesn't look like it is.~~

### DIFF
--- a/config/flags_payments.go
+++ b/config/flags_payments.go
@@ -79,11 +79,23 @@ var (
 		Usage:  "The duration we'll wait before giving up trying to fetch new balance.",
 		Hidden: true,
 	}
-	// FlagPaymentsZeroStakeUnsettledAmount determines the minimum amount of myst required before auto settling is triggered if zero stake is used.
+	// FlagPaymentsZeroStakeUnsettledAmount determines the minimum amount of myst that we will settle automatically if zero stake is used.
 	FlagPaymentsZeroStakeUnsettledAmount = cli.Float64Flag{
 		Name:  "payments.zero-stake-unsettled-amount",
 		Value: 5.0,
 		Usage: "The settling threshold if provider uses a zero stake",
+	}
+	// FlagPaymentsPromiseSettleMaxFeeThreshold represents the max percentage of the settlement that will be acceptable to pay in transaction fees.
+	FlagPaymentsPromiseSettleMaxFeeThreshold = cli.Float64Flag{
+		Name:  "payments.settle.max-fee-percentage",
+		Value: 0.05,
+		Usage: "The max percentage we allow to pay in fees when automatically settling promises.",
+	}
+	// FlagPaymentsUnsettledMaxAmount determines the maximum amount of myst for which we will consider the fee threshold.
+	FlagPaymentsUnsettledMaxAmount = cli.Float64Flag{
+		Name:  "payments.unsettled.max-amount",
+		Value: 20.0,
+		Usage: "The maximum amount of unsettled myst, after that we will always try to settle.",
 	}
 	// FlagPaymentsRegistryTransactorPollInterval The duration we'll wait before calling transactor to check for new status updates.
 	FlagPaymentsRegistryTransactorPollInterval = cli.DurationFlag{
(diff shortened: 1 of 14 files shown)


tomasmik (approved):
I would probably just remove the cache, but it's fine either way.

n10ty (approved):

config/flags_payments.go
```diff
@@ -79,11 +73,23 @@ var (
 		Usage:  "The duration we'll wait before giving up trying to fetch new balance.",
 		Hidden: true,
 	}
-	// FlagPaymentsZeroStakeUnsettledAmount determines the minimum amount of myst required before auto settling is triggered if zero stake is used.
-	FlagPaymentsZeroStakeUnsettledAmount = cli.Float64Flag{
-		Name:  "payments.zero-stake-unsettled-amount",
+	// FlagPaymentsAutoPromiseSettleMinAmount represents the minimum amount of myst that we will settle automatically.
+	FlagPaymentsAutoPromiseSettleMinAmount = cli.Float64Flag{
+		Name:  "payments.settle.min-amount",
```
tomasmik, line 78: This will break everything related to the config and this param. Which is old config and probably UI.

session/pingpong/hermes_promise_settler.go
```diff
@@ -1172,21 +1194,22 @@ func (ss settlementState) needsSettling(threshold float64, zeroStakeThreshold fl
...
-	}
-
-	floated := new(big.Float).SetInt(channel.availableBalance())
-	calculatedThreshold := new(big.Float).Mul(big.NewFloat(threshold), floated)
-	possibleEarnings := channel.UnsettledBalance()
-	i, _ := calculatedThreshold.Int(nil)
-	if possibleEarnings.Cmp(i) == -1 {
-		return false
+	unsettledAmount := channel.UnsettledBalance()
+	if unsettledAmount.Cmp(crypto.FloatToBigMyst(maxUnSettledAmount)) == 1 {
```
tomasmik, line 1198:
```suggestion
	if unsettledAmount.Cmp(crypto.FloatToBigMyst(maxUnSettledAmount)) > 0 {
```

session/pingpong/hermes_promise_settler.go
```diff
@@ -1172,21 +1194,22 @@ func (ss settlementState) needsSettling(threshold float64, zeroStakeThreshold fl
...
-	if channel.balance().Cmp(i) <= 0 {
-		return true
+	if unsettledAmount.Cmp(crypto.FloatToBigMyst(minSettleAmount)) >= 0 && settleFees != nil {
+		if time.Now().After(settleFees.ValidUntil) {
+			log.Warn().Msg("got invalid settlement fees, skipping check")
+		} else {
+			unsettledBalance := new(big.Float).SetInt(channel.UnsettledBalance())
+			calculatedFeesThreshold := new(big.Float).Mul(big.NewFloat(threshold), unsettledBalance)
+			calculatedFeesThresholdInt, _ := calculatedFeesThreshold.Int(nil)
+			if settleFees.Fee.Cmp(calculatedFeesThresholdInt) == -1 {
```
tomasmik, line 1209:
```suggestion
			if settleFees.Fee.Cmp(calculatedFeesThresholdInt) < 0 {
```

tomasmik (approved):

tomasmik (approved):

PR #5159, opened 2022-05-31
Save hermes fee in settler

tomasmik:
Hermes fee never changes but now we keep pinging our nodes and asking
for it. Ask once and keep until next restart, this is good enough.

### DIFF
--- a/session/pingpong/hermes_promise_settler.go
+++ b/session/pingpong/hermes_promise_settler.go
@@ -124,11 +124,34 @@ type hermesPromiseSettler struct {
 	paySettler                 paySettler
 	promiseStorage             promiseStorage
 	publisher                  eventbus.Publisher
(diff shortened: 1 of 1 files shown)


session/pingpong/hermes_promise_settler.go
```diff
@@ -124,13 +124,36 @@ type hermesPromiseSettler struct {
...
+}
+
+func (h *hermesFees) set(chainID int64, hermesID common.Address, fee uint16) {
+	h.m.Lock()
+	defer h.m.Unlock()
+	h.fees[h.key(chainID, hermesID)] = fee
+}
+
+func (h *hermesFees) get(chainID int64, hermesID common.Address) (uint16, bool) {
+	h.m.RLock()
```
n10ty, line 151: Locks are mixed up

PR #5186, opened 2022-06-14
Add initial support for wireguard provider in userspace

soffokl:


Snawoot:
As far as I know we use automatic switching between in-kernel (wg module) and userspace (TUN) implementation. Should we also use such autoswitching via probing for full userspace impl. as well?

soffokl:
@Snawoot sure, but there are many missing parts now like protection for accessing LAN. I want to keep it as an extra flag for now.

Snawoot (approved):

services/wireguard/endpoint/netstack/netstack.go
```diff
@@ -0,0 +1,359 @@
...
+	for _, ip := range tun.localAddresses {
+		if tcpip.Address(ip.AsSlice()) == remoteAddr {
+			return true
+		}
+	}
+
+	return false
+}
+
+func (tun *netTun) proxy(ctx context.Context, cancel context.CancelFunc, dst net.PacketConn, dstAddr net.Addr, src net.PacketConn, extend func()) {
```
Zensey, line 331: Can be done w/o using a time.AfterFunc by adding 2 SetReadDeadline()

```
func (tun *NetTun) proxy(ctx context.Context, cancel context.CancelFunc, dst net.PacketConn, dstAddr net.Addr, src net.PacketConn, extend func()) {
	defer cancel()
	buf := make([]byte, tun.mtu)

	for {
		select {
		case <-ctx.Done():
			return
		default:
			src.SetReadDeadline(time.Now().Add(idleTimeout))
			n, srcAddr, err := src.ReadFrom(buf)
			if e, ok := err.(net.Error); ok && e.Timeout() {
				return
			} else if err != nil {
				if ctx.Err() == nil {
					log.Printf("Failed to read packed from %s", srcAddr)
				}
				return 
			}

			_, err = dst.WriteTo(buf[:n], dstAddr)
			if err != nil {
				if ctx.Err() == nil {
					log.Printf("Failed to write packed to %s", dstAddr)
				}
				return
			}
			dst.SetReadDeadline(time.Now().Add(idleTimeout))
		}
	}
}
```

Zensey:
For some next revision:
packet source adress matching could added, so that unwanted packages could be dropped.

Example: 
https://github.com/xjasonlyu/tun2socks/blob/1536735456ae36fe3c9f1455bb4fe8fb10d77bc9/tunnel/udp.go#L120

Donatas-MN:
Missing parts:
- Protection for accessing LAN
- Traffic shaping for bandwidth limit

services/wireguard/endpoint/netstack-provider/private.go
```diff
@@ -0,0 +1,63 @@
...
+			panic(err)
+		}
+		ipnets[i] = ipnet
+	}
+	return ipnets
+}
+
+var privateIPv4Block []*net.IPNet
+
+var txtPrivateIPv4Block = []string{
```
soffokl, line 38: Please use the already existing flag `firewall.protected.networks` for this instead of hardcoded values. Some might want to adjust it. 
https://github.com/mysteriumnetwork/node/blob/master/config/flags_node.go#L96-L101

Zensey:
It's finally ready to merge.

services/wireguard/endpoint/netstack-provider/private.go
Zensey: Without allowing 10.0.0.0/8 web sites can not be accessed.
So it has to be overriden ?

soffokl: You need to allow only single IP like `10.182.0.2` which is coming from the session, the rest of the traffic should be blocked.

Zensey: yes, access to 10.182.0.1 is reqired

services/wireguard/endpoint/netstack-provider/private.go
```diff
@@ -0,0 +1,60 @@
...
+		_, ipnet, err := net.ParseCIDR(cidr)
+		if err != nil {
+			panic(err)
+		}
+		ipnets[i] = ipnet
+	}
+	return ipnets
+}
+
+var privateIPv4Block []*net.IPNet
```
soffokl, line 40: This is a global variable, there might be a race condition when several consumers connect at the same time, this will be concurrently overwritten. Don't use a global variable, you can just get it from the config for every session creation.

Zensey:
Switching to new gvisor version introduced  leak of go-routines: 
github.com/mysteriumnetwork/node/services/wireguard/endpoint/netstack-provider.(*netTun).proxy

But adding `defer dst.Close()` seems to solve it

Zensey:
It looks like now its ready to be merged

Snawoot:
Congrats!

PR #5198, opened 2022-06-14
Migration fixes

n10ty:
Migrate during connect
Add migration in suggestion commands

### DIFF
--- a/cmd/commands/cli/command.go
+++ b/cmd/commands/cli/command.go
@@ -291,6 +291,18 @@ func (c *cliApp) connect(args []string) (err error) {
 	}
 
 	consumerID, providerID, serviceType := args[0], args[1], args[2]
+	migrationStatus, err := c.tequilapi.MigrateHermesStatus(consumerID)
+	if migrationStatus.Status == contract.MigrationStatusRequired {
+		clio.Infof("Hermes migration status: %s\n", migrationStatus.Status)
+		clio.Info("Migration started")
+		err := c.tequilapi.MigrateHermes(consumerID)
+		if err != nil {
+			return err
+		}
+		clio.Info("Migration finished successfully")
+		clio.Info("Try to reconnect")
+		return nil
+	}
 
 	if !services.IsTypeValid(serviceType) {
 		return fmt.Errorf("invalid service type, expected one of: %s", strings.Join(services.Types(), ","))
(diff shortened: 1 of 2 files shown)


cmd/commands/cli/command.go
```diff
@@ -291,6 +291,18 @@ func (c *cliApp) connect(args []string) (err error) {
 	}
 
 	consumerID, providerID, serviceType := args[0], args[1], args[2]
+	migrationStatus, err := c.tequilapi.MigrateHermesStatus(consumerID)
+	if migrationStatus.Status == contract.MigrationStatusRequired {
```
Guillembonet, line 295: I think maybe we should just print an error?

n10ty: I thought about it. But what about programatically run nodes? @tomasmik

tomasmik: This is a double edged sword, both ways have their drawbacks, just printing `Migration required` and not doing it might be inconvenient for a consumer and doing the migration for the can result in some funky business by accident. So I'll let you chose.

cmd/commands/cli/command_identities.go
```diff
@@ -164,8 +164,8 @@ func (c *cliApp) getIdentity(actionArgs []string) (err error) {
 	if err != nil {
 		return err
 	}
-	clio.Info("Registration Status:", identityStatus.RegistrationStatus)
-	clio.Info("Channel address:", identityStatus.ChannelAddress)
+	clio.Info("Registration Status: ", identityStatus.RegistrationStatus)
```
tomasmik, line 167: Can we fix this for unlock as well: https://github.com/mysteriumnetwork/node/blob/2c1e2bf05fd137aaf3209164f78d394507f32281/cmd/commands/cli/command_identities.go#L215

tomasmik, line 167: Few more like this:
```
» identities new 
[SUCCESS] New identity created:0xea4b7c48b1181de4604aa52954964c5c7574be0f  <--------------- HERE
» identities unlock 0xea4b7c48b1181de4604aa52954964c5c7574be0f
[INFO] Unlocking0xea4b7c48b1181de4604aa52954964c5c7574be0f  <--------------- HERE
```

cmd/commands/cli/command.go
n10ty: To add: this is required migration, but not optional

cmd/commands/cli/command_identities.go
n10ty: Fixed

tomasmik (approved):

PR #5218, opened 2022-06-17
Hermes migration fix

n10ty:
Add record to local db to save migration status finished

### DIFF
--- a/consumer/migration/hermes_migrator.go
+++ b/consumer/migration/hermes_migrator.go
@@ -334,14 +334,14 @@ func (m *HermesMigrator) isRegistered(chainID int64, id string) (bool, error) {
 	return status == registry.Registered, nil
 }
 
-func (m *HermesMigrator) markAsMigrated(hermesIdm id string) {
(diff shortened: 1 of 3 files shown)


consumer/migration/hermes_migrator.go
```diff
@@ -308,3 +327,23 @@ func (m *HermesMigrator) isRegistered(chainID int64, id string) (bool, error) {
 	}
 	return status == registry.Registered, nil
 }
+
+func (m *HermesMigrator) markAsMigrated(id string) {
+	err := m.db.SetValue(hermesMigrationBucketName, m.getMigrationKey(id), true)
```
tomasmik, line 337: This is bad. Current hermesID should be part of the migration bucket key, so that we can continue migration to another hermes as it changes.

tomasmik (changes requested):

consumer/migration/hermes_migrator.go
n10ty: Good point

tomasmik (approved):

Guillembonet (approved):

PR #5225, opened 2022-06-20
beneficiary set with optional hermes id

Guillembonet:
hermes id is now fully optional, if none is provided it will use the one with the most unsettled funds.


n10ty (approved):

tomasmik (approved):
I assume this still requires the nodeUI changes or else it's gonna break again.

PR #5226, opened 2022-06-20
Hermes migration: add blockchain check for open channel

n10ty:
For freshly registered users:
* Add subscription to registration event. Mark as no registration is needed.
* If local db (boltdb) in some case not available and old Hermes down - skip migration and mark it as 'done'
* Add preceding call to blockchain: if channel for old Hermes not opened - skip
For users who are using old Hermes:
* if old Hermes is down - still open new payment channel in new Hermes

### DIFF
--- a/cmd/di.go
+++ b/cmd/di.go
@@ -1013,6 +1013,7 @@ func (di *Dependencies) bootstrapHermesMigrator() *migration.HermesMigrator {
 		di.IdentityRegistry,
 		di.ConsumerBalanceTracker,
 		migration.NewStorage(di.Storage, di.AddressProvider),
+		di.BCHelper,
 	)
 }
 
--- a/cmd/di.go
+++ b/cmd/di.go
@@ -28,16 +28,10 @@ import (
 
 	"github.com/ethereum/go-ethereum/accounts/keystore"
 	"github.com/ethereum/go-ethereum/common"
-	"github.com/mysteriumnetwork/node/consumer/migration"
-	"github.com/pkg/errors"
-	"github.com/rs/zerolog/log"
-
-	paymentClient "github.com/mysteriumnetwork/payments/client"
-	psort "github.com/mysteriumnetwork/payments/client/sort"
-
 	"github.com/mysteriumnetwork/node/communication/nats"
 	"github.com/mysteriumnetwork/node/config"
 	"github.com/mysteriumnetwork/node/consumer/bandwidth"
+	"github.com/mysteriumnetwork/node/consumer/migration"
 	consumer_session "github.com/mysteriumnetwork/node/consumer/session"
 	"github.com/mysteriumnetwork/node/core/auth"
 	"github.com/mysteriumnetwork/node/core/beneficiary"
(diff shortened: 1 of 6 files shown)


consumer/migration/hermes_migrator.go
```diff
@@ -111,6 +120,16 @@ func (m *HermesMigrator) Start(id string) error {
 	if err != nil {
 		return fmt.Errorf("could not get registry address: %w", err)
 	}
+
+	opened, err := m.isChannelOpened(chainID, common.HexToAddress(id), activeHermes, registryAddress)
```
Guillembonet, line 124: what if withdrawal failed but channel was open? I think this blockchain check should be used as some sort of fallback option

consumer/migration/hermes_migrator.go
```diff
@@ -111,6 +120,16 @@ func (m *HermesMigrator) Start(id string) error {
...
+	if err != nil {
+		return fmt.Errorf("could not check channel status: %w", err)
+	}
+	if opened {
+		m.st.markAsMigrated(chainID, id)
+		return nil
+	}
+
```
Guillembonet, line 132: also maybe instead of doing this checks we could just call isMigrationRequired?

consumer/migration/hermes_migrator.go
tomasmik: That is correct. This should be used interchangeably with the transactor call or it should be smarter and assure some withdrawal success.

consumer/migration/hermes_migrator.go
tomasmik: Is migration required breaks for client who never did a migration and never called these endpoints. We need a sure way of ensuring the migration is either done or was never required even if you call this 2 years from now and all these services dont exist anymore and cannot provide you with any data.

n10ty: Then it will execute some code twice. I suggest to remain it as is.

consumer/migration/hermes_migrator.go
```diff
@@ -80,43 +89,31 @@ type HermesClient interface {
 // Start begins migration from old hermes to new
 func (m *HermesMigrator) Start(id string) error {
 	chainID := config.GetInt64(config.FlagChainID)
-	if !m.st.isMigrationRequired(chainID, id) {
-		log.Info().Msg("Migration is already done")
-		return nil
-	}
-
-	registered, err := m.isRegistered(chainID, id)
```
tomasmik, line 88: We still want to check this as if you are not registered, migration is not required.

cmd/di.go
```diff
@@ -796,7 +794,11 @@ func (di *Dependencies) bootstrapNetworkComponents(options node.Options) (err er
 		TransactorPollTimeout:  options.Payments.RegistryTransactorPollTimeout,
 	}
 
-	if di.IdentityRegistry, err = registry.NewIdentityRegistryContract(di.EtherClientL2, di.AddressProvider, registryStorage, di.EventBus, di.HermesCaller, di.Transactor, registryCfg); err != nil {
+	migrationStorage := migration.NewStorage(di.Storage, di.AddressProvider)
+	registerCallback := func(chainID int64, identity string) {
```
tomasmik, line 798: I would much rather have the migration subscribe to `AppTopicIdentityRegistration` event and mark the state as needed. This seems like something the registration should not care about.

tomasmik (changes requested):

consumer/migration/hermes_migrator.go
```diff
@@ -85,65 +95,65 @@ func (m *HermesMigrator) Start(id string) error {
...
 	}
+
+	// try open channel only if it's not opened yet.
+	if err = m.openChannel(id, err, chainID, activeHermes, registryAddress); err != nil {
+		return fmt.Errorf("open channel error: %w", err)
+	}
+
+	// get data from old hermes
 	data, err := m.getUserData(chainID, oldHermes.Hex(), id)
 	if err != nil {
```
tomasmik, line 125: I think we can do the same here, if we have money or a promise on a new hermes and old hermes return no data, we can assume we're ok and migration was done.

tomasmik (approved):

PR #5252, opened 2022-06-28
Prevent code from a panic if dns list is empty

andrey-mysterium:
Signed-off-by: Andrey <andreyt@mysterium.network>


Snawoot (changes requested):
Will still panic if `cfg.DNS` slice nil or empty.

Snawoot (approved):

services/wireguard/endpoint/proxyclient/client_test.go
```diff
@@ -0,0 +1,65 @@
+/*
+ * Copyright (C) 2018 The "MysteriumNetwork/node" Authors.
```
soffokl, line 2: For new files, this should be set to the current date. 
```suggestion
 * Copyright (C) 2022 The "MysteriumNetwork/node" Authors.
```

soffokl (approved):

Issue #2053, opened 2020-04-14
Add android app to f-droid


wuniversales:

chompomonim:
Android related issues should be created in node repo. Moving issue there.

PR #5263, opened 2022-06-30
Publish mobile-node to Maven central

tadaskay:
Key motivation is that F-droid only accepts apps with dependencies in credible and long-established maven repositories. Unfortunately, maven.mysterium.network is not one of them. Jitpack was another alternative, but it doesn't allow to upload pre-built artifacts which we have (build by gomobile).
Drawback is that artifacts need to be promoted manually (similar flow as launchpad) in https://oss.sonatype.org/ by:
1) closing staging repository which is created automatically upon deploying the artifacts;
2) promoting the closed repository to release;
3) wait for central sync (takes some time, according to docs it may be hours).
But I don't see any other good alternative at the moment.

POC that it works (uploaded manually using commands in this PR to Sonatype staging and later promoted to release) - using a checked out `1.9.0` tag:
https://repo1.maven.org/maven2/network/mysterium/mobile-node/1.9.0/

Required for: https://github.com/mysteriumnetwork/node/issues/2053


Snawoot:
Maybe it is worth to upload to both maven repos?

tadaskay:
> Maybe it is worth to upload to both maven repos?

The only advantage we would have is faster rollout for mobile-node. 
Disadvantage would be maintaining different setup for app build flavors and not being able to drop our self-hosted repo.
But we could do either way.

ci/packages/package.go
```diff
@@ -140,11 +139,11 @@ func PackageWindowsAmd64() error {
...
-	job.Precondition(func() bool {
-		pr, _ := env.IsPR()
-		fullBuild, _ := env.IsFullBuild()
-		return !pr || fullBuild
-	})
+	//job.Precondition(func() bool {
+	//	pr, _ := env.IsPR()
+	//	fullBuild, _ := env.IsFullBuild()
+	//	return !pr || fullBuild
+	//})
```
soffokl, line 146: Are we keeping it commented?

ci/packages/package.go
```diff
@@ -160,19 +159,19 @@ func PackageAndroid() error {
...
+	//if err != nil {
+	//	return err
+	//}
 	pomTemplate, err := template.ParseFiles("bin/package/android/mvn.pom")
 	if err != nil {
 		return err
 	}
 
-	buildVersion := env.Str(env.BuildVersion)
+	buildVersion := "1.9.0.1"
```
soffokl, line 171: Is it just a debugging version or what?

ci/release/android_sdk.go
```diff
@@ -18,178 +18,80 @@
...
-		},
-	)
-}
+	//if err := env.EnsureEnvVars(env.TagBuild, env.BuildVersion); err != nil {
+	//	return err
+	//}
+	//if !env.Bool(env.TagBuild) {
+	//	log.Info().Msg("Not a tag build, skipping ReleaseAndroidSDK action...")
+	//	return nil
+	//}
```
soffokl, line 51: Commented again.

ci/release/android_sdk.go
```diff
@@ -18,178 +18,80 @@
...
-		repositoryOpts: repositoryOpts,
-	}
-}
-
-func (up *mavenPublisher) upload(filename string) error {
-	log.Info().Msg("Uploading: " + filename)
 
-	file, err := os.Open(filename)
-	if err != nil {
+	artifactBaseName := fmt.Sprintf("build/package/mobile-node-%s", "1.9.0.1")
```
soffokl, line 61: Is it a hardcoded version again?

ci/packages/package.go
tadaskay: @soffokl See commit message 👯

tadaskay: The last commit was just for testing it on PR.

tadaskay: I'm deleting that before merge.

Snawoot (approved):

PR #5282, opened 2022-07-06
Update identities export cli command. Generate escaped JSON string

andrey-mysterium:
Previously identities export command generated an unescaped double quotes string, and if the user copy-paste it to shell command as argument, we have an issue with unmarshalling. Because shells don't pass arguments with not escaped quotes.


soffokl (approved):

soffokl:
The only thing I'd change is a commit message to reflect what was changed. 
After many days it will be hard to track commit messages like `initial commit`.
https://github.com/mysteriumnetwork/devbook/tree/master/git#commit-messages

Guillembonet (approved):
+1 on Dmitry's comment

tomasmik (approved):

PR #5301, opened 2022-07-15
Show uptime value in proposals quality response

andrey-mysterium:


market/service_proposal_test.go
```diff
@@ -59,7 +60,8 @@ func Test_ServiceProposal_Serialize(t *testing.T) {
       "quality": {
         "quality": 2.0,
         "latency": 5,
-        "bandwidth": 100
+        "bandwidth": 100,
+		"uptime": 20
```
n10ty, line 64: broken alignment?

n10ty:
In what units is uptime measured? Seconds? Minutes? Some index?

market/service_proposal_test.go
andrey-mysterium: yes, it look like it broke a linter. will fix now

andrey-mysterium:
>

PR #5313, opened 2022-07-19
Recovery payments

n10ty:
While unlock - query hermes to get latest promise and recover if needed

### DIFF
--- a/session/pingpong/hermes_channel_repository.go
+++ b/session/pingpong/hermes_channel_repository.go
@@ -240,7 +240,7 @@ func (hcr *HermesChannelRepository) Subscribe(bus eventbus.Subscriber) error {
 	if err != nil {
 		return fmt.Errorf("could not subscribe to AppTopicHermesPromise event: %w", err)
 	}
-	err = bus.Subscribe(identity.AppTopicIdentityUnlock, hcr.handleIdentityUnlock)
+	err = bus.SubscribeAsync(identity.AppTopicIdentityUnlock, hcr.handleIdentityUnlock)
 	if err != nil {
 		return fmt.Errorf("could not subscribe to AppTopicIdentityUnlock event: %w", err)
 	}
--- a/session/pingpong/hermes_channel_repository_test.go
+++ b/session/pingpong/hermes_channel_repository_test.go
@@ -153,7 +153,7 @@ func TestHermesChannelRepository_Fetch_publishesEarningChanges(t *testing.T) {
 	mockBeneficiaryProvider := &mockBeneficiaryProvider{}
 	mockHermesCaller := &mockHermesCaller{}
 	addrProv := &mockAddressProvider{}
-	repo := NewHermesChannelRepository(promiseProvider, channelStatusProvider, mocks.NewEventBus(), mockBeneficiaryProvider, mockHermesCaller, addrProv, signerFactory, &mockEncryptor{})
+	repo := NewHermesChannelRepository(promiseProvider, channelStatusProvider, publisher, mockBeneficiaryProvider, mockHermesCaller, addrProv, signerFactory, &mockEncryptor{})
 
 	// when
 	promiseProvider.toReturn = expectedPromise1


session/pingpong/hermes_caller.go
```diff
@@ -288,16 +320,7 @@ func (ac *HermesCaller) GetConsumerData(chainID int64, id string) (HermesUserInf
...
-	if err != nil {
-		return HermesUserInfo{}, err
-	}
-	err = data.LatestPromise.isValid(id)
-	if err != nil {
-		return HermesUserInfo{}, fmt.Errorf("could not check promise validity: %w", err)
-	}
-
-	return data, nil
+	return ac.getProviderData(chainID, id)
```
n10ty, line 323: This is wrong condition, because promise for provider signed by Hermes (operator)

session/pingpong/hermes_channel_repository.go
```diff
@@ -356,3 +446,12 @@ func (hcr *HermesChannelRepository) updateChannel(chainID int64, new HermesChann
 		Current:  *earningsNew,
 	})
 }
+
+func generateR() []byte {
```
Guillembonet, line 450: maybe this we should make it part of payments repo

Guillembonet (approved):
Looks good I think. Some things to consider:
- Will this call hermes on each identity unlock, or is the `hcr.Fetch` call cached?
- If you tested this and it is able to settle a promise after deleting db it should be fine

session/pingpong/consumer_balance_tracker.go
```diff
@@ -242,11 +242,13 @@ func (cbt *ConsumerBalanceTracker) handleUnlockEvent(data identity.AppEventIdent
 	err := cbt.recoverGrandTotalPromised(data.ChainID, data.ID)
 	if err != nil {
 		log.Error().Err(err).Msg("Could not recover Grand Total Promised")
+		return
```
tomasmik, line 245: Even if these result in an error we would want to continue and start `go cbt.lifetimeBCSync(data.ChainID, data.ID)`

Registration status is also handled correctly with the switch. These errors are more for log purpose instead of detecting an actual error.

session/pingpong/hermes_caller.go
```diff
@@ -253,14 +253,46 @@ func (ac *HermesCaller) SyncProviderPromise(promise crypto.Promise, signer ident
...
+		RRecoveryData: common.Bytes2Hex(recoveryData),
+	}
+
+	req, err := requests.NewSignedPostRequest(ac.hermesBaseURI, "refresh_promise", toSend, signer)
+	if err != nil {
+		return res, fmt.Errorf("could not make promise sync request: %w", err)
+	}
+
+	err = ac.doRequest(req, &res)
+	if err != nil {
```
tomasmik, line 287: This doesn't seem properly handled

session/pingpong/hermes_channel_repository.go
```diff
@@ -252,6 +274,74 @@ func (hcr *HermesChannelRepository) handleNodeStart(payload nodeEvent.Payload) {
...
+		return
+	}
+	hermesChannel, err := hcr.Fetch(payload.ChainID, payload.ID, hermes)
+	if err != nil {
+		log.Err(err).Msg("failed to get hermes channel info")
+		return
+	}
+
+	unsettledBalance := hermesChannel.UnsettledBalance()
+	if unsettledBalance.Cmp(big.NewInt(0)) == 0 {
```
tomasmik, line 290: minor things but you could invert this if to `!= 0` and put a `return` in it. That way you wouldn't have to nest this big workflow inside of an `if`.

session/pingpong/hermes_channel_repository.go
```diff
@@ -252,6 +274,74 @@ func (hcr *HermesChannelRepository) handleNodeStart(payload nodeEvent.Payload) {
...
+		data, err := hcr.hermesCaller.GetProviderData(payload.ChainID, payload.ID.Address)
+
+		if err != nil {
```
tomasmik, line 293: probably not even worth to continue after this

session/pingpong/hermes_channel_repository.go
```diff
@@ -252,6 +274,74 @@ func (hcr *HermesChannelRepository) handleNodeStart(payload nodeEvent.Payload) {
 	hcr.fetchKnownChannels(config.GetInt64(config.FlagChainID))
 }
 
+func (hcr *HermesChannelRepository) handleIdentityUnlock(payload identity.AppEventIdentityUnlock) {
+	hermes, err := hcr.addressProvider.GetActiveHermes(payload.ChainID)
+	if err != nil {
+		log.Err(err).Msg("failed to get active Hermes")
+		return
+	}
+	hermesChannel, err := hcr.Fetch(payload.ChainID, payload.ID, hermes)
```
tomasmik, line 283: Fetch also calls BC and stuff, we only really care about local DB, so this is bit too much

session/pingpong/hermes_channel_repository.go
```diff
@@ -252,6 +274,74 @@ func (hcr *HermesChannelRepository) handleNodeStart(payload nodeEvent.Payload) {
...
+			R := generateR()
+			hashlock := ethcrypto.Keccak256(R)
+			details := rRecoveryDetails{
+				R:           hex.EncodeToString(R),
+				AgreementID: big.NewInt(0),
+			}
+
+			bytes, err := json.Marshal(details)
+			if err != nil {
+				log.Err(err).Msgf("could not marshal R recovery details: %s", err)
```
tomasmik, line 307: You're already printing the error in `Err(err)`, no reason to also include it in the message. Same for other places.

tomasmik (changes requested):

session/pingpong/consumer_balance_tracker.go
n10ty: Reverted

Issue #5262, opened 2022-06-30
Introduce new services for node pricing


Donatas-MN:
To better support B2B clients we need to introduce additional pricing services. Currently there is only one service called Wireguard available.

Tasks:
- [x] Keep existing Wireguard service as public service (accessed by all consumers). Name can be kept as well to retain compatibility.
- [x] Add "B2B VPN and Data transfer" service which could be accessed only by whitelisted consumers
- [x] Add "B2B Data scraping" service which could be accessed only by whitelisted consumers

mdomasevicius:
It would be good to have an endpoint in tequilapi that would list all of these servises plus any additional information related to them

Donatas-MN:
> It would be good to have an endpoint in tequilapi that would list all of these servises plus any additional information related to them

@yeszhanov95 can you add tequila api endpoint for this?

PR #5321, opened 2022-07-21
Introduce new services

yeszhanov95:
Added new service types, namely: scraping and data_transfer. These new services will be enabled later, when we prepare all components.
New services could be accessed only by whitelisted consumers.
As for now, all the three services have the same pricing, later I will set up for each individually.

According to https://github.com/mysteriumnetwork/node/issues/5262

### DIFF
--- a/cmd/commands/service/command.go
+++ b/cmd/commands/service/command.go
@@ -35,6 +35,8 @@ import (
 	"github.com/mysteriumnetwork/node/config/urfavecli/clicontext"
 	"github.com/mysteriumnetwork/node/core/node"
 	"github.com/mysteriumnetwork/node/services"
+	"github.com/mysteriumnetwork/node/services/data_transfer"
+	"github.com/mysteriumnetwork/node/services/scraping"
 	"github.com/mysteriumnetwork/node/services/wireguard"
 	"github.com/mysteriumnetwork/node/tequilapi/client"
 	"github.com/mysteriumnetwork/node/tequilapi/contract"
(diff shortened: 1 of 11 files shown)


n10ty (approved):

cmd/commands/service/command.go
```diff
@@ -111,9 +113,8 @@ type serviceCommand struct {
 // Run runs a command
 func (sc *serviceCommand) Run(ctx *cli.Context) (err error) {
 	arg := ctx.Args().Get(0)
-	// If no service type specified we are starting wireguard only.
-	// Other services could be started only explicitly.
-	serviceTypes := []string{wireguard.ServiceType}
+	// If no service type specified we are starting wireguard, scraping and data_transfer.
+	serviceTypes := []string{wireguard.ServiceType, scraping.ServiceType, data_transfer.ServiceType}
```
soffokl, line 117: Let's not start it automatically now. We need to prepare other components for it. 
Later we will enable it for sure.

mobile/mysterium/proposals_manager.go
```diff
@@ -208,9 +210,15 @@ func (m *proposalsManager) getFromRepository(req *GetProposalsRequest) ([]propos
 
 	// Ideally api should allow to pass multiple service types to skip noop
 	// proposals, but for now just filter in memory.
+	serviceTypesDict := map[string]bool{
```
soffokl, line 213: Just `serviceTypes` without `Dict` will be the same clear but shorter. We already know that this is a map, no need to put it into the variable name.

soffokl (approved):

Issue #5338, opened 2022-07-26
Raspberry Pi 4 - GNU/Linux Debian 11 - identities beneficiary-status returns "panic: runtime error: index out of range [0] with length 0"


blackmennewstyle:
Hello,

I installed mysteriumnetwork **Branch: 1.9.2. Build id: 570175622. Commit: a5cea4f0** on my **Raspberry Pi 4 8GB** running **GNU/Linux Debian 11** with kernel **5.10.0-16-arm64**.
I get the following error when i run the following command `identities beneficiary-status 0x************************`
```
identities beneficiary-status 0x************************
[INFO] Current beneficiary: 0x*************************
[INFO] Is a channel address: false
[ERROR] panic: runtime error: index out of range [0] with length 0

goroutine 1 [running]:
github.com/mysteriumnetwork/node/cmd/commands/cli/clio.sentenceCase(...)
	/build/myst-c3SwiB/myst-1.9.2+build570175622+focal/go/src/github.com/mysteriumnetwork/node/cmd/commands/cli/clio/clio.go:75
github.com/mysteriumnetwork/node/cmd/commands/cli/clio.Error({0x4000771988, 0x1, 0x1})
	/build/myst-c3SwiB/myst-1.9.2+build570175622+focal/go/src/github.com/mysteriumnetwork/node/cmd/commands/cli/clio/clio.go:63 +0x158
github.com/mysteriumnetwork/node/cmd/commands/cli.(*cliApp).handleActions(0x40001e0360, {0x40000300c0, 0x3, 0x4})
	/build/myst-c3SwiB/myst-1.9.2+build570175622+focal/go/src/github.com/mysteriumnetwork/node/cmd/commands/cli/command.go:276 +0x5b4
github.com/mysteriumnetwork/node/cmd/commands/cli.(*cliApp).Run(0x40001e0360, 0x4000219140)
	/build/myst-c3SwiB/myst-1.9.2+build570175622+focal/go/src/github.com/mysteriumnetwork/node/cmd/commands/cli/command.go:211 +0x3e8
github.com/mysteriumnetwork/node/cmd/commands/cli.NewCommand.func1(0x40002018c0?)
	/build/myst-c3SwiB/myst-1.9.2+build570175622+focal/go/src/github.com/mysteriumnetwork/node/cmd/commands/cli/command.go:86 +0xa0
[...]

PR #5340, opened 2022-07-27
#5338 beneficiary info

n10ty:
Print verbose error. If there are no beneficiary changes do not throw an error, but return not_found status instead.

### DIFF
--- a/cmd/commands/cli/command_identities.go
+++ b/cmd/commands/cli/command_identities.go
@@ -416,7 +416,7 @@ func (c *cliApp) getBeneficiaryStatus(actionArgs []string) error {
 	}
 
 	if st.State == beneficiary.NotFound {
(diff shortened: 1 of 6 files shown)


cmd/commands/cli/command_identities.go
```diff
@@ -415,6 +415,9 @@ func (c *cliApp) getBeneficiaryStatus(actionArgs []string) error {
 		return fmt.Errorf("could not get beneficiary change status: %w", err)
 	}
 
+	if st.State == beneficiary.NotFound {
+		return nil
```
tomasmik, line 419: Could we maybe do:
```suggestion
		clio.Info("No last change request information")
		return nil
```

or something like that

tomasmik (approved):

PR #5349, opened 2022-08-01
Node authenticated endpoint

andrey-mysterium:


core/quality/mysterium_morqa.go
```diff
@@ -291,9 +291,10 @@ func (m *MysteriumMORQA) ProviderSessions(providerID string) []ProviderSession {
 
 // ProviderStatuses fetch provider connectivity statuses from quality oracle.
 func (m *MysteriumMORQA) ProviderStatuses(providerID string) (node.MonitoringAgentStatuses, error) {
-	request, err := m.newRequestJSON(http.MethodGet, fmt.Sprintf("providers/statuses?provider_id=%s", providerID), "")
+	id := identity.Identity{Address: providerID}
+
+	request, err := requests.NewSignedGetRequest(m.baseURL, fmt.Sprintf("providers/auth-statuses?provider_id=%s", providerID), m.signer(id))
```
tomasmik, line 296: Just something interesting to mention, might be useful for the future: you should not be able to upgrade the API so that you would no longer need to send the `provider_id` field in URL as you can extract the signer from signature.

andrey-mysterium: Yes, you are right. In side effect of auth we can get it inside signature in context

core/quality/mysterium_morqa.go
```diff
@@ -291,9 +291,10 @@ func (m *MysteriumMORQA) ProviderSessions(providerID string) []ProviderSession {
 
 // ProviderStatuses fetch provider connectivity statuses from quality oracle.
 func (m *MysteriumMORQA) ProviderStatuses(providerID string) (node.MonitoringAgentStatuses, error) {
-	request, err := m.newRequestJSON(http.MethodGet, fmt.Sprintf("providers/statuses?provider_id=%s", providerID), "")
+	id := identity.Identity{Address: providerID}
```
tomasmik, line 294: We got a function for this: `identity.FromAddress(providerID)`

tomasmik (approved):

tomasmik:
Also another thing to mention. This PR has 3 commits, 2 of the are useless and 3rd is editing them. Please fix the commits properly before creating a PR. Commits should help to review, they aren't meant to track your development progress. This could have easily been a single commit.

core/quality/mysterium_morqa.go
soffokl: yep, let's just use the identity from the signature, it should not be possible to pass anything except it.

Issue #5322, opened 2022-07-21
Add support for different pricing per each service type.


soffokl:
Currently, prices are different only for node types. 

We need to extend it to support different prices for different service types. 

Prices:
**Wireguard**
```
"base_prices": {
        "residential": {
            "price_per_hour_usd": 0.00025,
            "price_per_gib_usd": 0.04
        },
        "other": {
            "price_per_hour_usd": 0.0001,
            "price_per_gib_usd": 0.0225
        }
    },
```

**B2B VPN and Data transfer**
```
"base_prices": {
        "residential": {
            "price_per_hour_usd": 0.00025,
            "price_per_gib_usd": 0.02
        },
        "other": {
            "price_per_hour_usd": 0.0001,
            "price_per_gib_usd": 0.01
        }
    },
```

**B2B Data scraping**
```
"base_prices": {
        "residential": {
            "price_per_hour_usd": 0.00025,
            "price_per_gib_usd": 0.2
        },
        "other": {
            "price_per_hour_usd": 0.0001,
            "price_per_gib_usd": 0.1
        }
    },
```

PR #5354, opened 2022-08-04
add support for different pricing

yeszhanov95:
Added support for different pricing. Migrated to Discovery V4 API (https://discovery.mysterium.network/api/v4).

According to https://github.com/mysteriumnetwork/node/issues/5322


soffokl (approved):
👍

PR #5366, opened 2022-08-09
add sessions list API endpoint

andrey-mysterium:


core/node/node_monitoring_agent_tracker.go
```diff
@@ -53,3 +59,26 @@ func (m *MonitoringAgentTracker) Statuses() (MonitoringAgentStatuses, error) {
...
+type SessionItem struct {
+	ID              string `json:"id"`
+	ConsumerCountry string `json:"consumer_country"`
+	Duration        int64  `json:"duration"`
+	StartedAt       int64  `json:"started_at"`
+	Earning         string `json:"earning"`
+	Transferred     int64  `json:"transferred"`
+}
+
+// MonitoringAgentSessions a object represent a sessions list of monitoring metrics.
```
soffokl, line 73: This is not related to the Monitoring Agent somehow. 
```suggestion
// ProviderSessions represents a list of provider sessions.
```

core/quality/mysterium_morqa.go
```diff
@@ -315,6 +316,32 @@ func (m *MysteriumMORQA) ProviderStatuses(providerID string) (node.MonitoringAge
...
+	id := identity.FromAddress(providerID)
+
+	request, err := requests.NewSignedGetRequest(m.baseURL, fmt.Sprintf("provider/sessions?range=%s", rangeTime), m.signer(id))
+	if err != nil {
+		return nil, err
+	}
+
+	response, err := m.client.Do(request)
+	if err != nil {
+		log.Err(err).Msg("Failed to request provider monitoring sessions list")
```
soffokl, line 330: This is not monitoring. 
```suggestion
		log.Err(err).Msg("Failed to request provider sessions list")
```

tequilapi/endpoints/node.go
```diff
@@ -87,6 +115,7 @@ func AddRoutesForNode(nodeStatusProvider nodeStatusProvider, nodeMonitoringAgent
 		{
 			nodeGroup.GET("/monitoring-status", nodeEndpoints.NodeStatus)
 			nodeGroup.GET("/monitoring-agent-statuses", nodeEndpoints.MonitoringAgentStatuses)
+			nodeGroup.GET("/monitoring-agent-sessions", nodeEndpoints.MonitoringAgentSessions)
```
soffokl, line 118: Can we make the same same structure in the node as we have in quality oracle?
```
/provider/monitoring-statues
/provider/sessions
/provider/any-new-data
```

tequilapi/docs/swagger.json
```diff
@@ -4525,6 +4562,40 @@
       },
       "x-go-package": "github.com/mysteriumnetwork/node/tequilapi/contract"
     },
+    "SessionItem": {
```
mdomasevicius, line 4565: At least our PO also had ideas about `Service Type` used during the session 

image

core/node/node_stats_tracker.go
```diff
@@ -0,0 +1,84 @@
...
+	id, ok := m.currentIdentity.GetUnlockedIdentity()
+	if ok {
+		return m.providerStatuses(id.Address)
+	}
+
+	return MonitoringAgentStatuses{}, errors.New("identity not found")
+}
+
+// SessionItem represents information about session monitoring metrics.
+type SessionItem struct {
```
mdomasevicius, line 64: Our PO also had ideas about `Service Type` of given session 

image

soffokl: yep, this is needed since we are close to having many service types.

tequilapi/endpoints/node.go
```diff
@@ -78,6 +80,32 @@ func (ne *NodeEndpoint) MonitoringAgentStatuses(c *gin.Context) {
...
 
+// GetProviderSessions A list of sessions metrics during a period of time
+// swagger:operation GET /node/provider/sessions GetProviderSessions
+// ---
+// summary: Provides Node sessions data during a period of time
+// description: Node sessions metrics during a period of time
+// parameters:
+//   - in: query
+//     name: range
+//     description: period of time ("1d", "7d", "30d")
```
mdomasevicius, line 91: Is it possible to have range as `[dateA, dateB]` ?

mdomasevicius: [Screencast from 08-11-2022 08:12:48 AM.webm](https://user-images.githubusercontent.com/20511486/184068432-72322a39-152f-4899-94a2-88ebda7bd1cd.webm)

soffokl: It's possible, but we don't want to give too much freedom to users since it might give us too big load.

mdomasevicius: what if you would limit this range to say max 7 day period? or something like that

soffokl: @mdomasevicius as we talked, free 7-day period might be abused as well. 
Let's stay with static limits for the last period of 1, 7, and 30 days now, and maybe later we will find a better way.

core/quality/mysterium_morqa.go
```diff
@@ -315,6 +316,32 @@ func (m *MysteriumMORQA) ProviderStatuses(providerID string) (node.MonitoringAge
...
```
tomasmik, line 330: Logging and then returning an error always makes it confusing on what you should do with it and might result in double triple logs of the same error

tequilapi/endpoints/node.go
```diff
@@ -78,6 +80,32 @@ func (ne *NodeEndpoint) MonitoringAgentStatuses(c *gin.Context) {
...
+// responses:
+//   200:
+//     description: Provider sessions list
+//     schema:
+//       "$ref": "#/definitions/ProviderSessionsResponse"
+func (ne *NodeEndpoint) GetProviderSessions(c *gin.Context) {
+	rangeTime := c.Param("range")
+	res, err := ne.nodeMonitoringAgent.Sessions(rangeTime)
+	if err != nil {
+		utils.WriteAsJSON(contract.ProviderSessionsResponse{Error: err.Error()}, c.Writer, http.StatusInternalServerError)
```
tomasmik, line 102: We're trying to force a switch to `"github.com/mysteriumnetwork/go-rest/apierror"` and this seems to be doing a complete opposite of that.

tequilapi/endpoints/node.go
```diff
@@ -78,6 +80,32 @@ func (ne *NodeEndpoint) MonitoringAgentStatuses(c *gin.Context) {
...
+// ---
+// summary: Provides Node sessions data during a period of time
+// description: Node sessions metrics during a period of time
+// parameters:
+//   - in: query
+//     name: range
+//     description: period of time ("1d", "7d", "30d")
+//     type: string
+// responses:
+//   200:
```
tomasmik, line 94: A 500 response is possible but is not described

core/node/node_stats_tracker.go
```diff
@@ -0,0 +1,84 @@
...
+	ID              string `json:"id"`
+	ConsumerCountry string `json:"consumer_country"`
+	Duration        int64  `json:"duration"`
+	StartedAt       int64  `json:"started_at"`
+	Earning         string `json:"earning"`
+	Transferred     int64  `json:"transferred"`
+}
+
+// SessionsList a object represent a sessions list.
+type SessionsList []SessionItem
```
tomasmik, line 74: This might just be me but I feel like these custom types do not add much clarity, I would even argue that they add additional complexity. There is nothing wrong with `[]SessionItem`, I can look at it and see that it's a slice of something, but when it's typed as `SessionsList` I get extra steps to do to figure that it's literally just a slice 🤷🏻

core/node/node_stats_tracker.go
```diff
@@ -0,0 +1,84 @@
...
+
+// Sessions retrieves and resolved monitoring status from quality oracle
+func (m *StatsTracker) Sessions(rangeTime string) (SessionsList, error) {
+	id, ok := m.currentIdentity.GetUnlockedIdentity()
+	if ok {
+		return m.providerSessionsList(id.Address, rangeTime)
```
tomasmik, line 80: So you do `id.Address` to a pass string, which you later instantly convert back to identity.Identity `id := identity.FromAddress(providerID)`. Could you just pass `id` and not convert anything later? seems a bit simpler.

tequilapi/endpoints/node.go
```diff
@@ -78,6 +80,32 @@ func (ne *NodeEndpoint) MonitoringAgentStatuses(c *gin.Context) {
...
+//     description: Provider sessions list
+//     schema:
+//       "$ref": "#/definitions/ProviderSessionsResponse"
+func (ne *NodeEndpoint) GetProviderSessions(c *gin.Context) {
+	rangeTime := c.Param("range")
```
tomasmik, line 99: This seem to not be validated anywhere and is described as being `period of time ("1d", "7d", "30d")`. Even if it's checked on the service you're calling, it will then produce a `http.StatusInternalServerError` which is not right, it should be a Bad Request error.

tequilapi/endpoints/node.go
mdomasevicius: discussed with Dmitry and our PO - let's keep this 1, 7, 30 day period for now

mdomasevicius (approved):

tequilapi/endpoints/node.go
```diff
@@ -78,6 +81,48 @@ func (ne *NodeEndpoint) MonitoringAgentStatuses(c *gin.Context) {
...
+//       "$ref": "#/definitions/APIError"
+//   500:
+//     description: Internal server error
+//     schema:
+//       "$ref": "#/definitions/APIError"
+func (ne *NodeEndpoint) GetProviderSessions(c *gin.Context) {
+	rangeTime := c.Param("range")
+
+	switch rangeTime {
+	case "1d", "7d", "30d":
```
tomasmik, line 111: Would be nice to have such a thing as a constant somewhere, but this is ok I guess.

tomasmik (approved):

tomasmik (approved):

PR #5367, opened 2022-08-10
provide new endpoints regarding new service types

yeszhanov95:
Update endpoint /services: add optional query param "include_all" (boolean). when include_all=true return all services, otherwise only running.
Add endpoint /v2/prices/current, which will return current prices for all service types.

According to https://github.com/mysteriumnetwork/node/issues/5262


core/service/manager.go
```diff
@@ -230,6 +233,35 @@ func (manager *Manager) List() map[ID]*Instance {
 	return manager.servicePool.List()
 }
 
+// ListAll returns array of all services
+func (manager *Manager) ListAll() []*Instance {
```
tomasmik, line 237: There are many reasons to use non pointer slices. Less memory, fewer allocations needed and it's also faster. Pointer slices also serve little practical use, especially in this case here. It's almost always more trouble than its worth.

I see it's been done here like that before, am I just missing something?

tequilapi/endpoints/service.go
```diff
@@ -70,7 +72,27 @@ func NewServiceEndpoint(serviceManager ServiceManager, optionsParser map[string]
 //     schema:
 //       "$ref": "#/definitions/APIError"
 func (se *ServiceEndpoint) ServiceList(c *gin.Context) {
-	instances := se.serviceManager.List()
+	includeAll := false
+	includeAllStr := c.Request.URL.Query().Get("include_all")
```
tomasmik, line 76: There is `c.Get` or `c.GetBool` wouldn't those do the same but better?

tequilapi/endpoints/service.go
```diff
@@ -70,7 +72,27 @@ func NewServiceEndpoint(serviceManager ServiceManager, optionsParser map[string]
...
+		var err error
+		includeAll, err = strconv.ParseBool(includeAllStr)
+		if err != nil {
+			c.Error(apierror.BadRequestField(fmt.Sprintf("Failed to parse request: %s", err.Error()), "include_all", contract.ErrCodeServiceList))
+			return
+		}
+	}
+
+	instances := make([]*service.Instance, 0)
+	if includeAll {
```
tomasmik, line 87: Could these be a method on the `serviceManager` itself? Would be nice to not have this `if / else` logic in the handler and the manager itself could decide if it should return all or some portion based on given params.

core/service/manager.go
yeszhanov95: Actually yes, there's similar method called List, which also uses pointer instead of struct. So, I decided to go this way

tequilapi/endpoints/service.go
yeszhanov95: as I understand those methods doesn't fetch data from query params. it looks inside `c.Keys` map[string]interface{}

tequilapi/endpoints/service.go
yeszhanov95: good point, I thought about it too

tequilapi/endpoints/service.go
n10ty: `c.Query("include_all")` ?

tomasmik (approved):

PR #5368, opened 2022-08-11
Better stability for netsack with proxy connections

soffokl:


Snawoot:
it's OK to supress linting for netstack.go as for imported "as is" file

tomasmik (approved):

PR #5396, opened 2022-08-23
Fix/pricing panic

yeszhanov95:
Some of our users has local configs, and it breaks migration to new version of Discovery API.

The first 8 commits was added by command `git cherry-pick <original commit sha>` (we had to revert them before).
New changes only in the last commit

### DIFF
--- a/config/flags_network.go
+++ b/config/flags_network.go
@@ -37,7 +37,7 @@ var (
 	// Deprecated: use FlagDiscoveryAddress
 	FlagAPIAddress = cli.StringFlag{
 		Name:  metadata.FlagNames.MysteriumAPIAddress,
-		Usage: "URL of Mysterium API",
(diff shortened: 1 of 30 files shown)


config/flags_network.go
```diff
@@ -34,11 +34,18 @@ var (
 		Value: string(Mainnet),
 	}
 	// FlagAPIAddress Mysterium API URL
+	// Deprecated: use FlagDiscoveryAddress
 	FlagAPIAddress = cli.StringFlag{
 		Name:  metadata.FlagNames.MysteriumAPIAddress,
 		Usage: "URL of Mysterium API",
```
soffokl, line 40: Add `Deprecated: use FlagDiscoveryAddress` into this usage too, so users will see it once it's compiled.

soffokl (approved):

PR #5425, opened 2022-09-13
Move stats tracking into connection manager

soffokl:
This improves performance for the high-loaded setups


tequilapi/endpoints/connection.go
```diff
@@ -263,6 +263,27 @@ func (ce *ConnectionEndpoint) GetStatistics(c *gin.Context) {
...
+// ---
+// summary: Returns connection traffic information
+// description: Returns traffic information about requested connection
+// responses:
+//   200:
+//     description: Connection traffic
+//     schema:
+//       "$ref": "#/definitions/ConnectionTrafficDTO"
+func (ce *ConnectionEndpoint) GetTraffic(c *gin.Context) {
+	n, _ := strconv.Atoi(c.Query("id"))
```
yeszhanov95, line 277: why not handle error here?

soffokl: Added a propper check there.

PR #5426, opened 2022-09-13
Fix a probable gvisor panic on peer disconnect

Zensey:


Snawoot (approved):

Snawoot:
I think you could write it like this instead:

```go

		wg.Add(2)
		go func() {
			defer wg.Done()
			tun.proxy(client, clientAddr, proxyConn) // loc <- remote
		}()
		go func() {
			defer wg.Done()
			tun.proxy(&wg, proxyConn, remoteAddr, client) // remote <- loc
		}
		wg.Wait()
```

to get rid of that useless wg parameter which is actually belong to outer function mechanics

Snawoot (approved):

PR #5456, opened 2022-09-21
allow increasing total promised in a single lock

Guillembonet:
We got the value, added, and then saved it. This could potentially lead to a broken state (though it's rare). Now we will allow increasing the value in a single lock to avoid this.


session/pingpong/consumer_totals_storage.go
```diff
@@ -85,6 +87,47 @@ func (cts *ConsumerTotalsStorage) Get(chainID int64, id identity.Identity, herme
...
+	if err != nil {
+		// wrap the error to an error we can check for
+		if errors.Is(storm.ErrNotFound, err) {
+			log.Debug().Msg("No previous invoice grand total, assuming zero")
+			value = big.NewInt(0)
+		} else {
+			return fmt.Errorf("could not get total promised to add : %w", err)
+		}
+	}
+	if value == nil || value.Cmp(big.NewInt(0)) == -1 {
```
tomasmik, line 107: Can it even be `nil`? https://go.dev/play/p/I5VFOAy5_kD

Guillembonet: fun fact: [yes](https://go.dev/play/p/_PH7DtQ2jed)

Snawoot:
@n10ty take a look as well, please

Snawoot (approved):

n10ty (changes requested):
This changes is useless until someone using same ID for the same node instance, which we are not supporting now.

PR #5506, opened 2022-10-04
store grand total in memory instead of boltdb

Guillembonet:


n10ty:
What about Update?
```
lock
get
store
unlock
```

PR #5521, opened 2022-10-05
add in single lock for total storage

Guillembonet:
Related to: https://github.com/mysteriumnetwork/node/pull/5456
I still think it makes sense to do the adding in a single lock just to be safe.


Snawoot:
Not sure if it is needed. Underlying database should provide proper transaction serialization on its own.

Snawoot:
@n10ty WDYT?

Guillembonet:
> Not sure if it is needed. Underlying database should provide proper transaction serialization on its own.

now underlying database is an in-memory map, also @n10ty was also suggesting to add this after moving to memory in the comments of [this PR](https://github.com/mysteriumnetwork/node/pull/5506)

Guillembonet:
> My main concern about this PR is following: if DB lock is being held and we're doing publish to internal bus and some handler on the internal bus tries to acquire same DB lock to do some update, it will cause deadlock situation. Maybe we can publish to internal bus without holding lock?

the publishing is done in a goroutine so it shouldn't lock I think

Snawoot (approved):
Okay then

n10ty:
1. We have Store and publish AppTopicGrandTotalChanged event now are wrapped with lock, so it wouldn't break now anything but it doesn't mean it will not break in a future. So I agree with Vlad it's better to publish outside a lock.
2. Number of "promise value too low" errors was dramatically dropped after moving from boltdb to storing into memory. I would rather understand why it still continue. Maybe it's because of some not updated nodes (like monitoring agent or something), while majority have this fix applied.

n10ty:
Ok, I see it publish in separate goroutine. Point 2 still would be nice to understand.

Guillembonet:
> 1. We have Store and publish AppTopicGrandTotalChanged event now are wrapped with lock, so it wouldn't break now anything but it doesn't mean it will not break in a future. So I agree with Vlad it's better to publish outside a lock.
> 2. Number of "promise value too low" errors was dramatically dropped after moving from boltdb to storing into memory. I would rather understand why it still continue. Maybe it's because of some not updated nodes (like monitoring agent or something), while majority have this fix applied.

Let's discuss sometime soon, its not urgent @n10ty

n10ty (approved):

Issue #5551, opened 2022-10-19
Node dropping connection and not updating stats, then freezing


tadaskay:
**Describe the bug**
When using node 1.19.0, consumer keeps disconnecting/connecting to provider. At the same time, session stats are not updating properly (every second) via /events/state endpoint. After a few connects and disconnects, without any user interaction, the node stops responding, cannot even disconnect.

**To Reproduce**
1. Run app with node 1.19.0
2. Connect to any provider

Expected: connects after a few seconds (<10)
Actual: connected state is updated only after 20s

Expected: connection stats (see duration timer) are updated every second
Actual: connection stats are not updated

Expected: connection is active
Actual: connection drops soon after connecting and the node keeps reconnecting (multiple times), then freezes and does not respond to disconnect call

**Screenshots**

![Oct-19-2022 10-02-46](https://user-images.githubusercontent.com/4745368/196619867-42fc9655-9214-4784-96ee-690b9f48e1a6.gif)

**Environment (please complete the following information):**
 - Node version: 1.19.0
 - OS: macOS
 - Desktop app version (if applicable): master
 - Your identity (if applicable): 0x1a16f8bff99c3d3cf743c803c10a51e75892d64c

**Additional context**

[mysterium-node.log](https://github.com/mysteriumnetwork/node/files/9817710/mysterium-node.log)

tadaskay:
With node version 1.19.4, state does not change at all, not even `Connecting` when the connection is initiated. After a few moments, an error is thrown: 
```
2022-10-31T14:34:01.857 INF ../../../../source/core/connection/manager.go:397    > Cancelling connection initiation:  error="could not start new connection: failed to start connection endpoint: could not configure device: failed to create wg interface: could not setup network: could not set DNS: exit status 4"
```

tadaskay:
Also, please link PRs when fixing issues. It helps much.

tadaskay:
On windows, connection appears to work. But state endpoints skips 'Connecting' status. When connection is initiated, it stays in 'NotConnected' status and then after 10s changes into 'Connected'. 'Disconnecting' works fine.

tadaskay:
```
 .\node_modules\@mysteriumnetwork\node\bin\win\x64\myst.exe --usermode daemon
 .\node_modules\@mysteriumnetwork\node\bin\win\x64\myst.exe cli
```

andrey-mysterium:
Fix "Connecting" state update

[...]

PR #5605, opened 2022-11-09
use Session.State instead Session.SessionID as attribute of worked co…

andrey-mysterium:
…nnection


Snawoot:
@andrey-mysterium What is the purpose of this change? What does it fixes or why it is needed? If there is some related bug, please mention if as "Closes #XXXX"

andrey-mysterium:
> @andrey-mysterium What is the purpose of this change? What does it fixes or why it is needed? If there is some related bug, please mention if as "Closes #XXXX"

This fix relates to [https://github.com/mysteriumnetwork/node/issues/5551](https://github.com/mysteriumnetwork/node/issues/5551)

Initially we use SessionID as  attribute to get connection state.  But during "Connecting" state not any SessionID assigned, so this state skipping during a connections states queue.  

My fix add "Connecting" to states queue in /events/state API call.

Snawoot (approved):

PR #5643, opened 2022-11-24
fix promise refresh

Guillembonet:
Only refresh if R is not revealed and we don't know it
Signed-off-by: Guillem Bonet <guillem.bonet@gmail.com>


session/pingpong/hermes_channel_repository.go
```diff
@@ -343,7 +356,32 @@ func (hcr *HermesChannelRepository) handleIdentityUnlock(payload identity.AppEve
 			HermesID:   hermes,
 			ProviderID: identity.FromAddress(data.Identity),
 		})
+
+		err = hcr.revealR(hermesPromise)
+		if err != nil {
+			log.Err(err).Msgf("failed to reveal R after promise refresh")
+		}
+		log.Debug().Bool("saved", err == nil).Msg("refreshed promise")
```
soffokl, line 364: Is it part of debugging or we going to keep it in the release?

soffokl (approved):

PR #5790, opened 2023-05-02
SSO Mystnodes

mdomasevicius:


tequilapi/endpoints/auth.go
```diff
@@ -142,13 +149,81 @@ func (api *authenticationAPI) Login(c *gin.Context) {
...
+// description: SSO init endpoint to auth via mystnodes
+//
+// responses:
+//
+//	200:
+//	  description: link response
+//	  schema:
+//	    "$ref": "#/definitions/MystnodesSSOLinkResponse"
+func (api *authenticationAPI) LoginMystnodesInit(c *gin.Context) {
+	link, err := api.ssoMystnodes.SSOLink(c.Request.Header.Get("X-Forwarded-Host"))
```
mdomasevicius, line 164: I am not sure about this.

I take this header which is set by tequilapi reverse proxy to be used for **redirectUrl** generation

tequilapi/sso/mystnodes.go
```diff
@@ -0,0 +1,154 @@
...
+		baseUrl: config.GetString(config.FlagMMNAddress),
+		ssoPath: "/login-sso",
+		signer:  signer,
+		client:  client,
+	}
+}
+
+// Subscribe unlocked identity is required in order to sign request
+func (m *Mystnodes) Subscribe(eventBus eventbus.EventBus) error {
+	if err := eventBus.SubscribeAsync(identity.AppTopicIdentityUnlock, m.onIdentityUnlocked); err != nil {
```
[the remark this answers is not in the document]
Guillembonet: I'm also not sure about this, there is other tequilapi calls to get the current identity, maybe we can just use the underlying methods of those to get this? but maybe this is also okay

tequilapi/sso/mystnodes.go
```diff
@@ -0,0 +1,181 @@
...
+
+	u = u.JoinPath(m.ssoPath)
+
+	info, err := pkce.New(128)
+	if err != nil {
+		return nil, err
+	}
+
+	m.lastCodeVerifierBase64url = info.Base64URLCodeVerifier()
+	messageJson, err := m.message(info, strings.Join([]string{"http://", host, "/#/auth-sso"}, "")).json()
```
Guillembonet, line 127: this would also be avoided by getting redirect url from the parameter as it would already contain the http or https prefix

Guillembonet (approved):

shawaj:
Not sure if this is possible already, but could the:

```
"redirectUrl": "http://localhost:4449/#/auth-sso"
```

Instead use the requesting URL? For example if it's on `192.168.1.64:4449` it could use that instead of localhost (or an alternative hostname perhaps?)

mdomasevicius:
@shawaj it is possible by design to pass redirect URL.

Please note that this implementation depends on mystnodes.com and SSO on mystnodes.com side is currently **disabled** because it is still in development.

PR #5801, opened 2023-05-11
Replace whitelist migration for service command

mdomasevicius:
First time start will enable all services.
This will also make services complete sticky. For example - if all disabled, none will be started after restart.

### DIFF
--- a/cmd/commands/service/command.go
+++ b/cmd/commands/service/command.go
@@ -35,9 +35,6 @@ import (
 	"github.com/mysteriumnetwork/node/config/urfavecli/clicontext"
 	"github.com/mysteriumnetwork/node/core/node"
 	"github.com/mysteriumnetwork/node/services"
-	"github.com/mysteriumnetwork/node/services/datatransfer"
-	"github.com/mysteriumnetwork/node/services/scraping"
-	"github.com/mysteriumnetwork/node/services/wireguard"
 	"github.com/mysteriumnetwork/node/tequilapi/client"
 	"github.com/mysteriumnetwork/node/tequilapi/contract"
 )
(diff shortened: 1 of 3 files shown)


config/config_test.go
```diff
@@ -232,6 +235,11 @@ func TestUserConfig_GetConfig(t *testing.T) {
 	)
 }
 
+func TestHardcodedServicesNameFlagValues(t *testing.T) {
+	// importing these constants into config package create cyclic dependency
+	assert.Equal(t, strings.Join([]string{wireguard.ServiceType, datatransfer.ServiceType}, ","), FlagActiveServices.Value)
```
yeszhanov95: shouldn't we add `scraping.ServiceType` here as well?

cmd/commands/service/command.go
```diff
@@ -114,16 +111,9 @@ type serviceCommand struct {
...
-	activeServices := config.Current.GetString(config.FlagActiveServices.Name)
-	if len(activeServices) > 0 {
-		serviceTypes = strings.Split(activeServices, ",")
-	} else { // start migration
-		if len(config.Current.GetString(config.FlagWireguardAccessPolicies.Name)) > 0 || len(config.Current.GetString(config.FlagAccessPolicyList.Name)) > 0 {
-			serviceTypes = []string{scraping.ServiceType, datatransfer.ServiceType}
-		} else {
-			serviceTypes = []string{wireguard.ServiceType, scraping.ServiceType, datatransfer.ServiceType}
-		}
+	fromConfig := config.Current.GetString(config.FlagActiveServices.Name)
```
yeszhanov95, line 114: i suggest to keep the name `activeServices`. `fromConfig` could be a bit confusing, cause value will be fetched from different sources, not only from config file

yeszhanov95 (approved):

n10ty (approved):

Issue #5793, opened 2023-05-04
Replace policy download with specific consumer request in background for trust oracle validation


soffokl:
## Description

We need to replace the download of the entire policy with a request about a specific consumer, performed in the background while the consumer is dialing in. Once the provider receives a consumer identity, it should start the validation process via the trust oracle in the background to not affect the connection time. The provider should wait for the validation before establishing the final connection.

### Requirements

- The provider should replace the download of the entire policy with a request about a specific consumer.
- The request should be performed in the background while the consumer is dialing in.
- Once the provider receives a consumer identity, it should start the validation process via the new trust oracle endpoint in the background.
- The validation process should not affect the total connection time.
- The provider should wait for the validation before establishing the final connection.

PR #5804, opened 2023-05-16
Replace policy fetching with specific identity check request

ScruffyPants:
Related issue: https://github.com/mysteriumnetwork/node/issues/5793

Trust oracle supports fetching trust policies with identity-value filtering through URL query parameters. This allows node to replace cron like fetching of all trust policies and storing them locally logic with a simple HTTP request. This change should reduce the memory footprint of the application. 
However, node's whitelist logic has support for types of whitelist entries other than identity, like "dns_zone" and "dns_hostname". To still have this support as an option (until trust oracle adds more filtering options) a new flag was added to disable continuous fetching of access policies. By default the fetching is disabled and new flow is used.

### DIFF
--- a/core/policy/provider.go
+++ b/core/policy/provider.go
@@ -1,7 +1,5 @@
-//go:build ignore
-
 /*
- * Copyright (C) 2019 The "MysteriumNetwork/node" Authors.
+ * Copyright (C) 2023 The "MysteriumNetwork/node" Authors.
  *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
@@ -17,16 +15,13 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-package main
+package policy
 
-import (
-	"os"
+import "github.com/mysteriumnetwork/node/identity"
 
-	"github.com/magefile/mage/mage"
-)
-
-// Zero install option.
-// Usage example:
-//
-//	go run mage.go test
-func main() { os.Exit(mage.Main()) }
+// Provider interface defines policy provider
+type Provider interface {
+	IsIdentityAllowed(identity identity.Identity) bool
+	HasDNSRules() bool
+	IsHostAllowed(host string) bool
+}
--- a/core/policy/requested/provider.go
+++ b/core/policy/requested/provider.go
@@ -1,0 +1,69 @@
+/*
+ * Copyright (C) 2023 The "MysteriumNetwork/node" Authors.
+ *
+ * This program is free software: you can redistribute it and/or modify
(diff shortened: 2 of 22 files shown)


core/service/session_manager.go
```diff
@@ -267,8 +280,14 @@ func (manager *SessionManager) startSession(session *Session, prices market.Pric
...
-	if !manager.service.Policies().IsIdentityAllowed(session.ConsumerID) {
-		return fmt.Errorf("consumer identity is not allowed: %s", session.ConsumerID.Address)
+	if manager.service.PolicyOracle().PeriodicFetchingEnabled() {
+		if !manager.service.Policies().IsIdentityAllowed(session.ConsumerID) {
+			return fmt.Errorf("consumer identity is not allowed: %s", session.ConsumerID.Address)
+		}
+	} else {
+		if !manager.service.PolicyOracle().IsIdentityAllowed(session.ConsumerID) {
+			return fmt.Errorf("consumer identity is not allowed: %s", session.ConsumerID.Address)
+		}
```
soffokl, line 290: Both `policies` and `policyOracle` implement the same `IsIdentityAllowed` method. 
Can we just add a single interface into the service manager to avoid all this duplicated code?

soffokl (changes requested):
e2e tests are not passing and indicate that something wrong with the code.

core/service/session_manager.go
ScruffyPants: Good suggestion :+1:

core/policy/provider.go
```diff
@@ -0,0 +1,27 @@
+/*
+ * Copyright (C) 2020 The "MysteriumNetwork/node" Authors.
```
soffokl, line 2: Once a new file is created, the date should be set to the current year. 
If you edit the file the date should not be changed. 
```suggestion
 * Copyright (C) 2023 The "MysteriumNetwork/node" Authors.
```

core/policy/requested/provider.go
```diff
@@ -0,0 +1,69 @@
+/*
+ * Copyright (C) 2020 The "MysteriumNetwork/node" Authors.
```
soffokl, line 2:
```suggestion
 * Copyright (C) 2023 The "MysteriumNetwork/node" Authors.
```

core/policy/requested/provider_test.go
```diff
@@ -0,0 +1,66 @@
+/*
+ * Copyright (C) 2020 The "MysteriumNetwork/node" Authors.
```
soffokl, line 2:
```suggestion
 * Copyright (C) 2023 The "MysteriumNetwork/node" Authors.
```

soffokl (approved):
Nice work!

PR #5811, opened 2023-05-26
Enable scraper service on the first run

Zensey:
Enable scraper service on the first run (after onboarding)


mdomasevicius (changes requested):
This looks like a hack. NodeUI is not responsible for automating node controls.

mdomasevicius (commented):
Looks weird

mdomasevicius (approved):

PR #5822, opened 2023-06-22
Save a list of stopped services for mobile provider

Zensey:


config/flags_service_start.go
```diff
@@ -69,6 +69,13 @@ var (
 		Usage: "Comma separated list of active services.",
 		Value: strings.Join([]string{"wireguard", "scraping", "data_transfer"}, ","),
 	}
+
+	// FlagStoppedServices a comma-separated list of stopped services.
+	FlagStoppedServices = cli.StringFlag{
+		Name:  "stopped-services",
+		Usage: "Comma separated list of stopped services.",
```
soffokl, line 76: Why do we need an extra flag for stopped services? All services without active services already list stopped services, right?

soffokl (approved):

config/flags_service_start.go
soffokl, line 76: Why do we need an extra flag for stopped services? All services without active services already list stopped services, right?

soffokl (changes requested):

config/flags_service_start.go
mdomasevicius: Yes, we are tracking which services to start so to me it seems like: 

### `all services` **-** `which to start` **=** `which are stopped` 

:)

Zensey: This flag is used in order to store a list of stopped services in a config file (user-config.toml).
On mobile app shutdown we store this list, so that on app start - we can start exactly the same set of services.

Zensey: @soffokl 
so, any objections ?

soffokl: yes, I still see no reason for storing extra flags for stopped services if you can get them from other flags.

Zensey: @soffokl 
not exactly. 
for instance, let's say I have A & B services active, C - is stopped (manually from UI).
1) then I stop provider, so no services are running, making active-services = "", stopped-services="A,B"
2) then I start provider again, A & B should start, C - no.

soffokl: There is a flag:
```
	// FlagActiveServices a comma-separated list of active services.
	FlagActiveServices = cli.StringFlag{
		Name:  "active-services",
		Usage: "Comma separated list of active services.",
		Value: strings.Join([]string{"wireguard", "scraping", "data_transfer"}, ","),
	}
```

If you stop the app, you do not overwrite that flag. If you start the app these services are started automatically. 
If you stop a single service from the UI it overwrites the flag to not start that service on the next restart.

Zensey: @soffokl 

> If you stop the app, you do not overwrite that flag. 
In the mobile provider use-case when we need to stop the node temporarilty (shortage of battery or traffic) we don't stop the node, we just stop all active services.
Yes, this is a workaround, the reason is - if I stop/start the whole node a few times it misbehaves.

soffokl: > if I stop/start the whole node a few times it misbehaves.
This sounds like we need to fix that problem instead of adding more flags/workarounds.

Zensey: @soffokl 
Ok. I opened the bug: https://github.com/mysteriumnetwork/node/issues/5840

config/flags_service_start.go
```diff
@@ -69,6 +69,13 @@ var (
 		Usage: "Comma separated list of active services.",
 		Value: strings.Join([]string{"wireguard", "scraping", "data_transfer", "dvpn"}, ","),
 	}
+
+	// FlagStoppedServices a comma-separated list of stopped services.
+	FlagStoppedServices = cli.StringFlag{
+		Name:  "stopped-services",
+		Usage: "Comma separated list of stopped services.",
+		Value: strings.Join([]string{}, ","),
```
soffokl, line 77: OK, I'm giving up. Mark this flag at least as hidden. 
```suggestion
		Value: strings.Join([]string{}, ","),
		Hidden: true,
```

soffokl (approved):

PR #5855, opened 2023-08-09
Revert "Use the latest version of wireguard-go (052af4a8 3/25/2023)"

Zensey:
This reverts commit 80de0274


Snawoot:
Any explaination why we would want that?

Zensey:
@Snawoot 

> https://github.com/mysteriumnetwork/node/compare/1.20.1...1.20.2
> here was introduced a bug that kills our monitoring agent/superproxy
> I see that wireguard package was updated, but looks like it’s not working as expected
> so monitoring agent creates a lot of connections, disconnects from it, and looks like it’s not releasing properly and the connection cannot be reused anymore

Snawoot (approved):

soffokl (commented):
What do these changes affect? Can we just revert it with no issues?

soffokl (approved):
I did a test, this seems to be solved the issue.

soffokl:
@Zensey please merge it.

PR #5867, opened 2023-08-29
Allow localhost-only routes to be used from host

Zensey:
Regarding the case when node is run on docker. When a request comes from 172.17.0.1 it could be sent not neccessarily from host, there are many possibilities.


Snawoot (changes requested):
Ask @mdomasevicius for details. Should work in docker with `127.0.0.1` as well.

Snawoot (approved):

PR #5919, opened 2023-11-14
set price for cellular nodes the same as for residential nodes

yeszhanov95:
Currently we have ~200 `android` nodes with node type `residential`. It's about to be changed to `cellular` node type.
Changes in this PR set price for cellular nodes the same as for residential nodes


soffokl (approved):

shawaj:
FYI the dockerhub release hasn't been pushed with the v1.29.1 tag see #5921

PR #5925, opened 2023-11-16
Add support of go1.21

Zensey:
Related changes of gvisor are here: https://github.com/mysteriumnetwork/gvisor/commits/fix/support-go1.21


Guillembonet:
few suggestions now that I can run this and provide feedback:
1. update `mockserver/mockserver` to `mockserver-5.15.0` tag. The slow start of this is causing tests to fail often both locally in my machine but I've also suffered this in the pipeline.
2. change `owlab/ipify` to `bunetz/ipify`. This is a docker image I created (maybe we can also just move this repo to mysterium's github and dockerhub) which supports mac os arm64 arch.

Also, after fixing this, does e2e test pass in your machine locally? for me it fails

Guillembonet:
> few suggestions now that I can run this and provide feedback:
> 
>     1. update `mockserver/mockserver` to `mockserver-5.15.0` tag. The slow start of this is causing tests to fail often both locally in my machine but I've also suffered this in the pipeline.
> 
>     2. change `owlab/ipify` to `bunetz/ipify`. This is a docker image I created (maybe we can also just move this repo to mysterium's github and dockerhub) which supports mac os arm64 arch.
> 
> 
> Also, after fixing this, does e2e test pass in your machine locally? for me it fails

Fixed in #5926

Guillembonet (approved):

PR #5926, opened 2023-11-17
ability to run e2e tests for arm64

Guillembonet:
here is what I needed to fix to run node e2e tests in arm64


Zensey (approved):
Looks good to me

PR #5934, opened 2023-12-11
Use mysteriumnetwork/wireguard-go with a stability fix

Zensey:
Fix to a previous fix #5912
This version fixes a problem with Mac. 
Tested in both provider and client modes for Mac, Windows, Linux platforms.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -269,6 +269,6 @@ require (
 )
 
 replace (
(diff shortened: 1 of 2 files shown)


Guillembonet (approved):
go mod tidy will remove old version from go.sum

Issue #5902, opened 2023-10-30
Disproportion between bandwith speed limiter and the actual internet speed of the client


m1hc1u:
**Describe the bug**
My ISP provides me with 450 Mb/s download speed and 50 Mb/s upload speed. We've conducted few tests via Mysterium Dark with my contact who is ~ 100km away within same country. He has 600 Mb/s download & 300 Mb/s upload speed.

We made 4 tests and I used "Limit bandwith to..." under Node UI\Settings\Traffic menu. ⁠#node-runners-general 

Test # 1 - Limit bandwith to 50 Mb/s
Test # 2 - Limit bandwith to 100 Mb/s
Test # 3 - Limit bandwith to 250 Mb/s
Test # 4 - Limit bandwith to 400 Mb/s

There were none other connections with my node during tests - only my contact.
As you can see, the difference between chosen limit and client's speeds are at least 9x lower than expected.

**To Reproduce**
1. Connect with my node
2. Use speedtest to check the bandwith

**Expected behavior**
Almost none disproportion between limited bandwith setup and contact's speed.

**Screenshots**
Test # 1 [https://cdn.discordapp.com/attachments/1096784581732093962/1167583754580148224/50mbps.png?ex=654ea81b&is=653c331b&hm=d2de60b1790d1a55b724e3f73d6cb06c35a7ceb5bc98812265f41cf899d12a4c&](https://cdn.discordapp.com/attachments/1096784581732093962/1167583754580148224/50mbps.png?ex=654ea81b&is=653c331b&hm=d2de60b1790d1a55b724e3f73d6cb06c35a7ceb5bc98812265f41cf899d12a4c&)
Test # 2 [https://cdn.discordapp.com/attachments/1096784581732093962/1167583754831790100/100mbps.png?ex=654ea81b&is=653c331b&hm=e628bb502d902d270b4309b153def0eb11ba368cc6a691cc2a97e370bef0529d&](https://cdn.discordapp.com/attachments/1096784581732093962/1167583754831790100/100mbps.png?ex=654ea81b&is=653c331b&hm=e628bb502d902d270b4309b153def0eb11ba368cc6a691cc2a97e370bef0529d&)
[...]

PR #5968, opened 2024-02-12
Convert bandwidth limits from Kbytes to Kbits

tzununbekov:
Issue #5902 is caused by the misalignment between the node config and network shaper component - node config uses Kbytes/sec as a network bandwidth unit, while shaper expects Kbits/sec. This fix simply converts values to Kbites before setting the limits.


tzununbekov:
superseded by #5969

PR #5970, opened 2024-02-13
Release IP address on session stop

Zensey:
Symptoms: 
> invalid config to consumer request with nil instead of ip address.
> like this:
> {"local_port":0,"remote_port":0,"ports":null,"provider":{"public_key":"iixI/brFnGk5JPADWVhzum2WltsMV2A8iabJKgJuWQo=","endpoint":"46.14.127.237:57757"},"consumer":{"ip_address":"\u003cnil\u003e","dns_ips":"\u003cnil\u003e"}}

Cause:
IP is allocated for consumer on session start, but never gets released. Given the limited size of IP pool, exhaustion can happen in a few days or hours.


services/wireguard/service/service.go
```diff
@@ -188,6 +188,7 @@ func (m *Manager) ProvideConfig(sessionID string, sessionConfig json.RawMessage,
 		if err := conn.Stop(); err != nil {
 			log.Error().Err(err).Msg("Failed to stop connection endpoint")
 		}
+		conn.ReleaseIP(config.Consumer.IPAddress)
```
soffokl, line 191: The same are done in the next line, no?

Zensey: Fixed

PR #5972, opened 2024-02-15
CI migration to Github Actions

tzununbekov:
This is part 1 of the CI migration to the GitHub Actions. After this PR is merged, we'll have the old CI workflows unchanged, and GitHub Actions that will run in parallel, with gradually increasing responsibilities scope. At this stage, the GitHub Actions workflow includes:
- Unit tests and lints - `mage check`, `mage testwithcoverage`,
- Codecov report,
- E2E basic and NAT scenarios.

Additionally:
- `build/env.sh` file generation dropped until there will be a need for a custom env setup,
- GitHub `production` environment created with vars/secrets from the current CI.

The proposed update triggers workflow run on the `pull_request` events - this will be tuned later on.
Part 2 of the migration process will extend the workflow jobs to the package build operations.

---

The previously existing workflow named `Released` was renamed to `Mobile-release` to better reflect its purpose. It did not pass the execution before, and won't be fixed or touched in the course of these updates.

### DIFF
--- a/.github/workflows/build-and-release.yml
+++ b/.github/workflows/build-and-release.yml
@@ -1,23 +1,8 @@
 name: Test and build
-on:
-  push:
-    branches:
-      - ci-migration
+on: [pull_request]
+
 env:
-  # following envs are added for the back compatibility
-  # we need to check if they are used at all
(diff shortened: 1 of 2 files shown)


Guillembonet (approved):
nice to finally have this

PR #5974, opened 2024-02-19
CI migration to Github Actions, part 2

tzununbekov:
Continuation of the CI migration to GitHub Actions started in [Part 1](https://github.com/mysteriumnetwork/node/pull/5972).

This PR includes the following changes:
- Packages build workflow added
  This workflow utilizes [strategy matrix](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs) to run multiple build operations in parallel - currently, 4 (chosen semi-randomly) platform build simultaneously. With this step, the original `go run mage.go -v GenerateEnvFile` operation had to be restored because the package build operations depend heavily on the internal CI framework, and changing that is not in the scope of this task.

- Single CI workflow is broken down into 3 separate pipelines: unit tests and linters, E2E, and build packages.
  The first version of the new workflow was triggered on every PR. With the new build packages job added it had to be changed, and having multiple pipelines that depend on each other is the most suitable approach. The proposed update makes the following CI job chain: 
  - _Unit tests and linters_ -> all PRs
  - _E2E (basic and NAT)_ -> successful unit tests and linters completion, push to `v*` tags (release)
  - _Build packages_ -> successful E2E workflow completion on _master_ branch 

  Release packages workflow will be added as the last step after the "Build packages" job completion.
  This CI pipeline logic is not final and is open for discussion.
  NOTE: The internal CI framework is still responsible for most of the triggering logic based on the env variables, so pipeline separation has a more cosmetic effect - we should see different jobs fired in different conditions instead of one huge workflow for every single push with the hidden skip/run switches.


tzununbekov:
**UPDATE**: After internal discussion, it was decided to trigger all workflows except release (linters, unit tests, E2E tests, all platforms build) on every single PR.

soffokl (approved):

Issue #5999, opened 2024-02-29
Data races


Zensey:
**Describe the bug**
A few data races have been found
```
hermes channels state race:
==================
WARNING: DATA RACE
Read at 0x00c00027c3c0 by goroutine 12549:
  reflect.typedmemmove()
      C:/Program Files/Go/src/runtime/mbarrier.go:193 +0x0
  reflect.packEface()
      C:/Program Files/Go/src/reflect/value.go:135 +0xc5
  reflect.valueInterface()
      C:/Program Files/Go/src/reflect/value.go:1520 +0x179
  reflect.Value.Interface()
      C:/Program Files/Go/src/reflect/value.go:1490 +0x3d99
  github.com/jinzhu/copier.copier()
      C:/Users/user/go/pkg/mod/github.com/jinzhu/copier@v0.3.5/copier.go:190 +0x3d27
  github.com/jinzhu/copier.copier()
      C:/Users/user/go/pkg/mod/github.com/jinzhu/copier@v0.3.5/copier.go:292 +0x3009
  github.com/jinzhu/copier.CopyWithOption()
      C:/Users/user/go/pkg/mod/github.com/jinzhu/copier@v0.3.5/copier.go:76 +0xec
  github.com/mysteriumnetwork/node/core/state.(*Keeper).announceState.func1()
      C:/Users/user/src/node/core/state/state.go:241 +0x98
  github.com/mysteriumnetwork/node/core/state.(*Keeper).announceState()
      C:/Users/user/src/node/core/state/state.go:244 +0xa4
  github.com/mysteriumnetwork/node/core/state.(*Keeper).announceState-fm()
      <autogenerated>:1 +0x47
  github.com/mysteriumnetwork/node/core/state.debounce.func1.1()
      C:/Users/user/src/node/core/state/state.go:614 +0x4f

Previous write at 0x00c00027c3c0 by goroutine 12543:
  github.com/mysteriumnetwork/node/session/pingpong.(*HermesChannelRepository).updateChannel()
      C:/Users/user/src/node/session/pingpong/hermes_channel_repository.go:471 +0x22b
  github.com/mysteriumnetwork/node/session/pingpong.(*HermesChannelRepository).updateChannelWithLatestPromise()
      C:/Users/user/src/node/session/pingpong/hermes_channel_repository.go:457 +0x347
  github.com/mysteriumnetwork/node/session/pingpong.(*HermesChannelRepository).handleHermesPromiseReceived()
[...]

PR #6003, opened 2024-03-04
Fix data races

Zensey:
Resolves https://github.com/mysteriumnetwork/node/issues/5999


soffokl (commented):
@Zensey could you please add a test that proves the data race exists before the PR? 
And after the PR fix, this race test starts to work and will protect us from appearing again in the future?

Zensey:
> @Zensey could you please add a test that proves the data race exists before the PR? And after the PR fix, this race test starts to work and will protect us from appearing again in the future?

@soffokl 
done

soffokl (approved):

PR #6040, opened 2024-04-26
Add command flag for setting wireguard mtu

Zensey:


config/flags_node.go
```diff
@@ -425,3 +434,19 @@ func ValidateAddressFlags(flags ...cli.StringFlag) {
...
+
+	v := Current.GetInt(FlagWireguardMTU.Name)
+	if v == 0 {
+		return nil
+	}
+	if v < 68 || v > 1500 {
+		msg := "Wireguard MTU value is out of possible range: 68..1500"
+		log.Error().Msg(msg)
+
+		return errors.New(fmt.Sprintf("Flag validation error: %s", msg))
```
soffokl, line 449:
```suggestion
		return fmt.Errorf("Flag validation error: %s", msg)
```

services/wireguard/wgcfg/device_config.go
```diff
@@ -38,12 +38,13 @@ type Stats struct {
...
 // DeviceConfig describes wireguard device configuration.
 type DeviceConfig struct {
-	IfaceName  string    `json:"iface_name"`
-	Subnet     net.IPNet `json:"subnet"`
-	PrivateKey string    `json:"private_key"`
-	ListenPort int       `json:"listen_port"`
-	DNSPort    int       `json:"dns_port,omitempty"`
-	DNS        []string  `json:"dns"`
+	IfaceName    string `json:"iface_name"`
+	WireguardMTU int
```
soffokl, line 42: I would name it without `wireguard` word, since it's already a wireguard device config.
Do we need to set `json:"mtu"`?

soffokl (approved):

services/wireguard/wgcfg/device_config.go
Zensey: OK

Issue #6022, opened 2024-03-20
Empty response.


Zensey:
**Describe the bug**
When connected to certain nodes all TCP responses are empty.

**To Reproduce**
1. Connect
2. Run command:
```
curl.exe -v http://hello-world:8080/index123
*   Trying hello-world:8080...
* Connected to hello-world (hello-world) port 8080
> GET /index123 HTTP/1.1
> Host: hello-world:8080
> User-Agent: curl/8.4.0
> Accept: */*
>
* Empty reply from server
* Closing connection
curl: (52) Empty reply from server
```

Nodes that are known to have an empty TCP responses:
windows:
0x07ca2cd1079490a005e551da5d2f8a2dc516222f 
0x0733df3b8cb3ed5f13ba860d19c686e88437189f
darwin:
0x1a52807858619c4acb2a075b659a86dbe459bc98
0x0f7689807782b66d0569783ea6e56883b02240b5

Zensey:
Logs from hellow-world show that it gets a correct request and reply is also done successfully,
so I suspect some firewall rule or app is what interferes.

Zensey:
Internal spreadsheet with node runners feedback on this issue
https://docs.google.com/spreadsheets/d/19HfNETFTQo2EBYDhFMsr2J2KbCpLhb5SeP29Ix1wVk4/edit?gid=0#gid=0

PR #6047, opened 2024-05-14
Provider checker API

Zensey:
Summry of task:
Essentially, the program should get the connection parameters from the API, raise the connection and make sure that it works.
Establishing a VPN connection to wireguard must be done through the wg+gvisor combination, because raising a real network interface is not convenient especially in could environments, so a VPN connection that works entirely from code would be very useful.

Implementation notes:
Check of provider's functioning is done by requesting to http://107.173.23.19:8080/test

How to build
go run mage.go buildProvChecker

Example of use
.\build\myst\myst.exe  --provchecker  --log-level=error daemon

Example of use #2 (with postgres - to save results of scan into table)
.\build\myst\myst.exe --provchecker --log-level=error --checker.dsn="host=yyy user=mypguser password=xxx dbname=myst_nodes port=5432 ss
lmode=disable" daemon

Api
http://localhost:4050/prov-checker?id=0x020dd683990895d0f81a47702161f1d29d209976

Response:
{"status":true,"error":null,"provider_id":"0x020dd683990895d0f81a47702161f1d29d209976"}

Batch mode (check many providers):
```
curl.exe -vvv -H "Content-Type: application/json" \
  http://localhost:4050/prov-checker-batch \
  -d '["0x020dd683990895d0f81a47702161f1d29d209976","0x1a52807858619c4acb2a075b659a86dbe459bc98"]'
```

Response:
```
[
  {
    "status": false,
    "error": "err_connect",
    "provider_id": "0x020dd683990895d0f81a47702161f1d29d209976"
  },
  {
    "status": true,
    "error": null,
    "provider_id": "0x1a52807858619c4acb2a075b659a86dbe459bc98"
  }
]
```

Batch mode #2 (check all providers of a given country or gloabally):
```
curl.exe -vvv -H "Content-Type: application/json" \
  http://localhost:4050/prov-checker-batch2?location=US
```


soffokl (commented):
Do we have a task for this with requirements? From now it looks like it duplicates what we have for providers checks in the monitoring agent.

Zensey:
> Do we have a task for this with requirements? From now it looks like it duplicates what we have for providers checks in the monitoring agent.

The difference from monitoring agent is the capability to function w/o using OS network interfaces, thus making it more cloud-friendly. The real application of that API is search of provider nodes affected by bug #6022

PR #6053, opened 2024-06-12
Bump NodeUI (dvpn-web v2.17.0)

mingaila:


Guillembonet (approved):
you can run `go mod tidy` to get rid of the previous version reference in go.sum

PR #6192, opened 2026-05-18
fix: add 16KB page size alignment for Android AAR builds

cvl:
Google Play requires apps targeting Android 15+ to support 16KB memory page sizes by Nov 2025. Add -extldflags=-Wl,-z,max-page-size=16384 to Go linker flags for both consumer and provider Android builds.


soffokl (approved):
Thank you, let's try.

soffokl:
@cvl it's released as 1.38.3. 
Please let us know if this solved the problem.

cvl:
yes, thank you for super speed.

